PR #21, opened 2018-03-07
Replace proprietary googleApi.js

Larivact:
* replace proprietary and obfuscated googleApi.js with youtubeApi.js & refactor the code
* replace toggleLoading with more descriptive startLoadingAnimation and stopLoadingAnimation
* fix endless loop in subscriptions.js: loadSubscriptions()
* remove duplication of apiKeyBank
* update README.md

See #20.


PrestonN:
This looks really good.  I'll have to take a closer look at it tonight but I don't see any issues at a glance.

PrestonN:
Hey there.  This seems to work fine, however it looks like loading up subscriptions run much slower versus keeping the API stored the way it is.  This becomes a big problem the more subscribers you end up having.  I want to merge this but I don't know if I'd want to take that big of a hit on performance.

PrestonN:
Sorry about this taking so long.  I wanted to make some changes to the code and get your opinion before I merged this.  I know that I don't have anything setup as far as coding standards but I'd like to keep this consistent if possible.  Here's a quick list of what I did.

* Merged my latest changes into your branch to resolve conflicts.

* Removed the request module.  This same functionality can be accomplished with jQuery/AJAX, which is already in the project.  I'd rather not include more modules if it's not needed.  Using AJAX as well allows the call to store in cache for a little bit, which can speed up navigation around the project.  If you know of a benefit with using request over AJAX then I'd be interested in hearing it.

* Removed a try catch in subscriptions.js.  Because the YouTube API function handles errors, this wasn't needed.  It also caused issues with the new changes which maxed out some of my API keys late last night during testing.

* Added comments to the function and slightly changed the formatting.

If you're okay with these changes then I'm more than happy to merge.  Thank you very much for doing this :)

Larivact:
I undid your commits as retroactively adding commits obscures history. I agree with changing my indentation to match yours but this pull request is not the place to establish new formatting styles.

PrestonN:
Thank you for understanding.  If issues come up in the future and it makes more sense to use request over jQuery for error handling, then I'd be willing to consider it.  For now, I'd like to keep the dependencies as low as possible.

I agree that this isn't the place for making standards.  I will write up some documentation for this during the weekend.

Could you change the variable in line 503 in videos.js from data to response?  Videos will not play because of this error.  Once that has been changed then we're good to merge :)

Larivact:
I don't see any error, video playing works fine for me.

PrestonN:
It's probably just some merge conflicts on my end.  Either way I'll get my stuff cleaned up before a new release.

Thanks for your help.  Merged :)

PR #33, opened 2018-03-10
Load templates with fs.readFileSync() instead of jquery.get()

Larivact:
I agree my contribution to be licensed under the GPLv3+.

Loading templates via HTTP requests doesn't make any sense and just clutters the network tab in the developer console.


PrestonN:
This looks good to me.  Thank you very much for you're work so far.  I really appreciate it :)

PR #40, opened 2018-03-12
Refactor code

Larivact:
Refactor code to make it more readable.

* Don't create variables if they are only used once.
* Use object.key instead of object['key'].
* videos.js: rename displayVideos() to displayVideo()
* remove comments pointing out obvious things
* don't display toast when trying to search without a term


PrestonN:
I don't want to ever claim that I'm perfect especially with big projects like this.  Things like this keeps the code looking clean and I'm very appreciative of it.  Thank you once again for your contribution :)

PR #43, opened 2018-03-12
Move background-color property in loading.css to *Theme.css

5had0w5talk3r:
This patch moves the background-color property for the .double-bounce1 and .double-bounce2 classes from loading.css into the css for both base themes. This will allow for the easier creation of themes by the community.


PrestonN:
Nice catch!  Definitely something that needed to be moved.  Merged and thank you :)

Issue #136, opened 2018-07-17
Removing/Replacing HookTube integration


2a747:
From the HookTube changelog: 

> - rest in pieces
> 
> - It was a good run, 1.5 years. Started as a quickly made addition to the norbot project, and within long the server had to be upgraded several times. Of course YouTube Legal was an inevitability at that point.
> 
> - July 16: YouTube api features are back but mp4 <video> is replaced with the standard YT video embed. HookTube is now effectively just a light-weight version of youtube and useless to the 90% of you primarily concerned with denying Google data and seeing videos blocked by your governments.

Since the latest HookTube updates there is really no point in supporting it in your app any longer. There is already a alternative: https://invidio.us/. I am not familiar with invidio but maybe someone else here is. Looks like a good alternative and other than HookTube it's open source: https://github.com/omarroth/invidious . Might be worth looking into. Thanks and keep up the good work.

PrestonN:
I've already had plans on doing this.  I mentioned it in our Matrix server a couple of days ago as everything was happening.  Switching support from HookTube to Invidious will only take a couple of lines to change so it will be no issue doing this for the next release.

thisisjustathrowaway:
So wait, what exactly happened with Hooktube?

PrestonN:
HookTube was given a cease and desist from YouTube for their use of YouTube's API. The site has been forced to use YouTube's official player and is no longer a viable option for those focused on privacy. Because of their changes, there isn't a reason for FreeTube to continue including links to HookTube. Invidious is a new website similar to HookTube that will replace the current HookTube links in the next release.

PrestonN:
I think the main reason has to do with the fact that HookTube became very popular which made YouTube aware of it's existence.  The other reason is that using YouTube's API in conjunction with not providing YouTube income is against their terms of use with their API.

FreeTube also uses YouTube's API, though I believe FreeTube is in a slightly better position.  Before I mention anything this is an obligatory "I am not a laywer, take everything I say with a grain of salt."

Because of HookTube being a website, HookTube provides a "service" to users by helping them bypass YouTube restrictions.  This puts HookTube as the person that is bypassing YouTube's restrictions.  Because FreeTube is a downloadable app, the user takes that responsibility instead of me as the developer.  FreeTube isn't a "service" like HookTube is and would be treated slightly different because of that.  I could be 100% wrong on everything I just said so please don't take anything here as fact.

[...]

Issue #139, opened 2018-07-22
Ability to unsave videos


K06RA:
I've noticed that there isn't a way to unsave videos. There's already a toggle function so it shouldn't be too hard to implement.

K06RA:
Issue will be fixed in my pull request... There actually was a way to unsave through the context menu, you just couldn't see it because of the css.

K06RA:
Fixed w/ Pull #140

Issue #122, opened 2018-07-09
reword the "save" option


grrrrr:
I'd like to suggest rewording the `save` option that is currently with the player as it feels a bit confusing.

`save` is more closer to `download` in meaning on computer.    I would like to suggest using `favorite` or `bookmark` instead as it is more equivalent to what it does.

PrestonN:
I can understand that as being confusing, especially for when proper download support comes in the future.  I'll change it to `Favorites` for the next release.

K06RA:
Fixed this issue.. You can close it when you see my push.

Issue #95, opened 2018-05-20
Muting?


SirKnight1234:
On Twitter there's a function to mute words and users so they don't appear on your feed. Can something like this be added on this app?

LexHimself:
Hi,

Can you specify a bit more? 
Do you just want to hide certain users from the comments section, or their channels? Or both? (block them)

Certain words could be blocked if specified by a user, by using a word filter for example. 

Where in FreeTube would you like to see the above blocked/hidden? 

I like the idea, not sure if the data can be parsed that way though.

SirKnight1234:
I would like to be able to hide users from the comments section and able to hide users' videos from the sidebar. Not merely the option of doing both like blocking does, but also the option of doing just one or the other.

With words and phrases, I also mean not only in comments sections but in videos on the sidebar as well. 

There could be like a page where you can manage these filtered users/phrases/words like there is on Twitter. Of course, it should not be like on Youtube where they guess what to hide from you based on your filtering choices. It should be more specific, like on Twitter.

MegaArthur:
Wouldn't that be kind of too echo-chambery?

SirKnight1234:
The way in which Youtube recommends you videos by tracking your data is echo-chambery. In this system, anything "echo-chambery" is knowingly self-imposed.

SirKnight1234:
@PrestonN Are you planning on making this an enhancement or not really?

PrestonN:
@SirKnight1234 I apologize for not responding to this issue.  Yes I would like to see this implemented in the future.  Unfortunately I do not know when I will be able to implement this as there a certain features that need to take priority for FreeTube.

I feel bad about this answer as I feel like I say the same thing on every issue so I'd like to explain a little more.  FreeTube is in a strange position right now where it is getting more popular, but there hasn't been anyone willing / able to help out towards contributing to the project.  This leaves myself as the sole person responding to issues, organizing, prioritizing, developing, testing, documentation, and anything else related to the project.  I don't mind doing this, I wouldn't have created the project if I didn't want to do this.  The problem is more so that I can only work on FreeTube in my spare time and progress takes a hit because of it.  The upside is that I don't have to ask for donations towards the project as FreeTube is not my source of income.

Progress would advance much quicker if there were even 1 or 2 people that could be dedicated to helping out.  I can't force anybody to help but I at least want everyone to know that the door is open.  I might not respond to everything but I do read everything on here.  I apologize for not responding but I hope you understand.

[...]

PR #140, opened 2018-07-23
Fixed Issues

K06RA:

I fixed issue #136 #139 #122. Also cleaned up the entire codebase to use a single formatting standard. I added the comment headers on every js and html file. I also made the console logs a lot prettier and gave a lot more information (time and such). 

I also think that issue #4 can be closed as an isolated incident and probably due to electron. I also think that issue #95 would be super hard to implement with our current way of using the YouTube API. 

I wanted to submit this pull request before I try to integrate tags and playlists etc. I know there is a LOT of changes but they're for the better!


PrestonN:
This is really good.  Everything looks much cleaner now!

When you use your API key for testing, I'd recommend using the UI to set your key instead of changing the hard coded ones.  This will prevent your key from leaking and you won't have to remember to change it back afterwards.

As we mentioned in the Matrix server, I plan on implementing Vue.js in the next update.  Because of the large amount of changes that we both made, there will definitely be several merge conflicts to look at.  I'll let you know when I plan on merging so that we can both look at it at the same time.

Thank you very much for this work.  I really appreciate it. :)

Issue #161, opened 2018-09-17
Tor broken with the appimage.


haydenhughes:
If you try to use tor with the appimage with tor it just constantly loads without doing anything. Just a guess but maybe appimage is blocking the connection? If I clone the repo and use `npm start` tor works as intended. BTW the distro Fedora 28. I haven't tried to replicate the issue on other distros.

PrestonN:
I was able to reproduce the issue.  I will take a look and see what I can do.  I'll update here if I figure anything out.

PrestonN:
This has been fixed and will be included in the next release.

PR #162, opened 2018-09-17
Fix flatpak building

haydenhughes:

Fix the build error with `npm run make:linux:x86:flatpak`. This was being caused because `electron-installer-flatpak` using out dated library versions by default.

Note: Similarly to the appimage, tor doesn't work in flatpak. See #161


PrestonN:
Hey I appreciate this.  I've actually been wanting to get Flatpak set up but I haven't had the time to do it.  I should be able to start supporting Flatpak in the near future now. :)

Thanks!

PR #191, opened 2018-11-04
Change "Favorited" to "Saved"

### LINKED ISSUE #182: Settings "Clear Saved Videos" button
In #122 the "saved video" option in the side bar was reworded to "Favorites".   This does not look to have been reflected on the settings page where the button is still called "Clear saved videos" .   For consistency and to avoid confusion it would be nice to have the name change reflected on the settings page.

ivanhercaz:

- As in #182 @grrrrr tell, it is necessary to change this string according to last changes: "Favorited/Favoritos" instead of "Saved/Guardados".

The commit said "fix #182" but it is fixed partially, because I could change only this two strings in English and Spanish. I might try to change in other languages like German, French and Italian, but it would be better if someone native or with a good level would do it.

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -82,7 +82,7 @@
 	"Export Subscriptions": "Export Subscriptions",
 	"Clear History": "Clear History",
 	"Are you sure you want to delete your history?": "Are you sure you want to delete your history?",
-	"Clear Saved Videos": "Clear Saved Videos",
+	"Clear Saved Videos": "Clear Favorited Videos",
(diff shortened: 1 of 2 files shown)


PrestonN:
I appreciate you looking through it.  The strings in the locales folder desperately need to be updated all around.  I think I'm going to try to work on updating them and getting them usable for 0.5.0.  I have been approved for Weblate so future translations should be easier to make.  I'll be making a blog post about it at some point however I'd like to clean up the current locales and start actually using them within FreeTube before I do that.

Thanks again!

Issue #208, opened 2018-12-31
Paid videos spam subscription feed


zorbakus:
One of my subs is "Vsauce" who has some non-free videos which you have to buy first. 
I attached a screenshot where you can see that those videos appear on top of the sub feed with 0 views and 0 or 2 seconds ago. I have to scroll further way down to see actual feed content.

![freetube_glitch](https://user-images.githubusercontent.com/43297999/50566656-39299e00-0d34-11e9-839d-3905d6325488.png)

MegaArthur:
Same here, also VSauce

MegaArthur:
But also, duplicate of https://github.com/FreeTubeApp/FreeTube/issues/198

PrestonN:
This should be fixed with #210

PR #210, opened 2019-01-18
Filter premium content from descriptions

### LINKED ISSUE #198: Youtube Originals cluttering subscriptions
I am aware you have a lot of stuff to do, mainly the syncing of audio files and video files to support 1080p+, but I would just like you to know that channels which have a youtube original series clutter at the top of my subscriptions tab, not sure if its just me but seems like it isnt. Thanks

omarroth:

Accurate info on when premium content is published is currently not available from the Invidious API, so this fix filters it out. 

Of note is that some premium content is still watchable without having a subscription, and can be filtered using the `paid` attribute. A better fix might be to cache premium content in some way in order to provide a consistent `published` time.

Before:
before
After:
after

Would close #198 and #208.

### DIFF
--- a/src/js/subscriptions.js
+++ b/src/js/subscriptions.js
@@ -131,6 +131,10 @@ function loadSubscriptions() {
 }
 
 function addSubsToView(videoList) {
+    videoList = videoList.filter(a => {
+        return !a.premium;
+    });
+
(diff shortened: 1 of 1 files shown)


PrestonN:
Looks good to me.

PR #213, opened 2019-01-27
Grammar and spelling fixes

rm-rf-home:
"Kündigen" is a better translation for "unsubscribe" because "Abmelden" usually means "log out".
No need to use formal "Sie", "Du" is now used for all translations

- [X] I have read and agree to the [Contribution Guidelines](https://github.com/FreeTubeApp/FreeTube/blob/master/CONTRIBUTING.md)
- [X] I can maintain / support my code if issues arise.


locales/de.json
```diff
@@ -53,7 +53,7 @@
 	"View": "Wiedergabe",
 	"Views": "Wiedergaben",
 	"Subscribe": "Abonnieren",
-	"Unsubscribe": "Abmelden",
+	"Unsubscribe": "Kündigen",
```
TobiGr: I'd say "Deabonieren". This is not contract or something similar

rm-rf-home: True, "kündigen" sounds more like cancelling a contract. Then I would suggest to use "Abo beenden". This is also what is used on YouTube.

```suggestion
	"Unsubscribe": "Abo beenden",
```

TobiGr: Sounds good 👍

PrestonN:
Great!  Thanks you!

PR #214, opened 2019-01-31
[Fix] Toggle saved video in videoListTemplate

ivan1548:

Updates toggleSave method on videos in videoListTemplate to actually remove videos from Favorites if it's already added.

It also solves an issue where there is no way to remove videos from Favorites without going to Video view page.

Update 02.18.2019: Only update this in savedView because it only really makes sense there.


PrestonN:
Looks good to me.

Issue #202, opened 2018-12-08
History duplicates


openffchrome:
Hello, it would be nice to check and remove the duplicate videos in the history and instead leave only the last played

thanks

PR #216, opened 2019-02-01
Remove duplicates from history

ivan1548:

Update to history functionality. Check if video already exist in history. If it does then update timestamp of history item instead of adding a new one.

Would close https://github.com/FreeTubeApp/FreeTube/issues/202


PrestonN:
Looks good.

PR #225, opened 2019-02-20
Update tor support with electron proxy

ivan1548:

Update tor support by using electron's [setProxy](https://electronjs.org/docs/api/session#sessetproxyconfig-callback) method

Settings are updated with a new field _proxy_. That field has a default value of 

> SOCKS5://127.0.0.1:9050

which is a default address for tor client.

Settings view is updated with new section for Tor and proxy fields.
https://i.ibb.co/mck7Hyg/settings.png
That section also includes **TEST PROXY** button which sends a request to https://ifconfig.co/json and displays information about user's IP.

I haven't worked with Tor before so it would be a good idea for someone who has some experience using tor to test this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -125,6 +125,7 @@
       "version": "4.2.0",
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-4.2.0.tgz",
       "integrity": "sha512-c+R/U5X+2zz2+UCrCFv6odQzJdoqI+YecuhnAJLa1zYaMc13zPfwMwZrr91Pd1DYNo/yPRbiM4WVf9whgwFsIg==",
+      "dev": true,
       "requires": {
         "es6-promisify": "^5.0.0"
       }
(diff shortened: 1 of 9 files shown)


PrestonN:
Tor support seems to work pretty well.  I definitely prefer this method compared to how it was done before.  There are some small changes I'd like to see made however.

- When trying to test a proxy, the test won't be made properly unless I save the settings prior. Could this be changed to do a test regardless of the user's settings?
- If I try to do a test and the settings are not saved properly, my input gets changed to `false`. Can you update this to not overwrite the user's input?
- Could the default proxy be filled into the input instead of `false`?
- Will videos get proxied as well through Tor?  We could change the wording for the setting if that's the case.  Change from "Use Tor for API Calls" to something like "Proxy Data Through Tor" or something similar.

Definitely looking forward to merging this. :)

ivan1548:
Updates
- You can test proxy without saving now.
- Input is no longer overwritten when testing proxy
- Default value for proxy address is changed from `false` to `SOCKS5://127.0.0.1:9050`
- `tor-request` removed
- Added new method [`proxyRequest`](https://github.com/FreeTubeApp/FreeTube/blob/bc7bea2bde62a51b8de1e72310d28e329635e8fe/src/js/layout.js#L222-L245). It is a wrapper that waits for proxy to become available and then executes ajax request. 

I think all videos will be proxied through tor but I'm not sure how to test that.

PrestonN:
Perfect!  Everything looks great.  I know a lot of people will be looking forward to this, thank you!

I'll have to try to find a way to test it, but we may be good to say that videos get proxied as well.  I'll keep the wording there for now but we may be able to change that.

Issue #125, opened 2018-07-09
Import NewPipe subscription exports


grrrrr:
Could it be possible to get a feature that you allow users of NewPipe to export their subscriptions from there and either import or merge into FreeTube.

This would help a long way to making the most use of these two great programs

tiotrom:
How's the progress with this? :)

PrestonN:
This has been finished thanks to #234

Slomo17:
> To add on to this, I think it would be pretty cool if we can have FreeTube and NewPipe sync together using Syncthing or something similar. This would take quite a bit of coordination with the NewPipe guys though so we could set everything up properly.

Syncing with Newpipe over Syncthing would be perfect, than there would really be no more reason to use a google account for youtube

PR #234, opened 2019-03-17
Implement importing of Newpipe Subscriptions 

kskarthik:

Enhancement https://github.com/FreeTubeApp/FreeTube/issues/125

### DIFF
--- a/src/js/settings.js
+++ b/src/js/settings.js
@@ -491,6 +491,7 @@ function importSubscriptions(){
       }
       else if( (fileType === '.json') && (data.includes("app_version")) ) {
         importNewpipeSubscriptions(data);
+        return;
       }
       else if ((fileType !== '.db') && (fileType !=='.json')) {
         showToast('Incorrect file type.  Import unsuccessful.');
@@ -517,7 +518,8 @@ function importNewpipeSubscriptions(data){
 
   progressView.seen = true;
   progressView.width = 0;
-    
+  showToast('Importing Newpipe Subscriptions, Please Wait.');
+
   let newpipe, n, link, newpipesub, counter;
       
       newpipe = JSON.parse(data);
@@ -533,16 +535,17 @@ function importNewpipeSubscriptions(data){
               channelName: data.author,
               channelThumbnail: data.authorThumbnails[2].url
             };
-            counter++;
               addSubscription(newpipesub, false);
+              counter++;
               progressView.progressWidth = (counter / newpipe.subscriptions.length) * 100;
+            
+              if ((counter + 1) == newpipe.subscriptions.length) {
+                showToast('Subscriptions have been imported!');
+                progressView.seen = false;
+                progressView.seen = 0;
+                return;
+        }
     });
-  }
-  if ((counter + 1) == newpipe.subscriptions.length) {
-    showToast('Subscriptions have been imported!');
-    progressView.seen = false;
-    progressView.seen = 0;
-    return;
   }
 }
 /**
--- a/src/js/settings.js
+++ b/src/js/settings.js
@@ -506,7 +506,65 @@ function importSubscriptions(){
     })
   });
 }
+/**
+* Import NewPipe Channels Subscriptions
+* @return {Void}
+*/
+function importNewpipeSubscriptions(){
+  // Open user's file browser.  Only show .json files.
+ dialog.showOpenDialog({
+  properties: ['openFile'],
+  filters: [
+    {name: 'JSON', extensions: ['*']},
+  ]
+}, function(fileLocation){
+  if(typeof(fileLocation) === 'undefined'){
+    console.log('Import Aborted');
+    return;
+  }
+  console.log(fileLocation);
+  let i = fileLocation[0].lastIndexOf('.');
+  let fileType = (i < 0) ? '' : fileLocation[0].substr(i);
+  console.log(fileType);
 
+  fs.readFile(fileLocation[0], function(readErr, data){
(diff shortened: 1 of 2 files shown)


src/js/settings.js
```diff
@@ -506,7 +506,61 @@ function importSubscriptions(){
...
+          link = newpipe.subscriptions[n].url.split("/");
+  
+          invidiousAPI('channels', link[4], {}, (data)=> {
+            newpipesub = {
+              channelId: data.authorId,
+              channelName: data.author,
+              channelThumbnail: data.authorThumbnails[2].url
+            };
+
+            showToast("Importing Newpipe Subscriptions. Please restart freetube when this message disappears, to see changes.");
```
PrestonN: FreeTube will not require a restart when importing in this way.  You can remove this line.

src/js/settings.js
```diff
@@ -506,7 +506,61 @@ function importSubscriptions(){
...
+      showToast('Unable to read file.  File may be corrupt or have invalid permissions.');
+      throw readErr;
+    }
 
+    if (data.includes("app_version")){
+      let newpipe, n, link, newpipesub;
+      
+      newpipe = JSON.parse(data);
+      
+      for (n in newpipe.subscriptions) {
```
PrestonN: This loop will likely take a long time.  Please use the loading bar to visualize the progress.  Take another look at the [OPML Import](https://github.com/FreeTubeApp/FreeTube/blob/3d29ce55a50ae1a3c3093338616e48859a86c61c/src/js/settings.js#L440) function for an example on how to do this.

src/js/settings.js
```diff
@@ -506,7 +506,61 @@ function importSubscriptions(){
...
+  console.log(fileLocation);
+  let i = fileLocation[0].lastIndexOf('.');
+  let fileType = (i < 0) ? '' : fileLocation[0].substr(i);
+  console.log(fileType);
+
+  fs.readFile(fileLocation[0], function(readErr, data){
+    if(readErr){
+      showToast('Unable to read file.  File may be corrupt or have invalid permissions.');
+      throw readErr;
+    }
```
PrestonN: Everything from here and above should be removed.  Please edit the existing Import functionality instead.

src/js/settings.js
```diff
@@ -506,7 +506,61 @@ function importSubscriptions(){
...
+              channelName: data.author,
+              channelThumbnail: data.authorThumbnails[2].url
+            };
+
+            showToast("Importing Newpipe Subscriptions. Please restart freetube when this message disappears, to see changes.");
+              addSubscription(newpipesub, false);
+        })
+      }
+    }
+    else if (fileType !== '.json'){
```
PrestonN: This should be checked in the main import function.  This can be removed.

src/templates/settings.html
```diff
@@ -75,6 +75,9 @@ <h4>Your Info:</h4>
         <div onclick='importSubscriptions()' class='settingsButton'>
             IMPORT SUBSCRIPTIONS
         </div>
+        <div onclick='importNewpipeSubscriptions()' class='settingsButton'>
+            IMPORT NewPipe SUBSCRIPTIONS <small>(experimental)</small>
+        </div>
```
PrestonN: This would be removed as well as the other import button will be used instead.

src/js/settings.js
```diff
@@ -506,7 +506,61 @@ function importSubscriptions(){
...
+  let fileType = (i < 0) ? '' : fileLocation[0].substr(i);
+  console.log(fileType);
+
+  fs.readFile(fileLocation[0], function(readErr, data){
+    if(readErr){
+      showToast('Unable to read file.  File may be corrupt or have invalid permissions.');
+      throw readErr;
+    }
 
+    if (data.includes("app_version")){
```
PrestonN: Add this check to `importSubscriptions()` instead and make a call to this function.

src/js/settings.js
```diff
@@ -506,7 +509,42 @@ function importSubscriptions(){
...
+              channelId: data.authorId,
+              channelName: data.author,
+              channelThumbnail: data.authorThumbnails[2].url
+            };
+            counter++;
+              addSubscription(newpipesub, false);
+              progressView.progressWidth = (counter / newpipe.subscriptions.length) * 100;
+    });
+  }
+  if ((counter + 1) == newpipe.subscriptions.length) {
```
PrestonN: Add this if statement into the invidiousAPI callback.  Node runs asynchronously so the statement is ran before the last API call is finished.

src/js/settings.js
```diff
@@ -506,7 +509,42 @@ function importSubscriptions(){
...
 }
+/**
+* Import NewPipe Channel Subscriptions
+* @return {Void}
+*/
+function importNewpipeSubscriptions(data){
 
+  progressView.seen = true;
+  progressView.width = 0;
+
```
PrestonN: Add a showToast call here using the same wording from the OPML import.

`showToast('Importing subscriptions, please wait.');`

src/js/settings.js
```diff
@@ -489,7 +489,10 @@ function importSubscriptions(){
         });
         return;
       }
-      else if (fileType !== '.db'){
+      else if( (fileType === '.json') && (data.includes("app_version")) ) {
+        importNewpipeSubscriptions(data);
```
PrestonN: Add a return statement here.  The code underneath will keep running otherwise.

PrestonN:
Perfect!  Looks good to me.

PR #243, opened 2019-03-25
Add support to export as Newpipe & OPML formats

kskarthik:

Support export as NewPipe & OPML formats.

### DIFF
--- a/src/js/settings.js
+++ b/src/js/settings.js
@@ -591,12 +591,56 @@ function exportSubscriptions() {
   const dateYear = date.getFullYear();
   const dateString = 'freetube-subscriptions-' + dateYear + '-' + dateMonth + '-' + dateDay;
 
-  // Open user file browser. User gives location of file to be created.
+    switch(document.querySelector('#exportSelect').value){
+      
+      case "NewPipe":
+        exportNewpipeSubscriptions(dateYear, dateMonth, dateDay); 
+      break;
+      case "OPML":
+        exportOpmlSubscriptions(dateYear, dateMonth, dateDay);
+      break;
+      default:
+        // Open user file browser. User gives location of file to be created.
+        dialog.showSaveDialog({
+          defaultPath: dateString,
+          filters: [{
+            name: 'Database File',
+            extensions: ['db']
+          }, ]
+        }, function(fileLocation) {
+          console.log(fileLocation);
+          if (typeof(fileLocation) === 'undefined') {
+            console.log('Export Aborted');
+            return;
+          }
+          fs.readFile(appDatabaseFile, function(readErr, data) {
+            if (readErr) {
+              throw readErr;
+            }
+            fs.writeFile(fileLocation, data, function(writeErr) {
+              if (writeErr) {
+                throw writeErr;
+              }
+              showToast('Susbcriptions have been successfully exported');
+            });
+          })
+        });
+    }
+}
+/**
+ * Export the subscriptions database compatable with NewPipe.
+ *
+ * @return {Void}
+ */
+function exportNewpipeSubscriptions(dateYear, dateMonth, dateDay){
+
+  const dateString = 'newpipe-subscriptions-' + dateYear + '-' + dateMonth + '-' + dateDay;
+
   dialog.showSaveDialog({
     defaultPath: dateString,
     filters: [{
-      name: 'Database File',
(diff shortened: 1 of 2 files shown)


src/js/settings.js
```diff
@@ -591,33 +591,145 @@ function exportSubscriptions() {
...
-    fs.readFile(appDatabaseFile, function(readErr, data) {
-      if (readErr) {
-        throw readErr;
-      }
-      fs.writeFile(fileLocation, data, function(writeErr) {
-        if (writeErr) {
-          throw writeErr;
+    returnSubscriptions().then((result)=>{
+        let newpipe = {
+          app_version: "0.15.1",
```
PrestonN: NewPipe's current version is 16.1.  Let's match that version number.

src/js/settings.js
```diff
@@ -591,33 +591,145 @@ function exportSubscriptions() {
...
+          app_version_int: 71,
+          subscriptions: []
         }
-        showToast('Susbcriptions have been successfully exported');
+        for (let i=0; i < result.length; i++) {
+          invidiousAPI('channels', result[i].channelId, {}, (iapi)=>{
```
PrestonN: All of the information needed to make this file should already be saved in the user's database.  We shouldn't need to make an API call here.

src/templates/settings.html
```diff
@@ -317,6 +317,18 @@ <h4>Your Info:</h4>
     </div>
     <br />
     <br />
+    <div class="select center">
```
PrestonN: I would add 2 break lines above here to space out the dropdown from the above button.

src/templates/settings.html
```diff
@@ -317,6 +317,18 @@ <h4>Your Info:</h4>
...
+            <option>FreeTube</option>
+            <option>NewPipe</option>
+            <option>OPML</option>
+        </select>
+        <span class="select-highlight"></span>
+        <span class="select-bar"></span>
+        <label class="select-label">Subscriptions Export Format</label>
+    </div>
+    <br />
+    <br />
```
PrestonN: Let's remove these 2 break lines to move the dropdown closer to the export button.  This makes it easier at a glance to know that the dropdown is associated with these buttons.

src/js/settings.js
```diff
@@ -591,33 +591,145 @@ function exportSubscriptions() {
...
+              showToast('Susbcriptions have been successfully exported');
+              progressView.seen = false;
+                progressView.seen = 0;
+                return;
+              }
+            });
+          })
+        }
+        progressView.seen = true;
+        progressView.width = 0;
```
PrestonN: Using a database call instead of an API call should be fast enough where we don't need to use the loading bar.  Let's remove these.

src/js/settings.js
kskarthik: The reason i made api call is that Newpipe subs have a mix of both /channel/ & /user/ links, Api resolves exactly as opposed to the db. That's the reason. Let me know your opinion on this

src/js/settings.js
kskarthik: can be done. I intentionally used 0.15.1 for backward compatibility

src/templates/settings.html
kskarthik: image

src/js/settings.js
PrestonN: Will NewPipe accept the export file if we only use the `/channel/` links?  I don't know of a reason as to why it wouldn't.  If it accepts the file that way, let's go ahead and use the information already in the database.  If NewPipe requires the `/user/` links for certain channels then this can be left the way it is.

src/js/settings.js
PrestonN: Is there any notable changes that would require the version number to be lower for backwards compatibility reasons?  If there isn't, let's match their version number.

src/templates/settings.html
PrestonN: :+1:

src/js/settings.js
kskarthik: Will check that out & report back to you soon

src/js/settings.js
kskarthik: Nothing ATM

src/js/settings.js
kskarthik: Works!

Issue #235, opened 2019-03-20
A few suggestions for UI


tiotrom:
1. A back button. Often I watch a channel and say I watch their 66 video, but now I have to click again on their channel and scroll down to find where I was at. 
2. Add a "open in miniplayer" button for listed videos - this way it makes it easier to quickly watch a video without losing track of the page you were at.
3. Maybe when you have a list of videos (search or channel videos) and a video is clicked, it opens a new column to to right side where it can play the video. Like most modern email clients work. This way you do not lose track of what you were watching.  Like so

![2019-03-21_00-32](https://user-images.githubusercontent.com/34804533/54725928-207d9700-4b71-11e9-8575-d55648e4887f.png)

I think option nr 3 could be super useful and it is enough space to the layout to integrate it.
Cheers!

PrestonN:
Closing as these are mostly fixed and / or should be a new issue if desired in the rewrite.

Issue #231, opened 2019-03-13
Feature request: add right click mini player


ghost:
adding right click to open in mini player would be nice. Sometime would like to open multiple videos and keep browsing through subscription without going to the main video.

PrestonN:
This will be included in the next release.

PR #246, opened 2019-03-28
Add Play video in mini player to video options

kskarthik:

Implement #231 & #235 (2nd suggestion)

### DIFF
--- a/src/js/player.js
+++ b/src/js/player.js
@@ -329,19 +329,13 @@ function openMiniPlayer() {
     });
 }
 /**
- * Right click on video thumbnail opens it in pop up player
- *
- * @param {string} videoThumbnail - The URL of the video thumbnail.  Used to prevent another API call.
+ * Opens the video directly in pop up player
  *
  */
-function rightClickMiniPlayer(contextImg, contextUrl){
+function clickMiniPlayer(contextUrl){
 
-  let vId = contextUrl.split('=');
-    
+  let vId = contextUrl.split('v=');
   invidiousAPI('videos', vId[1], {}, (video)=>{
-
-    let videoSUrl = video.formatStreams[0].url;
-
     // Create a new browser window.
     const BrowserWindow = electron.remote.BrowserWindow;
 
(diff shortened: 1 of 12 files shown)


src/js/player.js
```diff
@@ -329,19 +329,13 @@ function openMiniPlayer() {
...
+ * Opens the video directly in pop up player
  *
  */
-function rightClickMiniPlayer(contextImg, contextUrl){
+function clickMiniPlayer(contextUrl){
 
-  let vId = contextUrl.split('=');
-    
+  let vId = contextUrl.split('v=');
   invidiousAPI('videos', vId[1], {}, (video)=>{
```
PrestonN: Use `youtubedlGetInfo()` instead of the Invidious API.  FreeTube uses this instead so that we can avoid GeoLocation issues.  While the video might play for you currently, it's possible that it wouldn't play for someone else with this method.

src/js/player.js
```diff
@@ -355,10 +349,9 @@ function rightClickMiniPlayer(contextImg, contextUrl){
     $.get('templates/miniPlayer.html', (template) => {
         mustache.parse(template);
         const rendered = mustache.render(template, {
-            videoUrl: videoSUrl,
-            videoThumbnail: contextImg
+            videoUrl: video.formatStreams[0].url,
```
PrestonN: This URL won't always be valid, especially when changing to `youtubedlGetInfo()`.  Some older videos only max out at 480p and checks are needed to make sure that the video is valid.  There are [several](https://github.com/FreeTubeApp/FreeTube/blob/48436e50b3be179562b9a5ce54006d5152c2df88/src/js/player.js#L346) [different](https://github.com/FreeTubeApp/FreeTube/blob/48436e50b3be179562b9a5ce54006d5152c2df88/src/js/player.js#L92) [checks](https://github.com/FreeTubeApp/FreeTube/blob/48436e50b3be179562b9a5ce54006d5152c2df88/src/js/videos.js#L407) that are being made to ensure that the video plays properly and changes the URL if one doesn't work.

src/templates/videoTemplate.html
```diff
@@ -17,10 +17,11 @@
                                 <li>Open in Invidious</li>
                             </a>
                             <li v-on:click='copy("invidio.us", video.id)' onclick='showVideoOptions(this.parentNode.previousSibling);'>Copy Invidious Link</li>
+                            <li onclick='clickMiniPlayer(this.parentNode.querySelectorAll("a")[0].href)'>Open in Mini Player</li>
```
PrestonN: Clicking on this should close the 3 dot menu.

src/js/player.js
kskarthik: So, how do achieve this?

src/templates/videoTemplate.html
kskarthik: Done!

src/js/player.js
kskarthik: Switched to youtubedlGetInfo()

PR #254, opened 2019-04-11
Fix scheme in author thumbnail

omarroth:

Currently FreeTube adds an `https:` to `authorThumbnail` to correct an issue with the Invidious API that was fixed with omarroth/invidious@1fd7ff5. This produces `https:https://yt3.ggpht.com...` which doesn't resolve.


PrestonN:
:+1:

Issue #257, opened 2019-04-24
In Manjaro Gnome no proper icon or ability to add to favorite


tiotrom:
Like the title said, this new version doesn't seem to have a proper icon when the application is open, neither you can pin to the favourites bar.

Brn9hrd7:
I have noticed the same issue when using the AUR package, it doesn't happen when using the flatpak version though.

PrestonN:
This should be fixed in the latest release.  Please re-open this issue if the problem still exists.

PR #285, opened 2019-06-19
Explicitly set the window icon on Linux

mat8913:

This is the patch I wrote to fix flathub/io.freetubeapp.FreeTube#7. #257 might be related.

### DIFF
--- a/src/js/init.js
+++ b/src/js/init.js
@@ -73,7 +73,7 @@ let init = function () {
         webPreferences: {
           nodeIntegration: true,
         },
-        icon: path.join(__dirname, '..', 'icons', 'iconColor.png');
+        icon: path.join(__dirname, '..', 'icons', 'iconColor.png')
     });
 
     settingsDb.findOne({
--- a/src/js/init.js
+++ b/src/js/init.js
@@ -25,8 +25,7 @@ const {
     BrowserWindow,
     dialog,
     protocol,
-    ipcMain,
-    nativeImage
+    ipcMain
 } = require('electron');
 const path = require('path');
 const url = require('url');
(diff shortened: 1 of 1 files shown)


PrestonN:
It's likely that we may not need to check for the OS, as I believe that the Icon doesn't show up for other systems as well.  Let's remove that if check and everything should be good.

mat8913:
In that case, should I just pass the image directly to the `BrowserWindow` constructor? i.e.
```
win = new BrowserWindow({
    ...,
    icon: path.join(__dirname, '..', 'icons', 'iconColor.png')
});
```

PrestonN:
Sorry for the late response.

If that works out for all cases (and I don't see why it wouldn't) then that will likely be the best option.  Let's move forward with that.

mat8913:
It turns out this doesn't actually fix the Flatpak icon problem. I just thought that it did because I had the Debian package installed during testing which does display the icon correctly.

Does it happen to work on Windows or Mac?

PrestonN:
I'll test later tonight when I have access to a Windows device.  I'll update here when I've been able to test it.

diogogomes:
@mat8913,

I removed all configs, icons, menu entries, etc related with freetube and installed again and it **works**.

**FreeTube 0.6.0 Beta (flatpak)**
**Xubuntu 16.04**

PrestonN:
The change works on Windows.  Unfortunately I do not own a Mac to test on that end, so hopefully this fix will work over there as well.

PrestonN:
I will go ahead and merge this this everything seems to be well.  If I see any issues between now and the release that I'm working on then I will make adjustments as needed.

Thanks for the help!

PR #310, opened 2019-07-26
preserve window "maximized" state, closes #298

### LINKED ISSUE #298: Restore last window state at program start
Your program already saves the last program window size and restores it at the next program start.

Please also integrate saving and restoring of the last window state (normal or maximized).

vladimiry:

### DIFF
--- a/src/js/init.js
+++ b/src/js/init.js
@@ -80,18 +80,20 @@ let init = function () {
     settingsDb.findOne({
         _id: 'bounds'
     }, function (err, doc) {
-        if (doc !== null) {
-            if (doc.value !== false) {
-                let monitorList = screen.getAllDisplays();
-                let width = 0;
-                for (let i = 0; i < monitorList.length - 1; i++) {
-                    width = width + monitorList[i].size.width;
-                }
+        if (typeof doc !== "object" || typeof doc.value !== "object") {
+            return;
+        }
 
-                if (width <= doc.value.x) {
-                    win.setBounds(doc.value);
-                }
-            }
+        const {maximized, ...bounds} = doc.value;
+        const allDisplaysSummaryWidth = screen
+          .getAllDisplays()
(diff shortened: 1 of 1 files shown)


vladimiry:
PR got rebased so conflicts are gone now.

PrestonN:
Hey there.  This wasn't working for me at first and it took me a little bit to figure out why.

TL;DR, change `allDisplaysSummaryWidth <= bounds.x` to `allDisplaysSummaryWidth >= bounds.x`

The purpose of this logic is that a user was using FreeTube with multiple devices with shared settings.  One device had multiple monitors while the other had one.  If FreeTube was closed on that second monitor and opened on the device with only one monitor, FreeTube opened up in a position that was inaccessible to the user.

Your logic is slightly different in that you're calculating the max width of all monitors, while the previous logic grabs the width of all monitors minus the last one.

Besides this change, everything else looks good to me.  Thank you for your contribution!

vladimiry:
Dropped the last display from the array and turned the comparison as indicated.

vladimiry:
By the way, I noticed that quite often the bounds saving doesn't happen. I guess because in those cases the app manages to exit before the bounds saving call triggered in `window.onbeforeunload` gets finished in the main process. But that would be a different issue.

PrestonN:
Throughout my personal use I haven't experienced any issues with bounds not saving, but if you have an idea on how to fix it then I'm open to suggestions.

As for this PR, the required change needed to be either changing the less than logic to greater than, or to remove the last monitor within the calculation for width.  It looks like both was done which prevents the app from setting the bounds to the last position from being opened.  I apologize if I wasn't clear on this.

Please make the needed change and we should be good to merge.

vladimiry:
Amend has been made.

PrestonN:
Perfect!  Thank you for your help.  It's much appreciated.

Issue #332, opened 2019-08-14
Add "Copy video URL at the current time" in the share button


morfikov:
Currently there's no way to share a video link with "at the current time" position. You have to open the URL in a web browser and then right click on a YT player to get the "time link". It would be nice to have this kind of feature in FreeTube.

PR #468, opened 2020-04-24
Added 'current time' option to share menu as a checkbox.

kylejwatson:

Adds checkbox for #332

Wasn't too sure if there was a better styling option for this, the material 'switch' checkbox wouldn't fit in the share dropdown. 

Also the time parameter doesn't seem to work for youtube-nocookie so it was left out.


PrestonN:
Looks good to me.  Thank you for the contribution!

PR #565, opened 2020-10-03
Updated the Freetube screenshots, to the 0.8.0 rewrite UI 

PoorPocketsMcNewHold:
---
Updated the Freetube screenshot, to the 0.8.0 rewrite UI 
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Description**
As the previous one were directly hosted on the Freetube website, they may be re-uploaded later on.
I’ve also edited the previous width size of those, as they were taken in a different resolution, allowing for a wider showcase resolution.

**Additional context**
I’ll apologize, if the 0.8.0 redesign, was still supposed to be a work-in-progress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 FreeTube is an open source desktop YouTube player built with privacy in mind.
 Use YouTube without advertisements and prevent Google from tracking you with their cookies and JavaScript.
-Available for Windows, Mac & Linux thanks to Electron.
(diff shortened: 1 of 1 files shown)


PrestonN:
I believe GitHub caches these images which is why they haven't updated to new pictures on the site.  Switching to these pictures are fine though I may change them back at some point.

PR #571, opened 2020-10-03
Creating ft-intersection-observer component

VioletRose:
---
Creating ft-intersection-observer component
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [X] Feature Implementation

**Related issue**
N/A

**Description**
I have created a standalone component to observe viewport intersection. It can observe either itself (most useful for detecting if the user has reached a certain part of the page), or its parent element, controlled by props. It also has props to control standard IntersectionObserver options, as well as an option (defaulting to false) to run its check during the first execution which happens automatically when IntersectionObserver.observe() is called. When an intersection is detected, it emits an 'intersected' event, and emits an 'unintersected' event when intersection ceases.

Please add before and after screenshots if there is a visible change.

I temporarily added console.log statements to component methods, then temporarily added the component to the Settings view, observing the console outputs as I scrolled.

 - OS: Linux Xubuntu
 - OS Version: 20.04
 - FreeTube version: 0.8

**Additional context**
N/A

### DIFF
--- a/src/renderer/components/ft-intersection-observer/ft-intersection-observer.js
+++ b/src/renderer/components/ft-intersection-observer/ft-intersection-observer.js
@@ -40,11 +40,15 @@ export default Vue.extend({
   },
   methods: {
     handleIntersection(entries) {
-      if (this.runOnce || this.checkOnMount) {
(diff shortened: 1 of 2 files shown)


PrestonN:
This is awesome!  Thank you for the contribution!

Issue #554, opened 2020-10-02
Channel subscriber count is 0 in the search results


MPeti1:
**Behavior of the program**
The channel subscriber count is 0 in the search results.

**Expected behavior**
The channel subscriber count in the search results should be the same as on the channel main page, below the channel name

**To Reproduce**
Please add all steps to reproduce the behavior:
1. Go to the main menu
2. Type a channel's name in the search bar at the top
3. Click the arrow button to submit
4. See the channels with 0 subscribers

**Screenshots**
If applicable, please add screenshots to help explain your problem. Especially console logs from the development tools are very helpful.
Search results with the bug:
![image](https://user-images.githubusercontent.com/9215491/94974685-948f4e00-050f-11eb-943f-0c82066a137f.png)

Freetube can actually get the number of subscribers correctly:
![image](https://user-images.githubusercontent.com/9215491/94974728-ba1c5780-050f-11eb-84ef-b1916bc0c81c.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - OS Version: 1909
 - FreeTube version: 0.8.0

**Additional context**
Been trying out Freetube for a few minutes, and wow.. I like how much the UI is like the YT UI, that there's the thing when you hover the mouse over the video timeline, and that it's loading the small thumbnail VERY quickly even when just holding the mouse button down there and seeking and.. there's a lot of details, and it's fast, and it's actually hard to believe that it's been put together only in a year. I think I won't need to open YT in the browser anymore
Keep up the good work!

MPeti1:
It seems to be similar for viewer counts of live streams. It's probably only a problem with the search results menu
![image](https://user-images.githubusercontent.com/9215491/94975384-ca353680-0511-11eb-985f-c540f458b6ae.png)

syncng:
These two issues are related to the ytsr (Youtube Search) module dependency of the project.

The bad news is that the channel subscription count is unfortunately not possible to fix, since ytsr uses an endpoint that gives out information about the number of videos a channel has, but gives out no valid information on the subscriber count (even though there is a "followers" property obtained, it always returns 0).
However, it's completely possible to simply remove the subscribers indicator for channels from the app itself though, and maybe there's some other module out there that could be used in the future to replace this one that **does** contain that information, who knows, someone would have to do some digging on that.

The good news is that the live views count is actually provided, it's just a bug with the module itself and I've already opened a PR with a possible solution for this problem.

GilgusMaximus:
[...]

PR #602, opened 2020-10-06
Live video count fix

syncng:
---
Live video count fix
---

**Pull Request Type**
- [X] Bugfix
- [ ] Feature Implementation

**Related issue**
#554 partially

**Description**
Title and commit names says it all

**Screenshots**
Before:
Screenshot_4

After:
Screenshot_5

**Desktop**
 - Windows 10 1909
 - FreeTube version: latest commit

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -262,7 +262,7 @@ export default Vue.extend({
           publishText: this.data.publishedText,
           templateString: this.$t('Video.Publicationtemplate'),
           timeStrings: this.$t('Video.Published'),
-          liveStreamString: this.$t('Video.Watching').toLowerCase(),
+          liveStreamString: this.$t('Video.Watching'),
           upcomingString: this.$t('Video.Published.Upcoming'),
           isLive: this.isLive,
           isUpcoming: this.data.isUpcoming,
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -262,7 +262,7 @@ export default Vue.extend({
           publishText: this.data.publishedText,
           templateString: this.$t('Video.Publicationtemplate'),
           timeStrings: this.$t('Video.Published'),
-          liveStreamString: this.$t('Video.Watching'),
+          liveStreamString: this.$t('Video.Watching').toLowerCase(),
           upcomingString: this.$t('Video.Published.Upcoming'),
           isLive: this.isLive,
           isUpcoming: this.data.isUpcoming,
--- a/src/renderer/components/ft-list-video/ft-list-video.vue
+++ b/src/renderer/components/ft-list-video/ft-list-video.vue
@@ -78,12 +78,11 @@
         >
           <span>{{ channelName }}</span>
         </router-link>
-        <span
-          v-if="!isLive && !hideViews"
-          class="viewCount"
-        >• {{ parsedViewCount }}</span>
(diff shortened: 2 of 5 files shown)


src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -262,7 +262,7 @@ export default Vue.extend({
           publishText: this.data.publishedText,
           templateString: this.$t('Video.Publicationtemplate'),
           timeStrings: this.$t('Video.Published'),
-          liveStreamString: this.$t('Video.Watching'),
+          liveStreamString: this.$t('Video.Watching').toLowerCase(),
```
GilgusMaximus, line 265: The lower case has to be removed here. This is yet to be worked on.

src/renderer/store/modules/utils.js
```diff
@@ -292,51 +292,51 @@ const actions = {
     switch (strings[1].substring(0, 2)) {
       case 'se':
         if (singular) {
-          publicationString = publicationString.replace('%', payload.timeStrings.Second)
```
GilgusMaximus, line 295: The lower case has to be removed in the whole file, because of the same reason as before

GilgusMaximus (changes requested):

syncng:
I reverted that commit, but does that apply to the toLowerCase() in the first commit as well? Because those were mostly already there.

GilgusMaximus:
Sorry I might have missed that they were there already. It could have been that the lines were displayed because of the line shift. 
My bad. If it was there, then it was probably fine.

GilgusMaximus (approved):

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -78,12 +78,11 @@
         >
           <span>{{ channelName }}</span>
         </router-link>
-        <span
-          v-if="!isLive && !hideViews"
-          class="viewCount"
-        >• {{ parsedViewCount }}</span>
-        <span v-if="viewCount === 1">{{ $t("Video.View") }}</span>
-        <span v-else-if="parsedViewCount !== ''">{{ $t("Video.Views").toLowerCase() }}</span>
+        <template v-if="!isLive && !hideViews">
```
PrestonN, line 81: Not sure why you're using a template here.  Please switch this to something else like a `div` or another `span` unless I'm missing something here.

PrestonN (changes requested):
Just one small thing I'm curious about.  Once this is resolved then I'm good for merging.

src/renderer/components/ft-list-video/ft-list-video.vue
syncng: I'm only doing so the final result looks like this:
```html
<span class="viewCount">• (parsedViewCount here)</span>
<span>(View / Views string here)</span>
```
instead of this (example with div):
```html
<div>
  <span class="viewCount">• (parsedViewCount here)</span>
  <span>(View / Views string here)</span>
<div>
```
since the template tag gets interpreted as an invisible wrapper here
https://vuejs.org/v2/guide/conditional.html#Conditional-Groups-with-v-if-on-lt-template-gt

I assume you already know this, but I'm not 100% sure, so let me know if you still want a div or span to wrap them instead @PrestonN

PrestonN: Ahh, that's fair.  It just wasn't something I used too often, so I'm slightly unfamiliar with that method.  Yeah this is perfectly fine, no worries.

PrestonN (approved):

PR #652, opened 2020-10-12
Reconfigured Builder For Arch Package Format

BL4CKH47H4CK3R:


BL4CKH47H4CK3R:
Hello @PrestonN, there is no build error. Now you can merge and prepare 0.9.0 official release for Arch Linux & for automated builds too !

PrestonN:
I'm getting an unknown target error if I try to run this on my machine, which makes sense because this isn't an available option given in the [electron builder config](https://www.electron.build/configuration/linux#LinuxConfiguration-target).  The pacman build is the one meant for Arch Linux as you've been able to build it yourself.

I'm not familiar with too many projects that do official Arch builds so I'm not sure if .pacman builds are acceptable.  If they are then I don't mind using those as official builds, but if users prefer the `.pkg.tar.zst` then I need to find a way to properly convert it before I can offer it.

Issue #632, opened 2020-10-10
Alignment and typography


milotype:
Please check PR #631

**Behavior of the program**

- The div for **Duration** (Trajanje) does not align correctly. (Screenshot Alignment)
- Percentage and multiplication characters are not preceded by a non-breaking space. (Screenshot Typography)
- Multiplication character is not correct – you are using the _Latin Small Letter X_  (Screenshot Typography)
- `Space-Hyphen-Space` is used as separator.  (Screenshot Typography)

**Expected behavior**

- The div **Duration** should align correctly with other divs.
- Percentage and multiplication characters should be separated from the values by a non-breaking space.
- Correct the multiplication character
- I'd suggest to simply use `Collon-Space`, instead of `Space-Hyphen-Space` (I couldn't find out where this is specified, so please correct it accordingly … The hyphen makes it look as if it where a negative value …)

**To Reproduce**
1. Go to Filter/Settings

**Screenshot Alignment**

![Alignement](https://user-images.githubusercontent.com/43657314/95657372-80051400-0b14-11eb-96e0-2393fb84ddf7.png)

**Screenshot Typography**
![HyphenNobreakSpace](https://user-images.githubusercontent.com/43657314/95657376-885d4f00-0b14-11eb-93ec-85beabc7be07.png)

PrestonN:
Should be fixed with #653

milotype:
@PrestonN Just to inform you, 1st, 3rd and 4th item of this issue look good in the Build #174 on my MacOS 10.11.6

PR #653, opened 2020-10-12
Alignment and typography corrections

milotype:
**Pull Request Type**
- [x] Bugfix

**Related issue**
#632 

**Description**
Alignment and typography corrections - please see my comments in the commits

Not tested

 - OS: [Mac OS]
 - OS Version: [10.11.6]
 - FreeTube version: [0.9]

### DIFF
--- a/src/renderer/components/ft-slider/ft-slider.js
+++ b/src/renderer/components/ft-slider/ft-slider.js
@@ -39,7 +39,7 @@ export default Vue.extend({
       if (this.valueExtension === null) {
         return this.currentValue
       } else {
-        return `${this.currentValue}${this.valueExtension}`
+        return `${this.currentValue} ${this.valueExtension}`
       }
     }
   },
--- a/src/renderer/components/ft-slider/ft-slider.js
+++ b/src/renderer/components/ft-slider/ft-slider.js
@@ -39,7 +39,7 @@ export default Vue.extend({
       if (this.valueExtension === null) {
         return this.currentValue
       } else {
-        return `${this.currentValue} ${this.valueExtension}`
+        return `${this.currentValue}${this.valueExtension}`
       }
     }
   },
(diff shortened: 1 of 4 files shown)


src/renderer/components/ft-slider/ft-slider.js
```diff
@@ -39,7 +39,7 @@ export default Vue.extend({
       if (this.valueExtension === null) {
         return this.currentValue
       } else {
-        return `${this.currentValue}${this.valueExtension}`
+        return `${this.currentValue} ${this.valueExtension}`
```
PrestonN, line 42: This is the one change that I'm not so sure with.  Judging from what's included in the language that you're using, this is the preferred formatted for your language, though for most other languages, the expected formatting is how it currently is.  Let's revert this back for now, and I will create a solution at a later point.  There are many languages I have to account for and I'll likely open a general issue for finding and fixing all of these so that they look correct in all languages.

PrestonN (changes requested):
Please see my comment and make the requested change.  Once changed I will go ahead and merge.

src/renderer/components/ft-slider/ft-slider.js
milotype: I'm sorry to say, but you are wrong. It actually is the standard way of writing. Have a look at https://en.wikipedia.org/wiki/ISO_31-0, in the section **Typographic conventions**, subsection **Expressions.**

Of course if you definitely do not want to use it, I will change it (after all it's your project, so obviously you can decide). Please reconsider and let me know your final decision. Thank you.

PrestonN: And I will leave you with [this reference](https://en.wikipedia.org/w/index.php?title=Percent_sign#Form_and_spacing) as I stand by my original statement.  The formatting is different for several different languages and needs to be accounted for on a case by case basis.  This is something that goes above just this situation and is likely going to need it's own development cycle to get everything.

For now, I'd like to get a thread going that compiles all of these different edge cases together so that when we get to it, it'll be much easier to get these all resolved in one go.  Let's keep it the way it is for the time being and we will take a closer look at it once we get a good list put together.

milotype:
OK. I've changed it back.
When you do take care of it, please remember to use the 'NO-BREAK SPACE' (U+00A0) character as separator, if possible.

PrestonN:
Yes.  Will do.  Thank you very much for your PR.

Issue #549, opened 2020-10-02
Loop video


GilgusMaximus:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Related to a missing feature from the old FreeTube.

**Describe the solution you'd like to see implemented**
Add a button to VideoJS which automatically replays the video. Needs to use this [flag](https://docs.videojs.com/docs/guides/options.html#loop) in VideoJS.

mateusKoppe:
Need any help? I think I can work on this =]

PrestonN:
Hey there!  Feel free to work on it.  We already have [this](https://github.com/phhu/videojs-abloop) module installed in the project so if you could please use this in your work.  I'll take a look at your PR once it's ready.  Thank you for the offer!

PrestonN:
This has been implemented and will be included in the next release.

leszekbar:
if it is implemented, when it will be released?

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @leszekbar, you can find this functionality within the Cogwheel menu

<img width="1147" height="53" alt="Image" src="https://github.com/user-attachments/assets/13be911f-ca39-4a35-9da9-0e6cb50c1c4b" />

The one that says HD

PR #656, opened 2020-10-13
Implement videojs-abloop in the video player

mateusKoppe:
---
Implements a loop button in the Player
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
#549

**Description**
It adds a loop button by installing the [videojs-abloop](https://github.com/phhu/videojs-abloop) plugin in the player.

**Screenshots**
image

Click in the numbers to configure the loop and click in loop on.

 - OS: Linux
 - OS Version: Manjaro Gnome 20
 - FreeTube version: 0.9

**Additional context**
I'm not really sure if the videojs-abloop is the best plugin for this, probably most part of the users just want a simple loop button and abloop instead adds two more button, one to set the start of the loop and other to set the end of that, I read the documentation and couldn't find anything to hide just those two button, I can take another read if you guys prefer, just wanted to make sure that this is or is not the way that this is supposed to work before do any more work.

I had to add this line to make the plugin work
```js
abLoopPlugin(window, videojs)
```
I saw that in the others plugins there is no declaration like this one, if there is a better way to do it just tell me =]

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "freetube",
-  "version": "0.8.0",
+  "version": "0.9.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


PrestonN:
Judging just by your screenshot and not looking at your code, I definitely agree with you.  While I appreciated the added functionality of being able to define how a video is looped, it'd probably be preferred if it was one loop button that toggled it on and off like you mentioned.  If there's another plugin out there that accomplishes this, or if videojs-abloop has a toggle for this, then that direction is probably preferred.

I don't have a solution right away, but I'll likely be looking for one in the next day or so.  If you end up looking ahead of me, then I'm open to suggestions.

GilgusMaximus:
There is also the possibility to write a mini plugin that enables this option in VideoJs: https://docs.videojs.com/docs/guides/options.html#loop

mateusKoppe:
>There is also the possibility to write a mini plugin that enables this option in VideoJs


Yeah, that was my first attempt, the loop parameter works totally fine, I was trying to just add a loop button in the control bar but I could'nt find a way to add and control that, I'll take a better look at that later, to see a way to implement with or without abloop, if you have any material/article/documentation that could be helpful you can send me, I would appreciate

mateusKoppe:
Just for you information:
I'm discussing about the possibility of using ab-loop with just the loop button.

You can follow the discussion here: https://github.com/phhu/videojs-abloop/issues/2
I hope to get an answer from the maintainers of the project.

PrestonN:
Apologies for not responding to this in so long.

Thanks to a PR for different functionality, we had a base for creating simple button add-ons within the code now. I was able to use that to create a new loop button that behaves the way we'd expect it to. You can look up the code for it in the latest commit.

I appreciate you taking the time to work on this and reaching out to the original maintainer. It appears we have a solution that works now so I'll go ahead and close this.

Issue #660, opened 2020-10-13
Live videos in the recommmended section show up as undefined


thansk:
Expected behaviour is to show up as "LIVE" as they do on the search page.
![image](https://user-images.githubusercontent.com/53181514/95893832-bf7f6a80-0d77-11eb-96b6-bc2d25fdd87f.png)

PrestonN:
This is an upstream issue, I'll create the issue to the appropriate repository and link back here.  When the fix has been made, I will update the module on my end and we should be good.

Synkky:
What do intend to suggest upstream? A boolean "isLive" field for each recommended video?

PrestonN:
That' pretty likely.  I imagine it might not be a priority to include though.

PR #665, opened 2020-10-14
Recommended live video shows live tag, instead of undefined

Synkky:
**Pull Request Type**
- Bugfix

**Related issue**
#660


PrestonN:
This isn't a very good long term fix, as I mentioned that this is an upstream issue.  I'll merge this for now, but I'd really like to see this properly fixed before the next release.

Synkky:
Remember to make a note somewhere to remember it's there when the fix arrives so you don't have to spend time looking.

PR #669, opened 2020-10-15
Fix freetube:// protocol handling on macOS.

### LINKED ISSUE #618: FreeTube Doesn't Register it's Protocol Correctly on Mac
Since I love **FreeTube** very much, I would like to use it to open automagically YT links in any kind of situation. I use Mac OS but I think can be applied to Linux and Windows as well.

**ANALYSIS**
I can identify 3 kinds of "open a link" situations:
1. **inside the browser in the address bar**
2. **inside the browser when I click on a link**
3. **outside the browser**

**CHALLENGES/PROBLEMS**
1. needs an extension to stop loading the video and ask me what I want to do or get the option to always convert the link to freetube:// (like https://github.com/SimonBrazell/privacy-redirect)
2. needs an extension like https://github.com/FreeTubeApp/freetube-redirect (currently I get the "_TypeError: Cannot read property 'simpleText' of undefined_" error)
3. needs an app like https://github.com/will-stone/browserosaurus or https://github.com/johnste/finicky to catch the interaction and manage it. Currently **FreeTube** doesn't seems to **manage the http protocol in the first place**, so I cannot see the App in browserosaurus menu.

image

Any solution/fix?

Thank you! 🙏

cmf:
1. Handle open-url events.
2. Make protocol handling more robust on Win/Linux based on SO post.
3. Change 'ping' message to more descriptive 'openUrl' message.
4. Remove freetube:// protocol in main, and unify URL handling logic.

---
Title
---

**Important note**

**Pull Request Type**
- [X] Bugfix
- [ ] Feature Implementation

**Related issue**
Fixes #618.

**Description**
This fixes the freetube:// links that the Firefox redirect plugin uses on macOS, both when the app is closed and when it is already open.

I've tested this in the above situations on the macOS version below. However the flow for the equivalent functionality in Windows/Linux has changed a bit, but I don't have a Windows/Linux machine to test that on. If someone could do a sanity check on that that would be great.

 - OS: macOS
 - OS Version: 10.15.7
 - FreeTube version: 0.9

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -25,13 +25,25 @@ const path = require('path')
 const isDev = process.env.NODE_ENV === 'development'
 const isDebug = process.argv.includes('--debug')
 let mainWindow
+let startupUrl
 
 // CORS somehow gets re-enabled in Electron v9.0.4
 // This line disables it.
 // This line can possible be removed if the issue is fixed upstream
 app.commandLine.appendSwitch('disable-features', 'OutOfBlinkCors')
 
-app.setAsDefaultProtocolClient('freetube')
+// See: https://stackoverflow.com/questions/45570589/electron-protocol-handler-not-working-on-windows
+// remove so we can register each time as we run the app.
+app.removeAsDefaultProtocolClient('freetube')
+
+// If we are running a non-packaged version of the app && on windows
+if (isDev && process.platform === 'win32') {
+  // Set the path of electron.exe and your app.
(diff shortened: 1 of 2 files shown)


timharek:
I just tested these changes myself, and it works like a charm! 😄

PrestonN:
This still works for me on Linux, so I'll take the test that @timharek did as a confirmation that everything is good.  Thank you very much for helping with this!

Issue #639, opened 2020-10-10
Use shorter youtu.be links instead of youtube.com for YouTube share links


dennis1248:
Currently when copying the link of a YouTube video you will be provided with the longer `https://youtube.com/watch?v=abcdefgh` URL which is less visually pleasing then the shorter `https://youtu.be/abcdefgh`. 

The reasons for me to suggest these changes are that the shorter URL is more visually pleasing as I already mentioned and it is to remain consistent with what YouTube does, when on YouTube you copy a video URL it will provide you with the shortened version. These shorter URLs are also easier to work with when copying them to for example a document or blog post.

fatpowaranga:
I think I'll try to work on this

GilgusMaximus:
> 
> 
> I think I'll try to work on this

Okay then I will assign you to this for the moment. If you have any questions, you can ask them here, or for a faster reply in our [Matrix](https://riot.im/app/#/room/#freetube:matrix.org) channel

PrestonN:
Closed with #671

PR #671, opened 2020-10-15
Use shorter youtu.be links instead of youtube.com for YouTube share links

fatpowaranga:
---
Changes the share YouTube link option to use short YouTu.be links instead
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
[#639 ](https://github.com/FreeTubeApp/FreeTube/issues/639)

**Description**
Very simply makes a change so that when a video URL is copied it uses the shorter YouTube link instead. Eg. old link: https://www.youtube.com/watch?v=imaginaryURL is now new link: https://youtu.be/imaginaryURL

No visible changes, URL is different now.

Fairly straightforward. 

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.9

**Additional context**
The only note I have on this change is that perhaps people don't always want the shorter URL, but I couldn't think or find any reasons why this might be the case, whether it be privacy-related or usability-related. Feel free to correct if I'm missing something.

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -93,6 +93,10 @@ export default Vue.extend({
       return `https://www.youtube.com/watch?v=${this.id}`
     },
 
+    youtubeShareUrl: function () {
+      return `https://youtu.be/${this.id}`
+    },
+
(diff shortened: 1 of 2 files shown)


PrestonN:
LGTM.  Thank you!

PR #674, opened 2020-10-15
Comment scraper child process [Help needed]

GilgusMaximus:
---
Comment scraper child process
---

**Pull Request Type**
- [x] Bugfix
- [x] Feature Implementation

**Description**
The usage of the yt-comment-scraper module is put into it's own handler and forked from the main process as a native NodeJs process. This stops the application from freezing when loading comments.

Code does not work yet. Problem with the require() in the comment manager

 - OS: [Win
 - OS Version: 10
 - FreeTube version: 0.9


Synkky:
*I'm a noob at this, so don't throw me to the wolves if I got anything wrong*

I tried to understand what you meant with ***problem with the require() in the comment manager***, but, as for I can tell, there's no problem with that whatsoever.
There's only a couple of things not making this work and both of them have to do with the options you're providing to the **fork()** function, namely:
- `cwd: path.join(__dirname, '../../')` is only messing up the process and making the process return 1 immediately.
I'm **assuming** you think the default working directory is either the directory of the file calling the fork (*watch-video-comments.js*, in which case, `../../` will only take you as far as `src/renderer`) or something else that I failed to perceive.
Either way, that's wrong, the `cwd` by default is the root of the project, which makes overriding `cwd` unnecessary.
- `stdio: 'pipe'` will not allow the processes to communicate between each other through IPC, it will only initiate the standard file descriptors for the process itself (*stdin, stdout, stderr*).
To allow IPC, `stdio` should be `['pipe', 'pipe', 'pipe', 'ipc']`

GilgusMaximus:
I implemented the changes you suggested, however to no avail.
The child process is still not working in a built version

Synkky:
Change this:
```javascript
node: {
  __dirname: isDevMode,
  __filename: isDevMode,
  fs: 'empty',
  net: 'empty',
  tls: 'empty',
}
```
to this:
```javascript
node: {
  __dirname: isDevMode,
  __filename: isDevMode,
  fs: 'empty',
  net: 'empty',
  tls: 'empty',
  child_process: 'empty'
}
```
in the `_scripts/webpack.web.config.js` file

Synkky:
bumping @GilgusMaximus

GilgusMaximus:
Yeah thanks for the code.
Unfortunately I now discovered that the whole child process is not working at the moment. Probably because of wrong paths in the built version.

I boiled its code down to 'Send a message to the main thread when you get a message' and it didn't even do that. That also explains why there is an error every time the beforeDestroy() method is executed.

So that needs to be fixed first. However I'm not really that sure how to do that

Synkky:
I tested the build yesterday and it seemed fine.
Let me check again.

Synkky:
Yeah, you're right, this alone isn't enough. This is gonna take more work than I thought...

Synkky:
I don't know what approach to take here at all, fucking hell.
I understand why the working directory is all messed up in builds, but at the same time, I have no idea how to fix it.
Should I fiddle with webpack and try to find a way for a loader to convert these paths properly?
Should I fiddle with the electron builder to see if I can unpack that one file and then use a relative path to get it working? And if so, **how**?
I don't know a damn thing about these two techs, we need a more knowledgeable individual on this...

Synkky:
Wait, isn't it possible to spawn a process just for the JSON parsing (which is obviously the most clock intensive task) in the module itself?

GilgusMaximus:
In the end, we'd likely have to get rid of the awaits in the comment scraper and make everything work with promises. Because these are the ones blocking the main thread from my understanding. However I am not too sure how to return the data from multiple promises to the main thread

Synkky:
I have a working solution. Is it possible for you to give me write permissions to your fork?

GilgusMaximus:
Merging because it is working rn

Issue #704, opened 2020-10-21
No watching history on live videos


Mech0z:
**Behavior of the program**
If I watch a video that was broadcasted as a live video, but is now finished https://www.youtube.com/watch?v=-3jzyOpb2F0 and watch 10 minutes, the next time I start the video, it will start at 0:00

**Expected behavior**
I would expect it to remember that I watched 10:00 already and start there

**To Reproduce**
Please add all steps to reproduce the behavior:
1. Go to https://www.youtube.com/watch?v=-3jzyOpb2F0
2. Jump 10:00 in
3. Close video
4. Watch video again
5. Video starts at 0:00

**Desktop (please complete the following information):**
 - OS: Win10
 - OS Version:1809
 - FreeTube version: v0.9.0Beta

DuaneJaspers:
I've played around with this,
currently live broadcasts are checked like this: 
`this.isLive = result.player_response.videoDetails.isLiveContent || result.player_response.videoDetails.isLive`

past live broadcasts are considered live by `result.player_response.videoDetails.isLiveContent`.

In my quick testing I used a different check
`this.isLive = player_response.microformat.playerMicroformatRenderer.liveBroadcastDetails.isLiveNow`

This resulted in past live broadcasts being considered as normal videos and current live broadcasts being considered live. 
This fixed the watch history as well. I'm not 100% this broke anything though, but it seemed fine.

DuaneJaspers:
I've looked at this.

Currently the way a live video is determined is by `this.isLive = result.player_response.videoDetails.isLiveContent || result.player_response.videoDetails.isLive` 

what happens for a past broadcast is that `result.player_response.videoDetails.isLiveContent` returns true so it's handled like a live video.
I've tried to use a different check: 
`this.isLive = result.videoDetails.liveBroadcastDetails ? result.videoDetails.liveBroadcastDetails.isLiveNow : false`
This check only returns true if the broadcast is currently live. 
I'm not 100% this wouldn't break anything but I've quickly tested it for current live broadcasts, past broadcasts, and normal uploads. I haven't encountered any issues.

GilgusMaximus:
Thanks for looking into it. If you create a PR with that change, more people can test it and it is easier to verify whether it works correctly 👍

PrestonN:
[...]

PR #728, opened 2020-10-26
Change to only consider currently live broadcasts as livestreams.

DuaneJaspers:
---
Change to only consider currently live broadcasts as livestreams.
---

**Important note**

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
#704 

**Description**
In the current situation, both live streams that are already finished and live streams that are still streaming are considered as live content. So for live streams that are finished things like watch history and comments aren't available, and the live chat is empty. I changed it so that past broadcasts are considered normal uploads so watch history, downloads and comments become available.

***before***
Before
***after***
After

I've tested this for normal videos, past broadcasts, and live broadcasts. This all seems to work as expected. 

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.9.1


GilgusMaximus:
Thanks for the contribution. So the code itself looks fine and works fine.

However with this change in place, it would be the perfect chance to add a string saying 'Streamed on the x. of x xxx'. Just to make it more consistent with YouTube. Something similar to what is in place for normal videos. 

If you want to do it, go ahead, otherwise I'll look into it.

DuaneJaspers:
That makes sense!
I'll try to add other variants like "Premiered", "Started streaming on", and "Streamed on" to be totally consistent with YouTube.

DuaneJaspers:
I've added the "Started streaming on" (for streams that are still live), and "Streamed on" (for streams that are finished). 
I could not find a way to determine whether a video was once a premiere to change the string to "Premiered Oct 19, 2020".

PrestonN (approved):
I wanted to do a bit of testing myself before I approved this, as I recalled a reason I did it this way.  Something about the formats of a recent stream not behaving properly.  Testing it out here though I can't seem to find an issue, so it looks like we're good.  Thank you very much!

PR #739, opened 2020-10-27
Changed title of Watch view to title of the video

### LINKED ISSUE #702: Video name instead of "watch" in the title bar
It may be a detail, but on the title bar, instead of the expected video name, is "watch".

The name of the actual video instead of "watch".

**Screenshots**
Screenshot_20201020_181402

TDDR:
---
Changed Title of Watch view 
---
**Pull Request Type**
- [ ] Bugfix
- [ ] Feature Implementation
- [x] Enhancement

**Related issue**
Fixes #702 

**Description**
I have changed the title of the Watch view from the generic "Watch - FreeTube" to now show the title of the video instead

pr

 - OS: Windows 10
 - OS Version: 1909
 - FreeTube version: 0.9

**Additional context**
I wasn't sure exactly where to call my updateTitle function. I decided to call it in addToHistory because 
that will get called once for every video. Let me know if you have a better spot for it or want me to change anything.

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -413,6 +413,7 @@ export default Vue.extend({
           }
 
           this.isLoading = false
+          this.updateTitle()
         })
         .catch(err => {
           const errorMessage = this.$t('Local API Error (Click to copy)')
@@ -578,6 +579,7 @@ export default Vue.extend({
           }
 
           this.isLoading = false
+          this.updateTitle()
         })
         .catch(err => {
           const errorMessage = this.$t('Invidious API Error (Click to copy)')
@@ -618,7 +620,6 @@ export default Vue.extend({
         type: 'video'
       }
 
-      this.updateTitle()
       this.updateHistory(videoData)
     },
 
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -992,14 +992,6 @@ export default Vue.extend({
 
     updateTitle: function () {
       document.title = `${this.videoTitle} - FreeTube`
-
-      // simulates a reload to update the title after fetching it
-      this.$router.go({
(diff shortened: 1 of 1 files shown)


GilgusMaximus:
Thanks for the contribution. Looks good to me, excepty as you mentioned, the place where you call the function. 
The problem here is that when you deactivate the video history in the settings, the title will not change, because the addToHistory function is not called.

So it would be better to call your function at the end of getVideoInformationInvidious() and getVideoInformationLocal(), which in both cases can be done after the line `this.isLoading = false`, because then the video title is available and can bet set correctly

TDDR:
I originally thought that's where it should go as well, but when I tried to place it there it didn't work. Turns out that was because I was putting it in the .catch() portion instead of .then()... I made the call in both places you suggested. Let me know if anything needs to be changed!

src/renderer/views/Watch/Watch.js
```diff
@@ -988,6 +990,18 @@ export default Vue.extend({
       return Math.floor(this.getWatchedProgress())
     },
 
+    updateTitle: function () {
+      document.title = `${this.videoTitle} - FreeTube`
+
+      // simulates a reload to update the title after fetching it
+      this.$router.go({
```
PrestonN, line 997: Is there a reason this is needed?  Just changing the title without this router chunk works fine for me.

PrestonN (changes requested):
Please respond to the comment I left in my review and then I can merge depending on the discussion.

src/renderer/views/Watch/Watch.js
TDDR: I added that when I first couldn't get it working. That was when I was calling updateTitle in the wrong location. You are correct, that portion is not needed. I have amended the code to get rid of it.

PrestonN (approved):
LGTM.  Thank you!

Issue #730, opened 2020-10-27
Frame by frame advance


squinonescolon:
Would like to have frame by frame advance like in the Youtube player.
In the Youtube player you can advance frame by frame by pressing the , and . keys. I'm an animator and I use that feature everyday when looking at reference.

GilgusMaximus:
There are already a lot of speeds available, so implementing a frame by frame speed should be relatively easy to do.
Adding speed would be enough for the moment, as we do not have any hotkeys available for it, and might need to revamp them a bit in order to make them fully customizable.

PrestonN:
This wouldn't really be a speed and more so be something that can be done while the video is paused.  I'm assuming video.js has a plugin to allow this and if that's the case then it should be a pretty simple feature to include.

PrestonN:
This has been implemented in the latest release, closing...

PR #758, opened 2020-11-01
Implement frame by frame advance using keyboard shortcuts

Hiers:
---
Implement frame by frame advance using keyboard shortcuts
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
#730 

**Description**
Enables advancing and returning to the next or previous frame by pressing Z or X, respectively.

Open a video and press Z or X to move ahead or behind one frame.
Tested on both local and Invidious APIs, resolutions equal and above 720p and those below, dash and legacy file formats, on videos without a 60fps option for higher resolutions and those with it.

**Desktop**
 - OS: Artix
 - OS Version: N/A
 - FreeTube version: v0.9.2

**Additional context**
I settled on Z and X because those were the free keys that felt most natural to me.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -671,12 +671,12 @@ export default Vue.extend({
             event.preventDefault()
             this.changeDurationByPercentage(0)
             break
-          case 90:
+          case 188:
             // Z Key
             // Return to previous frame
             this.framebyframe(-1)
             break
-          case 88:
+          case 190:
             // X Key
             // Advance to next frame
             this.framebyframe(1)
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -445,18 +445,13 @@ export default Vue.extend({
 
     framebyframe: function (step) {
       this.player.pause()
+      const qualityHeight = this.useDash ? this.player.qualityLevels()[this.player.qualityLevels().selectedIndex].height : 0
       let fps
       // Non-Dash formats are 30fps only
-      if (!this.useDash) {
-        fps = 30
-      } else if (this.player.qualityLevels()[this.player.qualityLevels().selectedIndex].height <= 480) {
-        fps = 30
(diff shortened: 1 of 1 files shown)


Hiers:
After re-reading the issue, I changed the bindings to , and . to keep them consistent with YouTube.

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -251,6 +253,19 @@ export default Vue.extend({
...
     },
 
+    determineMaxFramerate: function() {
+      if (this.dashSrc[0] !== undefined) {
+        fs.readFile(this.dashSrc[0].url, (err, data) => {
+          if (err) throw err
+          if (data.includes('frameRate="60"')) {
+            this.maxFramerate = 60
+          } else {
+            this.maxFramerate = 30
```
PrestonN, line 263: `maxFramerate` should be initialized within the `data` object

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -251,6 +253,19 @@ export default Vue.extend({
       }
     },
 
+    determineMaxFramerate: function() {
+      if (this.dashSrc[0] !== undefined) {
+        fs.readFile(this.dashSrc[0].url, (err, data) => {
+          if (err) throw err
```
PrestonN, line 259: Since this will always error when the Invidious API is selected, we probably just don't want to throw an error into the console.  There's nothing catching this on the other side as well, so throwing doesn't help us too much in this situation and can prevent code execution depending on where `determineMaxFramerate` is called in the future.

We'll want to change this to do a couple of things. First, we'll let it default the `maxFramerate` to 60 (as per our discussion). Next, we'll want to `return` so that the next statement block isn't executed, since at this point, `data` isn't defined.

We can still log this error if we want with `console.log()` or with `console.error()` if we feel like seeing this error needs to be a little more visible, though a `console.log` or simply not logging anything is fine, since this would error out on every video if Invidious is the default.

PrestonN (changes requested):
Everything seems to work fine, however there's a couple stylized changes I'd like to see first.  Please make these changes and we'll be good to merge.

PR #789, opened 2020-11-15
Closes: Title of vid in fullscreen

### LINKED ISSUE #694: Title of vid in fullscreen
In YT i noticed when in fullscreen and move ur mouse in any direction u will see the Title of the vid in the left top cornor. It bugs me it is not there. For example if i watch a music vid in fullscreen of an unknown artist and i want to see who it is or what the name of the song is then i have to leave fullscreen.

I would like this to be implemented as the same visual way like YT.

None

**Screenshots**
None

**Additional context**
None

TDDR:
---
Title of video in full screen
---

- [ ] Bugfix
- [x] Feature Implementation

Closes #694 

**Description**
This PR adds an overlay to the application that displays the title of the video. For it to work, the app must be in full screen mode and it will active when you move the mouse. The overlay will hide itself when the `userinactive` event happens from the controlBar.

Screenshot (23)

 - OS: Windows 10
 - OS Version: 1909
 - FreeTube version: 0.9

**Additional context**
I am stuck on the last step. I am unable to manipulate the css for the overlay. I want to make the font larger and for the container it's in to span the top of the screen instead of causing the title to wrap underneath with bigger titles. I have tried putting the css for it inline in the `.vue` file as well as the `content` part of the overlay where the h1 tags are. 

Screenshot (24)

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -5,6 +5,8 @@ import $ from 'jquery'
 import videojs from 'video.js'
 import qualitySelector from '@silvermine/videojs-quality-selector'
 import fs from 'fs'
+import 'videojs-overlay/dist/videojs-overlay'
+import 'videojs-overlay/dist/videojs-overlay.css'
 import 'videojs-vtt-thumbnails-freetube'
 import 'videojs-contrib-quality-levels'
 import 'videojs-http-source-selector'
@@ -205,6 +207,8 @@ export default Vue.extend({
 
         this.player.on('volumechange', this.updateVolume)
         this.player.controlBar.getChild('volumePanel').on('mousewheel', this.mouseScrollVolume)
+
+        this.player.on('fullscreenchange', this.fullscreenOverlay)
 
         const v = this
 
@@ -521,6 +525,35 @@ export default Vue.extend({
       }
     },
 
+    fullscreenOverlay: function () {
+      const v = this
+      const title = document.title.replace('- FreeTube', '')
+
+      if (this.player.isFullscreen()) {
+        this.player.ready(function () {
+          v.player.overlay({
+            overlays: [{
+              showBackground: false,
+              content: title,
+              start: 'mousemove',
+              end: 'userinactive'
+            }]
+          })
+        })
+      } else {
+        this.player.ready(function () {
+          v.player.overlay({
+            overlays: [{
+              showBackground: false,
+              content: ' ',
+              start: 'play',
+              end: 'loadstart'
+            }]
+          })
+        })
+      }
+    },
+
     keyboardShortcutHandler: function (event) {
       const activeInputs = $('.ft-input')
 
@@ -594,7 +627,7 @@ export default Vue.extend({
             break
           case 40:
             // Down Arrow Key
-            // Descrease Volume
+            // Decrease Volume
             event.preventDefault()
             this.changeVolume(-0.05)
             break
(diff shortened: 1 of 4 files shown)


PrestonN:
The one thing I notice without looking at the code yet, I'd like to see the title of the video show regardless if the video is full screen or not. This makes it more in line with how YouTube does it. I'm only mentioning this now on case you have time to make this change before I can take a proper look at everything. Like I said in the other issue, I'll take a better look at this tomorrow when I'm free and help out with the issues you're having.

TDDR:
Yea, I can for sure have it display when not in full screen, that was the default behavior. I only made it display in full screen because I was trying to mimic the original as best I could. The original does not always show the title in an overlay, it only does so when in full screen. 

Please confirm that you still want me to display the overlay at all times with a mousemove event (regardless of full screen state) and I can make those changes.

PrestonN:
Ahh okay.  That was my mistake.  I just checked on YouTube and you are correct.  The video title only shows up when fullscreen is enabled.  We'll go ahead and keep it the way it is now to stay in line with how YouTube does it.  Thanks for the correction!

src/renderer/components/ft-video-player/ft-video-player.css
```diff
@@ -5,3 +5,9 @@
 .ftVideoPlayer {
   max-height: 50vh;
 }
+
+.relative.vjs-overlay {
```
PrestonN, line 9: Styling for the video player itself is a little odd, in that the styling isn't done within this component.  The styling for the video player is within `videoJS.css` which is located in the root of the `renderer` folder.

In this file (preferably towards the bottom), we'll want to add the following block.

```css
.vjs-overlay {
  font-size: xx-large;
  max-width: 100% !important;
}
```

The `!important` tag has to be included because the `vjs-overlay` CSS is forcing the max-width to be 33%, so this makes sure that this is overridden.

Once this has been done, you can completely remove the `.relative.vjs-overlay` block within `ft-video-player.css`

src/renderer/components/ft-video-player/ft-video-player.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative">
+  <div class="relative vjs-overlay">
```
PrestonN, line 2: In my testing, this class declaration wasn't needed.  Please remove this.

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -521,6 +525,34 @@ export default Vue.extend({
...
 
+    fullscreenOverlay: function () {
+      const v = this
+
+      if (this.player.isFullscreen()) {
+        this.player.ready(function () {
+          v.player.overlay({
+            overlays: [{
+              showBackground: false,
+              content: `<h1>${document.title}</h1>`,
```
PrestonN, line 536: Using the document title is fine, however please remove the " - FreeTube" string that's at the end of the title so that only the title is shown in the overlay.

PrestonN (changes requested):
I've left some comments on how to fix the issue you've mentioned along with some small changes I'd like to see made.  Please make these changes and then we will be good to merge.

src/renderer/components/ft-video-player/ft-video-player.js
TDDR: Do you want me to get rid of the "- Freetube" postfix in the `document.title`? That would also affect the title of the application that [I added](https://github.com/FreeTubeApp/FreeTube/pull/739).

Or do you still want the postfix in the application title but not the overlay title?

PrestonN: Let's keep the postfix in the document title, but remove it for the overlay.

PrestonN (approved):
LGTM.  Thank you very much for the contribution!

TDDR:
Thanks for the assistance!

PR #819, opened 2020-11-30
Fixed comment loading issue upstream

GilgusMaximus:
---
Fixed comment loading issue upstream
---

**Important note**

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Description**
The comments module only returned empty objects due to a problem with encoding in the XSRF token received from the comments page.

Tested in the module itself and when running FreeTube in dev mode

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.9.2 nightly

**Additional context**
For anyone searching similar problems, here a short description to get it indexed by search engines: The HTML page of Youtube encoded the '=' sign in the XSRF token, which is required to access comments, in Unicode for some reason (so \u003d) which then is not converted back when sending the Ajax request for the comment data. This leads to a wrong token and therefore no data

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18429,9 +18429,9 @@
       }
     },
     "yt-comment-scraper": {
-      "version": "1.3.9",
-      "resolved": "https://registry.npmjs.org/yt-comment-scraper/-/yt-comment-scraper-1.3.9.tgz",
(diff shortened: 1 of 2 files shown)


GilgusMaximus:
I messed up so big in my branch I'll resubmit in a bit

Issue #637, opened 2020-10-10
Keybindings for the videoplayer are not successfully overriden


DGoedtkindt:
While browsing the source code I found some code which is supposed to (I believe) override the default keybindings of the video player. However my experience shows that they are not overridden.    
  
The code is located in file src/renderer/components/ft-video-player/ft-video-player.js, from line 464 (keyboardShortcutHandler()).
```
      if (this.player !== null) {
        switch (event.which) {
          case 32:
            // Space Bar
            // Toggle Play/Pause
            event.preventDefault()
            this.togglePlayPause()
            break
         [...]
```
If I understand the code correctly, this means that when a video is playing and the user presses the spacebar then the video should toggle between playing and paused. This is not the case:  
Steps to reproduce:  

1. open any video
2. put the video in full screen using the button in the lower right corner of the video
3. press spacebar
4. Notice that the video exits fullscreen

The code does not do what it was intended (I believe) to do and I would love the keybinding overriding to be correctly implemented.

v0.9.0 Beta on Manjaro Linux GNOME

PrestonN:
Will look more into this.  In the mean time, if you click on the video while in fullscreen, spacebar to play / pause will once again work.

I am able to reproduce this issue though so I'll look into a proper fix.

Synkky:
[Probably not going to be possible](https://github.com/videojs/video.js/issues/6884#issuecomment-706737247)

TDDR:
Could this [plugin](https://github.com/ctd1500/videojs-hotkeys) be a possible solution? They seem to have found a way around the auto focus. 

"
Note: clicking any of the control buttons such as Play/Pause, Fullscreen, or Mute, will remove focus on the player which appears to "break" the hotkeys. This is for accessibility reasons so that people who do not use or know about the hotkeys can still properly use the Tab key to highlight the control buttons and press space to toggle them.

To restore focus, just click on the video, or an empty part of the control bar at the bottom of the video player.

To override this behaviour, set the flag alwaysCaptureHotkeys to true. This will "fix" hotkeys. For accessibility, the Tab key may be used in combination with the Enter/Return key to navigate and activate control buttons.
"

TDDR:
@PrestonN or @GilgusMaximus, just wondering if you saw my above post?

GilgusMaximus:
Hi sorry. I missed that apparently.  Thanks for the reminder.
[...]

PR #828, opened 2020-12-03
Fixes: Keybindings for the videoplayer are not successfully overriden

TDDR:
---
Fixes:  Keybindings for the videoplayer are not successfully overriden
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
Fixes: #637 

**Description**
This PR is intended to fix the auto focus issue when changing to full screen so that hotkeys will still work without having to manually give focus back to the player. I have it mostly working but there are still a few bugs that I was hopping to get a hand with.

Issues -
- The `alwaysCaptureHotkeys` property kind of works, If you click on fullscreen, then let the `userInactive` event happen, focus will return to the player, allowing the use of hotkeys without having to manually click the player again. I know this is not the optimal behavior, but I'm not sure if the behavior I'm getting is because of the warning that came with using it, namely `Note: This feature may break accessibility, and cause unexpected behavior`. Or if the behavior is because I am not capturing the correct elements on [this line](https://github.com/FreeTubeApp/FreeTube/compare/development...TDDR:issue-637?expand=1#diff-02a41b6ac31886e222921f371aad34b0ff9cf7f41f72d095c9cc760b5e0faf99R575)
- Three different implementations and I cannot successfully override the 'm' key
- the 'c' hotkey does not work, but I could not toggle the captions with a click either so maybe a different issue?

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.9

**Additional context**
I have replicated most of the old hotkey functionality. The only difference, aside from the bugs, is that I bundled `l and right arrrow` and `j and left arrow`. the old functionality was that `j and l` would change the playtime by `10sec` and the arrows would only change it by `5sec`. They both change it by 5 seconds now but I can change it to the old way if requested.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -1,5 +1,4 @@
 import Vue from 'vue'
-import { mapActions } from 'vuex'
 import FtCard from '../ft-card/ft-card.vue'
 
 import $ from 'jquery'
@@ -11,6 +10,7 @@ import 'videojs-overlay/dist/videojs-overlay.css'
 import 'videojs-vtt-thumbnails-freetube'
 import 'videojs-contrib-quality-levels'
 import 'videojs-http-source-selector'
+import 'videojs-hotkeys'
 
 export default Vue.extend({
   name: 'FtVideoPlayer',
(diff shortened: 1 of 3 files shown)


PrestonN:
Hello! Apologies for taking so long to look at this. Could you please update this branch to reflect upstream? There's a lot of out od date packages now and it's preventing me from looking into this further.

Thank you!

TDDR:
Hello @PrestonN! No worries, it was the holiday season when I first did the PR.

I have rebased my branch and pushed the changes. The same issues still persist as detailed above. 

I did not implement the new keybindings that were added this past month ('D' ,'esc') or the `...mapActions` block of code. I can implement those if you think this plugin is worth pursuing.

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -740,194 +740,317 @@ export default Vue.extend({
...
+    //        // S Key
+    //        // Toggle Full Window Mode
+    //        this.toggleFullWindow()
+    //        break
+    //      }
+    //   }
+    //    ...mapActions([
+    //      'calculateColorLuminance'
+    //    ])
+    //  }
```
manuee, line 1054: I would prefer these lines removed if we don't need them, we can always revert using git if we need to restore them.

manuee (commented):
No comments on the proposed solution since I am not familiar with the codebase... Here are my 2cents for fixing this issue :)

Funnily enough, if you hit the f key the full-screen comes on and keyboard shortcuts work. Its only when you use your mouse to go full-screen that the shortcuts stop working. 

Its as if the player (or whatever has the shortcuts bind to it) does not contain the full screen button, and thus it goes out of focus. I wonder if bringing in hierarchically the full-screen button inside the player would fix the problem once and for all?

efb4f5ff-1298-471a-8973-3d47447115dc:
Please resolve conflicts and update us otherwise close this request.

PrestonN:
Hey there!

Thanks for understanding.

Issue #667, opened 2020-10-14
Windowed FullScreen


galfwender:
Please add a button similar to the Chrome extension "Windowed FullScreen" to fully maximize playback within the application's window.

https://chrome.google.com/webstore/detail/youtube-windowed-fullscre/gkkmiofalnjagdcjheckamobghglpdpm

CristianMartin:
Hi,
I have a little knowledge with videojs and I would like to contribute to the project by adding this functionality.
thanks.

PrestonN:
Hello!  Feel free to work on this if you'd like.  I will review your pull request once it is ready.  Let me know if you have any questions.

Thanks!

CristianMartin:
HI,
Thanks for letting me contribute, and uploaded the pull request. I hope you can see it soon.

Thanks!!

PrestonN:
Closing as this has been implemented

PR #829, opened 2020-12-03
Windowed FullScreen

CristianMartin:
---
Windowed FullScreen
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/667

**Description**
I bring full window functionality to the video player via a dedicated button (similar to full screen).

**Screenshots**

**No full window**
freetube-no-fullwindow

**With full window**
freetube-fullwindow

 - OS: Linux mint
 - OS Version: 20 “Ulyana”
 - FreeTube version: 0.9.2


src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -509,6 +502,61 @@ export default Vue.extend({
       }
     },
 
+    toggleFullWindowed: function() {
```
PrestonN, line 505: Let's avoid creating a function within an existing function.  This should be possible to branch out into a `clickFullWindow` function within the component.  Alternatively, the new function can be called `toggleFullWindow` because of my next comment.

This function seems like it isn't strictly handling the toggling of the new button and it's more so creating the button so that it can be rendered on to the video player.  This makes it confusing to see this function get called in the initializing function.  With the above change, this function should be renamed to something like `createFullWindowButton` so that it better describes what the function is doing.  With this renamed, it can reference the new `toggleFullWindow` function or whatever it gets renamed to.

PrestonN (changes requested):
Hello!  This PR seems to work well and I like it a lot!

Please see my comments however as there's some organizational changes I'd like to see made to the code itself.

I've also noticed that the icon within the video player is ever so slightly higher than the rest of the icons.  Please adjust this so that the new icon lines up with the rest of the icons.

Also if possible, could you please place the new button to the left of the full screen button instead of the right?  This makes it more in line with how YouTube structures their buttons.

Let me know if you have any questions or concerns.

josx:
Looks good, simple and usefull!

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -200,7 +200,7 @@ export default Vue.extend({
           }, 200)
         }
 
-        $(document).on('keydown', this.keyboardShortcutHandler)
+        $(document).on('keyup', this.keyboardShortcutHandler)
```
PrestonN, line 203: This can't be `keyup` as then the spacebar will scroll down the page before the video toggles between play / pause, so we need to keep this as `keydown`.

PrestonN:
I've left a comment on something just to make you aware of it, but besides that, everything else looks good to me.

I'm going to go ahead and merge and make the change myself, as I'll be making some changes to this that are of no fault of you or your PR. I was hoping that this button would go between the quality selector and the full screen toggle, however the quality selector forces itself to go next to the full screen toggle when initialized. I'll go ahead and work on making this change.

If you'd like something else to add into a new PR, I'd love to combine the Theatre Mode button into this functionality to have a list selector similar to the quality and video speed that allows you to choose between Theatre Mode and the Full Screen Window. This would remove the existing Theatre Mode button in favor of this new functionality. I don't expect you to work on this but it's an idea if you'd like to make another PR.

Thank you very much for this functionality, it's much appreciated!

PrestonN (approved):

PR #851, opened 2020-12-09
Croatian translation of countries.json

milotype:


GilgusMaximus:
Wow, thanks for the great translation work. We really appreciate it : )

PR #862, opened 2020-12-10
Czech language is added, but not enabled. Enable Czech translation

alium:
Czech language is added, but not enabled. Enable Czech translation

---
Title
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [ X] Feature Implementation


GilgusMaximus:
Thanks for adding it. It probably wasn't added because it didn't have enough words translated to be usable. But thanks for checking on that

alium:
it's only 100%, more i can't  do :) thanks

PR #802, opened 2020-11-21
Adding playlist database logic

VioletRose:
---
Adding playlist database logic
---

**Important note**
Please note that only PrestoN is able to merge Pull Requests into master.

**Pull Request Type**
Please select what type of pull request this is:
- [ ] Bugfix
- [X] Feature Implementation

**Related issue**
N/A

**Description**
Updated playlist.js with state, getters, mutations, and actions

**Screenshots (if appropriate)**
N/A

**Testing (for code that is not small enough to be easily understandable)**
The updates have not been tested.
This file is not currently used.

**Desktop (please complete the following information):**
- OS: Linux Xubuntu
- OS Version: 20.04
- FreeTube version: 0.9.2

**Additional context**
The purpose of this pull request is to get feedback on the store/database business logic and align on expected functionality before proceeding further with playlists development.

PR #907, opened 2021-01-01
Saving favorite videos for later

GilgusMaximus:
---
Saving favorite videos for later
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Description**
At the moment no function to save favorite videos is available due to the absence of the major playlist feature. Because this feature will 100% come in the future, I added the possibility to use the favorite button already to save the videos in an intermediate JSON format, which then can be imported later when the playlists are available

Writing to the file was tested in dev and non-dev mode. 

 - OS: Windows
 - OS Version: 10
 - FreeTube version: v10.0.0

**Additional context**
File lies in the user data folder as favorites.json

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -170,8 +170,9 @@ export default Vue.extend({
   methods: {
     toggleSave: function () {
       console.log('TODO: ft-list-video method toggleSave')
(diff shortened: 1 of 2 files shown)


PrestonN:
This is something I've been planning on doing, so thanks for starting on this.

Let's make sure this logic works with #802 as that's what I will be using for the base for full playlist support.

My plan for the initial implementation is to store everything in a "Favorites" playlist and then instead of expanding on the playlist functionality, the playlists page will simply list out the videos in this playlist instead. This way once playlist functionality can be fully finished, all of the content is already where it needs to be and we won't have to refactor anything as we move forward.

I'll go ahead and merge #802 so that we can use it as a base.

PrestonN:
Going to close this since the basic functionality is ready now.

PR #916, opened 2021-01-09
Sidebar fixes

cloudrac3r:
See the commit messages for information. I tested this in my own way and I think it's good, but you might want to double check with the issue author that everything is solved.

Outstanding issues: What happens to the playlist sidebar? How can I detect if it's being displayed or not?


PrestonN (approved):
LGTM

Issue #908, opened 2021-01-01
User having the ability to modify the side bar (Subscriptions) and a new tab for the same


voiceofvraja:
**Describe the solution you'd like to see implemented**
Ability to remove the side bar of subscriptions channel icons. It would be good if the user can choose to see those or not because we already have the single area dedicated to subscriptions. So we can just add another tab just like NewPipe for instance
 which says "Whats New" and in there user can see what they can currently see in their subscriptions tab and in the place of subscriptions they can actually see all their subs ready to be managed lilke bulk unsubscribe.

![image](https://user-images.githubusercontent.com/50140643/103444067-29c6ed80-4c8b-11eb-8549-426cd37d4647.png)

PrestonN:
This would be fairly simple to add to the "Distraction Free Settings" of the app. I'll label this as a good first issue in case someone would like to make a PR for it.

I've also considered adding more links to the subscriptions manager page around the app as it seems that some users are having difficulty finding it.

voiceofvraja:
Wow, thanks a lot @PrestonN really!

PrestonN:
Added as of #918

PR #918, opened 2021-01-10
Hide active subscriptions

pablodekota:
---
Hide active subscriptions
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [X] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/908

**Description**
It allows users to hide their subscriptions, so the avatars dont show up, this setting is under the : Distraction Free Settings

image
image

Has this pull request been tested? Yes
I used the npm dev run to compile fresh code to spot any issues, currently all working well!

 - OS: Windows 10
 - OS Version: Latest
 - FreeTube version: v0.10.0 Beta

**Additional context**
simple feature and my first real pull request :)

### DIFF
--- a/src/renderer/components/side-nav/side-nav.vue
+++ b/src/renderer/components/side-nav/side-nav.vue
@@ -97,25 +97,23 @@
       </div>
       <hr>
       <div
-        v-for="(channel, index) in activeSubscriptions"
-        :key="index"
-        class="navChannel mobileHidden"
-        :title="channel.name"
-        @click="goToChannel(channel.id)"
+        v-if="!hideActiveSubscriptions"
       >
         <div
-          v-if="!hideActiveSubscriptions"
+          v-for="(channel, index) in activeSubscriptions"
+          :key="index"
           class="navChannel mobileHidden"
+          :title="channel.name"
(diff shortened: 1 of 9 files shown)


src/renderer/components/side-nav/side-nav.vue
```diff
@@ -104,11 +104,16 @@
         @click="goToChannel(channel.id)"
       >
         <div
-          class="thumbnailContainer"
+          v-if="!hideActiveSubscriptions"
```
PrestonN, line 107: This new `<div>` container here doesn't seem to be needed. With this current logic, Vue will still do a loop on all of the channels in the subscription list, but will display nothing. Instead, we should make it so that the loop doesn't run at all if the new setting is enabled.

Simply adding the `if` logic to the existing `<div>` will accomplish this. This is the only line that needs to be added to this file.

PrestonN (changes requested):
Everything looks good for the most part. Please make this one change and fix the linting issues and we'll be good to merge. :)

PrestonN:
LGTM. Thank you for the PR! It's much appreciated. :)

Issue #647, opened 2020-10-11
Add setting to always play videos in fullscreen


Brier1104:
A setting to always play videos in full screen mode would be great.

saideepesh000:
@PrestonN, Hi, I would want to work on this, just let me know, from where(file) can I start working on this, should we use [BrowserWindow](https://www.electronjs.org/docs/api/browser-window#new-browserwindowoptions) for this. Thanks

PrestonN:
Hi @saideepesh000 we would be more than happy to have a PR for this feature.  To help you get started, I'll let you know what needs to be done in order for us to accept a PR for this.

1. A new setting will need to be create in the player settings.  Follow the syntax of the other toggle switches to create this setting.  A suggested label should be "Full Screen Video on Start" for the toggle switch.  This string will need to be included in the English translation file within your PR.
2. All of the logic you'll need is already located in the [ft-video-player](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/ft-video-player/ft-video-player.js) component.  The component should check the setting within the `initializePlayer` function and full screen the player.  You can call `this.player.requestFullscreen()` to enable full screen.

Feel free to ask here or submit a draft PR if you have any questions with this.

saideepesh000:
@PrestonN, It's so nice of you to help others by explaining the details to get started, definitely I would want to solve this issue, but I need some time, thanks.

saideepesh000:
Hey @PrestonN, Should I add the translation of the string "Enable FullScreen Mode by Default" for every locale file.

PrestonN:
For adding new locale strings, you only have to add the new strings to the en_US file. Weblate will automatically add the string to the rest of the files for you.

saideepesh000:
> For adding new locale strings, you only have to add the new strings to the en_US file. Weblate will automatically add the string to the rest of the files for you.

Oh I didn't know that, so is that string (Enable FullScreen Mode by Default) OK for you.

PrestonN:
Yes, that's a good string for the new option. Apologies for not answering that as well.

saideepesh000:
Hey @PrestonN @GilgusMaximus , you could add this to in-progess column of 0.12.0 roadmap.

PrestonN:
Done!

jasonhenriquez:
Consider #1629 when implementing this to account for the case when this future setting is disabled.

Gevaarlijk:
[...]

PR #954, opened 2021-01-18
feat: added setting to always play videos in fullscreen

saideepesh000:
---
feat: added setting to always play videos in fullscreen
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
#647 

**Description**
Now users can set the default mode to watch videos in fullscreen mode. This setting will be available on Player Settings. 

fullscreen

 - OS: Windows
 - OS Version:10
 - FreeTube version: Latest

**Additional context**

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -141,6 +141,10 @@ export default Vue.extend({
       return this.$store.getters.getDefaultVideoFormat
     },
 
+    defaultFullScreenMode: function () {
+      return this.$store.getters.getDefaultFullScreenMode
+    },
+
     autoplayVideos: function () {
       return this.$store.getters.getAutoplayVideos
     }
@@ -203,6 +207,11 @@ export default Vue.extend({
             }
           }
         })
+
+        if(this.defaultFullScreenMode){
+          this.player.requestFullscreen();
+        }
+       
 
         this.player.volume(this.volume)
         this.player.playbackRate(this.defaultPlayback)
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -208,10 +208,6 @@ export default Vue.extend({
           }
         })
 
-        if (this.defaultFullScreenMode) {
-          this.player.requestFullscreen()
-        }
-
         this.player.volume(this.volume)
         this.player.playbackRate(this.defaultPlayback)
 
@@ -259,6 +255,7 @@ export default Vue.extend({
 
         this.player.on('ended', function () {
           v.$emit('ended')
+          v.player.exitFullscreen()
         })
 
         this.player.on('error', function (error, message) {
@@ -266,6 +263,9 @@ export default Vue.extend({
         })
 
         this.player.on('play', function () {
+          if (v.defaultFullScreenMode) {
+            v.player.requestFullscreen()
+          }
           if (this.usingElectron) {
             const { powerSaveBlocker } = require('electron')
 
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -208,10 +208,9 @@ export default Vue.extend({
           }
         })
 
-        if(this.defaultFullScreenMode){
-          this.player.requestFullscreen();
+        if (this.defaultFullScreenMode) {
+          this.player.requestFullscreen()
         }
-       
 
         this.player.volume(this.volume)
         this.player.playbackRate(this.defaultPlayback)
--- a/src/renderer/components/player-settings/player-settings.js
+++ b/src/renderer/components/player-settings/player-settings.js
@@ -82,6 +82,10 @@ export default Vue.extend({
       return this.$store.getters.getDefaultTheatreMode
     },
 
+    defaultFullScreenMode: function () {
+      return this.$store.getters.getDefaultFullScreenMode
+    },
+
     hideRecommendedVideos: function () {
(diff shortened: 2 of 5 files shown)


saideepesh000:
Damn U linting :(

PrestonN:
I likely won't be able to properly test this today as I'm trying to focus on a fix for some outstanding issues at the moment.

The one thing I do notice is please make sure this works when you navigate to a video from the video page. Navigation works slightly differently for pages that uses the same view so that will also need to be tested. If it already works with that then there's nothing to worry about.

saideepesh000:
@PrestonN, I tried every possible way of watching the video(Eg: through searching, through the trending page, through saved videos, through subscriptions, etc), it seems like working fine, do check from your side too.

saideepesh000:
Hey @PrestonN, sorry for rushing this a bit, I'm afraid that this may end up like [this](https://github.com/gkoberger/stacksort/pull/4#issuecomment-15081351), lol

PrestonN:
Hey there, apologies for taking so long to look into this. I was finally able to do some testing and here's what I found.

- Having FreeTube autoplay a video (either via "Play Next Video" / "Autoplay or by watching a playlist) doesn't set the video to full screen unless the user clicks on the window again while the next video is loading. Please take a look into this and if you have any troubles I'll see what sort of solution can be made.
- Switching between Legacy / Dash / Audio formats forces full screen when the toggle is enabled. Let's not have the video full screen in this case even when the setting is enabled.

This one is a nice to have but I won't require it for this PR.
- Let's close out of full screen when a video ends so that the user can see the notification that the next video is about to play. Doing this may even help with having the full screen come back on when the next video plays.

The code structure itself looks fine. Please fix the above problems and we'll be good to merge.

PrestonN (changes requested):
Hey there, apologies for taking so long to look into this. I was finally able to do some testing and here's what I found.

- Having FreeTube autoplay a video (either via "Play Next Video" / "Autoplay or by watching a playlist) doesn't set the video to full screen unless the user clicks on the window again while the next video is loading. Please take a look into this and if you have any troubles I'll see what sort of solution can be made.
- Switching between Legacy / Dash / Audio formats forces full screen when the toggle is enabled. Let's not have the video full screen in this case even when the setting is enabled.

This one is a nice to have but I won't require it for this PR.
- Let's close out of full screen when a video ends so that the user can see the notification that the next video is about to play. Doing this may even help with having the full screen come back on when the next video plays.

The code structure itself looks fine. Please fix the above problems and we'll be good to merge.

saideepesh000:
Okay, I'll update it soon

saideepesh000:
Hey @PrestonN, I could able to solve this(somewhat), now this `Enable fullscreen by default setting` only works when the video starts playing, so in order to understand this I added a `tooltip` for the setting, please make necessary changes to `tooltip` text to make it as intuitive as possible.

saideepesh000:
> 
>     * Switching between Legacy / Dash / Audio formats forces full screen when the toggle is enabled. Let's not have the video full screen in this case even when the setting is enabled.

I'm not sure I understood this.

PrestonN:
When you're watching a video, there is a button in between the download and share button that allows you to swap between the DASH / Legacy / Audio formats. When this button is pressed, it initializes the player again and will force the player to be full screen if the setting is enabled. I would like it so that the player does not enable full screen when these buttons are used even if the setting is enabled.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @saideepesh000 would it be possible for u to implement the same functionality for 'Fullwindow'? That would be really nice.

peepopoggers:
Please resolve conflicts and update us otherwise close this request. @saideepesh000

jasonhenriquez:
> Please resolve conflicts and update us otherwise close this request. @saideepesh000

I am bumping this comment. We understand that it is a busy time, so you may not have the time to work on this any further. However, given that this is an in-demand issue, please respond before the end of the week or we will have to re-assign the issue and close this PR.

ChunkyProgrammer:
I didn't actually test this code but from what I understand by looking at it, it looks like both defaultFullScreenMode and defaultTheatreMode can be enabled at the same time which I don't think should be the case. So another change might be needed to unset theatremode when full screen is made default and to unset fullscreen when theatre is made default.

jasonhenriquez:
> I didn't actually test this code but from what I understand by looking at it, it looks like both defaultFullScreenMode and defaultTheatreMode can be enabled at the same time which I don't think should be the case.

Why can't this be the case? I think you're wrong on this one.

ChunkyProgrammer:
> Why can't this be the case? I think you're wrong on this one.

Wouldn't one of the settings just end up overriding the other?  (Fullscreen over theatre mode)

efb4f5ff-1298-471a-8973-3d47447115dc:
Quick side note: I'm not sure if i interpreted the discussion correctly 

I agree with @ChunkyProgrammer i think its better  to do the following

If the toggle Enable theater mode by default is enabled

And u try to

Enable fullscreen by default by clicking on the toggle

The toggle of theater mode must turn off.

This must also happen when fullscreen by default is enabled and u want to turn on theater mode by default.

With this u wont get in any weird situation where we must make choices what default has the highest priority to override when multiple toggles are enabled.

jasonhenriquez:
> Wouldn't one of the settings just end up overriding the other? (Fullscreen over theatre mode)

Full screen would definitely be the dominant feature, but when full screen is exited, theater mode would be active by default. Similarly, since these settings are overridden by their last state when watching a set of videos, all videos switched into in the session after exiting full screen would automatically open in theater mode.

ChunkyProgrammer:
Good point! I guess this change would need to be kept in mind for default
fullwindow mode and default theatre mode

PR #1041, opened 2021-02-18
YT-Like show save-icon on hover on thumbnails

saideepesh000:
---
YT-Like show save icon on hover
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
None, this is just a proposal kind of a thing

**Description**
So on Youtube, we can only see the watch-later and add-to-queue icons only we hover on the thumbnail, this PR clones that type of behaviour, currently on FreeTube there's the save-video icon visible on every video thumbnail by default.

Before 

Screenshot (26)

After

savevideo

 - OS: Windows
 - OS Version:10
 - FreeTube version:0.11.3

Should I create an Issue for this?

### DIFF
--- a/src/renderer/sass-partials/_ft-list-item.sass
+++ b/src/renderer/sass-partials/_ft-list-item.sass
@@ -89,12 +89,16 @@ $thumbnail-overlay-opacity: 0.85
       @include is-watch-playlist-item
         font-size: 12px
 
-    .favoritesIcon
-      position: absolute
-      top: 3px
-      right: 3px
-      font-size: 17px
-      opacity: $thumbnail-overlay-opacity
+  .favoritesIcon
+    display: none
+
+  &:hover .favoritesIcon
+    display: inline-block
+    position: absolute
+    top: 3px
+    right: 3px
+    font-size: 17px
+    opacity: $thumbnail-overlay-opacity
 
     .watchedProgressBar
       height: 2px


PrestonN:
My only issue with doing this is making sure that the button is still accessible with a mobile device / interface. A touch screen wouldn't be able to use this button or it'd be rather frustrating attempting to do so. We'd have to add logic to add the button to the 3 dot menu when the mobile view is being used.

saideepesh000:
> 
> 
> My only issue with doing this is making sure that the button is still accessible with a mobile device / interface. A touch screen wouldn't be able to use this button or it'd be rather frustrating attempting to do so. We'd have to add logic to add the button to the 3 dot menu when the mobile view is being used.

@PrestonN, Oh, so to make it more simple we can add the save-video option to the three-dot menu by default, then this would be just an additional enhancement, what do you think. or otherwise, we'll go with your decision.

PrestonN:
I'm fine with that, as we'll likely need to do this anyways when the user playlists feature is finished. It may be best to break out all the copy / open functions to another section in an effort to de-clutter the 3 dot menu. Even simply making a breakout section that shows up on hover would probably be good enough. Again whatever solution we go with we should make sure it functions well for mobile as well.

efb4f5ff-1298-471a-8973-3d47447115dc:
Please resolve conflicts and update us otherwise close this request.

PrestonN:
Hey there!

Thanks for understanding.

Aspireve:
Hey @saideepesh000  could you give me a hint of how you did that in the second video, cause I am currently working on a similar issue (#1746) but cant find a way to do it
I am fairly new to this.
Really sorry for disturbing you but if you could give me a hint it would be a great help. 
Thanks!!

PR #1083, opened 2021-03-03
App string suggestions

### LINKED ISSUE #1212: Correction: Spelling error "Theatre" mode
If u hover above the button in the screenshot u'll see that the label says "Theatre".

"... theater is a movie theater while a theatre is a drama venue."

Theatre should be changed into "Theater".

**Screenshots**
Capture

comradekingu:
Not sure about all of this. Indentation seems intentional in some places, and produces doublespaces in strings other places. Not sure about the escaping either.

There are a lot of smaller issues than this for another day.

closes #1212


static/locales/en-US.yaml
```diff
@@ -187,24 +187,24 @@ Settings:
...
     Watch history has been cleared: Watch history has been cleared
     Remove All Subscriptions / Profiles: Remove All Subscriptions / Profiles
     Are you sure you want to remove all subscriptions and profiles?  This cannot be undone.: Are
-      you sure you want to remove all subscriptions and profiles?  This cannot be
-      undone.
+  you sure you want to remove all subscriptions and profiles? This cannot be
+  undone.
   Subscription Settings:
     Subscription Settings: Subscription Settings
     Hide Videos on Watch: Hide Videos on Watch
```
comradekingu, line 207:
```suggestion
    Hide Videos on Watch: Hide Watched Videos
```
?

GilgusMaximus: The thing here is that it only hides watched videos on the watch page. So that was probably the intention behind 'Watch' as in Watch page back when profiles were not back in again. But I would say that this now fits better.  Once could add a tooltip which shortly states 'Only hides watched videos on subscription page' in order to make it clear.

static/locales/en-US.yaml
```diff
@@ -187,35 +187,32 @@ Settings:
...
     Fetch Feeds from RSS: Fetch Feeds from RSS
     Manage Subscriptions: Manage Subscriptions
   Distraction Free Settings:
-    Distraction Free Settings: Distraction Free Settings
+    Distraction Free Settings: Distraction-Free Settings
     Hide Video Views: Hide Video Views
-    Hide Video Likes And Dislikes: Hide Video Likes And Dislikes
+    Hide Video Likes And Dislikes: Hide Video 👍 / 👎
```
GilgusMaximus, line 213: The intention is good, although the consequences are not that great. There are are people who might be color blind or have problems with emoticons, so simple text will have to stay here.

static/locales/en-US.yaml
```diff
@@ -187,35 +187,32 @@ Settings:
...
     Hide Channel Subscribers: Hide Channel Subscribers
-    Hide Comment Likes: Hide Comment Likes
+    Hide Comment Likes: Hide Comment 👍 / 👎
```
GilgusMaximus, line 215: Same as for Video Likes/Dislikes

static/locales/en-US.yaml
```diff
@@ -226,35 +223,32 @@ Settings:
...
+    Export NewPipe: NewPipe Export
     Import History: Import History
     Export History: Export History
     Profile object has insufficient data, skipping item: Profile object has insufficient
-      data, skipping item
-    All subscriptions and profiles have been successfully imported: All subscriptions
-      and profiles have been successfully imported
+  data, skipping item
+    All subscriptions and profiles have been successfully imported: Imported all
+  subscriptions and profiles
```
GilgusMaximus, line 239: I'd still add the 'successfully' at the end

static/locales/en-US.yaml
```diff
@@ -226,35 +223,32 @@ Settings:
...
     All subscriptions have been successfully imported: All subscriptions have been
-      successfully imported
+  imported
```
GilgusMaximus, line 241: The successfully can be moved behind the imported, but as before, I would leave it in. Imo, it just makes the text feel more positive from a users perspective

static/locales/en-US.yaml
```diff
@@ -226,35 +223,32 @@ Settings:
...
-      exported
+    Subscriptions have been successfully exported: Subscriptions exported
     History object has insufficient data, skipping item: History object has insufficient
-      data, skipping item
-    All watched history has been successfully imported: All watched history has been
-      successfully imported
-    All watched history has been successfully exported: All watched history has been
-      successfully exported
+  data, skipping item
+    All watched history has been successfully imported: Imported all watching history
```
GilgusMaximus, line 250: Same as before with successfully

static/locales/en-US.yaml
```diff
@@ -226,35 +223,32 @@ Settings:
...
+    All watched history has been successfully exported: Exported all watching history
```
GilgusMaximus, line 251: Same as before with successfully

static/locales/en-US.yaml
```diff
@@ -380,11 +373,11 @@ Channel:
 Video:
   Mark As Watched: Mark As Watched
   Remove From History: Remove From History
-  Video has been marked as watched: Video has been marked as watched
-  Video has been removed from your history: Video has been removed from your history
+  Video has been marked as watched: Video marked as watched
+  Video has been removed from your history: Videoremoved from your history
```
GilgusMaximus, line 377: Missed a space between Video and removed

static/locales/en-US.yaml
```diff
@@ -519,7 +512,7 @@ Comments:
   Comments: Comments
   Click to View Comments: Click to View Comments
   Getting comment replies, please wait: Getting comment replies, please wait
-  There are no more comments for this video: There are no more comments for this video
+  There are no more comments for this video: No comments for this video
```
GilgusMaximus, line 515: The 'more' should stay, as there is a difference between No comments available at all and you have reached the end of comments. So something like 'No more comments available for this video'

static/locales/en-US.yaml
```diff
@@ -530,65 +523,63 @@ Comments:
   Hide: Hide
   Replies: Replies
   Reply: Reply
-  There are no comments available for this video: There are no comments available
-    for this video
-  Load More Comments: Load More Comments
+  There are no comments available for this video: No comments for this video
```
GilgusMaximus, line 526: If we go with 'available' above, then it would improve consistency if we add it here as well

static/locales/en-US.yaml
```diff
@@ -530,65 +523,63 @@ Comments:
...
-      a frame of the video instead of the default thumbnail
-    Invidious Instance: The Invidious instance that FreeTube will connect to for API
-      calls. Clear the current instance to see a list of public instances to choose
-      from
-    Region for Trending: The region of trends allows you to pick which country's trending
-      videos you want to have displayed. Not all countries displayed are actually
-      supported by YouTube
+  a frame of the video instead of the default thumbnail.
+    Invidious Instance: Invidious instance FreeTube connects to for API
+  calls. Clear the current selection see a list of public instances to choose
```
GilgusMaximus, line 543: A 'to' is missing

GilgusMaximus:
Hi, 
so thanks for putting in the work of refining the text file. It hasn't been worked on for some time, as you probably saw.
I added some notes at certain parts, where I think it is necessary to changes strings or a space is missing, etc... 

Also it'll be necessary to fix the linter issues and conflicts before we can merge the file. If you need any help, please let us know :)

@PrestonN Should the Strings which are referenced in the code be adjusted to their new wording where they changed ?

static/locales/en-US.yaml
comradekingu: How does colour-blindness factor into seeing outlines of shapes? YouTube uses it.

static/locales/en-US.yaml
comradekingu: There is (presumably) no import that can be unsuccessful, and success is to be expected(?).

static/locales/en-US.yaml
comradekingu: It makes for more text to read.

static/locales/en-US.yaml
comradekingu: Not sure how it fails though.

static/locales/en-US.yaml
comradekingu: So, what should I opt to put in?

static/locales/en-US.yaml
```diff
@@ -534,64 +527,62 @@ Comments:
...
 Shuffle is now enabled: Shuffle is now enabled
-The playlist has been reversed: The playlist has been reversed
+The playlist has been reversed: The playlist was reversed
 Playing Next Video: Playing Next Video
 Playing Previous Video: Playing Previous Video
 Playing Next Video Interval: Playing next video in no time. Click to cancel. | Playing next video in {nextVideoInterval} second. Click to cancel. | Playing next video in {nextVideoInterval} seconds. Click to cancel.
 Canceled next video autoplay: Canceled next video autoplay
-'The playlist has ended.  Enable loop to continue playing': 'The playlist has ended.  Enable
-  loop to continue playing'
+'The playlist has ended.  Enable loop to continue playing': 'The playlist ended. Click
```
comradekingu, line 584: ' is only used to mark up some strings.…

static/locales/en-US.yaml
GilgusMaximus: Yes, you can leave it like it is. I'll add the tooltip afterwards :)

static/locales/en-US.yaml
GilgusMaximus: Yes, but it might be a problem if no alternative is added.
Here a few resources. Additionally we had a request from someone to make the app usable with screen readers. Unfortunately we were not able to implement that yet due to the many changes that happen all the time. But here are some resources to read on. The last one is behind a paywall bc it is a scientific paper, but the abstract is still available:

https://www.ashleysheridan.co.uk/blog/Emoji+and+Accessibility
https://uxdesign.cc/accessibility-matters-632fa2fb0672
https://dl.acm.org/doi/10.1145/3313831.3376267

static/locales/en-US.yaml
GilgusMaximus: I mean it can fail if something in the data is wrong.

static/locales/en-US.yaml
GilgusMaximus: It simply displays the text to my knowledge, if it does not error out with an error code

static/locales/en-US.yaml
GilgusMaximus: As I said it is debatable and my opinion. If some more people would like to add to that or discuss it and the majority says one way or the other I am open to leaving your change as it is. But these text changes are always dependent on what people want

static/locales/en-US.yaml
comradekingu: Yes, but in the event that it would, it would say so(?).
There is no reason to question the default, and that is how it is said.
"Successfully" isn't added unless there is reason to.
It is just not good, and worse yet when translated directly.

static/locales/en-US.yaml
comradekingu: It is never a problem to remove it, because it never does anything useful.

GilgusMaximus:
Yeah so, I would say we leave it like it is if there are no more fields you'd like to edit. We can discuss for weeks which either of us finds better, but in the end it really doesn't matter that much. 

But the final thing I request for merging is the change of the emoticons. As mentioned this is not very accessible, so after this has been done and having a quick discussion with PrestoN about the names of the fields for the actual code, I'll go ahead and merge it.

Again thanks a lot for the contribution,
the file really needed it and we are always happy to take contributions from the community :)

static/locales/en-US.yaml
```diff
@@ -189,35 +189,32 @@ Settings:
...
     Subscription Settings: Subscription Settings
     Hide Videos on Watch: Hide Videos on Watch
     Fetch Feeds from RSS: Fetch Feeds from RSS
     Manage Subscriptions: Manage Subscriptions
   Distraction Free Settings:
-    Distraction Free Settings: Distraction Free Settings
+    Distraction Free Settings: Distraction-Free Settings
     Hide Video Views: Hide Video Views
-    Hide Video Likes And Dislikes: Hide Video Likes And Dislikes
+    Hide Video Likes And Dislikes: No 👍 / 👎 for videos
```
comradekingu, line 215:
```suggestion
    Hide Video Likes And Dislikes: No video likes and dislikes
```

static/locales/en-US.yaml
```diff
@@ -189,35 +189,32 @@ Settings:
...
     Hide Channel Subscribers: Hide Channel Subscribers
-    Hide Comment Likes: Hide Comment Likes
+    Hide Comment Likes: No 👍 / 👎 for comments
```
comradekingu, line 217:
```suggestion
    Hide Comment Likes: Hide commend likes and dislikes
```

PrestonN:
Apologies for not looking too much into this. I still need to spend some time looking through the changes however my time has been limited recently.

The biggest problem I see with the changes is the fact that the field names would not match the string associated with it. This can cause some confusion in the future whenever we need to reference these strings within our code. If we're going to be revamping the wording on a lot of these strings, I'd expect to see the field names match as well.

This unfortunately means that there will be an annoying and tedious process of updating those strings within the code itself to match our locale file. I don't really have an easy or simple solution for doing this. There's basically two different directions we can go about doing this.

- We can update the field names in both the locale file and in the code when a change is made
- We update every field name to instead be a quick description of the text instead of a carbon copy of the string

With our current structure, the second option will take much longer to do, though it means that we wouldn't have this problem again in the future when changes are being made. The first option would be much quicker however has more short term work to do whenever changes are made.

I'm open to suggestions on what to do moving forward, but these are my thoughts. Apologies for not being more active in this discussion. I'll try to get some time in to properly review everything in the coming days. I would like to discuss the best method forward before merging everything however.

comradekingu:
@PrestonN Having keys that say "successful" is actually useful, since it helps translators with some context.
Pretty sure it will work mismatched, and I don't see a reason to change that. Could be done later if needed.
These are just suggestions. Use as many as you want, whenever you want :)

comradekingu:
@PrestonN @GilgusMaximus This passes again. Fixes the issues translators have, and run into.
Could give it another go once this passes.

static/locales/en-US.yaml
```diff
@@ -465,12 +458,12 @@ Video:
     Year: Year
     Years: Years
     Ago: Ago
-    Upcoming: Premieres on
-  Published on: Published on
-  Streamed on: Streamed on
-  Started streaming on: Started streaming on
+    Upcoming: Premieres on
```
comradekingu, line 461: All these are for https://hosted.weblate.org/translate/free-tube/translations/he/?checksum=dfc62afb07713e1f
Not very elegant, but alas.

peepopoggers:
@comradekingu @GilgusMaximus Is this request needed? Why has it been left? Thx.

comradekingu:
@peepopoggers I think so, on account of being more consistent and easier to navigate. Makes it a whole lot easier to translate FreeTube.

static/locales/en-US.yaml
```diff
@@ -18,69 +18,69 @@ Delete: Delete
...
 Open New Window: Open New Window
 
-Version $ is now available!  Click for more details: Version $ is now available!  Click
-  for more details
+Version $ is now available!  Click for more details: Version $ is now available! Click
+  for details.
 Download From Site: Download From Site
-A new blog is now available, $. Click to view more: A new blog is now available, $.
-  Click to view more
+A new blog is now available, $. Click to view more: New blogpost available, $.
```
Hiers, line 36: I looked up and didn't find example of blogpost as a word, should be "blog post".

static/locales/en-US.yaml
```diff
@@ -131,26 +130,26 @@ Settings:
     Current Invidious Instance: Current Invidious Instance
     # $ is replaced with the default Invidious instance
     The currently set default instance is $: The currently set default instance is $
-    No default instance has been set: No default instance has been set
-    Current instance will be randomized on startup: Current instance will be randomized on startup
+    No default instance has been set: No default instance  set
```
Hiers, line 133: This line has a double space.

Hiers (changes requested):
The file has a lot of lines which had their indentation removed which breaks the English localisation. The first example of this is in line 82.

static/locales/en-US.yaml
Hiers: Asides from these, I'm ready to approve these changes.

jasonhenriquez:
> The biggest problem I see with the changes is the fact that the field names would not match the string associated with it.

A tenable solution to this problem needs to be implemented before this PR can be merged.

peepopoggers:
@jasonhenriquez if that is the case, that stops scalability. We shouldn't use unnecessary variable names to match to the strings IMO.

static/locales/en-US.yaml
comradekingu: Good catch :)

static/locales/en-US.yaml
comradekingu: Changed it around

comradekingu:
@jasonhenriquez How is that not a tangential (and esoteric?) problem?
I would understand it if I changed it all for the proposed string content, (which I can do if needed.)

I like it better when the keys grant some context.

jasonhenriquez:
> @jasonhenriquez How is that not a tangential (and esoteric?) problem?
> I would understand it if I changed it all for the proposed string content, (which I can do if needed.)
> 
> I like it better when the keys grant some context.

Having only some of the labels be different, sometimes in minor ways, could be confusing and inconsistent. Any substantial change to the label system, such as presented in this PR, should be done clearly and consistently. It's as simple as that.

Hiers:
While I understand the concerns proposed by Jason and Preston, I still think that this isn't a blocking problem, nor is it something that justifies the effort that would be required to rename all the keys to something more generic in the locale files and then in the source code where they're used.

The en-US locale file as is already contains some minor discrepancies from the labels to the keys, and the confusion that could hypothetically arise from having more of these discrepancies would never become a blocking problem itself, or even substantially delay development.

I can't see a situation where writing `const message = this.$t('A new blog is now available, $. Click to view more')` and seeing "New blog post available, $. Click to view." in the application itself would be a problem worth fixing, especially when that is what already happens in all the other locale files by design.

Hiers (dismissed):
I just fixed the indentation problem so this doesn't break, aside from that this looks good to me.

peepopoggers (dismissed):
Looked through this, looks good and all issues look resolved.
I was wondering if 'successfully' should be added to some areas but it is not a big deal, just a minor inconvenience so this is approved :)

jasonhenriquez (changes requested):
I have no immediate feedback and have softened on the issue since, but I still would like this to be discussed more thoroughly with the team before using the contributor approval process.

comradekingu:
@jasonhenriquez That is the same request, plus more in the second. Did as much as possible without adding new strings.

jasonhenriquez:
> @jasonhenriquez That is the same request, plus more in the second. Did as much as possible without adding new strings.

The main request of #1212 is to change "theatre" to "theater" because that is the proper option for American English (en-us).

Hiers (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@comradekingu please resolve merge conflicts in order to get this mergerd.

efb4f5ff-1298-471a-8973-3d47447115dc:
This PR has been closed because there are unnecessary changes being made to the locales file.

comradekingu:
Like what?
There was a complaint over "successfully" removed where it is redundant.
Changes are not lost in translation since Weblate handles fuzzy strings directly.
Do you want me to comment on every change to explain what I think the benefit is?

PR #1095, opened 2021-03-08
Add channel verification indicator

### LINKED ISSUE #944: Verified mark(Right mark) beside channel name

Suppose if I want to check whether a YouTube channel is verified or not, there is no way to do it.

I think it would be nice if you add Verified mark(Right mark) beside channel name as youtube website does.

In YouTube website we can whether a channel has verified mark or not.

**Screenshots**

Screenshot from YouTube website

Verification mark

**Additional context**

Verified mark beside channel name helps in deciding whether channel is authentic or not.

lambdaclan:
---
Title
---

Display channel verification status in *some pages. 

**Pull Request Type**

Feature Implementation

**Related issue**

closes #944  

**Description**

The channel verification status can be shown in multiple locations but not all of them can be supported without adding extra requests (depending on the page being viewed) since the data available to that page might not include the verified status. This PR adds the verification mark in all places where it can be added without any additional requests (with the exception of the subscriptions page, see below for more info).

**Notes**
- channel verification mark has been added to following pages:
  - subscription list (additional request  -enabled for discussion purposes)
  - channel page
  - user playlist page including watched
  - history page
  - search page
    - no filter
    - channel filter
    - video filter
    - playlist filter 
  - video view including recommendations  
- channel verification is not shown on most popular, trending and playlist view
- channel verification mark does not work when using invidious backend: https://github.com/iv-org/invidious/issues/444
- channel verification mark does not work in the subscriptions page when using RSS to fetch feeds (data not included)
- channel verification mark uses the checkmark icon: https://fontawesome.com/icons?d=gallery&q=check-circle&m=free

**Screenshots - SHOWN**
verification_mark_subs
verification_mark_channel
verification_mark_urplaylists
verification_mark_history
verification_mark_search_all
verification_mark_search_channel
verification_mark_search_playlist
verification_mark_video_watch_off
verification_mark_video_watch_on

**Screenshots - NOT SHOWN**

verification_mark_trending_off
verification_mark_most_popular_off
verification_mark_playlist_off

**Testing**

The overall functionality has been tested:

- [x] ensure verification mark is displayed as expected in the various pages
- [x] ensure using invidious and RSS for feeds (where verification is not accessible) does not cause any issues
- [x] verify lint and prettier produce no warnings

**Desktop:**
 - OS: Arch Linux
 - FreeTube version: latest

**Additional context**

As mentioned above and based on previous conversations we had on Matrix, I am assuming that we want to avoid extra requests just to get the channel verification status. If that is the case do we:

- just display where is already obtainable?
- add it in all pages and allow the user to control its apperance via an option?
- some other way that requires further work such as updating ytch itself?

I am happy to discuss further and tackle any of the approaches mentioned above so let me know.

It will also be good to have a list of the places we actually want and can display the icon to verify that has been added.

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -120,10 +120,16 @@ export default Vue.extend({
         let textDecode = new TextDecoder('utf-8').decode(data)
         textDecode = textDecode.split('\n')
         textDecode.pop()
-        textDecode = textDecode.map(data => JSON.parse(data))
+        textDecode = textDecode.map((data) => JSON.parse(data))
 
         const firstEntry = textDecode[0]
-        if (firstEntry.channelId && firstEntry.channelName && firstEntry.channelThumbnail && firstEntry._id && firstEntry.profile) {
+        if (
+          firstEntry.channelId &&
+          firstEntry.channelName &&
+          firstEntry.channelThumbnail &&
+          firstEntry._id &&
+          firstEntry.profile
+        ) {
           // Old FreeTube subscriptions format detected, so convert it to the new one:
           textDecode = await this.convertOldFreeTubeFormatToNew(textDecode)
         }
@@ -156,56 +162,85 @@ export default Vue.extend({
           })
 
           if (Object.keys(profileObject).length < requiredKeys.length) {
-            const message = this.$t('Settings.Data Settings.Profile object has insufficient data, skipping item')
+            const message = this.$t(
+              'Settings.Data Settings.Profile object has insufficient data, skipping item'
+            )
             this.showToast({
               message: message
             })
           } else {
-            if (profileObject.name === 'All Channels' || profileObject._id === 'allChannels') {
-              primaryProfile.subscriptions = primaryProfile.subscriptions.concat(profileObject.subscriptions)
-              primaryProfile.subscriptions = primaryProfile.subscriptions.filter((sub, index) => {
-                const profileIndex = primaryProfile.subscriptions.findIndex((x) => {
-                  return x.name === sub.name
-                })
-
-                return profileIndex === index
-              })
-              this.updateProfile(primaryProfile)
-            } else {
-              const existingProfileIndex = this.profileList.findIndex((profile) => {
-                return profile.name.includes(profileObject.name)
-              })
-
-              if (existingProfileIndex !== -1) {
-                const existingProfile = JSON.parse(JSON.stringify(this.profileList[existingProfileIndex]))
-                existingProfile.subscriptions = existingProfile.subscriptions.concat(profileObject.subscriptions)
-                existingProfile.subscriptions = existingProfile.subscriptions.filter((sub, index) => {
-                  const profileIndex = existingProfile.subscriptions.findIndex((x) => {
-                    return x.name === sub.name
-                  })
+            if (
+              profileObject.name === 'All Channels' ||
+              profileObject._id === 'allChannels'
+            ) {
+              primaryProfile.subscriptions = primaryProfile.subscriptions.concat(
+                profileObject.subscriptions
+              )
+              primaryProfile.subscriptions = primaryProfile.subscriptions.filter(
+                (sub, index) => {
+                  const profileIndex = primaryProfile.subscriptions.findIndex(
+                    (x) => {
+                      return x.name === sub.name
+                    }
+                  )
 
                   return profileIndex === index
-                })
+                }
+              )
+              this.updateProfile(primaryProfile)
+            } else {
+              const existingProfileIndex = this.profileList.findIndex(
+                (profile) => {
+                  return profile.name.includes(profileObject.name)
+                }
+              )
+
+              if (existingProfileIndex !== -1) {
+                const existingProfile = JSON.parse(
+                  JSON.stringify(this.profileList[existingProfileIndex])
+                )
+                existingProfile.subscriptions = existingProfile.subscriptions.concat(
+                  profileObject.subscriptions
+                )
+                existingProfile.subscriptions = existingProfile.subscriptions.filter(
+                  (sub, index) => {
+                    const profileIndex = existingProfile.subscriptions.findIndex(
+                      (x) => {
+                        return x.name === sub.name
+                      }
+                    )
+
+                    return profileIndex === index
+                  }
+                )
                 this.updateProfile(existingProfile)
               } else {
                 this.updateProfile(profileObject)
               }
 
-              primaryProfile.subscriptions = primaryProfile.subscriptions.concat(profileObject.subscriptions)
-              primaryProfile.subscriptions = primaryProfile.subscriptions.filter((sub, index) => {
-                const profileIndex = primaryProfile.subscriptions.findIndex((x) => {
-                  return x.name === sub.name
-                })
+              primaryProfile.subscriptions = primaryProfile.subscriptions.concat(
+                profileObject.subscriptions
+              )
+              primaryProfile.subscriptions = primaryProfile.subscriptions.filter(
+                (sub, index) => {
+                  const profileIndex = primaryProfile.subscriptions.findIndex(
+                    (x) => {
+                      return x.name === sub.name
+                    }
+                  )
 
-                return profileIndex === index
-              })
+                  return profileIndex === index
+                }
+              )
               this.updateProfile(primaryProfile)
             }
           }
         })
 
         this.showToast({
-          message: this.$t('Settings.Data Settings.All subscriptions and profiles have been successfully imported')
+          message: this.$t(
+            'Settings.Data Settings.All subscriptions and profiles have been successfully imported'
+          )
         })
       })
     },
@@ -285,16 +320,22 @@ export default Vue.extend({
           this.setProgressBarPercentage(progressPercentage)
 
           if (count === textDecode.length) {
-            primaryProfile.subscriptions = primaryProfile.subscriptions.concat(subscriptions)
+            primaryProfile.subscriptions = primaryProfile.subscriptions.concat(
+              subscriptions
+            )
             this.updateProfile(primaryProfile)
 
             if (subscriptions.length < count) {
               this.showToast({
-                message: this.$t('Settings.Data Settings.One or more subscriptions were unable to be imported')
+                message: this.$t(
+                  'Settings.Data Settings.One or more subscriptions were unable to be imported'
+                )
               })
             } else {
               this.showToast({
-                message: this.$t('Settings.Data Settings.All subscriptions have been successfully imported')
+                message: this.$t(
+                  'Settings.Data Settings.All subscriptions have been successfully imported'
+                )
               })
             }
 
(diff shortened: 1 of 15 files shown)


Josegto1 (approved):

GilgusMaximus:
Hi, 
so great work you've been doing here. Also the level of detail of the pr is really great. 
I mean the discussion about the extra requests were already made as you said. I'm always a friend of leaving people the option to configure the app how they want it. But at the moment the cost of doing an extra request, compared to not doing it probably too high. A while back I had the idea of adding default setting profiles to the app, so users can choose between 'Full YouTube experience', 'Fastest loading times', 'No Rate Limiting' and so on. But that is currently on ice, when looking at the large number of things to work on. So the additional request would fit right into that at some point. 

But for the moment not doing the request is probably the better choice.
I'll take a look at the code and then come back and merge it if no other person has anything to comment

Best Regards,
Gilgus

lambdaclan:
Hello Luca thank you very much for your kind words.

>  I'm always a friend of leaving people the option to configure the app how they want it. But at the moment the cost of doing an extra request, compared to not doing it probably too high. A while back I had the idea of adding default setting profiles to the app, so users can choose between 'Full YouTube experience', 'Fastest loading times', 'No Rate Limiting' and so on.

Yes indeed, this is what I originally had in mind as well, happy to see that this could potentially be an option. The profiles you mention above in terms of naming make sense but I feel that the overall work will be much more than simply including a verified mark in the "full experience". I guess when the time comes and if we actually decide to implement this feature then we can start with adding the options and add/remove features as necessary over time.

>But for the moment not doing the request is probably the better choice.
I'll take a look at the code and then come back and merge it if no other person has anything to comment

No objections from my side, as mentioned above I did mostly to see how the teams think we should proceed which you kindly explained above.

Let me know if any further modifications are necessary and will be happy to assist.

> Only question I have is, what kind of license does the icon come with? A quick look onto their Q&A section mostly bombarded me with info about buying their pro license

Good question, I believe the icon is part of the [free font awesome pack](https://fontawesome.com/license/free) that we are already using in the app.

GilgusMaximus:
> Good question, I believe the icon is part of the free font awesome pack that we are already using in the app.

Then it'll be fine. I just always want to ask about that, not that we add something we cannot use easily.

Okay the code looks good so far. Gowever in the history page videos that were watched beforehand are not displayed with the mark. 

image


For obvious reasons that was not expected, but it will be a major inconsistency in the program if we add it like this. Because the verified attribute is added to the history database, it would be beneficial to go over all history videos once and add the verified logo. However that obviously is not as easy as going through all videos and requesting the verified status of that channel.

An idea would be that we keep track of the channel that owns newly watched videos. If the channel was not yet updated in the history, we go through all the videos in the history database and update the verified status for that channel. This way, all videos from channels that are watched again, are displayed as verified, For this we would probably need a new database file or add it to an existing database file. 


However a new database file plays right into an idea I had yesterday that would effectively eliminate the problem with the subscription page. So the current state is that we need a new request for the subscription page to check which channels are verified and which not. 

But, let's take a step back. How is a new channel added to the subscriptions? Well, this can happen through 3 ways:
1) The channel page
2) A video page
3) Importing subscriptions and channels that a user subscribed to before merging this code

So, 2 out of 3 paths are already covered by your implementation and do not require a new request. And the third path usually only occurs once. So for all new channels a users adds, we can eliminate the need for a new request, if we store whether a channel is verified or not in a database table. 

For the third path I see two choices, which are not mutually exclusive. 
1) Add them step by step like with the history videos above
2) Add a button in the settings which requests the verified mark for as many channels as possible and might rate limit the user for a day or two


So I believe that keeping track of the channels is probably the best bet to give users the best experience. If we combine the watch history step by step approach with the table for all subscriptions, then each user builds up a list of channels he/she watched and he/she subscribed to incrementally. And additionally we are consistent with previous subscribed and watched channels.

So this might need some refinement, and I'd also like to hear your opinion about that @lambdaclan 

Best Regards,
Gilgus

lambdaclan:
Hello @GilgusMaximus, apologies for the delayed reply but got swamped with work. I love technical discussions so thank you for asking my opinion on how to proceed even though I am a new contributor. I will be happy to comply.

> However in the history page, videos that were watched beforehand are not displayed with the mark.

Indeed, I did notice that and it is an interesting problem to solve.

> For obvious reasons that was not expected, but it will be a major inconsistency in the program if we add it like this. Because the verified attribute is added to the history database, it would be beneficial to go over all history videos once and add the verified logo.

Yeap, I agree with you. I also do not like the inconsistency where we show the verified mark in some pages but we don't in others. Ideally, I would prefer to display the mark *in all locations where it is expected to be visible* - more or less the list "not shown" above but that of course as you also mentioned will require additional requests which can be taxing:

> However that obviously is not as easy as going through all videos and requesting the verified status of that channel.

with the addition of other pages such as trending, most popular etc so the question is should it be available as an option?

> Add a button in the settings which requests the verified mark for as many channels as possible and might rate limit the user for a day or two

this might make sense but will add complexity since we will need to track the option and have a lot of conditionals for just displaying an icon. We can extend this to show everywhere (heavy - same as youtube) or show recommended (lighter - where we think it should be visible) if we need finer control but I am not sure how important it is to actually show the mark so it is subjective I guess.

> An idea would be that we keep track of the channel that owns newly watched videos. If the channel was not yet updated in the history, we go through all the videos in the history database and update the verified status for that channel. This way, all videos from channels that are watched again, are displayed as verified, For this we would probably need a new database file or add it to an existing database file.

> However a new database file plays right into an idea I had yesterday that would effectively eliminate the problem with the subscription page. 

I more or less had the same idea but after briefly discussing this matter with @PrestonN  over at Matrix:



Not sure if you mean a different kind of implementation by `database` instead of `cache` so please let me know and will be happy to discuss further.

Another way to approach it might be an option to update the history at launch once the feature gets added, at that point if the user agrees we through all watched videos and requesting the verified status of that channel (keeping track to avoid requesting verification for video from the same channel). This might be slow but it is a once and done operation since future videos added to history will include the verified information already.

> So for all new channels a users adds, we can eliminate the need for a new request, if we store whether a channel is verified or not in a database table.

I do like the idea of a database/cache of keeping track things to avoid unnecessary requests but beyond what Preston mentioned above we also need to keep in mind that we will need to update that once in a while to ensure channels appearing as verified still are (not sure how common it is for a channel to go from verified to unverified and vice versa). A database might make sense if we need to keep track of other things beyond the verification of a channel since adding a new database only for that purpose might be an overkill.

> For this we would probably need a new database file or add it to an existing database file.

I was not aware there are databases already, if we already have one then it might make more sense to add this info to an existing file rather than making a new one. Can you point me to an example?

> So I believe that keeping track of the channels is probably the best bet to give users the best experience. If we combine the watch history step by step approach with the table for all subscriptions, then each user builds up a list of channels he/she watched and he/she subscribed to incrementally. And additionally we are consistent with previous subscribed and watched channels.

I agree in terms of consistency this sound the logical thing to do but it should probably be optional as this will take over resources. if we decide to go the extra mile to keep track of channel verification we should also think how to approach other pages such as trending and popular which might complicate the process.

One last thing we should decide is whether or not this keeping track of channel verification and all the tasks it entails should be part of this PR of be separated into a different maybe even multiple tickets that we can keep track off. 

Happy to proceed either way to just let me know.

Best Regards,

λ

shivasagarrao:
Also consider implementing verified mark beside comments under comment section of a video. 

YouTube_Comments

GilgusMaximus:
> Also consider implementing verified mark beside comments under comment section of a video.

I'll check that over at the module. If the field is provided we can easily do that

GilgusMaximus:
> Hello @GilgusMaximus, apologies for the delayed reply but got swamped with work.

As I said, so problems. We are happy about everyone who contributes, and no one is required to do feature x in y period of time.


> with the addition of other pages such as trending, most popular etc so the question is should it be available as an option?

 I am not too sure whether you now agree with me that it is too taxing to request the verified marker for each channel in the history or not.

> this might make sense but will add complexity since we will need to track the option and have a lot of conditionals for just displaying an icon.

If else will only be required on the pages where the data is not provided anyways. But yeah, I do not think this options is so viable. I'd rather stick to the incremental build up.


> Not sure if you mean a different kind of implementation by database instead of cache so please let me know and will be happy to discuss further.

I meant that. But tbh I do not see that problem. People don't run around with 4KB of disk space. I mean we literally do not clear dash and storyboard files at the moment and no one is storming down the issue page that the app accumulates a few MB. In my opinion a good experience should be worth a few MB. 

I mean let's assume we have a maximum channel id of 30 characters. We have 30 bytes in the worst case + 1 byte as indicator whether it is verified or not. Now let's assume a very broad user catches around 10k different channels. Which is a lot. And with a lot I mean a lot. Then we need some overhead for the JSON like database file format, so I'll add a realistic amount of bytes.

10.000 * 31 Bytes + 10000B = 311 000 Bytes = 321 KB. 

This is not filling up any disk space. Otherwise you probably have problems you should first sort out on your device,

We can even make further improvements and only add channels that **are** verified to the database. I'll make a very generous assumption and say that 85% of the channels are verified. 

8500 * 30 Bytes + 10000 B = 266 KB

So this is probably defendable. However I'll talk with PrestoN about this to check what he thinks about this.

>  A database might make sense if we need to keep track of other things beyond the verification of a channel since adding a new database only for that purpose might be an overkill.

At the moment we have nothing specifically planned for another usecase. However there were requests for a block list for certain channels. So this might come in handy here later.

> I was not aware there are databases already, if we already have one then it might make more sense to add this info to an existing file rather than making a new one. Can you point me to an example?

Yes we have a database. E.g. the history.db file, which is fueled through [this](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/store/modules/history.js) code. We use nedb cause that is what PrestoN already started with. (Maybe, and I mean very big maybe, one day we might switch to something like sqlite or similar, as this db is always causing some problems the way we want to work with it. But the task is not important at all at the moment.)

lambdaclan:
Hello Luca

Thank you breaking down the scenario with an example. Indeed, when you look at the numbers having a file storing channel verification info it does not look bad at all which is why I was also a bit confused by Preston's statement. Maybe he was more worried about the extra requests rather than the file getting too large.

> I am not too sure whether you now agree with me that it is too taxing to request the verified marker for each channel in the history or not.

I do agree and your suggestion of doing it incrementally sounds good. If we had to deal only with history then we could have done it all at once (optionally) but since we also want to tackle other pages such as subscriptions then doing it incrementally for everything is a better option.

> So this is probably defendable. However I'll talk with PrestoN about this to check what he thinks about this. Just as you mentioned, a setting is always possible.

> I mean if we do this in multiple PR's I would probably leave out the history page for now. This makes it really confusing for users when it suddenly starts adding the channels in two different ways. Or at least the incremental updating of the history is added in this pr, so when a new video is added. Although I am not too sure how much performance that will cost users

Agreed, provided there are no objections we can proceed with the implementation.

> From verified to non verified is probably not happening. I mean why would it happen, except for special cases. But of course we can check whether the current channel we got information from is still verified or not. But I'd be very careful with this. Because if something changes in YouTube's layout and this information is not parsed anymore, it would knock out all channels. So I'd rather stay with the assumption that a channel that was once verified will stay verified for the moment. We'll likely hit 99.9999% of all channels with that assumption.

Understood, sounds reasonable.

OK, to make sure we are in line let me try to put things in order so that we know what is left to do.

- we go through all the videos in the history database and update the verified status for that channel once a new video of the same channel is watched - this means we cross check that the history db contains a video from the same channel 
- we combine the watch history step by step approach with the table for all subscriptions, then each user builds up a list of channels he/she watched and he/she subscribed to incrementally
-  we add the channel verification status to a database that we can use to avoid future requests

does this sound right?

I also want to clarify a few things:

1. To update the watch list in an incremental fashion we agreed to check the video being watched and if that channel is in history db but not in the verification status db we update all the videos of the channel that are in the history db. How do we do that with subscriptions? Do we just check if the video being played is part of a channel the user is subscribed to and follow the same pattern? If that is the case we sometimes might have to update both history and subs at the same time which is good because we can do that with a single request.
2. Should we use a different db file for channel verification and if yes where should it be stored. I guess it will follow a similar scheme to the history db you mentioned above.
3. Do we only store verified channels:
 > We can even make further improvements and only add channels that are verified to the database 

  or everything?
4. Do we make this optional or add it as a feature? (provided there are no performance issues it shouldn't be a problem)
5. Do we skip other pages (trending, popular, playlists etc) for now or? Do we even want to display it there?

More questions might pop up but that should be enough to get started.

Regarding the verification status for comments I am assuming you are talking about your yt-comment-scrapper?

GilgusMaximus:
Sounds good on paper.
However I am still in discussing this with PrestoN.

I made the suggestion to first check whether we can add the verified extraction to as many modules as possible. Due to time zones I did not get a response yet. But I went ahead and updates the trending module to support the verified check as well. 

There you'll simply get a isVerified field which is either true or false. But I'll bump the packages later, when I have an overview where it can be added as well.

So I'll get back to you as fast as I can

GilgusMaximus:
> channel verification is not shown on most popular, trending and playlist view
channel verification mark does not work when using invidious backend: iv-org/invidious#444
channel verification mark does not work in the subscriptions page when using RSS to fetch feeds (data not included)


Okay I went ahead and checked different things.

1) Most popular is not our fault, as this is Invidious.
2) Trending has been added, as said before.
3) Playlist view does not show it on YouTube either, so we are good here.
4) Invidious is Invidious.
5) RSS is known to have less data, so this is fine.


So in the end I'd say that the extra database is not needed because we are on par with YouTube now. 

So I'll bump the trending module and only the trending page needs an update before we can merge this.

peepopoggers:
Hi, what is the progress on this? Is this still an option or are there still problems with this request. Thanks.

PrestonN:
Hey there!

Thanks for understanding.

Issue #898, opened 2020-12-23
Default profile changes back to All profile after closing app


ghost:
What determines which profile is loaded on startup? I have a profile set as default, but despite this, the "All Channels" will sometimes load instead.

GilgusMaximus:
Your default profile usually. But just as you said it, it seems that this does not always stay between sessions and reverts to the All profile for some reason.

I'm gonna change this into a bug issue in order to fix this

saideepesh000:
@PrestonN , I think the [updateActiveProfile ](https://github.com/FreeTubeApp/FreeTube/blob/9e73c5be8e81e0c8c82a84481a39be091995ce61/src/renderer/store/modules/profile.js#L138)is not commiting to database, Am I right, can I try this

saideepesh000:
@PrestonN, And also we need to be able to differentiate the UI of the default Profile from the rest of the profiles.

saideepesh000:
@PrestonN, just realized we need to tweak the profiledb a little bit, currently, it's only storing the profile data, activeprofile data should be separately stored, so how should I approach it ?.

PrestonN:
Apologies for taking so long in this. The `updateActiveProfile` function isn't meant to save data permanently and it's only used for the active session. The part that saves the default profile is in [settings.js](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/store/modules/settings.js#L44).

I do agree that both the default profile as well as the "All Channels" profile needs to have some differentiation in the UI. That'd be a separate discussion from this issue however.

For fixing this, my only guess is that there's a runtime issue where the default profile is being set before FreeTube loads all profiles that have been created. We need to make sure that the profiles are loaded before it sets the default profile.

You're still welcome to tackle this if you'd like, feel free to contact me if you need any help.

PR #1114, opened 2021-03-14
Profile loading inconsistency

GilgusMaximus:
---
Profile loading inconsistency
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
#898 

**Description**
On start not always the default profile is loaded, but the allChannel profile might be loaded in some cases.

No

 - OS: Win
 - OS Version: 10
 - FreeTube version: 0.12.0

**Additional context**
As already discussed in the issue #898 there seems to be a problem with loading of files. My investigation so far:

The problem is that the settings file is not processed fast enough before the profile script is called in order to set the active profile as seen in the two images below.

Wrong execution:
wrong_exec

Right execution:
right

However I adjusted the code in the App.js file to the following, so that the other parts wait for the execution of the first dispatch to finish:
```
mounted: function () {
    this.$store.dispatch('grabUserSettings').then(() => {
      this.$store.dispatch('grabHistory')
      this.$store.dispatch('grabAllProfiles', this.$t('Profile.All Channels'))
      this.$store.dispatch('grabAllPlaylists')
      this.$store.commit('setUsingElectron', useElectron)
      this.checkThemeSettings()
      this.checkLocale()

      if (useElectron) {
        console.log('User is using Electron')
        this.activateKeyboardShortcuts()
        this.openAllLinksExternally()
        this.enableOpenUrl()
        this.setBoundsOnClose()
      }

      setTimeout(() => {
        this.checkForNewUpdates()
        this.checkForNewBlogPosts()
      }, 500)
    })
  },
```

But this does not work yet. Sometimes it still goes to the allChannel profile instead. So any input would be appreciated


GilgusMaximus:
Note: I edited some things in other files to get some output and a better understanding of which names are related to which outputs. Don't be confused by this

GilgusMaximus:
Thanks to the amazing @Svallinn for helping me out here. Especially finding the problems in my approaches.

However there seems to be a certain problem with Windows that simply installing the new version over the old version might not solve the problem every time. To solve this, the best way is to export all your freetube data, then uninstall Freetube, delete the FreeTube data folder in %appdata% and then reinstall and import all the data. Otherwise it might occur that the fix is not working all the time.

PR #1121, opened 2021-03-17
Overhaul of the captions subsystem

### LINKED ISSUE #763: Captions dont always work
**Behavior of the program**
Sometimes when playing a video with captions (not auto generated captions) the captions simply does not appear

**Expected behavior**
Captions must appear when watching a vid

**To Reproduce**
1. Go to any vid with captions
2. Turn on Captions in any language 
3. Captions dont appear

**Screenshots**
If applicable, please add screenshots to help explain your problem. Especially console logs from the development tools are very helpful.

 - OS: [Win 10 home]
 - OS Version: [1909]
 - FreeTube version: [0.9.2]
 - FreeTube build: [Release]

**Additional context**
none

Svallinn:
**Pull Request Type**
- [X] Bugfix
- [X] Feature Implementation

**Related issue**
Closes #763 (this was all I could find, let me know in the comments if any other issues exist)

**Description**
This PR marks an overall refactor of how `video.js` obtains and handles the caption data
- There should no longer be any inconsistencies with the process of providing the captions to the player
- The player is now provided a richer version of the caption data (e.g. font styles, font coloring, window positioning, etc.)

_TODO:_
- [x] ~Provide an auto-generated caption for the selected locale of the user to the player instead of providing English by default~
- [x] Provide a **translated** caption for the selected locale of the user to the player instead of providing English by default

**Details**
I initially planned to make use of the current way things are set up, provide direct YouTube links that provide WebVTT from the get-go to the `<track>` tags and get rid of the need to use `xml2vtt` parser and to create `data` URIs entirely

Unfortunately, the WebVTT provided by YouTube needs some work before it can be fed into `video.js`, so I was only able to get rid of `xml2vtt` parser, but this step is most likely still substantially faster than before

There was also a certain bottleneck related to the text tracks provided to the player in which, if there were a lot of text tracks provided, the player would stall because all tracks would have to be **fully parsed and preloaded** and, due to what I can only imagine to be limitations of the player, they can only be loaded sequentially

Therefore, the captions are now provided asynchronously to the player, so that the player doesn't have to wait for the captions' data to be processed for the video to start. Also, the `data` URIs are now only called when the corresponding caption is selected by the user

**Testing**
All implemented fixes and features were manually tested

**Desktop**
 - FreeTube version: development


Svallinn:
The force push modifies the previous commit in order to do the following:
- makes the app favor translating an existing manual English caption over an auto-generated English caption
- checks if auto-generated captions are enabled for the current video (by checking if an English auto-generated one exists)

If auto-generated captions are disabled **and** no manual English caption is provided, the app doesn't try to obtain a locale caption

Issue #1144, opened 2021-03-29
Typography in FreeTube


sveinki:
**Topic**
Please select what topic your discussion is about.
- [X] Coding
- [X] Languages

**Discussion**
In a few places 'DASH' is treated as a proper name, e.g. 'Dash Formats'. Since 'DASH' is an acronym for 'Dynamic Adaptive Streaming over HTTP' it really should be treated like 'HTTP' og 'API' and should be capitalized like any other acronym.
Another thing that bugs me is that the tooltips never end with a '.' full stop period; even if they consist of several sentences, where the first ones end with a period. Perhaps there is some coding logic for this I'm unaware of.

Some may regard these as 'cosmetic' issues, but I know several people who regard such a software as 'juvenile' and 'unprofessional'.
Correcting these would result in plenty of 'fuzzy' strings in all existing translations, so I'm proposing this should be changed while stepping up a major release. 

Any thoughts?

GilgusMaximus:
Hi,
this is not a code related problem. Many things are artifacts from when we worked during the rewrite of FreeTube. Because everything had to be done quickly and tested, we simply ignored it here and there as long as it was working and understandable. Some other things like the tooltips were introduce later and with the first few being merged into the program like this, it just became the way to do it, so that the app is consistent. Also we are using Weblate as a translation service, so for these translations there is no from us checking in on whether all sentences in all languages are consistent. 

If you want, you can go ahead and open a pull request and edit the following [file](https://github.com/FreeTubeApp/FreeTube/blob/development/static/locales/en-US.yaml). We'll take a look and if it is fine, we can merge it in and it will be brought to all translations in weblate automatically, so the translators can adjust their translations if needed.

If you want to do it and need any help, feel free to ask, otherwise I'll add the good first issue, so that other new contributors could take on this

sveinki:
Marked it as code-related because it's about original UI-messages. 
I could modify the strings if people agree on the timing (see my last paragraph above) and if it's OK to modify the tooltip strings.

GilgusMaximus:
Thanks for doing so.
At the moment we do not plan major releases. We simply wait until enough has accumulated to justify a new release and then we write it together and that is it. The next one will not be much further away tho (maybe a week or so, depends on how much PrestoN has to do outside of FT), because the Sponsorblock pull request will likely be the major feature in the next release, as we are getting some stuff done in the background and on the backend to make life a bit easier.

[...]

PR #1147, opened 2021-03-29
Update en-US.yaml

sveinki:
As discussed in #1144


GilgusMaximus:
Okay thanks for doing the work. We appreciate it a lot :)
I just merged something, cause along the way I added another setting, which was conflicting. But it is now resolved.

PR #1153, opened 2021-04-01
Add "open new window" button

### LINKED ISSUE #1122: Feature Request: Add "Create New Window" Menu Item
**Important note**
Please check other open and closed issues for a similar feature request before proposing a new one. 
Also please leave all headers except this one in the issue, so that it is clear what belongs where.

I'm always a bit annoying when I am playing a video / browsing search result, and wanted to start another search / play some background music/video in another window.

In menu bar (Webkit Menu Bar?), probably "File", add "New Window" menu item.
Pressing it should open a new window just like when app starts up.

I know I can open a random video with middle click and use that new window.

**Screenshots**
My expected location to add the new menu item
image

**Additional context**
Add any other context about the feature request here.

PikachuEXE:
**Pull Request Type**
- [X] Feature Implementation

**Related issue**
- Closes #1122 

**Description**
Adds a button to top nav
A new window is created when this new button is pressed
This button is only shown on wide windows (reverse of search container toggle button)

image
image

Tested via these steps
- Run app via `npm run debug`
- Press button to open new window(s)
- Close old window(s), especially the initially created one
- Ensure no error raised in main process

Issues encountered during development:

**1. `Object has been destroyed`**
caused by accessing `mainWindow.webContents`, due to `mainWindow` points to a closed window

**2. Duplicate Application Menu Items**
The menu created via `setMenu` > `Menu.setApplicationMenu`
calling `setMenu` each time a new window is opened would cause it
This a conditional statement is added

 - OS: MacOS
 - OS Version: 10.15.7

**Additional context**
**Side Effect 1**
This change might affect how window state are saved
On each window close/unload `onbeforeunload` callback is run and would send event `'setBounds'` to `ipcMain`
And `mainWindow` state is saved
But with this change there can be multiple windows and when event `'setBounds'` is sent only 1 window's state would be saved.

**Side Effect 2**
Potential issue might be introduced since events on `ipcMain`
- `disableSmoothScrolling`
- `enableSmoothScrolling`
- `disableProxy`
- `enableProxy`
rely on `mainWindow`
And I have not tested these events

Changes to `package-lock.json` not committed
Let me know if they should be (I assume those changes are MacOS only)

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -31,6 +31,7 @@ function runApp() {
   const isDev = process.env.NODE_ENV === 'development'
   const isDebug = process.argv.includes('--debug')
   let mainWindow
+  let openedWindows = []
   let startupUrl
 
   // CORS somehow gets re-enabled in Electron v9.0.4
@@ -219,11 +220,11 @@ function runApp() {
     }
   }
 
-  function createWindow (useProxy = false, proxyUrl = '') {
+  function createWindow (useProxy = false, proxyUrl = '', replaceMainWindow = true) {
     /**
      * Initial window options
      */
-    mainWindow = new BrowserWindow({
+    const newWindow = new BrowserWindow({
       backgroundColor: '#fff',
       icon: isDev
         ? path.join(__dirname, '../../_icons/iconColor.png')
@@ -241,14 +242,18 @@ function runApp() {
       },
       show: false
     })
+    openedWindows.push(newWindow)
+    if (replaceMainWindow) {
+      mainWindow = newWindow
+    }
 
-    mainWindow.setBounds({
+    newWindow.setBounds({
       width: 1200,
       height: 800
     })
 
     if (useProxy) {
-      mainWindow.webContents.session.setProxy({
+      newWindow.webContents.session.setProxy({
         proxyRules: proxyUrl
       })
     }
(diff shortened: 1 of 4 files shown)


GilgusMaximus:
Thanks for putting this together and apologies for taking so long to take a look. A lot's been going on recently.

In regards to the side effects:
As we do not support multiple windows as our own features but simply rely on the basic implementation of Electron in the way that we did not disable it, there are always gonna be side effects. For example the concurrency in the history database is not specified for this multi use scenario. It seems to work, but we are not guaranteeing it. 

Also the changes to the settings seem to take effect after completely closing all windows. So that should be fine.

Only request I have is that the button is available all the time. If someone with a smaller screen, or splitting the screen in two halfs,  still wants to have multiple windows that is not possible with the new icon. So therefore it should be visible as long as possible.

After that I'm happy to merge this :)

Best Rgeards,
Gilgus

PikachuEXE:
I updated the style to make it appear when "History" menu item appear
Let me know if there are other changes required

Screen Shot 2021-04-15 at 17 30 12
Screen Shot 2021-04-15 at 17 30 18

PikachuEXE:
I think there are still issues like not all windows closed when setting like smooth scrolling changed

GilgusMaximus:
Thanks for making the changes. Personally I do not have problems with the setting after closing all windows. So yeah, if someone has a problem with this, we can probably add the info to the restart popup

PR #1203, opened 2021-04-16
* Update main process to handle all opened windows on some setting change

PikachuEXE:
**Pull Request Type**
- [x] Feature Implementation

**Related issue**
Follow up change for #1153

**Description**
In #1153 some settings change does not apply on all opened windows
This PR's change would ensure the following settings are applied to all opened windows
- smooth scrolling enabled (by closing all opened windows instead of just the "main" window)
- proxy URL (by applying the proxy setting to all opened windows instead of just the "main" window)

(Not Applicable)

"Smooth scrolling": opening multiple windows via the button introduced in #1153, update "smooth scrolling enabled", all windows are closed and a new window reopened
"Proxy URL": NOT tested (No idea how to inspect proxy value of a window visually)

 - OS: MacOS
 - OS Version: 10.15.7

**Additional context**
(Not Applicable)

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -423,7 +423,7 @@ function runApp() {
   })
 
   app.on('activate', () => {
-    if (mainWindow === null) {
+    if (mainWindow === null || mainWindow === undefined) {
       createWindow()
     }
   })


PikachuEXE:
Added a fix for an issue that might be introduced due to #1153

`var === null` cannot handle `undefined` but `var == null` can

src/main/index.js
```diff
@@ -423,7 +427,7 @@ function runApp() {
   })
 
   app.on('activate', () => {
-    if (mainWindow === null) {
+    if (mainWindow == null) {
```
Svallinn, line 430: Either use simplified checking
```js
if (mainWindow) {
```
Or use strict equality, even if it's more verbose
```js
if (mainWindow === null || mainWindow === undefined) {
```

Svallinn (changes requested):
Nice work.
However, there's one big problem with this PR.

Technically speaking, the windows are successfully letting each other know when the proxy setting changes, but only the windows are aware of it, the store state isn't.
So what ends up happening is that if the proxy is turned off and you turn it on in one window

image

Every other windows' settings will look like this

image

because their store states weren't updated. The same applies if the host or port number were to be changed.
So you'll need to implement an ipc call that can handle that and propagate every other window when the proxy settings change.

Of course, this most likely applies to **every other setting** that we have, so making it as generic as possible now so that we can cover every other setting in the future would be ideal.

PikachuEXE:
I simply give up on multi-window state sync
For this PR I will just include the fix for new window no opened when all window closed on MacOS

Issue #1218, opened 2021-04-21
Search box doesn't handle channel type urls appropriately


Svallinn:
**Behavior of the program**
Search box doesn't redirect channel type urls

**Expected behavior**
Search box should redirect channel type urls to the approppriate channel view

**To Reproduce**
1. Paste a channel type url into the search box
2. It assumes it's a search query just like any other

**Environment Information (please complete the following information):**
 - FreeTube version: development

**Additional context**
The way to properly fix this is to move the `handleYoutubeLink` function (present in the `src/renderer/App.js` file) to one of the Vuex stores, in order to make it accessible to the `top-nav` component as well
A few modifications to the function itself will also be necessary

https://github.com/FreeTubeApp/FreeTube/blob/78324ff775349b78ad52066d49e864b409cb3b0f/src/renderer/App.js#L298-L401

PikachuEXE:
I just finished https://github.com/FreeTubeApp/FreeTube/pull/1221
Before I found this issue
I don't even remember why I started it

PR #1221, opened 2021-04-22
* Update top nav search input to direct to channel panel when channel URL provided

PikachuEXE:
**Pull Request Type**
- [X] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/1218

**Description**
This PR makes channel URL input in search input box in top nav handled by redirecting to channel view.
Instead of searching with it
Channel URL expected to be like `/channel/{channelID}`

https://user-images.githubusercontent.com/1018543/115644811-c48a2b80-a351-11eb-9504-39f0982c24b8.mp4

- Enter channel URL (e.g. https://www.youtube.com/channel/UCkRfArvrzheW2E7b6SVT7vQ)
- Press Enter/"Go To" button
- Ensure channel view shown instead of search result view

 - OS: MacOS
 - OS Version: 10.15.7

**Additional context**

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -297,129 +297,75 @@ export default Vue.extend({
     },
 
     handleYoutubeLink: function (href) {
-      const v = this
+      this.$store.dispatch('getYoutubeUrlInfo', href).then((result) => {
+        switch (result.urlType) {
+          case 'video': {
+            const { videoId, timestamp } = result
 
-      // Assume it's a video
-      this.$store.dispatch('getVideoParamsFromUrl', href).then(({ videoId, timestamp }) => {
-        if (videoId) {
-          v.$router.push({
-            path: `/watch/${videoId}`,
-            query: timestamp ? { timestamp } : {}
-          })
-        } else {
-          // Could be a playlist, channel, search query or hashtag
-          // If it's none of these, do nothing
-          //
-          // There's a limitation where some unknown URL types will be
-          // determined to be channels
-          // This is due to the ambiguity of some of the existing
-          // channel URL formats and there's not much that can be
-          // done to remedy it
-
-          const url = new URL(href)
-          let urlType = 'unknown'
-
-          const channelPattern =
-            /^\/(?:c\/|channel\/|user\/)?([^/]+)(?:\/join)?\/?$/
-
-          const typePatterns = new Map([
-            ['playlist', /^\/playlist\/?$/],
-            ['search', /^\/results\/?$/],
-            ['hashtag', /^\/hashtag\/([^/?&#]+)$/],
-            ['channel', channelPattern]
-          ])
-
-          for (const [type, pattern] of typePatterns) {
-            const matchFound = pattern.test(url.pathname)
-            if (matchFound) {
-              urlType = type
-              break
-            }
+            this.$router.push({
+              path: `/watch/${videoId}`,
+              query: timestamp ? { timestamp } : {}
+            })
+            break
           }
 
-          switch (urlType) {
-            case 'playlist': {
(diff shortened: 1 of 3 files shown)


src/renderer/store/modules/utils.js
```diff
@@ -260,6 +260,28 @@ const actions = {
...
+        if (urlObject.pathname.startsWith('/channel/') && !urlObject.pathname.endsWith('/channel/')) {
+          // Remove `/channel/` from `/channel/{channelID}`
+          return urlObject.pathname.slice('/channel/'.length)
+        }
+      }
+    ]
+
+    return extractors.reduce((a, c) => a || c(), null) || false
+  },
+
```
GilgusMaximus, line 284: Good idea, unfortunately 2 out of 4 channel links are not covered by this.
Channel links can be constructed the following ways:
/channel/id
/user/userName
/c/name
/name

That is the reason @Svallinn wanted to reuse the existing extraction code in #1218

GilgusMaximus (changes requested):

PikachuEXE:
Extracted a function (but not the entire `handleYoutubeLink`) which tries to extract info from input text if it's Youtube URL
I am assuming the handling of different Youtube URL types might be slightly different in different context

src/renderer/App.js
```diff
@@ -299,127 +299,76 @@ export default Vue.extend({
     handleYoutubeLink: function (href) {
       const v = this
```
Svallinn, line 301: Since you're using an arrow function in the callback, you can get rid of the `v` identifier and replace it with `this`.
Right now, there's a mixture of the two.

src/renderer/components/top-nav/top-nav.js
```diff
@@ -90,32 +90,76 @@ export default Vue.extend({
...
+
+            this.$router.push({
+              path: `/channel/${channelId}`
+            })
+            break
+          }
+
+          case 'invalid_url':
+          default: {
+            router.push({
```
Svallinn, line 149: Should be `this.$router`.

Svallinn (commented):
It looks pretty good to me :)

The only thing that's really itching at me is the fact that a lot of the result logic is repeated in the two callbacks used in the `getYoutubeUrlInfo` dispatches.

But the only thing that came to mind in order to "fix" it was gimmicky hacks like passing `this` to the dispatch and making use of errors and `catch` callbacks, so I'm just gonna ignore the itching entirely.

GilgusMaximus:
> The only thing that's really itching at me is the fact that a lot of the result logic is repeated in the two callbacks used in the getYoutubeUrlInfo dispatches.

In theory yes, but when looking at the proportion of code deleted and code added, I would argue this is fair enough. Also it is not I have not seen stuff like that in other parts of the code / modules before

GilgusMaximus:
Code looks good in total. So if you could add the changes @Svallinn annotated, I would test it a last time and then merge it

PikachuEXE:
Requested changes made
The callbacks for `getVideoParamsFromUrl` in each component are slightly different when handling `default` and `invalid_url` cases
e.g. Clicking on a link vs entering a valid but not youtube URL in top nav

Generally I keep this kind of "duplicate" code around (with slight difference) until there is too much copy & paste happening.


For `router.push` (which can be replaced by `this.$router.push`) 
there is still one occurrence in 
https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/side-nav/side-nav.js#L63

GilgusMaximus (approved):

GilgusMaximus:
Thanks for putting in the work and changes, appreciate it a lot. Works great for all channel types :)

Issue #1005, opened 2021-02-01
Add Reload button


Iusy:
I think a reload button is a basic functionality , but very much needed.
I looked everywhere and there isn't a way to manually reload the curent video.
The reload button is to be added at the top of the page, as on every browser, next to the previous/next arrows.

saideepesh000:
Hey @Iusy , thanks for pointing out, until it's implemented, you can use ctrl+r, but this reloads the entire app.

PikachuEXE:
#1223 opened

PikachuEXE:
After reading [a review comment for #1223](https://github.com/FreeTubeApp/FreeTube/pull/1223#pullrequestreview-646031239)
Now I don't understand the meaning of "reloading the current video"

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE maybe issue #1500 helps u out.

jcartner-young:
IMHO, adding a refresh button is just hiding existing problems and it starts to become a 'YouTube web browser'... Is there not just a 'replay' or restart button?

I have posted a comment about this on related thread: https://github.com/FreeTubeApp/FreeTube/pull/1223#issuecomment-874865868

jasonhenriquez:
Continuing the discussion under #1223:
> Or, as a last resort, initiate the reload routine when this error occurs, with a counter so after 3 tries it can fail with an error?

I think this is a good idea, but auto-reloading is a bit jarring to the user IMO, and it might be irksome if it's a client-side issue they're working on (e.g., they turned off their Internet connection). One possibility might be to add a polite "An error occurred. Click here to reload the video" alert that appears upon the detection of an error that impedes the video player.

> IMHO a 'reload button' feels like a bit of a workaround for some unresolved problems laying beneath. In the same way that turning it off and back on is not actually a solution.

This is technically correct, but I would argue that this rationale leans on the nirvana fallacy. There will always be new bugs popping up, bugs that can't be fixed easily, errors that can't be detected easily, upstream bugs outside of our direct purview, and bugs that most users won't see fixed until the next official release. For example, sometimes I pull up a video on FT, and it takes forever to load. The bug doesn't show up as a detectable error, and the problem usually goes away after reloading the video. Maybe there's an upstream fix being worked on, maybe the upstream fix is out but just not yet merged into FT, maybe I should just work on solving it myself because open source ideology etc etc, but realistically I'm just going to be annoyed as a user. 

[...]

PR #1223, opened 2021-04-22
+ Add Reload Video button on video page

PikachuEXE:
**Pull Request Type**
- [X] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/1005

**Description**
Adds a button to view video page, next to share button (can be repositioned)
Which reload the view/page actually on click

image

- View a video 
- Press the new button

 - OS: MacOS
 - OS Version: 10.15.7

### DIFF
--- a/src/renderer/components/watch-video-info/watch-video-info.js
+++ b/src/renderer/components/watch-video-info/watch-video-info.js
@@ -341,6 +341,10 @@ export default Vue.extend({
       }
     },
 
+    reloadVideo: function () {
+      this.$router.go(this.$router.currentRoute)
+    },
+
     addToPlaylist: function () {
       const videoData = {
         videoId: this.id,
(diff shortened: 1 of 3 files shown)


Svallinn (changes requested):
As far as I can tell, this doesn't actually work.
It's trying to refresh the page rather than reload the components.
So it ends up like this after clicking the button.
image

peepopoggers:
yeah reloading breaks components, not sure y

jcartner-young:
IMHO a 'reload button' feels like a bit of a workaround for some unresolved problems laying beneath.

Maybe build some logic into the play button to deal with these 'disconnections' in the background?

Or,  as a last resort, initiate the reload routine when this error occurs, with a counter so after 3 tries it can fail with an error?

Likely need to find out what is causing the error to start with - TCP/server/session timeout? TCP/Session keep-alive??

PikachuEXE:
@jcartner-young 

Agree
I think the current workaround (reloading) is good enough as a workaround and this is no much better (and it's not working)

Issue #1205, opened 2021-04-17
Locale Preference: System


ovari:
1. Please add a `System` preference to the `Locale Preference` drop-down, i.e. `Settings` → `General Settings` → `Locale Preference` → `System`.
2. Please make `System` the default `Locale Preference`.
3. In the event that the `System` locale translations are unavailable in FreeTube, please show the strings in `English (US)`.

What do you think?

Thank you

GilgusMaximus:
This issue could be implemented by a new user, as it is potentially easy to do.
A quick google search lead me to this documentation page:

https://www.electronjs.org/docs/api/locales

If someone wants to get going with this and needs some help, feel free to ask :)

milupol:
Hey I am new to Open Source and at the moment I'm also fairly new to electron and vue (I know what they are doing and get used for). But i want to try implementing this feature. Is that fine

ovari:
Thank you for implementing. Is the label to "System Language" translatable? Thank you once again

Svallinn:
No, not yet.
Thanks for reminding me of that.
One of us will implement that ourselves when we have the time (unless someone else does it first).

ovari:
Perhaps this has been done here:
https://github.com/FreeTubeApp/FreeTube/commit/8e4fe4eacb2de6a831d11eed84a0d6a8104d5bba#diff-4eaadcc25bfbd7ab1ba912e36d5e1bcbab346cab59546a970df7546d5fb7a145R116

With the string now showing on Weblate at:
https://hosted.weblate.org/translate/free-tube/translations/hu/?checksum=43cef30db93cb0b1

Thanks to everyone for helping with this feature.

Issue #673, opened 2020-10-15
Locale Issues Megathread


PrestonN:
Hey everyone,

I'll be the first to admit that the way that locales are setup isn't perfect.  There are a few inconsistencies between languages that need to be fixed.  Some smaller, some larger.

I'd like to compile a list of all of the problems that are out there.  Fixing these issues will likely end up being done all at once in their own development cycle.  I'd like this thread to be the location for all issues related to localization changed.

Note:  If the issue is about a certain wording of a translation, it may be best to have that discussion over in [Weblate](https://hosted.weblate.org/engage/free-tube/) as I won't be too helpful with that.  This thread is more for stylizing and formatting issues.

What I need from translators / native speakers is the following.

1. The language that needs changes
2. The formatting / stylized change that needs to be made
3. How it's currently formatted
4. How it needs to change
5. Anything else you feel may be helpful

Once the problem has been fixed, I will hide your post to make you aware that it's resolved.

Thanks for the help!

milotype:
@PrestonN Should I simply link my issues/merge request here one by one in order not to repeat myself, and so you could remove them if needed? Or am I missinterpretating the intention for this thread?

PrestonN:
Go ahead and link the issues that you've created.  Also go ahead and please close them if they are not already.

milotype:
Alignment and typography corrections https://github.com/FreeTubeApp/FreeTube/pull/653#issuecomment-708464187 
The issue about a separator charater between "value" and "extension" remained unsolved.

Elaborendum:
- Selecting System Default when the computer is using any Latin American Spanish (excluding, I assume, Mexican Spanish) results in the Locale not being found and the app defaulting to English. I assume this happens for other locales that are specified to a region (like maybe Belgian French?). This happened on Windows 10.
- I don't really agree with the idea of having multiple Spanish translations in projects that aren't very big, as I've seen that a lot of the time one ends up being very under worked because translators probably feel like it's tedious to do most of the same work twice. But since this project has both UK and US English, I'll assume this shall follow suit in this case. 
  - Anyway, onto the actual issue. I don't really think it's necessary to have a third Spanish translation (the Argentinian Spanish one), as the Mexican Spanish translation could cover Latin American Spanish in general, plus, it's already under worked in comparison with the other two.

Is this the right place, or should I file a separate issue for each one?

GilgusMaximus:
[...]

PR #1241, opened 2021-04-30
Adding a 'Follow System' option to the 'Locale Preference'

milupol:
---
# Adding a 'Follow System' option to the 'Locale Preference'
---

 ## **Pull Request Type**

- [ ] Bugfix
- [x] Feature Implementation

## **Related issue**
#1205 

## **Description**
- New locale option created
- Option sets the local to the same local that the system running freetube is using  
- Option is the new standard Locale Preference 
- Option only shows up in case of a supported system locale if the system locale is not support `en-US` gets set 

### Before
image

### After
image

## **Testing (for code that is not small enough to be easily understandable)**
I tested it with system locale: English(US) `en-US.yaml` most of the time and it works fine. But if I change to just English without (US) it tries to access `en.yaml`, which is not existing. Same would be for `fr-FR`, `de-DE`. For `pt-Pt` and `pt` it should be working fine, because there is a  `pt.yaml` file. 

Maybe something for a new Issue or for this Issue #673

## **Desktop:**
 - OS: Manjaro Linux 
 - OS Version: 21.0.3
 - FreeTube version: 7.11.1

**Feel free to correct the code :D**

### DIFF
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -103,12 +103,15 @@ export default Vue.extend({
       const names = []
 
       Object.keys(this.$i18n.messages).forEach((locale) => {
-        const localeName = this.$i18n.messages[locale]['Locale Name']
-        if (typeof localeName !== 'undefined' && locale !== 'system') {
-          names.push(localeName)
-        } else if (typeof localeName !== 'undefined' && locale === 'system') {
+        if (locale === 'system') {
           names.push('Follow System')
-        } else {
+          return
+        }
+        const localeName = this.$i18n.messages[locale]['Locale Name']
+        if (typeof localeName !== 'undefined') {
+          names.push(localeName)
+        }
+        else {
           names.push(locale)
         }
       })
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -104,8 +104,10 @@ export default Vue.extend({
 
       Object.keys(this.$i18n.messages).forEach((locale) => {
         const localeName = this.$i18n.messages[locale]['Locale Name']
-        if (typeof localeName !== 'undefined') {
+        if (typeof localeName !== 'undefined' && locale !== "system") {
           names.push(localeName)
+        } else if (typeof localeName !== 'undefined' && locale === "system") { // adds Follow System to localeNames if system locale was found
+          names.push("Follow System")
         } else {
           names.push(locale)
         }
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -106,7 +106,7 @@ export default Vue.extend({
         const localeName = this.$i18n.messages[locale]['Locale Name']
         if (typeof localeName !== 'undefined' && locale !== 'system') {
           names.push(localeName)
-        } else if (typeof localeName !== 'undefined' && locale === 'system') { // adds Follow System to localeNames if system locale was found
+        } else if (typeof localeName !== 'undefined' && locale === 'system') {
           names.push('Follow System')
         } else {
           names.push(locale)
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -104,7 +104,7 @@ export default Vue.extend({
 
       Object.keys(this.$i18n.messages).forEach((locale) => {
         if (locale === 'system') {
-          names.push('Follow System')
+          names.push('System Language')
           return
         }
         const localeName = this.$i18n.messages[locale]['Locale Name']
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -110,8 +110,7 @@ export default Vue.extend({
         const localeName = this.$i18n.messages[locale]['Locale Name']
         if (typeof localeName !== 'undefined') {
           names.push(localeName)
-        }
-        else {
+        } else {
           names.push(locale)
         }
       })
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -104,10 +104,10 @@ export default Vue.extend({
 
       Object.keys(this.$i18n.messages).forEach((locale) => {
         const localeName = this.$i18n.messages[locale]['Locale Name']
-        if (typeof localeName !== 'undefined' && locale !== "system") {
+        if (typeof localeName !== 'undefined' && locale !== 'system') {
           names.push(localeName)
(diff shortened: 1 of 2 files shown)


src/renderer/components/general-settings/general-settings.js
```diff
@@ -104,8 +104,10 @@ export default Vue.extend({
...
       Object.keys(this.$i18n.messages).forEach((locale) => {
         const localeName = this.$i18n.messages[locale]['Locale Name']
-        if (typeof localeName !== 'undefined') {
+        if (typeof localeName !== 'undefined' && locale !== 'system') {
           names.push(localeName)
+        } else if (typeof localeName !== 'undefined' && locale === 'system') {
+          names.push('Follow System')
         } else {
           names.push(locale)
         }
```
Svallinn, line 113: A better structure for this would be:
```js
if (locale === 'system') {
  names.push('Follow System')
  return
}

const localeName = this.$i18n.messages[locale]['Locale Name']
if (typeof localeName !== 'undefined') {
  names.push(localeName)
} else {
  names.push(locale)
}

File: src/renderer/main.js [hunk 2]
```diff
@@ -27,26 +27,39 @@ Vue.component('FontAwesomeIcon', FontAwesomeIcon)
...
-const activeLocales = ['en-US', 'en_GB', 'ar', 'bg', 'cs', 'da', 'de-DE', 'el', 'es', 'es-MX', 'fi', 'fr-FR', 'gl', 'he', 'hu', 'hr', 'id', 'is', 'it', 'ja', 'nb_NO', 'nl', 'nn', 'pl', 'pt', 'pt-BR', 'pt-PT', 'ru', 'sk', 'sl', 'sv', 'tr', 'uk', 'vi', 'zh-CN', 'zh-TW']
+const activeLocales = ['system', 'en-US', 'en_GB', 'ar', 'bg', 'cs', 'da', 'de-DE', 'el', 'es', 'es-MX', 'fi', 'fr-FR', 'gl', 'he', 'hu', 'hr', 'id', 'is', 'it', 'ja', 'nb_NO', 'nl', 'nn', 'pl', 'pt', 'pt-BR', 'pt-PT', 'ru', 'sk', 'sl', 'sv', 'tr', 'uk', 'vi', 'zh-CN', 'zh-TW']
 const messages = {}
 /* eslint-disable-next-line */
 const fileLocation = isDev ? 'static/locales/' : `${__dirname}/static/locales/`
 
 // Take active locales and load respective YAML file
 activeLocales.forEach((locale) => {
+  // Import elctrons app object to access getLocale function
+  const { remote } = require('electron')
```
Svallinn, line 38: The `remote` API module is deprecated.
Use `@electron/remote` instead.

src/renderer/main.js
```diff
@@ -27,26 +27,39 @@ Vue.component('FontAwesomeIcon', FontAwesomeIcon)
...
+      try {
+        const doc = yaml.load(fs.readFileSync(`${fileLocation}${app.getLocale()}.yaml`))
+        messages[locale] = doc
+      } catch (e) {
+        console.log(e)
+      }
+    } else {
+      const doc = yaml.load(fs.readFileSync(`${fileLocation}${locale}.yaml`))
+      messages[locale] = doc
+    }
```
Svallinn, line 53: The aforementioned issue of having certain "missing" locales can be somewhat mended here.
You can first check if the locale exists.
If it doesn't, you can then check through the `activeLocales` to check if there's any locale that simply starts with those characters and, if there is, choose that one (or, if multiple exist, the first one).

Example:
if the app has locale `en`, you'll verify that there is no `en.yaml` (if there was, you'd load that one)
you then go through the locales and see that both `en-US` and `en_GB` locales exist (both start with `en`)
`en-US` occurs first, so you choose to load `en-US.yaml`

Svallinn (changes requested):

src/renderer/components/general-settings/general-settings.js
milupol: Thanks for the review. I wasn't aware of the `return` statement skipping the loop-run :see_no_evil:. 
I will clean that mess up

src/renderer/main.js
milupol: changed it

src/renderer/main.js
milupol: So i created a Regular Expression 
```javascript
const regex = new RegExp(`${systemsLocale}-\\w+`, 'g')
```
So if systemsLocale is `fr` it matches with `fr-FR`

80e8466

src/renderer/main.js
```diff
@@ -27,26 +27,40 @@ Vue.component('FontAwesomeIcon', FontAwesomeIcon)
...
-    messages[locale] = doc
+    if (locale === 'system') {
+      const systemsLocale = app.getLocale()
+      if (activeLocales.indexOf(systemsLocale) !== -1) {
+        const doc = yaml.load(fs.readFileSync(`${fileLocation}${systemsLocale}.yaml`))
+        messages[locale] = doc
+      } else {
+        const regex = new RegExp(`${systemsLocale}-\\w+`, 'g')
+        const doc = yaml.load(fs.readFileSync(`${fileLocation}${activeLocales.join(' ').match(regex)[0]}.yaml`))
+        messages[locale] = doc
```
Svallinn, line 49: Ok, I'm gonna ask you to change this entirely, but not because this wouldn't work with some adjustments, it's just not as robust.
First, I'll let you know why this is not good enough as it is now:
1. `-` alone is not enough because some locales have `_` instead. Right now, only `nb_NO` exists,  but there might be more added later on, so it needs to be future proof
2. You're assuming that there will always be at least one existing match (no `try/catch` block), not sure if you just forgot
3. Regex totally works for this but it's not necessary, there's a much simpler way

The best way to do this is to use `filter` on the `activeLocales` array and use a `currentValue.startsWith(systemsLocale)` as the condition for it to pass the test in the arrow function.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

Then you can simply check if the new array is empty or not.
If it isn't, then you can load the yaml using the first element of that array.
There won't be a need for a `try/catch` block, because you're already sure the file exists since the array isn't empty :)

Svallinn (changes requested):

src/renderer/main.js
GilgusMaximus: Additional Request from my side: Please change the string 'Follow System' to something like 'System Language' or similar. 'Following System' seems a bit out of place here and might confuse people translating it over at Weblate

milupol: Good ideas implemented them in 3ea9f0e and 9a04312 it's amazing how much you can learn through contributing :D
Thank you for the help :+1:

Svallinn (approved):

Svallinn:
LGTM. Thank you for working on this feature :)

Issue #882, opened 2020-12-17
Allow Controling the volume with the scroll wheel on the entire video window


shivasagarrao:
Add a feature to Control the volume level with the mouse wheel when hovering the video player with the mouse cursor.
This feature is not natively available in YouTube but I use [Enhancer extension](https://chrome.google.com/webstore/detail/enhancer-for-youtube/ponfpcnoihfmfllpaingbgckeeldkhle) to get this feature.

PrestonN:
This is already implemented and in the latest release.

shivasagarrao:
I'm currently using v0.9.3, am unable to control volume with mouse. Is there any setting that I need to change?

PrestonN:
You need to hover over the volume slider. It currently doesn't work if you hover over the video itself.

shivasagarrao:
My feature request was controlling the volume when you hover over the video. May I open another issue?

PrestonN:
That's fair. I'll just reopen this issue.

Albinexotech:
Hey @PrestonN , I was wondering if I would be able to take on this issue?

GilgusMaximus:
Sure, if you need some help feel free to ask here or once you opened a PR in the PR :)

PR #1254, opened 2021-05-05
Allow changing volume by scrolling over any part of video player

Albinexotech:
**Pull Request Type**
- [x] Feature Implementation

**Related issue**
#882 

**Description**
This PR allows the ability to control the volume of the video by scrolling over it. Currently, the main way of changing the volume is either scrolling or dragging over the volume slider. When this setting is enabled, the volume can be changed by scrolling over any part of the video. The setting can be found in Player Settings, and is turned off by default.

playerSettings

Tested using DASH, Legacy, and Audio formats, as well using Local API and Invidious API. Ensured setting functioned as intended, both when disabled and enabled.

 - OS: Manjaro Linux
 - OS Version: 5.12.0-1-MANJARO
 - FreeTube version: 0.12 Beta (up to date with development branch)

**Additional context**
I settled on the wording `Scroll Volume Over Video` for the setting. However, I am on the fence about it, and worry that it might not be descriptive enough. As this is my first PR, please let me know if there is anything that I missed, or should be changed. Thanks!


Albinexotech:
Changed the wording of the setting to `Scroll Volume Over Video Player`, as it better describes the behaviour.

Albinexotech:
Theater Mode, Full Window, and Fullscreen all work as intended. With Picture-in-Picture, it only allows changing the volume from the original video area, and not the Picture-in-Picture player itself. 

areaOfEffect2
(The red dashed area is where the volume can be changed, in this example).

efb4f5ff-1298-471a-8973-3d47447115dc:
Do u feel comfortable to try take on the challenge to implement this for PIP?

Albinexotech:
I went through to see if it is possible to add better support for PiP. However, I wasn't able to find an easily-accessible reference point for the PiP window itself. 

I'm pretty new to a lot of this, so I could be completely wrong! However, the PiP window that Electron creates seems to use an interface that doesn't allow direct customization, apart from properties/events relating to the window size.

efb4f5ff-1298-471a-8973-3d47447115dc:
Ah thanks for looking though.
Im also new to allot of this so i know where ur coming from :)

PikachuEXE:
Let's do PiP part in a separate PR if too difficult
Don't let perfect be a blocker of being better

Albinexotech:
I agree! This could be a problem to look further into in its own PR

efb4f5ff-1298-471a-8973-3d47447115dc:
Agreed. Will open a issue later today.

GilgusMaximus:
> Let's do PiP part in a separate PR if too difficult
Yes, everything PiP related should probably go into their own PR. Then it is easier to keep track of which PR's interacted with the PiP functionality.

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -229,7 +233,11 @@ export default Vue.extend({
         this.player.on('mouseleave', this.removeMouseTimeout)
 
         this.player.on('volumechange', this.updateVolume)
-        this.player.controlBar.getChild('volumePanel').on('mousewheel', this.mouseScrollVolume)
+        if (this.videoVolumeMouseScroll) {
```
SebTota, line 236: I would recommend using the on `wheel` event rather than the `mousewheel` event due to `mousewheel` being [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event). 
This would only really matter if the application transitions to a web native app rather than an electron app, but I figured it may be worth pointing out.

efb4f5ff-1298-471a-8973-3d47447115dc: Could u place this in #1284?

Albinexotech:
Resolved the changes, and updated the event to `wheel`. Seems to work well!

GilgusMaximus:
So thanks for putting this together. Works like a charm. Only thing left to do before merging is fixing the linter issues. Otherwise Github and the local linters will complain all the time.

Please run `npm run lint` in the freetube folder´to fix it, or fix the issues from the linter check manually :)

Best regards,
Gilgus

GilgusMaximus:
Thanks for fixing the linter issue and putting this feature together. 
We appreciate the help a lot :)

Best Regards,
Gilgus

Closes #882

Issue #1251, opened 2021-05-04
Changing the color of the filter button when a filter is apply


e471wo:
Hi, it's a stupid thing but each time I use the filter (for example find videos longer than 20 minutes), I totally forget there is a filter and so I m astonesh when I have to find something and only find video longer than 20 minutes after. 

Could be cool if the filter button could change color when it's used. (lets say red). And so we know if there is an active filter in the research result. 

thanks.

GilgusMaximus:
Definitely a good idea. I do not use filters myself, but I can see that this might be a good visual clue.
An alternative would be to change the icon to make it more accessible as well

PikachuEXE:
Implemented but not sure what effect should be used
https://github.com/FreeTubeApp/FreeTube/compare/development...PikachuEXE:feature/filter-change-indicator

![Screen Shot 2021-05-07 at 11 55 35 ](https://user-images.githubusercontent.com/1018543/117400121-96514200-af34-11eb-857a-e3ef9396bdd9.png)
![Screen Shot 2021-05-07 at 11 55 38 ](https://user-images.githubusercontent.com/1018543/117400131-99e4c900-af34-11eb-9e4f-a8f63de1b373.png)

PR #1257, opened 2021-05-07
Implement search filter changed indicator

PikachuEXE:
**Pull Request Type**
- [X] Feature Implementation

**Related issue**
#1251 

**Description**
When search filter is updated and value is different from default
Update UI to indicate that

Actual effect to be decided

Screen Shot 2021-05-07 at 11 55 35
Screen Shot 2021-05-07 at 11 55 38

- Open filter panel
- Change any filter from default value and back
- Observe visual change

 - OS: MacOS
 - OS Version: 10.15.7

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -399,7 +399,8 @@ export default Vue.extend({
       'grabHistory',
       'grabAllPlaylists',
       'getRegionData',
-      'getYoutubeUrlInfo'
+      'getYoutubeUrlInfo',
+      'getLocale'
     ]),
 
     ...mapMutations([
--- a/src/renderer/components/ft-search-filters/ft-search-filters.js
+++ b/src/renderer/components/ft-search-filters/ft-search-filters.js
@@ -42,6 +42,15 @@ export default Vue.extend({
       return this.$store.getters.getSearchSettings
     },
 
+    filterValueChanged: function() {
+      return [
+        this.$refs.sortByRadio.selectedValue !== this.sortByValues[0],
+        this.$refs.timeRadio.selectedValue !== this.timeValues[0],
+        this.$refs.typeRadio.selectedValue !== this.typeValues[0],
+        this.$refs.durationRadio.selectedValue !== this.durationValues[0]
+      ].some((bool) => bool === true)
+    },
+
     sortByLabels: function () {
       return [
         this.$t('Search Filters.Sort By.Most Relevant'),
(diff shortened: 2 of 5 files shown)


PikachuEXE:
Here is how it looks in different colors

Screen Shot 2021-05-12 at 11 47 38
Screen Shot 2021-05-12 at 11 47 44
Screen Shot 2021-05-12 at 11 47 48
Screen Shot 2021-05-12 at 11 47 53
Screen Shot 2021-05-12 at 11 47 57
Screen Shot 2021-05-12 at 11 48 04

PikachuEXE:
I think it's better to use secondary color theme for the effect

Screen Shot 2021-05-12 at 13 14 46
Screen Shot 2021-05-12 at 13 14 50
Screen Shot 2021-05-12 at 13 14 55
Screen Shot 2021-05-12 at 13 15 04

efb4f5ff-1298-471a-8973-3d47447115dc:
In my opinion it is better to go with the primary color. When u open the filter window u will see that the color of the filters matches the primary color, because of that i think it would be more recognizable for the user if the color is the same as the main color, because it is a familiar color. 

In short the primary color would look more uniform and recognizable to the user.

PikachuEXE:
Changed back to use primary color

GilgusMaximus:
Okay, thanks taking on this. I finally had some time to test it out and it works great.
However three things are required to merge this:

1) We need a solution for when the user enables the setting 'Match Top Bar with Main Color'. Because then you do not see whether the circle is actually displayed, cause it is the same color as the top bar. This could be, that the ring is now displayed in the same color as the other icons in the top bar. They change from white to dark, so a similar switch for the ring should be doable.

2) The filter icon should be moved only a bit to the right, or the ring's radius should be reduced, so it is not cut of by the search box on it's left hand side.

3) The linter issues in your files must be resolved, otherwise the linter will complain all the time. Which is very annoying

But other than that, I really like the solution.

Best regards,
Gilgus

PikachuEXE:
1. Updated to handle "Match Top Bar with Main Color". See screenshots below.
2. Moved the button via `margin-left` and also reduced effect radius 20px > 10px
3. Linter warnings resolved
Screen Shot 2021-05-25 at 13 43 44
Screen Shot 2021-05-25 at 13 43 49
Screen Shot 2021-05-25 at 13 44 27
Screen Shot 2021-05-25 at 13 44 30

efb4f5ff-1298-471a-8973-3d47447115dc:
Nice solution for the match with top bar color

GilgusMaximus:
Thanks again for making the changes.
Really like how it looks now.

Gonna merge this :)

PR #1258, opened 2021-05-07
Update "Open Link" button to include playlist ID

PikachuEXE:
**Pull Request Type**
- [X] Feature Implementation

**Related issue**
None

**Description**
Update Video Info Panel -> Share Button -> Youtube Side -> "Open Link" button
To open webpage for current video **with** active playlist if there is one in FreeTube

Just screenshot for specifying the button updated
image

Test case A (video without playlist):
- Input URL https://www.youtube.com/watch?v=eBZ2vA9D2RA into FT
- Press "Open Link" button
- Inspect URL

Test case B (video with playlist):
- Input URL https://www.youtube.com/playlist?list=PLzFTGYa_evXiOHCB_cb1Sqsk-LIARmLgd into FT
- View one of videos
- Press "Open Link" button
- Inspect URL

 - OS: MacOS
 - OS Version: 10.15.7

**Additional context**
Not sure if other buttons should be updated too, but let's leave them for later

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -399,7 +399,8 @@ export default Vue.extend({
       'grabHistory',
       'grabAllPlaylists',
       'getRegionData',
-      'getYoutubeUrlInfo'
+      'getYoutubeUrlInfo',
+      'getLocale'
     ]),
 
     ...mapMutations([
--- a/src/renderer/components/ft-share-button/ft-share-button.js
+++ b/src/renderer/components/ft-share-button/ft-share-button.js
@@ -19,6 +19,10 @@ export default Vue.extend({
       type: String,
       required: true
     },
+    playlistId: {
+      type: String,
+      required: true
+    },
     getTimestamp: {
       type: Function,
       required: true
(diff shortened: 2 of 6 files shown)


GilgusMaximus:
Okay, again thanks for putting in the work. Works good, but a few things are missing here as well unfortunately:

1) The open link must then also work for Invidious. Wherever possible Invidious should be kept to the same standard as YouTube. The playlist link for the playlist you provided looks like this: `https://invidious.tube/watch?v=KBAeb-xChG4&list=PLzFTGYa_evXiOHCB_cb1Sqsk-LIARmLgd` . So it is basically the same.

2) If it works for 'open video', I think it should also work for 'copy url'. Otherwise these two options would be inherently different.


Other than these two points, it works great, and makes it a lot easier to share playlists.
So thanks again,

Best Regards,
Gilgus

PikachuEXE:
I will work on point 1 soon.
But for point 2, the current copy button returns a short URL (`https://youtu.be/video_id`) and does not support playlist. I wonder if returning the longer version is fine (as I have no idea why short URL was used at first).

PikachuEXE:
Point 1: done
Point 2: updated to return short URL like before except when playlistID is present

GilgusMaximus:
Ah yeah, good point in regards to point two. But I believe your change is very well done.
Checked everything and it is working fine.

Thanks for making the changes and putting this together :)

Issue #966, opened 2021-01-21
Picture-in-picture - Returning to normal player after changing tab


Jimmymax297:
**Behavior of the program**
When User tries to come back to normal player with picture-in-picture player is opened after changing tab the player is in the background.

**Expected behavior**
Application should return to normal player even if User changed tabs with picture-in-picture player is opened.

**Preconditions**
- User has videos in their Subscriptions tab

**To Reproduce**
Please add all steps to reproduce the behavior:
1. User watches any video from their Subscriptions
2. User clicks on the player bar to change to picture-in-picture player
3. User changes the tab of the application, e.g. clicks on Subscriptions
4. User clicks on the picture-in-picture player to return to the normal player
5. See issue

**Environment Information (please complete the following information):**
 - OS: [Windows]
 - OS Version: [Windows 10]
 - FreeTube version: [0.11.1]
 - Primary API Used: [Local API]

PrestonN:
This is working as intended and was recently talked about in our Matrix channel. FreeTube is not able to differentiate between the user clicking on the "Close" button and the "Back to video player" button if you are on a different page. Because of this, we have no way of identifying the user's intention when PiP is closed. If we forced FreeTube to always navigate back to that video, then it would more than likely not be the user's intention. It's always possible the user simply wanted to stop watching the video. Because of this, closing the video is the better option and will not be changed.

Jimmymax297:
@PrestonN I understand your reasoning here, but you are expecting user to remember that if they changed the tab while having PiP opened when they they can only close it. The main issue here is that the player is somewhere in the background and still playing, only closing the app can close this player or waiting for it to end playback.

PrestonN:
The issue you're mentioning is something that I believe should be fixed in the latest nightly builds. There is logic that prevents the PiP video from playing in the background and there was a bug where this logic wouldn't run if your history is disabled. If you do keep history enabled and you're still experiencing this bug on the latest nightly builds, then please provide step by step instructions to reproduce and I will look into it.

SebTota:
I'm sorry for creating a PR before continuing this issue here. 

I propose, and already created a PR for, a solution that will check if the video is currently paused or playing to determine if the PiP window should close or return to the regular video window.

[...]

Issue #624, opened 2020-10-09
'Back to video player' doesnt work when not on video page


NotTheReal-Jamie:
Pressing the 'back to video player' button in the pop-up video player doesn't do anything besides closing the PIP-player when you're not on the video's page in the Freetube app.

[See here](https://i.imgur.com/diRmnYd.gif) for an example of this in action.

GilgusMaximus:
Thanks for the reminder. This is definitely an issue that has to be sorted out. I'm going to add it to the target list for release 0.10.0

PrestonN:
I think I might consider this working as intended.

There isn't any way to differentiate wanting to close the PiP player and wanting to continue the video in the main player.  Since you've left the video screen, we'd have to re-grab that information so that you could place it in the main player again.

If you leave the main video page and come back to that same video, then the PiP will have the video continue where you left off once closed.

I'd consider this a feature and probably a low priority one for me at least.  Either a toggle will need to be created in the settings to always load the video back up, or a toast message will need to be displayed that when clicked on will take you back to that video.  Both methods will need logic added to allow setting the desired timestamp on a route change.

NotTheReal-Jamie:
Since the button is labelled as "back to video player" I can't imagine that this it's working as intended like this, which would also be in line with how PIP functions in Firefox and other applications.

PrestonN:
As I mentioned, in the background there is nothing different between clicking on "Back to video player" and clicking on the exit button, so we can't use that information to determine the user's intentions when PiP is closed.  What if the user didn't want to keep watching that video?  What if the user had PiP paused, was watching a different video, and decided to close the PiP player?  We don't want it to take over and navigate the user away from that other video in that situation.

The only way we can make this work is if we know 100% that the user's intention is that they want to continue watching the video.  This will need to be done via a prompt or a setting toggle.  If you come back to the video while it's still in PiP, then we can comfortably say that the video should skip to where it was in PiP, which is how it currently behaves.

We need new logic to determine the user's intention when we close a video, which is why it's a new feature and not a bug.  In it's current iteration, since we cannot determine the user's intention, it is working as intended, though that doesn't mean it can't be changed.

SebTota:
Copied from https://github.com/FreeTubeApp/FreeTube/issues/966#issuecomment-835499932

I'm sorry for creating a PR before continuing this issue here. 

[...]

PR #1262, opened 2021-05-08
Implemented the ability to return to the regular viewing mode from the PiP window after navigating away from the video

SebTota:
---
Added the ability to return to the regular viewing mode from the PiP window after navigating away from the video
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
#966 ?
#624 

**Description**
When using Picture In Picture mode and the user navigates away from the video page:
* Return to the video page if the user clicks on the 'Back to video player' button and the PiP video is not paused
* Close the PiP view if the user clicks the 'Close' button or if they click the 'Back to video player' button and the video is paused

1.  Open any video
2. Click to watch the video in Picture in Picture mode
3. Navigate away from the video by either starting another video or going to Settings/Subscriptions/etc. page

Test returning to regular viewing mode from PiP mode
* Perform steps 1-3
* Make sure the PiP video is not paused
* Click 'Back to video player' button 

Test closing PiP mode
* Perform steps 1-3
* Click 'Close' button
OR
* Perform steps 1-3
* Make sure the PiP video is paused
* Click 'Back to video player' button

 - OS: MacOS
 - OS Version: 10.15.7
 - FreeTube version: Latest development branch
 
**Additional context**
Due to the inability to properly track 'Close' vs 'Back to video player' button clicks, the Picture in Picture window will treat both the 'Close' and 'Back to video player' buttons as close buttons when the video is paused.

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -967,7 +967,7 @@ export default Vue.extend({
 
                 if (!videoPaused) {
                   if (window.location.href !== videoUrl) {
-                    window.location.assign(videoUrl)
+                    this.$router.push({ path: `/watch/${this.videoId}` })
                   } else {
                     const activePlayer = $('.ftVideoPlayer video').get(0)
                     activePlayer.currentTime = watchTime
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -54,6 +54,7 @@ export default Vue.extend({
       videoTitle: '',
       videoDescription: '',
       videoDescriptionHtml: '',
+      videoUrl: window.location.href,
       videoViewCount: 0,
       videoLikeCount: 0,
       videoDislikeCount: 0,
@@ -940,24 +941,43 @@ export default Vue.extend({
 
       if (!this.isUpcoming && !this.isLoading) {
         const player = this.$refs.videoPlayer.player
+        const videoUrl = this.videoUrl
 
         if (player !== null && !player.paused() && player.isInPictureInPicture()) {
           const playerId = this.videoId
           setTimeout(() => {
-            player.play()
-            player.on('leavepictureinpicture', (event) => {
-              const watchTime = player.currentTime()
-              if (this.$route.fullPath.includes('/watch')) {
-                const routeId = this.$route.params.id
-                if (routeId === playerId) {
-                  const activePlayer = $('.ftVideoPlayer video').get(0)
-                  activePlayer.currentTime = watchTime
-                }
-              }
+            try {
+              player.play()
+              player.on('leavepictureinpicture', (event) => {
+                const videoPaused = player.paused()
+                const watchTime = player.currentTime()
 
-              player.pause()
-              player.dispose()
-            })
+                // Update watch progress of Picture In Picture video
+                if (this.rememberHistory && !this.isUpcoming && !this.isLoading && !this.isLive) {
+                  if (this.saveWatchedProgress) {
+                    const payload = {
+                      videoId: playerId,
+                      watchProgress: watchTime
+                    }
+                    this.updateWatchProgress(payload)
+                  }
+                }
+
+                player.pause()
+                player.dispose()
+
+                if (!videoPaused) {
+                  if (window.location.href !== videoUrl) {
+                    window.location.assign(videoUrl)
+                  } else {
+                    const activePlayer = $('.ftVideoPlayer video').get(0)
+                    activePlayer.currentTime = watchTime
+                  }
+                }
+              })
+            } catch (e) {
+              console.log(e)
+            }
           }, 200)
         }
       }
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -54,7 +54,6 @@ export default Vue.extend({
       videoTitle: '',
       videoDescription: '',
       videoDescriptionHtml: '',
-      videoUrl: window.location.href,
       videoViewCount: 0,
       videoLikeCount: 0,
       videoDislikeCount: 0,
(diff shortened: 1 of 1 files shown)


SebTota:
After thinking more about this feature, it seems like maybe it should not pertain to navigating to another video. Rather it should only have this functionality when navigating to the subscription, settings, or any other non video page. I was using this locally and it seems quite distracting to have the PiP video retain on screen when navigating to another video. 

What's the input on this? Should it stay as is and keep the PiP video up no matter where the navigation is going? Or should it only retain the PiP video mode when navigating to another page that is not a video. 

This would make things a bit more complex since the PiP video should probably close again when a video from the subscription (or any other) page is chosen, but it seems like it might be worth the change.

GilgusMaximus:
Okay, so thanks for approaching this and sorry for taking so long to test this.
But unfortunately there are some problematic errors int the implementation at the moment:

1) We have a dedicated router setup in place in order to refer to new pages which handles everything necessary to redirect to a new page. So this should definitely be used over setting the window variable. You only need the video id, which you have access to Watch.js 

2) When opening a video in pip and I then move to another video page, multiple things can happen:
  - If I click on the pip icon of the new video, then the new video opens in pip and starts playing from the timestamp the previous pip video was playing
  - If I am in pip and open another video and click back to full screen, the pip video closes and the newer video starts playing in normal player at the timestamp of the pip video



In regards to 1), it looks something like this:
```
this.$router.push({
              path: `/watch/${videoId}`,
              query: timestamp ? { timestamp } : {}
            })
```
You can find many examples in the code for that.

In regards to 2), we have to find a consistent solution here.
First the new video should not play at the timestamp of the pip video at all times. I would imagine that there is a problem with the association of the timestamp fix you added and the video player FT is trying to apply it to. So removing the timestamp for pip again would also be a valid solution and could be the original reason that was not done (without any research though)

If one clicks on the 'go back' button in pip, I am open to what happens then. Does it close itself and the newer video is played, or does it go all the way back to the original video? The second solution would obviously be nicer and more consistent, but I can see a lot of struggles trying to keep the videoId across Watch instances. One solution could be to use a store variable for last pip video, but that would require some more changes. If this is not done and instead the currently open video starts playing in the normal player, I am fine with this as well.

SebTota:
@GilgusMaximus You brought up a really great use case I didn't think of: when a PiP window exists and the user tries to open another video in PiP. 

I am struggling to find a way to differentiate between a PiP window being closed by a user clicking the "X" button in the PiP window and a user opening up another video in PiP. Both of these events send a "closepictureinpicture" event but have no way of distinguishing how the event was delivered. I'm open to ideas of how to differentiate between the two if anyone has one. 

Feel free to close this pull request for now and I can reopen it at a later date if I am able to find a way to achieve this.

GilgusMaximus:
is it possible to recognize whether the mouse pointer is over the pip player? 
In that case one could check that to determine the difference.

But yeah, pip has no defined standard, so everyone is doing what they want with it.

SebTota:
Not as far as I can tell. The way the pip "element" is treated is quite weird. It doesn't actually refer to the PiP window itself, but rather to the original video player on the screen. As the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API#properties_on_the_document_or_shadowroot_interfaces) states

> The pictureInPictureElement property tells you which Element is currently being displayed in the floating window.

I tried also just using the `mousemove` event and trying to see if it would recognize the PiP window as a parentElement, but that also did not work. The only other idea I had was to try to get the coordinates of the PiP window and compare them to where the mouse is. But this introduces a bunch of issues, a major one being that I wasn't able to find a way to get the coordinates of the PiP window. All these issues stem from the fact that you can't reference the actual PiP window as a regular element.

GilgusMaximus:
Yeah this is really not a good way to implement this. Let alone the complexity of this. 
But yeah, if no other idea is left open for the moment I'd close this pull request.

PiP really seems like not so good to support at the moment

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @SebTota just checking in. What is the progress of this? Is this even going to be reviewed because lack of support?

SebTota:
@efb4f5ff-1298-471a-8973-3d47447115dc I wasn't able to find any new implementation of using PiP in a way that would reliably let us know all the different events that are coming from the PiP window. I think as @GilgusMaximus said, we should probably close this for now and maybe reopen #966 to keep track of new ideas there.

efb4f5ff-1298-471a-8973-3d47447115dc:
Its very unfortunate that there is lack of support for this functionality. If @GilgusMaximus agrees, we can close the PR and maybe reopen #966.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Not as far as I can tell. The way the pip "element" is treated is quite weird. It doesn't actually refer to the PiP window itself, but rather to the original video player on the screen. As the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API#properties_on_the_document_or_shadowroot_interfaces) states
> 
> > The pictureInPictureElement property tells you which Element is currently being displayed in the floating window.
> 
> I tried also just using the `mousemove` event and trying to see if it would recognize the PiP window as a parentElement, but that also did not work. The only other idea I had was to try to get the coordinates of the PiP window and compare them to where the mouse is. But this introduces a bunch of issues, a major one being that I wasn't able to find a way to get the coordinates of the PiP window since it only provides the an x and y size property with no location property. All these issues stem from the fact that you can't reference the actual PiP window as a regular element. 

I saw the updated their documentation. Im not sure what it looked like before but maybe there is something new in there that might help us out.

SebTota:
@efb4f5ff-1298-471a-8973-3d47447115dc  If [this](https://developer.mozilla.org/en-US/docs/Web/API/Picture-in-Picture_API#properties_on_the_document_or_shadowroot_interfaces) is the documentation you are referring to, I do not see anything new that would help in this case.

efb4f5ff-1298-471a-8973-3d47447115dc:
Yes i was referring that link. Well thats too bad. I had so much hope :(

efb4f5ff-1298-471a-8973-3d47447115dc:
@SebTota I'll take the liberty to close this PR because Luca will be unavailable for a while https://github.com/FreeTubeApp/FreeTube/pull/1568#issuecomment-932809018

Issue #1273, opened 2021-05-12
Centering default playback rate to 1.0x speed


Tiltowait2:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Please describe clear and concise of what the problem is. Ex. I'm always frustrated when [...]

I'm always frustrated when I click the speed icon to change the playback rate of a video.  It starts at the top of the list, 3.0x speed and only goes down to 1.75x speed.  I have to scroll down to get to the 0.75x - 1.5x range (which is what I guess most people would ever use).

**Describe the solution you'd like to see implemented**
Please give a clear and concise description of what you want to happen.

I would like when the speed icon is clicked, for it to center on 1.0x.  This seems like it would be most useful for the most people.

**Describe alternatives you've considered**
Are there other ways to solve the problem? Please give a clear and concise description of any alternative solutions or features you might have considered.

Alternatively, it could center on the current speed, if it is not 1.0x in preferences.

**Screenshots**
Sometimes people have screenshots from mock-ups or other projects. Add them here.

**Additional context**
Add any other context about the feature request here.

I think few people ever listen to videos at 2.0x and above, and changing this default would save everyone a few clicks.

PR #1274, opened 2021-05-13
Update playback rate menu to put 1x in view when opened

PikachuEXE:
**Pull Request Type**
- [x] Feature Implementation

**Related issue**
#1273

**Description**
Update playback rate menu to put 1x in view when opened

https://user-images.githubusercontent.com/1018543/118444804-33c02900-b720-11eb-9e33-774ae80d7468.mp4

- Open a video
- Open playback rate menu
- Observe the menu items

 - OS: MacOS
 - OS Version: 10.15.7

**Additional context**
`video.js` provide no option to control menu behaviour  
Already check source code about `PlaybackRateMenuButton` & its parent `MenuButton`

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -222,6 +222,18 @@ export default Vue.extend({
         this.player.volume(this.volume)
         this.player.playbackRate(this.defaultPlayback)
 
+        // region playback rate menu focus hack
+        // Makes the playback rate menu focus on last item on  mouse hover
+        // last item is "0.25x" (video.js put lowest rate at last)
+        // This ensures "1x" is visible on menu open
+        const playbackRateMenuButton = this.player.controlBar.getChild('playbackRateMenuButton')
+        playbackRateMenuButton.on(playbackRateMenuButton.menuButton_, 'mouseenter', () => {
+          // Assuming this happens after original hover event handler
+          // Which shows the menu
+          playbackRateMenuButton.menu.focus(999)
+        })
+        // endregion playback rate menu focus hack
+
         if (this.storyboardSrc !== '') {
           this.player.vttThumbnails({
             src: this.storyboardSrc,


efb4f5ff-1298-471a-8973-3d47447115dc:
So im not sure if i read the code correctly but this will make 1x show as the highest at the top and go down to .25?

efb4f5ff-1298-471a-8973-3d47447115dc:
Isn't it preferable to let it look more like this?



It looks a little weird to scroll down for higher playback speeds.

PikachuEXE:
I intended but there is no way to auto scroll down when menu opened
There is no such option in `video.js`

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmmm my bad overlooked that u mentioned that. Now it makes sense why u inverted the array. Instead of searing auto scrolling down maybe search for stick to top or bottom or something like that?

PikachuEXE:
There is **no** option that can alternate the behaviour (as I said I checked the source code)
Now I found [a SO answer about stick to bottom](https://stackoverflow.com/questions/18614301/keep-overflow-div-scrolled-to-bottom-unless-user-scrolls-up) and I can try that later
But the current change will be the fallback if those solutions don't work

efb4f5ff-1298-471a-8973-3d47447115dc:
My apologies i was a bit tired when i typed that😅 
Nice that u found something about that.
Hope it works.
Don't get me wrong, i like ur implementation but my OCD thinks otherwise.

PikachuEXE:
I am fine with your OCD ;)
But if we can make merge this first then try to improve it later
Especially when the next release might be made soon

efb4f5ff-1298-471a-8973-3d47447115dc:
Yeah im fine with that. :)

PrestonN:
We used to have the playback rates in this order at one point, however I had to revert the change shortly after.

Currently, you can directly click on the playback speed underneath the pop-up to cycle through the different rates. When these are reversed, they immediately go to 3x and no longer cycle. I'm not sure why it does this but this is why I haven't changed the order. If we can fix this then I'd be more than happy to order them this way, else I'll likely have to close this PR.

PikachuEXE:
Oh the behaviour is caused by https://github.com/videojs/video.js/blob/v7.12.1/src/js/control-bar/playback-rate-menu/playback-rate-menu-button.js#L113-L128

I guess I will have to look for a better solution then

PikachuEXE:
With CSS only solution: failed
```css
.vjs-playback-rate .vjs-menu-content {
  display: flex;
  flex-direction: column-reverse;
}
```
image

PikachuEXE:
OK code updated and screenshot updated to video to show how it looks now
A bit hacky but hope `video.js` won't change the internal structure that soon

PikachuEXE:
Feel free to test the latest version

Issue #691, opened 2020-10-18
Dynamically adapt theme to system preference


Mynacol:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Most of the applications on my (Linux) system can switch between a light and dark theme. FreeTube bugs me, as it doesn't change the theme if I toggle the system-wide theme.

**Describe the solution you'd like to see implemented**
FreeTube should add a theme option, `System`, which selects a light theme if the system-wide theme is a light one and a dark theme (if dark or black could be another option) if the system has a dark theme.
This should ideally work during usage, so no restart is required to "automatically" change the theme.

**Describe alternatives you've considered**
I see no way to change the theme with a command line tool.
Editing `theme` in `settings.db` in the configuration directory seems to change nothing.

**Screenshots**
(No screenshots by me)

**Additional context**
I think [this](https://www.electronjs.org/docs/api/native-theme) is the related electron manual.

gracjankn:
I hope it'll be implemented. Same applies to macOS and Windows.

Mennaruuk:
Agreed! Switching based on system theme would be awesome.

julianfairfax:
I'd also appreciate this feature being added. If it would follow the system theme and change to the light theme or the dark theme automatically based on what the system theme is, it would improve the integration with the system, and it would be a welcome change for me. Thanks for all the work you've put into this wonderful app!

lothar-cell:
I'd like to see this one on MacOS too!

f-x1-1:
Additionally, You can add light/dark mode theme option, allowing users to choose built in themes for light/dark mode.

https://file.coffee/u/7yF7-eNKkGvZDa461OrY0.png

PR #1318, opened 2021-05-23
Add option to follow system theme

julianfairfax:
---
Add option to follow system theme
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/691

**Description**
This adds an option to follow the system theme. More specifically: it writes a new set theme value of `setTheme` to the user's preferences which is the theme they set, writes the base theme value of `this.baseTheme` to the set theme. If the set theme value is the follow system option of "system", then the base theme value is written though "nativeTheme.shouldUseDarkColors". A new current set theme value of `currentSetTheme` is also written from the set theme value. This is used in the app though "App.js" and the theme settings through "theme-settings.js".

Setting the theme to the follow system option shows the set theme as "System" and sets the theme to the system theme. When changing the system theme and reopening the app, the set theme still shows as "System" and the theme is the new system theme.

 - OS: Ubuntu
 - OS Version: 21.04
 - FreeTube version: 0.13.0


julianfairfax:
---
Add option to follow system theme
---



**Pull Request Type**
- [x] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/691

**Description**
This adds an option to follow the system theme. More specifically: it writes a new set theme value of `setTheme` to the user's preferences which is the theme they set, writes the base theme value of `this.baseTheme` to the set theme. If the set theme value is the follow system option of "system", then the base theme value is written though "nativeTheme.shouldUseDarkColors". A new current set theme value of `currentSetTheme` is also written from the set theme value. This is used in the app though "App.js" and the theme settings through "theme-settings.js".



Setting the theme to the follow system option shows the set theme as "System" and sets the theme to the system theme. When changing the system theme and reopening the app, the set theme still shows as "System" and the theme is the new system theme.

 - OS: Ubuntu
 - OS Version: 21.04
 - FreeTube version: 0.13.0

julianfairfax:
I'm no longer using this so I'm no longer providing it on my APT repo. I'm keeping this pull request open in case you still want to add the other feature. (I still love the app. I just don't want to use YouTube anymore)

Svallinn:
Just writing this down for clarity and closure's sake
This'll still work, you just need to use the IPC mechanism to communicate with the main process and you can use the `nativeTheme` module there

If anyone picks this up, there's a bunch of IPC calls on the `src/main/index.js` file for you to reference to, I also recommend reading about `ipcMain` and `ipcRenderer` in the Electron docs.

PR #1320, opened 2021-05-23
Escape HTML in chat & comments

### LINKED ISSUE #1307: Comments' text isn't escaped (Security & Privacy Vulnerability)
**Behavior of the program**
YouTube comments contain HTML code is parsed and displayed as an html element

**Expected behavior**
The HTML code should not be parsed and should be shown as plain text.

**To Reproduce**
1. Go to https://www.youtube.com/watch?v=AWzKGOAru-8 on YouTube
2. Go to https://www.youtube.com/watch?v=AWzKGOAru-8 on FreeTube
3. See error

**Screenshots**
YouTube:
image

FreeTube:
image

**Environment Information (please complete the following information):**
 - OS: Windows
 - OS Version:  Windows 10
 - FreeTube version: 13.0 nightly
 - Primary API Used: Local API

**Additional context**
Showing a minion meme and text that is slightly larger isn't super malicious but this security flaw can definitely be exploited. ([XSS ](https://owasp.org/www-community/attacks/xss/) vulnerabilities). This shouldn't be a problem for video descriptions as YouTube blocks angle brackets from being used there.

ChunkyProgrammer:
---
Escape HTML in chat & comments
---

**Important note**

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
closes #1307

**Description**
It replaces angle brackets, quotation marks and ampersands with html char codes.

image

Mostly, invidious api & local api for comments has been tested but live chat has not been tested.
I used this page for testing comments https://www.youtube.com/watch?v=AWzKGOAru-8 & sorted by 'new'

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13

### DIFF
--- a/src/renderer/components/watch-video-comments/watch-video-comments.js
+++ b/src/renderer/components/watch-video-comments/watch-video-comments.js
@@ -7,6 +7,14 @@ import FtTimestampCatcher from '../../components/ft-timestamp-catcher/ft-timesta
 import autolinker from 'autolinker'
 import ytcm from 'yt-comment-scraper'
 
+const SYMBOLS = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+  '"': '&quot;',
+  "'": '&#039;'
+}
+
 export default Vue.extend({
   name: 'WatchVideoComments',
   components: {
@@ -249,7 +257,9 @@ export default Vue.extend({
         if (this.hideCommentLikes) {
           comment.likes = null
         }
-        comment.text = autolinker.link(comment.text)
+        comment.text = autolinker.link(comment.text.replace(/[&<>"']/g, function(index) {
+          return SYMBOLS[index]
+        }))
 
         return comment
       })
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer:
Using a library to sanitize might be a better idea than what I did. Thoughts?

GilgusMaximus:
If the image you provided is the actual look of it, then the PR is really not necessary. I researched a bit in the last few days. Vue does this on it's own. You'd just have to move the comment code out of the v-html element and into brackets. 

Additionally blacklists are not really safe, as you can miss things and there are many encodings possible to solve that.

Additionally we want to avoid the comments looking like this. It looks completely off. So therefore, if I understand the linked library, that is also no solution.

My best solution atm, and I want to point out that this is probably not the best out there, would be to remove all html tags. So basically removing all <...> </> elements in the text, so the text cannot be stylized anymore, but it doesn't look off either.

But I'll read up some more on that

ChunkyProgrammer:
Those are some good points. I'm new to Vue and didn't know about the brackets and I think that is a pretty good option.

About whether to remove the tags or not:
YouTube doesn't remove the tags
image

I can't really think of good use cases for why the tags should remain other than YouTube has them and maybe a coding channel might post html code in their comments(very unlikely as a github link would be a better option)?

I agree that a mix of the brackets and the filtering of <..> </> would be the best option. So, I will close this request.

GilgusMaximus:
Yeah now worries. I didn't know that either. I just googled around and eventually found out about that in one thread somewhere.

ChunkyProgrammer:
> Yeah now worries. I didn't know that either. I just googled around and eventually found out about that in one thread somewhere.

PR #1321, opened 2021-05-23
Translate window titles

### LINKED ISSUE #1310: Window title in English for all locales
**Behavior of the program**
Window title is in English, even though the language is set as `System Default` which is currently in Hungarian.

**Expected behavior**
`About - FreeTube` should be `Névjegy - FreeTube`.

`Subscriptions`, `User Playlists`, `History`, `Settings` also show in English in the window title instead of the `System Default` which is currently `Hungarian`.

**To Reproduce**
1. Change the Language to anything other than English.

**Screenshots**
kép

**Environment Information:**
 - OS: Linux Mint
 - OS Version: Linux Mint 20.1 Cinnamon
 - FreeTube version: 0.13.0 Beta
 - Installation Method: FlatHub
 - Primary API Used: Not sure

Thank you

ChunkyProgrammer:
---
Translate window titles
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
closes #1310 

**Description**
- Setting the document title was moved to the "Views" (Translations could not be done directly in router, router instead passes a key as the title)
- Both Channels & Profile Settings didn't have a good translation string for the title:
   - For channel, I changed it to become the channel's name followed by " - FreeTube" because video's were doing video name followed by " - FreeTube"
   - For Profile Settings, I added a new translation string to en-US and en-GB

Title when viewing Adam Neely's YT channel:
image

Title when viewing "Subscriptions" (French):
image

Title when viewing "About" (French):
image

Title when viewing "History" (French):
image

Title when viewing "Settings" (French):
image

Title when viewing "User Playlists" (French):
image

Change "locale" in settings and navigate to different pages

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13
 
**Additional context**
~~Titles don't get updated when using the "Forward" & "Backwards" navigation (this is the case in the current release of FreeTube though)~~ fixed
A positive consequence of this implementation is that it updates window titles when using forward/backwards navigation buttons which is not the case in the current version of FreeTube


ChunkyProgrammer:
- Go to trending
- Go to a channel
- Click 'Back' button
- Title = "[object Object]"
I think this is caused by 'Channel' in router getting passed to 'Trending' .
This is an issue with the title not being properly updated on forwards/backwards navigation, however which I feel like should be a separate issue

src/renderer/views/About/About.js
```diff
@@ -13,6 +13,7 @@ export default Vue.extend({
     'ft-button': FtButton
   },
   data: function () {
+    document.title = `${this.$t(this.$route.meta.title)} - ${process.env.PRODUCT_NAME}`
```
ChunkyProgrammer, line 16: Changing this.$route.meta.title to "About.About" and repeating for all other cases should fix the window not being properly updated when back button is pressed. Thoughts?

ChunkyProgrammer:
Ok, I am now happy with how this feature turned out. (it is no longer hacky and bad). Let me know if there's anything that should be changed.

src/renderer/router/index.js
```diff
@@ -55,7 +55,7 @@ const router = new Router({
       path: '/settings/profile/edit/:id',
       name: 'editProfile',
       meta: {
-        title: 'Edit Profile',
+        title: 'ProfileEdit',
```
PrestonN, line 58: This needs to be `Profile.Edit Profile`

PrestonN (changes requested):
Looks good to me for the most part. Please make this one change and then I'm good to merge.

src/renderer/router/index.js
ChunkyProgrammer: fixed

PrestonN:
LGTM. Thank you!

Issue #1317, opened 2021-05-23
Remove spellchecking for search


peepopoggers:
Spellchecking is not needed as most things people will search will not be words, but names/nicknames.

Example:
![image](https://user-images.githubusercontent.com/72892531/119257284-79677080-bbbc-11eb-8f7d-b46073a7910a.png)

Can be removed  with 
spellcheck="false" in html element.

Fix:
![image](https://user-images.githubusercontent.com/72892531/119257321-a0be3d80-bbbc-11eb-9c60-5ad02d0a8948.png)

peepopoggers:
btw, not really sure how this works, but would probably be an edit here in `src/renderer/components/top-nav.vue`
here:
![image](https://user-images.githubusercontent.com/72892531/119257453-322daf80-bbbd-11eb-88da-6595bc57afce.png)
with `:spellcheck = "false"` or something
but I am not sure how project is structured

PikachuEXE:
@peepopoggers 
See #1346

peepopoggers:
@PikachuEXE <<<3333

peepopoggers:
if #1346 is accepted, this will be closed.

peepopoggers:
Not pulled yet :L

PikachuEXE:
@peepopoggers 
#1346 merged now

peepopoggers:
:+1:

PR #1346, opened 2021-05-27
Update top nav search input to disable spellcheck

PikachuEXE:
**Pull Request Type**
- [x] Feature Implementation

**Related issue**
#1317 

**Description**
Update top nav search input to disable spellcheck

image

- Enter `joyner` into top nav search input

 - OS: MacOS
 - OS Version: 10.15.7
 - FreeTube version: 13.0.0 (latest on `development` actually)
 
**Additional context**

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -35,6 +35,10 @@ export default Vue.extend({
       type: Boolean,
       default: false
     },
+    spellcheck: {
+      type: Boolean,
+      default: true
+    },
     dataList: {
       type: Array,
       default: () => { return [] }
--- a/src/renderer/components/top-nav/top-nav.vue
+++ b/src/renderer/components/top-nav/top-nav.vue
@@ -61,6 +61,7 @@
           :is-search="true"
           :select-on-focus="true"
           :data-list="searchSuggestionsDataList"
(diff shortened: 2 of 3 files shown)


src/renderer/components/ft-input/ft-input.js
```diff
@@ -35,6 +35,10 @@ export default Vue.extend({
       type: Boolean,
       default: false
     },
+    spellcheck: {
+      type: String,
+      default: ''
+    },
```
Svallinn, line 41: Make this a default `true` `Boolean` instead, `String` makes it look like it can have more than 2 states.
```suggestion
    spellcheck: {
      type: Boolean,
      default: true
    },
```

src/renderer/components/top-nav/top-nav.vue
```diff
@@ -61,6 +61,7 @@
           :is-search="true"
           :select-on-focus="true"
           :data-list="searchSuggestionsDataList"
+          spellcheck="false"
```
Svallinn, line 64: You'll need to make this a dynamic assignment after making it expect a Boolean.
```suggestion
          :spellcheck="false"
```

Svallinn (changes requested):

PikachuEXE:
Updated

PR #1407, opened 2021-06-12
Fix emoji initials in Profile Previews

### LINKED ISSUE #1396: Emojis are replaced with � in Profile Previews
**Behavior of the program**
The profile preview gets replaced with a � if an emoji was used at the start of the profile name.

image

**Expected behavior**
The emoji to be normally displayed in profile previews. (Shown below is how it looks if you manually 'inspect element' and replace � with the emoji.)

image

**To Reproduce**
1. Set a profile name to one which contains an emoji as the beginning character.
2. Observe that the � symbol appears in its place as the profile preview.

**Environment Information:**
```
OS: Microsoft Windows 10 Pro
OS Version: 10.0.19042 Build 19042
FreeTube version: Release 0.13.1 Beta
Installation Method: Windows x64 exe
Primary API Used: Local API
```

**Additional context**
It is possible to temporarily replace the � with the emoji by using 'inspect element' (as done above) but it doesn't persist through sessions.

ChunkyProgrammer:
---
Fix emoji initials in Profile Previews
---

**Important note**

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
closes #1396

**Description**
Modifies profileName.slice(0,1) to use Array.from(profileName)[0] instead

image

image

I tested this by Editing, Adding and viewing a profile

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13.1

**Additional context**

### DIFF
--- a/src/renderer/components/ft-profile-bubble/ft-profile-bubble.js
+++ b/src/renderer/components/ft-profile-bubble/ft-profile-bubble.js
@@ -22,7 +22,12 @@ export default Vue.extend({
   },
   computed: {
     profileInitial: function () {
-      return this.profileName.slice(0, 1).toUpperCase()
+      if (this.profileName !== undefined) {
+        if (this.profileName.length > 0) {
+          return Array.from(this.profileName)[0].toUpperCase()
+        }
+      }
+      return ''
     }
   },
   methods: {
--- a/src/renderer/components/ft-profile-bubble/ft-profile-bubble.js
+++ b/src/renderer/components/ft-profile-bubble/ft-profile-bubble.js
@@ -22,12 +22,7 @@ export default Vue.extend({
   },
   computed: {
     profileInitial: function () {
-      if (this.profileName !== undefined) {
-        if (this.profileName.length > 0) {
-          return Array.from(this.profileName)[0].toUpperCase()
-        }
-      }
-      return ''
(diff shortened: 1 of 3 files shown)


src/renderer/components/ft-profile-bubble/ft-profile-bubble.js
```diff
@@ -22,7 +22,12 @@ export default Vue.extend({
   },
   computed: {
     profileInitial: function () {
-      return this.profileName.slice(0, 1).toUpperCase()
+      if (this.profileName !== undefined) {
+        if (this.profileName.length > 0) {
+          return Array.from(this.profileName)[0].toUpperCase()
+        }
+      }
+      return ''
```
Svallinn, line 30: Simplify this to:
```js
return this?.profileName?.length > 0
      ? Array.from(this.profileName)[0].toUpperCase()
      : ''
```
Do it similarly on the other cases.

Svallinn (changes requested):

src/renderer/components/ft-profile-bubble/ft-profile-bubble.js
ChunkyProgrammer: Didn't realize optional chaining was a thing pretty cool.

Issue #418, opened 2019-11-01
Open video in external player


n1trux:
It would be awesome if you could have a setting and a right-click menu entry where you open a video in an external player from a list of videos.

MPV can play youtube (and other!) videos with youtube-dl.

dojima:
This is probably my most desired feature.  Having the option to have FreeTube simply pass the YouTube URL to a configured player when clicking a video thumbnail would be amazing.

pirate486743186:
yep, that.

PeterAlabaster:
Until this gets implemented on freetube - my workaround on linux is click the drawer icon, click `Copy YouTube Link`, then have a keybind preset for the following:
`mpv "$(xclip -o -selection clipboard)"`

Requirement on debian based distro:
`sudo apt install xclip`

Then just press the keybind combo and video opens in external player

vensko:
More use cases:
- Open video in any browser with specified arguments (incognito, without extensions, and so on).
- Send URL to youtube-dl or a shell script.

trostboot:
I'll add my voice of support for a feature like this. Ideally it would be configurable to pass the applicable URL to the external application of choice with a single click on the video.

I'll spare you the details, but this would be a huge improvement for me and I would love to see this implemented.

palikacska:
Another reason why this feature should be a must have.
https://github.com/FreeTubeApp/FreeTube/issues/912

ghost-420:
This feature is great if you use filters in your media player to make videos look higher resolution than they are, like MPC-HC + madVR.

feynmania:
Hi guys! Just discovered FreeTube and I love it! I prefer playing videos in mpv though so I downloaded the source code and edited it so that if I middle-mouse click a thumbnail or video title I open it in mpv. I've never used js before so my implementation is very very ugly but it works so if anyone else is interested I thought I'd post it:
All I did was open FreeTube/src/renderer/components/ft-list-video/ft-list-video.js and add the following function to the methods:

```js
activateMiddleMouseClickShortcut: function() {
  // I loop through all the links on the page and change
  // their middle-click function. Doing the entire for loop 
  // seems very silly to me but for some reason I couldn't 
  // get the function to work without it.
  let links_on_page = document.getElementsByClassName('ft-list-video');
  for(let i=0; i<links_on_page.length; i++) {
    links_on_page[i].onauxclick = (e,t) => {
      let youtube_id = ''
      // If we clicked on a title target.src will be undefined 
[...]

PR #1409, opened 2021-06-13
Add command line argument mapping for the VLC player

Randshot:
---
Add command line argument mapping for the VLC player
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
#418 

**Description**
This PR adds command line argument mapping for the VLC player.

- Opening videos
- Opening videos at a specified start offset
- Opening videos with different playback rates

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13.1 Beta

**Additional context**
VLC doesn't seem to have support for YouTube playlists out-of-the-box.
While I did add flags for  playlist shuffle/reverse, they are essentially useless, since opening playlists isn't supported.
Just wanted to note this here.

### DIFF
--- a/static/external-player-map.json
+++ b/static/external-player-map.json
@@ -19,5 +19,21 @@
             "playlistShuffle": "--shuffle",
             "playlistLoop": "--loop-playlist"
         }
+    },
+    {
+        "name": "VLC",
+        "value": "vlc",
+        "cmdArguments": {
+            "defaultExecutable": "vlc",
+            "supportsYtdlProtocol": false,
+            "videoUrl": "",
+            "playlistUrl": null,
+            "startOffset": "--start-time=",
+            "playbackRate": "--rate=",
+            "playlistIndex": null,
+            "playlistReverse": null,
+            "playlistShuffle": "--random",
+            "playlistLoop": "--loop"
+        }
     }
 ]


GilgusMaximus:
In theory thanks for adding, but it seems like VLC cannot play videos? 
I added the path to the exe and it opens, but then keeps hanging and does not play at all

Randshot:
Hm, that's weird. It just worked when I tried it with my new VLC install.
Which version are you using?

Svallinn:
Works fine for me, but VLC tends to pick a pretty low quality video stream.
Is there any way to manipulate that behavior?

ChunkyProgrammer:
> Works fine for me, but VLC tends to pick a pretty low quality video stream.
> Is there any way to manipulate that behavior?
Looks like there's a `--preferred-resolution` argument
https://askubuntu.com/questions/796575/how-to-force-command-line-vlc-to-stream-youtube-videos-of-specific-quality

Randshot:
> 
> 
> > Works fine for me, but VLC tends to pick a pretty low quality video stream.
> > Is there any way to manipulate that behavior?
> 
> Looks like there's a `--preferred-resolution` argument
> https://askubuntu.com/questions/796575/how-to-force-command-line-vlc-to-stream-youtube-videos-of-specific-quality

Yeah, `--preferred-resolution` seems to work. Otherwise, you can also set the preferred resolution globally in the VLC settings (like described in the same post above).

Svallinn:
If there's a global setting you can set in the program, this is fine as is then.
I'll let @GilgusMaximus figure out what's wrong on his end and let him merge this.

rootedetc:
@Randshot I would like to add a command line argument for [iina](https://github.com/iina/iina) (native macOS player based on mpv, thus being the prefered choice for macOS users). 

Since I'm new to using Git, should I fork this PR and add my change, or should I fork the main development branch (I don't want to have a conflict with your PR)?

Svallinn:
@rootedetc Fork the project, make changes in your fork's `development` branch (or create a new branch and do it there) and then make a PR to merge it back to the original project's `development` branch.

Randshot:
> 
> 
> @Randshot I would like to add a command line argument for [iina](https://github.com/iina/iina) (native macOS player based on mpv, thus being the prefered choice for macOS users).
> 
> Since I'm new to using Git, should I fork this PR and add my change, or should I fork the main development branch (I don't want to have a conflict with your PR)?

@rootedetc 
According to the [iina wiki page](https://github.com/iina/iina/wiki/MPV-Options-and-Properties), it does support most of `mpv`'s command line arguments, which means you can probably just point FreeTube to the relevant iina executable, and it *should* theoretically work, without having to change anything (haven't tested this though).

rootedetc:
Thank you guys for answering, @Svallinn and @Randshot 

I have been able to get iina working, and it does infact support most of mpv's command line arguments.
The only issue I am facing is iina can only run from third party programs if started with an argument --no-stdin.

Would there be a way to hard-code that argument so the end user doesn't have to use the custom arguments input in the settings? For the life of me, I am not able to hard-code it, my knowledge just isn't good enough (I don't want to give up though).

Randshot:
Hm, hadn't thought about that when implementing the external player feature.
How about adding a mapping for `defaultCustomArguments`, which would be used to populate the custom arguments field, if the user doesn't enter their own custom arguments?

rootedetc:
> Hm, hadn't thought about that when implementing the external player feature.
> 
> How about adding a mapping for `defaultCustomArguments`, which would be used to populate the custom arguments field, if the user doesn't enter their own custom arguments?
> 
> Or we could just add a separate mapping for `playerArguments` (similar to what you did).
> 
> *The latter one would be easier to implement.*
> 
> 
> 
> Tbh, I prefer the first choice, since that gives the user the option to not use those "hard-coded" arguments, for what ever reason they might have.
> 
> 
> 
> Edit:
> 
> I have a working prototype of the first option. Just need to figure out how to display those default custom arguments to the user.

That first option sounds nice since it would give the user the feeling of having full control. However, would that be too much work?

Moreover, in the case of iina specifically, it would be counter productive since the end user would probably remove the default custom argument upon seeing it - making the issue I mentioned appear once again. 

However, maybe there could be a warning displayed to not to disable an argument, depending on what player is selected?

Randshot:
Since the label is tied to the placeholder, I thought about adding the `defaultCustomArguments` to the help tooltip instead, so the user is at least made aware that there are default custom arguments.

Randshot:
This is what I came up with:
image

Randshot:
@rootedetc
Created a PR for the above: #1415

rootedetc:
> This is what I came up with:
> 
> image
> 
> 

It's probably for the best if the end user has the transparency, after all. Quick question, every contributor will need to reference their default custom arguments in the tooltip, if they choose to add one, correct?

I suppose that could be included as a guideline within the [docs](https://docs.freetubeapp.io/development/getting-started/).

rootedetc:
> @rootedetc
> 
> Created a PR for the above: #1415 
> 
> 
> 
> > Edit: throwing out an idea - what if the default custom arguments were displayed within the custom arguments input area, but it was a label (so the user knows it's needed)?
> 
> 
> 
> That's what I initially tried, but like I mentioned, the `placeholder` property is tied to the label it shows above the input field.
> 
> So without changes to the `ft-input` component, that doesn't seem likely.

Never mind, I think the final implementation was the better choice anyway.￼ Thank you for your hard work! :)

GilgusMaximus:
Well I was on 3.0.12, seems like it has some kind of bug. With 3.0.14 it works fine.

GilgusMaximus:
Well then thanks for adding vlc and if any questions / discussions regaring the other PR are coming up, feel free to post them over there :)

Best Regards,
Gilgus

PR #1411, opened 2021-06-14
Updated frame-by-frame function to cover previous edge cases.

Hiers:
---
Updated frame-by-frame function to cover previous edge cases.
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [x ] Feature Implementation

**Description**
Recent changes now allow the distinction between 30fps and 60fps videos for 720p+ qualities. Method was changed to reflect this and catch edge cases that were left behind last time.

 - OS: [Artix Linux]
 - OS Version: [N/A]
 - FreeTube version: [0.13.1]

**Additional context**
The adaptiveFormats object does not specify which quality is being used, and it has a different size from the object that does tell you. To verify which one it was I had to run a for loop that felt unnecessary, I'm all hears for a better solution.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -848,14 +848,20 @@ export default Vue.extend({
 
     framebyframe: function (step) {
       this.player.pause()
-      const qualityHeight = this.useDash ? this.player.qualityLevels()[this.player.qualityLevels().selectedIndex].height : 0
-      let fps
+      const quality = this.useDash ? this.player.qualityLevels()[this.player.qualityLevels().selectedIndex] : 0
+      let fps = 30
       // Non-Dash formats are 30fps only
-      if (qualityHeight >= 480 && this.maxFramerate === 60) {
-        fps = 60
-      } else {
-        fps = 30
+      if (quality.height >= 480 && this.maxFramerate === 60) {
+        for (let i = 0; i < this.adaptiveFormats.length; i++) {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @Hiers pinging u here in case u didnt get the message on matrix. Can i add the label 'waiting for review'?

Hiers:
You can, this has been ready for a while now.

efb4f5ff-1298-471a-8973-3d47447115dc:
You can, this has been ready for a while now.

@Hiers i added the label 'waiting for review' to this to make sure everyone sees that.

jasonhenriquez:
@Hiers Can you give an example explaining what new behavior this PR provides as compared to before? I do not understand.

Hiers:
@jasonhenriquez This function as it stands right now in the actual code can't distinguish a 1080p video that's running at 60fps from a 1080p video running at 30fps, which means on a 30fps 1080p it won't actually advance 1 frame, but 2 frames. When Preston changed how the different qualities are fetched to allow things like HDR and qualities above 1080p, it changed the backend in such a way that now that distinction is possible.

To answer some code questions I anticipate you might have:

1. `this.maxFramerate` is effectively checking if the video is using DASH or not
2. The for loop you see is there because the `this.player.qualityLevels()` object (the one that tells you the quality you selected) and the `this.adaptiveFormats` object (with the information about the fps values of the qualities) have different sizes, so you have to compare until you find it.

Hiers:
I changed the default value to an empty object. It should never actually be evaluated itself since if it is empty the max framerate will always be 30.

PR #1415, opened 2021-06-14
Add support for the 'defaultCustomArguments' external player mapping field

Randshot:
---
Add support for the 'defaultCustomArguments' external player mapping field
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
While there isn't a directly related issue, this PR was created in response to a discussion in PR #1409, which pointed out the need for being able to pass default arguments to specific external players (in this case iina).

**Description**
This PR adds support for another external player argument mapping: `defaultCustomArguments`
Any arguments defined via this field (arguments should be separated by semicolons `;`, like user-defined custom arguments), will be only used, if the user hasn't defined any custom arguments themselves. I decided on this approach, since that gives the user more control on what is passed to the external player. Additionally, if defined, the default custom arguments will be displayed in the help tooltip for the custom arguments text input in the external player settings (see screenshot below).

image

- Tested the new `defaultCustomArguments` field

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13.1 Beta

### DIFF
--- a/src/renderer/components/external-player-settings/external-player-settings.js
+++ b/src/renderer/components/external-player-settings/external-player-settings.js
@@ -40,7 +40,21 @@ export default Vue.extend({
     },
     externalPlayerCustomArgs: function () {
       return this.$store.getters.getExternalPlayerCustomArgs
-    }
+    },
+    externalPlayerCustomArgsTooltip: function () {
+      const tooltip = this.$t('Tooltips.External Player Settings.Custom External Player Arguments')
+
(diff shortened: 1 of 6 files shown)


Hiers (approved):
Seems good to me, unfortunate that people using iina will have to write out --no-stdin to use custom arguments, but it's fair. I agree that if in the future we add defaults to other players too, people should have the option to not use those defaults if they want to use their own custom arguments, even if in the case of iina the default is always necessary.

Issue #1293, opened 2021-05-18
Dont show text on the sidebar in compact mode and increase size of channel names 


MaxiMii05:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
1. The sidebar looks horrendous in compact mode. The text that is displayed next to the icons in large mode is just displayed smaller below the icons in compact mode. This reduces space for the icons which in turn leads to small icons with hardly readable text. This is especially bad in some languages like German, where long words get cut off by the scroll bar.
2. In large mode, channel names can be very small if some of the subscribed channels have long names. It makes the names hard to read. The font size is also inconsistent with the text next to the icons on top.

**Describe the solution you'd like to see implemented**
1. I suggest completely hiding text on the sidebar in compact mode as it is unnecessary. This allows the icons to be much larger which in turn improves the user experience.
2. Instead of adjusting all channel names because of one long name, I suggest using the same font size as for everything on top. Long names should be cut off as it is not necessary to read the entire name of a channel. Maybe, the full channel name could be shown as a tooltip. 

**Describe alternatives you've considered**
-

**Screenshots**
Bad 5 minute mock-ups:
1. 
![small](https://user-images.githubusercontent.com/77237423/118669987-c080ea80-b7e5-11eb-8565-adc7e62deef6.jpg)
2. 
![big](https://user-images.githubusercontent.com/77237423/118670015-c676cb80-b7e5-11eb-91d7-367da1a83f2f.jpg)

**Additional context**
I understand that it may be helpful to see the text of the icons. But as there are only 7 buttons with very descriptive icons, the text is not needed all the time.

GilgusMaximus:
I agree that the text does not look very nice. But, I would add this as an option to remove the text in this mode. There might be users on something like a Pinephone, where when they first download now text is displayed. This is then not very user friendly. 

Text sizes are not fully 100%, so changing the CSS there would solve that issue

JInfantesC:
Hi, i would like to collaborate in this issue. 

After fiddling around inside the side-nav i propose something like this:

![FreeTube_#1293_title](https://user-images.githubusercontent.com/7262829/119223448-0a6d1780-bae9-11eb-920c-d8fa0771e5ea.png)
Default font size for channel and on long names, an ellipsis will be visible. The HTML title attribute displays the full name on hover.

[...]

PR #1462, opened 2021-06-27
Sidebar compact issue1293

JInfantesC:
---
Issue #1293. Sidebar text labels and sizes
---
**Pull Request Type**
- [ ] Bugfix
- [X] Feature Implementation

**Related issue**
https://github.com/FreeTubeApp/FreeTube/issues/1293

**Description**
This commits revise the size of the icons in the sidebar and how the text labels displays using ellipsis on long texts, making use of CSS properties and the HTML title atributte.

Freetube_1293_ex1
Freetube_1293_ex2

As a CSS and HTML modification without touching anything more complex or critical, code changes are easily understandable. No test required

 - OS: Ubuntu Linux
 - OS Version: 20.04
 - FreeTube version: 0.13.0

### DIFF
--- a/src/renderer/components/side-nav/side-nav.vue
+++ b/src/renderer/components/side-nav/side-nav.vue
@@ -6,16 +6,21 @@
   >
     <div class="inner">
       <div
-        class="navOption topNavOption mobileShow"
+        class="navOption topNavOption mobileShow "
         role="button"
         tabindex="0"
+        :title="$t('Subscriptions.Subscriptions')"
         @click="navigate('subscriptions')"
       >
-        <font-awesome-icon
-          icon="rss"
-          class="navIcon"
-          fixed-width
-        />
+        <div
+          class="thumbnailContainer"
+        >
+          <font-awesome-icon
+            icon="rss"
+            class="navIcon"
+            fixed-width
+          />
+        </div>
         <p class="navLabel">
           {{ $t("Subscriptions.Subscriptions") }}
         </p>
@@ -25,14 +30,19 @@
         class="navOption mobileHidden"
         role="button"
         tabindex="0"
+        :title="$t('Trending')"
         @click="navigate('trending')"
         @keypress="navigate('trending')"
       >
-        <font-awesome-icon
-          icon="fire"
-          class="navIcon"
-          fixed-width
-        />
+        <div
+          class="thumbnailContainer"
+        >
+          <font-awesome-icon
+            icon="fire"
+            class="navIcon"
+            fixed-width
+          />
+        </div>
         <p class="navLabel">
           {{ $t("Trending") }}
         </p>
@@ -42,14 +52,19 @@
         class="navOption mobileHidden"
         role="button"
         tabindex="0"
+        :title="$t('Most Popular')"
         @click="navigate('popular')"
         @keypress="navigate('popular')"
       >
-        <font-awesome-icon
-          icon="users"
-          class="navIcon"
-          fixed-width
-        />
+        <div
+          class="thumbnailContainer"
+        >
+          <font-awesome-icon
+            icon="users"
+            class="navIcon"
+            fixed-width
+          />
+        </div>
         <p class="navLabel">
           {{ $t("Most Popular") }}
         </p>
@@ -59,14 +74,19 @@
         class="navOption mobileShow"
         role="button"
         tabindex="0"
+        :title="$t('Playlists')"
         @click="navigate('userplaylists')"
         @keypress="navigate('userplaylists')"
       >
-        <font-awesome-icon
-          icon="bookmark"
-          class="navIcon"
-          fixed-width
-        />
+        <div
+          class="thumbnailContainer"
+        >
+          <font-awesome-icon
+            icon="bookmark"
+            class="navIcon"
+            fixed-width
+          />
+        </div>
         <p class="navLabel">
           {{ $t("Playlists") }}
         </p>
@@ -78,14 +98,19 @@
         class="navOption mobileShow"
         role="button"
         tabindex="0"
+        :title="$t('History.History')"
         @click="navigate('history')"
         @keypress="navigate('history')"
       >
-        <font-awesome-icon
-          icon="history"
-          class="navIcon"
-          fixed-width
-        />
+        <div
+          class="thumbnailContainer"
+        >
+          <font-awesome-icon
+            icon="history"
+            class="navIcon"
+            fixed-width
+          />
+        </div>
         <p class="navLabel">
           {{ $t("History.History") }}
         </p>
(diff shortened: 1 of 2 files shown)


peepopoggers:
@JInfantesC link the pull request pls.

We were discussing this on the element matrix. I really think this is a good idea for compact mode.

JInfantesC:
Hi @peepopoggers, 

I'm afraid i don't understand what you mean by "link the pull request". Github does not allow me to use the "Linked issues" tool if that's what you mean.

Please, let me know what should i do. Thank you!

ChunkyProgrammer:
I think they want you to type "closes #1293" instead of "#1293" for the related issue of your description.

JInfantesC:
Thanks! Done!

src/renderer/components/side-nav/side-nav.vue
```diff
@@ -25,14 +30,19 @@
         class="navOption mobileHidden"
         role="button"
         tabindex="0"
+        :title="$t('Trending')"
```
ChunkyProgrammer, line 33: This should be changed to `$t('Trending.Trending')` before merging (#1638 changed this)

ChunkyProgrammer:
Hi, I just tested your PR and it works fine for the most part.
It looks like this when using the mobile view though (I think "scr/renderer/components/side-nav-more-options" needs to be updated):
image

This is what the mobile view looks like without using this PR:
image

JInfantesC:
I've revised the more options component:
Freetube more-options

Thank you!

jasonhenriquez:
Getting rid of text labels & relying on the title attribute to provide the information is generally discouraged for accessibility reasons. I in fact believe that FreeTube should be moving toward the opposite direction. 
https://www.tpgi.com/html5-accessibility-chops-title-attribute-use-and-abuse/
https://www.mediacurrent.com/blog/dont-rely-title-attribute-accessibility-seo/
https://www.w3.org/WAI/tutorials/forms/labels/

PikachuEXE:
I agree with @jasonhenriquez.
I have seen UI with mostly icons only and it's really annoying and confusing on some of them.
I rather see text only buttons without the icons.
On PC I can still hover to see the title but annoying.
On Mobile there is no way to "hover".

efb4f5ff-1298-471a-8973-3d47447115dc:
Does this mean that we have to close the issue related to this?

ChunkyProgrammer:
https://github.com/FreeTubeApp/FreeTube/issues/1293#issuecomment-844996542
Should probably be an optional setting that is disabled by default?

PikachuEXE:
@ChunkyProgrammer 
It's that's how it's implemented I have no problem with that.
This PR is not implemented that way currently.

efb4f5ff-1298-471a-8973-3d47447115dc:
> https://github.com/FreeTubeApp/FreeTube/issues/1293#issuecomment-844996542
> Should probably be an optional setting that is disabled by default?

And maybe add something like a warning that if u enable it that it will not improve the UX for people with issues stated in comments above.

peepopoggers:
I agree that this should be an opt-in feature, but it is still cool for those who want a more minimal UI but have used FT enough to know their way round.

JInfantesC:
Thanks everyone for the comments!

I've encouraged myself to try this feature as an opt-in setting as discussed here.
The setting "Hide Compact Side Bar text" is available in settings/Distraction Free Settings. 

A few screenshots:
sidenav_setting_2
sidenav_setting_3

sidenav_setting_4
sidenav_setting_5

Because this pull request is now more complex than initially defined, let me know if the changes made are going in the right direction.

PD. The `settings.js` instructions are wonderful and very helpful :)

Thank you!

jasonhenriquez:
@JInfantesC I appreciate your initiative in fixing this -- there are three things you will have to fix first.
1. Use a template string in en-us.yaml for the name of this control. This will allow it to be translatable into other languages.
2. Change the name to "Show Labels in Side Bar" throughout your code & as the value of the template string. As a result of this naming change, you will have to change the default value to "true" and reverse your logic for whether it should be displayed. The reason for this change is that the meaning of the original text is not very clear.
3. Move the toggle from Distraction Free Settings to Theme Settings. The reason for this change is that Distraction Free Settings are meant for hiding content that will distract someone. Labels are not distracting and more just explain the meaning of something.

After these changes, there should not be many additional changes left to make. Again, thank you so much for your good work.

ChunkyProgrammer:
> @JInfantesC I appreciate your initiative in fixing this -- there are three things you will have to fix first.
> 
> 1. Use a template string in en-us.yaml for the name of this control. This will allow it to be translatable into other languages.
> 2. Change the name to "Show Labels in Side Bar" throughout your code & as the value of the template string. As a result of this naming change, you will have to change the default value to "true" and reverse your logic for whether it should be displayed. The reason for this change is that the meaning of the original text is not very clear.
> 3. Move the toggle from Distraction Free Settings to Theme Settings. The reason for this change is that Distraction Free Settings are meant for hiding content that will distract someone. Labels are not distracting and more just explain the meaning of something. You will also have to put your template string in en-us.yaml under the Theme Settings group (seen on line 155 of en-us.yaml). Your label in the Vue code with look like this: `:label="$t('Settings.Theme Settings.Show Labels in Side Bar')"`
> 
> After these changes, there should not be many additional changes left to make. Again, thank you so much for your good work.

Should it not be "Hide Labels in Side Bar" iirc all the distraction free settings default to false so it'd be a bit odd having one default to true

peepopoggers:
I agree with Chunky here, it should be named "Hide labels in side bar" -> sounds better.

jasonhenriquez:
> Should it not be "Hide Labels in Side Bar" iirc all the distraction free settings default to false so it'd be a bit odd having one default to true

> > 3. Move the toggle from Distraction Free Settings to Theme Settings.

JInfantesC:
Thank you all for your comments. I've moved the switch to theme-settings and renamed it "Hide labels in Side Bar"

Please, take a look and do not hesitate to change anything if needed :)

src/renderer/components/theme-settings/theme-settings.vue
```diff
@@ -21,6 +21,11 @@
         :default-value="disableSmoothScrollingToggleValue"
         @change="handleRestartPrompt"
       />
+      <ft-toggle-switch
+        :label="$t('Settings.Theme Settings.Hide labels in Side Bar')"
```
peepopoggers, line 25: make it Hide Labels in Side Bar

peepopoggers:
Some feedback I want to give as I inspected your changes and tested the PR:

- In my opinion the new default (icon with text) icons are too big. I feel like they could be made a little bit smaller. (In no-label mode you can keep the icon sizes as you have changed them to be)
- You should also make the side bar a bit less wide when you hide the labels. That would make it more compact.

Other than that this is a great new addition

efb4f5ff-1298-471a-8973-3d47447115dc:
> In my opinion the new default (icon with text) icons are too big. I feel like they could be made a little bit smaller. (In no-label mode you can keep the icon sizes as you have changed them to be)

> You should also make the side bar a bit less wide when you hide the labels. That would make it more compact.

In my opinion both do not have to be addressed to me looks great the way it is.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

peepopoggers:
I would like some more feedback on my ideas before this gets merged

JInfantesC:
Thanks for the feedback @peepopoggers.

I've implemented those changes, the icons will keep the current size by default. When the hide label option is active those icons will get bigger.

And when the navbar is closed and the hide label option is active, the navbar width is set to 60px instead of 80px.

i've user computed properties and CSS selectors for these changes.

src/renderer/components/theme-settings/theme-settings.vue
```diff
@@ -21,6 +21,11 @@
         :default-value="disableSmoothScrollingToggleValue"
         @change="handleRestartPrompt"
       />
+      <ft-toggle-switch
+        :label="$t('Settings.Theme Settings.Hide Labels in Side Bar')"
```
peepopoggers, line 25: One final suggestion: Change to:
Hide Side Bar Labels

peepopoggers:
LGTM just 3 things

JInfantesC:
Done!
Also, i've fixed a getter on theme-settings.js.

peepopoggers (approved):
LGTM

ChunkyProgrammer (approved):
LGTM!

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

PR #1482, opened 2021-06-29
Update colours (search results & select boxes) to match theme

### LINKED ISSUE #1466: Dropdown boxes in setting must match theme setting
At this moment the dropdown boxes are white even if another theme is enabled like dark theme.

Match the color to the theme.

None.

**Video**

https://user-images.githubusercontent.com/73130443/123651322-e3f27880-d81a-11eb-8d94-fe9c2d2bb4b3.mp4

**Additional context**
None.

### LINKED ISSUE #1467: Searchbox must match theme settings
At this moment the searchbox is white even if another theme is enabled like dark theme.

Match theme setting.

None.

**Video**

https://user-images.githubusercontent.com/73130443/123651676-35026c80-d81b-11eb-9ae3-25b90f72e881.mp4

**Additional context**
None.

ChunkyProgrammer:
---
Update colours(search results & select boxes) to match theme
---

**Important note**

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
closes #1467 
closes #1466

**Description**
Updated some css to use variables instead of hardcoded values

## Search:
Light
image

Dark
image

Black
image

Match Top Bar With Main Color
image

## Select Boxes: (Hover color can't change)
Light
image

Dark
image
Black
image

Yes

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13.2


ChunkyProgrammer:
@efb4f5ff-1298-471a-8973-3d47447115dc thoughts on the screenshots?

efb4f5ff-1298-471a-8973-3d47447115dc:
Looks good imo.

Just to be sure i ping @dragosnfy

dragosnfy:
I agree, looks good, but I usually use the same background color for the select/search window as the select/search input so that the amount of colors used is reduced and a css variable can be used for all elements (you did that for half of the screenshots). I see you added borders as well, which probably wouldn't need to be added if the window matches the input's bg and I'd also consider something like `box-shadow: 0 0 20px #0003;` (just an example, could be something else) to kinda have a border but in a more subtle & modern way that creates an almost 3d separation between layers, especially on light mode, on dark mode the opacity of the shadow is usually higher if you want it to be as visible as on light mode.

And for the hover on select boxes I like that you mentioned that it can't be changed, so that people won't ask haha! Cuz that's depending on the theme of the os. In my case it's green, but for most will stay blue. The only option I'm aware of would be to create a "manual select" made out of html elements that are easier to style (like li, div etc).

ChunkyProgrammer:
> I agree, looks good, but I usually use the same background color for the select/search window as the select/search input so that the amount of colors used is reduced and a css variable can be used for all elements (you did that for half of the screenshots). I see you added borders as well, which probably wouldn't need to be added if the window matches the input's bg and I'd also consider something like `box-shadow: 0 0 20px #0003;` (just an example, could be something else) to kinda have a border but in a more subtle & modern way that creates an almost 3d separation between layers, especially on light mode, on dark mode the opacity of the shadow is usually higher if you want it to be as visible as on light mode.
> 
> And for the hover on select boxes I like that you mentioned that it can't be changed, so that people won't ask haha! Cuz that's depending on the theme of the os. In my case it's green, but for most will stay blue. The only option I'm aware of would be to create a "manual select" made out of html elements that are easier to style (like li, div etc).

Changing the background color to match the search definitely looks nice
image

image

image

It's not as consistent when using "Match Top Bar with Main Color", however. (The search box's background color is somewhat transparent).

image

(It also effects the background color depending on what's under it)
image

 I am curious as to your thoughts on this. I can easily make it so the background color matches a different variable for when main color is used for search or I can keep it like this.

dragosnfy:
I agree 10000%. I personally don't like the colored top bar... it looks kinda outdated and brings focus to the bar instead of the video or whatever people wanna focus on. Plus creates this issue that you just mentioned haha. So I wouldn't have that as an option in the first place, but if it's gonna stay, then either a new color variable could be introduced or you could do the same thing you mentioned before my comment

peepopoggers:
i have to say, a gradient does look :fire: but it isnt intended ;{

peepopoggers:
also +1 i rate this

jasonhenriquez:
Looks great! Two points:
1. What about the dropdown lists on ft-profiile-selector and ft-list-dropdown? 
2. This isn't your own doing, but If you do wish to address the latter, I just discovered that ft-list-dropdown list items don't change appearance on hover with black theme. I think it would make sense to knock that out here as well.

peepopoggers:
@ChunkyProgrammer The comment jason made is the only thing that I would like for it to be verified.

ChunkyProgrammer:
@jasonhenriquez I believe I addressed the points you've made
ft-list-dropdown:
image

image

jasonhenriquez (approved):
LGTM!

PikachuEXE (commented):
> Match Top Bar With Main Color

I actually dislike this change on
Having the search text suggestion dropdown to have the same color as top bar
I am using this setting and it was looking fine before this PR
Now it looks red with the same setting and becomes very hard to read

jasonhenriquez:
In fairness to @ChunkyProgrammer, this insufficient color contrast issue is not their fault. This problem already exists with, for example, ft-buttons when the secondary theme is changed to some of these colors. However, I do think that you should nip this problem in the bud in this PR, given the relevance of the subject matter & how the PR will make this accessibility problem much more salient.

PikachuEXE:
All I care is to avoid introducing a bigger issue when fixing something.

But "Match Top Bar With Main Color" should not make UI looks worse than now.
I think we can exclude the search bar for now, or somehow make it follow the "base theme color" instead of "primary theme color".

ChunkyProgrammer:
it now uses the base theme color instead of forcing primary color

jasonhenriquez (approved):
LGTM

PikachuEXE (approved):
The issue I found is fixed
LGTM in other cases

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
👍👍

peepopoggers:
great improvement

jasonhenriquez:
Dracula search result hover color will have to be changed in a later PR -- but very good improvement all around nonetheless.

PR #1502, opened 2021-07-06
Don't blind dark theme users (Dark loading screen)

### LINKED ISSUE #1107: Dark background when starting the app
When starting the app there's a white flash before goin in the main menu of freetube

Remove the whit flash

**Screenshots**
Sometimes people have screenshots from mock-ups or other projects. Add them here.

**Additional context**
Manjaro KDE fully updated
Freetube appimage

output

peepopoggers:
It is better to start the temporary background dark so it doesnt flashbang dark theme users late at night.
Just happened to me :cry: 
This makes the loading background color the normal dark theme background color.

---
Dark Introduction when loading
---

**Pull Request Type**
- [x] Bugfix
- [x] Feature Implementation

**Related issue**
closes #1107

**Description**
Changes loading background color to dark.

Changes Color


efb4f5ff-1298-471a-8973-3d47447115dc:
Is this only for dark theme?

Or both dark and night theme?

peepopoggers:
I don't think theme variables can be called from the electron js file. a dark theme is miles better than light for loading in any case. most users use dark so black is not overly important

dragosnfy:
I agree with the idea, but I think it should be something like hiding the entire body content right before the page loads, then when it does, depending on the dark/light mode it will change dynamically. Here's an example:

Before Page Loads
```
html,body{background:#555;}
@keyframes pageload { 0% { opacity: 0; } 100% { opacity: 1; } }
body{animation: pageload 0.1s;}
```
After Page Loads

`html,body{background:var(---light-&-dark-background-color-for-this-page-)};`

On the current implementation what could be targeted is `.dark`, `.light` and `.black`

peepopoggers:
Dragos, I don't think you understand, no css is used to set loading color, just a hex code in the electron js file inside browserwindow backgroundclor

dragosnfy:
Alright, I get you now. I still think that there could be a way to make that dynamic, not sure how tho. That css for setting the color and opacity was from a platform I built in the past, none of it is in freetube. My point was the idea, not the code necessarily. JS can inject css too or change values like `yourelement.style.background = 'red';` but there's probably a better way to do it. Someone more skilled than me would know.

peepopoggers:
look at the commit so you get an idea of what changed

dragosnfy:
I did lol... before as well... you changed the color, it's one small change

efb4f5ff-1298-471a-8973-3d47447115dc:
btw u have to add the word "closes" before the issue number to link the issue.

Before: #1107 
After: closes #1107

dragosnfy:
I think that here you commented on the wrong one lol. But I added what you said.

efb4f5ff-1298-471a-8973-3d47447115dc:
no i didnt haha. I was being a bot and adding this line to almost every PR that didn't link the issue properly.

dragosnfy:
ohhhhh my bad. I forgot it was @peepopoggers who created this pr for a second, so I was confused lmao

peepopoggers:
i don't think you can inject the electron creation values to the css values. I do not know this project and electron's capabilities enough to know if this is possible (or needed, effort-wise) but for now this change helps alot.

dragosnfy:
It does, I feel ya. If you wanna target the theme with js, then there's a localstorage variable called `baseTheme` which can be `black`, `light` or `dark`. And with css (and js as well), the `<body>` will have a class called either `.dark`, `.black` or `.light`.

I haven't checked the order in which css and js load, cuz it could be that the some things load after and before that line, but when I tested your version a few seconds ago it was opening with the dark background, but then it was light for 1s... that's freetube figuring out what theme is set (cuz light is the default while it recognizes your localstorage variable's value)... and then went dark again, which is normal cuz I was on dark... which means your change works, but only 1/3 times. The other time it doesn't work is if I refresh the app, then it spends 1 second again in light mode figuring out what theme I have. Although to be fair, I don't think that most users refresh the app. So then your change works 1/2 times haha

PrestonN:
I think the best solution would be to have the background colour be set based on which base theme is set. `index.js` should have access to the DB so it should be able to access which theme is set and can change the background colour accordingly. I'd like to see this method implemented instead so that the app starts with whatever theme the user has set.

peepopoggers:
@PrestonN What about the issue mentioned by dragos? Might have to look into that. The way to make the bg color not inconsistent would be to load the color when app is not yet open

PrestonN:
Yes, My solution addresses that.

`index.js` is what runs before the window is loaded, so this is the appropriate location for loading the theme settings and setting the background colour accordingly. This file already accesses the database in a few areas so those could be used as a reference for knowing what theme the user has set and updating the background colour.

adityab98:
Thanks for this. Was about to submit one myself. The app always blinds me when I open it haha.

peepopoggers:
I think light being theme loading color needs to change too.

peepopoggers:
The reason this happens is due to two issues:
Electron index.js BrowserWindow background color being white.
Light theme as placeholder when theme is not loaded in yet.

peepopoggers:
Updated and there is no white flash for me anymore. I tried to use localstorage in index.js but FreeTube would not launch with any errors, so I used the default dark mode for the beginning flash. The main change is moving the checkThemeSettings() to before the other functions so that it is not delayed.

PikachuEXE:
I still see a flash when using `npm run dev` but that might be Electron issue

I don't see any other fixes can be done besides the [`BrowserWindow#backgroundColor`](https://www.electronjs.org/docs/api/browser-window#setting-the-backgroundcolor-property)

PikachuEXE:
I was wrong
I am able to fix the final flash via the following changes

```diff
diff --git a/src/index.ejs b/src/index.ejs
index 7cb00ad6..11d81f7d 100644
--- a/src/index.ejs
+++ b/src/index.ejs
@@ -16,7 +16,7 @@
     <% } %>
   </head>
 
-  <body class="light mainRed secBlue">
+  <body class="dark mainRed secBlue">
     <div id="app"></div>
     <!-- Set `__static` path to static files in production -->
     <script>
diff --git a/src/renderer/App.js b/src/renderer/App.js
index 46787903..504782cc 100644
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -169,7 +169,7 @@ export default Vue.extend({
       let secColor = localStorage.getItem('secColor')
 
       if (baseTheme === null) {
-        baseTheme = 'light'
+        baseTheme = 'dark'
       }
 
       if (mainColor === null) {
```

peepopoggers:
We might need to discuss if having dark mode as the default option is good. What do you think? I think dark mode has basically no downsides and is better on eyesight.

peepopoggers:
@PikachuEXE when looking further into it, the change you suggested does not make any changes in consistency, at least for me. It flashes for a very little time around half the time with the current changes.

PikachuEXE:
@peepopoggers 
Have you tried using `npm run build` and running the built app?

peepopoggers:
I will document the flashing difference with dev and without dev if any later today

peepopoggers:
With that change it looks consistent when built but it needs testing

PikachuEXE:
With this change the app flash issue has become at least less noticeable.
I would like see this to be included in the next release if possible.

PikachuEXE (approved):

PikachuEXE:
@peepopoggers 
Hey I just spotted that
https://github.com/peepopoggers/FreeTube/blob/patch-1/src/index.ejs#L19
still has `light mainRed secBlue` instead of `dark mainRed secBlue` like my patch suggested
Is this intended?

PikachuEXE (dismissed):
One change left for https://github.com/FreeTubeApp/FreeTube/blob/development/src/index.ejs#L19 to fix the flash

`light mainRed secBlue`

https://user-images.githubusercontent.com/1018543/135240848-c9ebb22a-531a-4058-bb05-2afead417c3d.mov

`dark mainRed secBlue`

https://user-images.githubusercontent.com/1018543/135240890-86d83f87-9c48-4092-ad56-6637a990e136.mov

peepopoggers:
@PikachuEXE I did not see that change! My mistake. Will change promptly.

PikachuEXE (approved):
Tested via `npm run debug`

https://user-images.githubusercontent.com/1018543/135375142-738a3ab5-05b8-43bc-956b-a87e49663678.mov

ChunkyProgrammer (approved):

PikachuEXE:
1 more to go!

peepopoggers:
@efb4f5ff-1298-471a-8973-3d47447115dc Could you test this :)

efb4f5ff-1298-471a-8973-3d47447115dc:
already did waiting for approval to merge this

PikachuEXE:

image

Issue #1542, opened 2021-07-24
How to import subscriptions?


ZormeinYT:
So i exported my file from Youtube, but FrreTube accepts only .json files but i have a .csv file. Is this possible to import?

efb4f5ff-1298-471a-8973-3d47447115dc:
Did u read this?


https://docs.freetubeapp.io/usage/importing-subscriptions/

ZormeinYT:
Yes i exported and its a .csv file

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm do u have newpipe?

If u do maybe its faster to do an import of the newpipe file.

ZormeinYT:
i do okay i will try that

efb4f5ff-1298-471a-8973-3d47447115dc:
Lemme know if it worked out. If u were able to successfully import it please close this issue.

ZormeinYT:
On newpipe it laso requires a .json. Maybe i have too many subscritions

ChunkyProgrammer:
I think YouTube updated the way subscriptions are exported and now exports a csv instead of a json

laundmo:
getting a CSV too, made sure to select json everywhere i could.

efb4f5ff-1298-471a-8973-3d47447115dc:
PR for this has to be merged, after that u guys can  import ur subscriptions.

https://github.com/FreeTubeApp/FreeTube/pull/1543

JeeVee11:
Until the issue is fixed, for anyone still running into this:
https://www.reddit.com/r/NewPipe/comments/oprw6d/google_takeouts_exporting_in_csv_format/h694cxs?utm_source=share&utm_medium=web2x&context=3

With this guide you can download your youtube subscribtion .csv file and convert it into a Newpipe .json one, and then import that.
Also an automated tool from the same thread: http://jsbin.com/pegiloviza/2

efb4f5ff-1298-471a-8973-3d47447115dc:
I have some good news for u guys. In this commit we made it possible for u to import ur subscriptions. All u need to do is download the nightly build.

https://github.com/FreeTubeApp/FreeTube/commit/044e5bf907ddd60e0d02cdd6b007e0a90ec41b3b

PR #1543, opened 2021-07-25
[Feature] Add support for importing/exporting csv YouTube subscriptions + improve speed of reimporting subscriptions

ChunkyProgrammer:
---
Add support for importing/exporting csv YouTube subscriptions
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
#1542

**Description**
This allows FreeTube to import/export new YouTube subscriptions (csv file instead of json now).
This PR will also improve the speed of 'syncing'/reimporting with opml (yt) and newpipe by only making the web requests if you are not subbed to the channel already. Thjis can also potentially prevent some rate limiting

#### Importing CSV
Yes. I imported a csv & exported one, unsubbed from a channel then reimported the exported one and it worked fine.

#### Optimizing reimporting
I exported my subscriptions as "newpipe" & "opml" files. Unsubbed from a few channels then reimported. Speed was a lot quicker.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13.2 (latest dev)

**Additional context**
Additional requests need to be made to get the thumbnail of a channel (importing)

When exporting from YT, headers change based on the language that's exporting them, this only exports the English headers


src/renderer/components/data-settings/data-settings.js
```diff
@@ -228,6 +234,77 @@ export default Vue.extend({
...
+
+          if (typeof channelInfo.author !== 'undefined') {
+            const subscription = {
+              id: channelId,
+              name: channelInfo.author,
+              thumbnail: channelInfo.authorThumbnails[1].url
+            }
+
+            const subExists = primaryProfile.subscriptions.findIndex((sub) => {
+              return sub.id === subscription.id || sub.name === subscription.name
```
ChunkyProgrammer, line 276: I copied this line from the other imports; is there a reason why there's a comparison of the subscription names? (Couldn't someone be subbed to different channels with the same name?). If that's the case (and it can be removed) then should "subExists" be checked before channelInfo is set to prevent unneeded requests(this goes for the other imports)?

ChunkyProgrammer: I added this to the PR, let me know if I should remove it

Techngro:
So is this a solution to allow importing of CSV files? I assume this will be implemented in next update, but is there a workaround right now? I just started using Freetube and I have a lot of subscriptions I would like to import. Don't want to do them manually.

ChunkyProgrammer:
> So is this a solution to allow importing of CSV files? I assume this will be implemented in next update, but is there a workaround right now? I just started using Freetube and I have a lot of subscriptions I would like to import. Don't want to do them manually.

You can use this as a workaround:
page: https://chunkyprogrammer.github.io/newpipe-csv-json-fixer/
source code: https://github.com/ChunkyProgrammer/newpipe-csv-json-fixer

jasonhenriquez (approved):
Tested everything -- it works perfectly. Exquisite job.

cloudrac3r:
Please don't request reviews from me without consulting first.

jasonhenriquez:
Understood, thank you for letting me know.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/data-settings/data-settings.js
```diff
@@ -228,6 +234,75 @@ export default Vue.extend({
       this.handleFreetubeImportFile(filePath)
     },
 
+    handleYoutubeCsvImportFile: function(filePath) { // first row = header, last row = empty
```
PikachuEXE, line 237: I wonder if this comment should be placed just before `for (let i = 1; i < (youtubeSubscriptions.length - 1); i++)` 
As it would explain the starting value being `1` not `0`

PikachuEXE (approved):
Just one comment about the position of a code comment
But I still approves it

src/renderer/components/data-settings/data-settings.js
ChunkyProgrammer: I agree, that comment should be moved to be before the for loop.

PR #1579, opened 2021-08-12
Color clues for history navigation arrows

### LINKED ISSUE #1058: [User Interface] arrow colour clue for history
Navigating away to another page, the "forward" or right-arrow should be pale and not change with mouse pointer hover, to indicate that I am at the head of the history.

This is current standard browser UI.

Change also has to aid those that are colour blind and those with vision impairment.

jasonhenriquez:
---
Color clues for history navigation arrows
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
closes #1058 

**Description**
Implements the standard browser behavior of graying and disabling the history navigation arrows as appropriate.

**Video**
https://user-images.githubusercontent.com/84899178/129220012-d0c0f3c9-2f81-4433-a163-b756f6ecc10b.mp4

Tested combinations of pushing, forwarding, and backing window history states. See video for example.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.13.2 Beta


ChunkyProgrammer:
Window title is no longer updating on back/forward navigation
(Shows Settings on Subscriptions page)
image

I'm not sure the best way to solve this but I think it's caused by the afterEach in App.js

jasonhenriquez:
> Window title is no longer updating on back/forward navigation

I can't replicate this (OpenSUSE). This is odd.

ChunkyProgrammer (approved):
Tested again and am no longer getting the issue (might have been an electron being electron moment). LGTM

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

peepopoggers (changes requested):
I think the transition could be more smooth with a <0.5s animation color change. It would make it less immediate. Thoughts?

jasonhenriquez:
> I think the transition could be more smooth with a <0.5s animation color change. It would make it less immediate. Thoughts?

I do not feel too strongly about this one way or the other, but personally I think the immediacy looks fine & matches the way most browsers I've seen do it. I welcome other thoughts about this, however.

efb4f5ff-1298-471a-8973-3d47447115dc:
> 
> 
> > I think the transition could be more smooth with a <0.5s animation color change. It would make it less immediate. Thoughts?
> 
> I do not feel too strongly about this one way or the other, but personally I think the immediacy looks fine & matches the way most browsers I've seen do it. I welcome other thoughts about this, however.

I think its also fine like this.

peepopoggers (approved):

PR #1525, opened 2021-07-16
Accessibility Improvements

jasonhenriquez:
---
Accessibility Improvements
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
closes #693 
closes #1580

**Description**
Currently, most of FreeTube's controls do not use the proper semantic elements (e.g., `button`, `a`). These "fake controls" (i.e., mostly `<div>`s, `<span>`s, and custom FreeTube elements with a `@click` or `@change` event handler) seem to work fine for sighted users, but they lack much of the functionality & screen reader interaction provided by the native control elements.

To minimize the effect on the look and feel of the app, this pull request uses a combination of roles (e.g., `button`, `link`), accessibility attributes (e.g., `aria-label`, `aria-selected`), and event handlers (e.g., `@keydown`) to enhance FreeTube's controls (e.g., `<ft-icon-button>`, `<ft-notification-banner>`) & assorted existing "fake controls" with the proper behavior. Almost all controls should now be both keyboard focusable and keyboard interactive.

The appearance of the app is virtually unaffected, but some elements have new styling on being focused (and hovered for some). For people who do not use screen readers or keyboard navigation, most of the focusing style changes should be insignificant. Sighted users will appreciate that some buttons, links, and tabs now appropriately change appearance on hover.

While most users do not use their keyboard to navigate the app, the user experience for all users should be noticeably improved. The profile selection tab, channel page tabs, and all drop-down lists can now be navigated with arrow keys. All video lists can now be navigated with arrow keys as well.
 
**Screenshots (if appropriate)**
The visual appearance of the app is the same. The main differences relate to interactivity.

**Testing (for code that is not small enough to be easily understandable)**
I am neither Blind nor competent with a screen reader. I did not use a screen reader in my testing, but I did thoroughly test keyboard navigation. Although these improvements should be unilaterally positive, I would be welcome to hear the input of people who regularly use screen readers / keyboard navigation.

Here are two videos testing/showcasing most of the new functionality in five separate clips. The only changes not displayed in the videos are comment navigation, live chat navigation, and new benefits for screen readers (e.g., new button labels): 

 - https://user-images.githubusercontent.com/84899178/125984326-abbbc140-9046-407f-81d5-6008d1893491.mp4

 - https://user-images.githubusercontent.com/84899178/125984380-67135c0e-e383-4909-a577-dab587cf120e.mp4

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: 10
[...]

PR #1613, opened 2021-08-25
Make trending type selector not change background for current selection

peepopoggers:
Bit hard to explain, so TLDR previews are below

Before:
image

After:
image

Basically active selection has no background color change. I made this pr because it looks strange next to hover.

### DIFF
--- a/src/renderer/views/Trending/Trending.css
+++ b/src/renderer/views/Trending/Trending.css
@@ -32,7 +32,7 @@
   transition: background 0.2s ease-out;
 }
 
-.tab:hover, .tab:focus {
+.tab:hover {
   background-color: var(--side-nav-hover-color);
   -moz-transition: background 0.2s ease-in;
   -o-transition: background 0.2s ease-in;


peepopoggers:
@jasonhenriquez It looks strange having the hover color being the active tab color too. The active tab is underlined so I don't see the need for the background change too on the underlined one.

jasonhenriquez:
> @jasonhenriquez It looks strange having the hover color being the active tab color too. The active tab is underlined so I don't see the need for the background change too on the underlined one.

You are a bit confused on the implementation. The active tab is not any particular color. The hovered and/or focused tab(s) change color. I am saying that the focus color should be a different color than the hover color.

peepopoggers:
@jasonhenriquez look at my preview..

jasonhenriquez:
> @jasonhenriquez look at my preview..

It's that color because it's focused my friend.

efb4f5ff-1298-471a-8973-3d47447115dc:
@peepopoggers could u explain what causes ur confusion?

peepopoggers:
@efb4f5ff-1298-471a-8973-3d47447115dc in my preview image the active tab (THE ONE I AM CURRENTLY IN SO LIKE DEFAULT) and the one i am hovering over (WITH MY MOUSE) have the same background color which looks weird when they are next to each other. I am confused why the ACTIVE TAB I AM CURRENTLY ON has a background color change as it also has an underline change.

jasonhenriquez:
> @efb4f5ff-1298-471a-8973-3d47447115dc in my preview image the active tab (THE ONE I AM CURRENTLY IN SO LIKE DEFAULT) and the one i am hovering over (WITH MY MOUSE) have the same background color which looks weird when they are next to each other. I am confused why the ACTIVE TAB I AM CURRENTLY ON has a background color change as it also has an underline change.

The active tab is given focus programmatically after the videos load (This is useful for enabling arrow navigation of the tabs). Because it is focused, it changes color. Click somewhere else, or navigate away from that tab with the arrow keys, and that active tab will not have that coloration, because it's not focused anymore. I think a good solution would be to have the :focus and :hover colors be different.

peepopoggers:
@jasonhenriquez Is that design anywhere else in the app?

jasonhenriquez:
> @jasonhenriquez Is that design anywhere else in the app?

There is little to no focus styling changes in the app currently, and I am asking if you would like to use this PR (or close and create another) to 1. implement a css color variable for focus styling and 2. use it here on the trending page. Once merged, I can add this variable to a ton of places in #1525 where I'm using the same styling for :hover and :focus. This would solve the issue you were trying to solve and be a great help for so many other purposes as well.

peepopoggers:
@jasonhenriquez Sure

efb4f5ff-1298-471a-8973-3d47447115dc:
So...ready to review @peepopoggers or are u going to make the suggested changes?

jasonhenriquez:
Addressed this grievance in #1736 -- this PR will no longer need to exist, but you can open a new PR addressing the issue discussed here (the need for :focus styling) @peepopoggers.

peepopoggers:
@jasonhenriquez Beautiful work :)

PR #1634, opened 2021-09-03
Allow large changelog messages to scroll

jasonhenriquez:
---
Allow large changelogs to scroll
---

**Pull Request Type**
- [x] Bugfix

**Related issue**
N/A

**Description**
Fixes the changelog message -- you know, the one that pops up when there is a new release out -- to scroll when there is a sizable enough amount of text.

**Screenshots**
Before:
fix-changelog-before
After:
fix-changelog

 - OS: OpenSUSE
 - OS Version: Tumbleweed
 - FreeTube version: bleeding edge

**Additional context**
This pull request about large changelogs is, of course, just a spontaneous fix & is completely unrelated to anything. 😉

### DIFF
--- a/src/renderer/App.css
+++ b/src/renderer/App.css
@@ -32,6 +32,12 @@ body {
   margin-bottom: -75px;
 }
 
+#changeLogText {
+  overflow-y: scroll;
+  height: 40vh;
+  display: block
+}
+
 .fade-enter-active, .fade-leave-active {
   transition: opacity .15s;
 }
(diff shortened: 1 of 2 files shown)


PrestonN:
:+1:

peepopoggers:
Just a question, but are ID's standard for element's css? I've only seen classes before. :)

jasonhenriquez:
> Just a question, but are ID's standard for element's css? I've only seen classes used for most things, thanks. :)

Yes, it is sensible to use an id (and use it as a CSS selector, etc) for objects of which there can only be one in existence at a given time. To put it into perspective, another id is used as a CSS selector for styling in the same App.css file that I modified here.

peepopoggers:
@jasonhenriquez Makes sense :) :+1:

PR #1661, opened 2021-09-06
feat: Optimize caption sort order

### LINKED ISSUE #1444: [Enhancement] Optimize what FreeTube selects as the first caption

Yes, this feature request is related to a usage problem.  (I don't think this is a bug.)

I would like FreeTube to improve the default set of captions that it chooses by default.  There are two categories of scenarios, small cases and large cases.  For the examples, they will be videos in English, but this could apply to any other language.  (Ignore cases with either 0 captions or only auto-generated captions, even if YouTube detected the wrong spoken language.)

YouTube somehow manages to always pick the correct default caption, which is provided English captions, but FreeTube does not.  (I'm not sure how FreeTube picks the default caption list, though it's probably rather deterministic; on the other hand,  I'll surmise that YouTube is always correct due to Google's privacy invasive browser tracking.)

### Small Cases

Small cases" will be when there are only 2 captions: channel provided or auto-generated.

* Sometimes, when the captions are first activated in FreeTube, FreeTube currently defaults to the auto-generated captions - even when channel-supplied captions exist.  For the simplest example case, assume the language is English with [this video](https://www.youtube.com/watch?v=r1bUcNeyMpM).

For example, when I press `C` in FreeTube to turn on the captions in FreeTube for the following video, the captions default to `english auto-generated` - even though provided `english (united states)` exists.

However, on YouTube, doing the same action will automatically choose `English (United States)` first, instead of `English (auto-generated)`.

* Sometimes, the default behavior of captions match in FreeTube and YouTube.  [This video](https://www.youtube.com/watch?v=AUaJ8pDlxi8) is an example of this.

There isn't a clear pattern (from just 2 examples) why FreeTube behaves this way.

* I'm not sure how FreeTube should handle cases in which YouTube has provided captions and  auto-generated captions in the wrong language.  (There are legitimate cases I've seen but can't remember where it was clear there were intentionally no English subtitles.)

### Large Cases

Large cases are when there are 2 or more provided caption tracks.  To illustrate the behavior of FreeTube, there are 2 types of large cases.

* The first type of large case is when all other provided captions are alphabetically after English (at least in English), so FreeTube and YouTube are in agreement (but probably not the most robust method for FreeTube).  For example, [this video](https://www.youtube.com/watch?v=xU1ffHa47YY) has provided captions for English and 10 other languages.  However, FreeTube displays English first because it is the first provided caption language in the caption menu.

* The second type of large case is when there is any language that appears before English alphebetically (usually, either Arabic or either form of Chinese).  In [this example](https://www.youtube.com/watch?v=yVwAodrjZMY) where the first caption in FreeTube doesn't match YouTube, FreeTube shows provided Arabic but YouTube shows provided English.

FreeTube should prioritize using the channel's provided soft captions, instead of defaulting to YouTube's auto-generated captions.  For example, in the "smallest" case of 2 choices of either auto-generated English or provided English captions, FreeTube should be able to default to the provided English captions when I press `C` to activate the captions.

No, I don't think there are any other alternatives (besides leaving FreeTube the same).

**Screenshots**

(I didn't think this was necessary, since the description was more important.)

**Additional context**

* I am not sure how this would be solved/implemented for more complex cases, or 
if other unaccounted cases exist.

* I understand if this is out of scope for FreeTube, since all cases of behavior  that I noticed over time when I used to use the official instance of Invidious and downloading captions with `youtube-dl`.  If this GH issue is barking up the wrong tree, then I could suggest the optimization suggestion in the appropriate location.

ChunkyProgrammer:
---
feat: Optimize caption sort order
---

**Important note**

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
Closes #1444

**Description**
This PR will place captions with a matching locale (language +country) at the top (full string ex: english (United States)) followed by locales for the same language (ex: English(United Kingdom), English (Canada) ) then locales for that language that are auto-generated (ex: English (auto-generated) ) and finally it will go alphabetically (ex: Albanian)

The [example video](https://www.youtube.com/watch?v=r1bUcNeyMpM) (provided by in the issue) will look like this: 
image

German as selected language:
image

English (GB) as selected language: (English GB=> English => English (CA) => English (US) => Albanian => Arabic)
image

I found videos with multiple languages & tried out different locales

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.14.0

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -1107,7 +1107,7 @@ export default Vue.extend({
     sortCaptions: function (captionList) {
       return captionList.sort((captionA, captionB) => {
         const aCode = captionA.languageCode.split('-') // ex. [en,US]
-        const bCode = captionB.languageCode.split('-') 
+        const bCode = captionB.languageCode.split('-')
         const aName = (captionA.label || captionA.name.simpleText) // ex: english (auto-generated)
         const bName = (captionB.label || captionB.name.simpleText)
         const userLocale = this.currentLocale.split(/-|_/) // ex. [en,US]
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -140,6 +140,10 @@ export default Vue.extend({
       return this.$store.getters.getUsingElectron
     },
 
+    currentLocale: function () {
+      return this.$store.getters.getCurrentLocale
+    },
+
     defaultPlayback: function () {
       return this.$store.getters.getDefaultPlayback
     },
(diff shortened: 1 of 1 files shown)


peepopoggers:
That's a really cool feature! There may be a better / more efficent algorithm for sorting captions so I will wait for another pair of eyes on this, otherwise :+1:

jasonhenriquez:
I know much of our code base is not well documented, but I think this would be a good place to start helping with reversing that norm.

PikachuEXE (commented):
I can't understand the content for `sortCaptions` at all

ChunkyProgrammer:
I added some comments and simplified some of the logic. It should be a lot clearer now. @jasonhenriquez were you thinking about using jsdocs comments? or do you think that this is fine?

jasonhenriquez:
> I added some comments and simplified some of the logic. It should be a lot clearer now. @jasonhenriquez were you thinking about using jsdocs comments? or do you think that this is fine?

I think your comments are perfectly sufficient in the context of this PR. Outside of the scope of this PR, I do think FreeTube needs to systematically improve documentation. But that's neither here nor there, you did a good job. I'll test it out soon.

jasonhenriquez:
No sorting by linguistic distance?? Terrible, 0/10, I will never approve.

jasonhenriquez (approved):
Just kidding, this is great! This is an amazing convenience.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Worked great! Please accept this approval as token of my gratitude.

PikachuEXE (approved):
Good job!

PR #1392, opened 2021-06-04
* Improve style of top nav search input

PikachuEXE:
**Pull Request Type**
Please select what type of pull request this is:
- [x] Feature Implementation

**Related issue**
Closes #1387

**Description**
Update top nav search to add clear search input text button
Also update the padding of input text box so that input text won't appear under the "arrow-right" button

**Screenshots (if appropriate)**
https://user-images.githubusercontent.com/1018543/120764623-ebd83900-c54a-11eb-95da-a50112ef01a7.mp4
image

**Testing (for code that is not small enough to be easily understandable)**
- Enter text very long into input text box
- Press the new button

**Desktop (please complete the following information):**
 - OS: MacOS
 - OS Version: 10.15.7
 - FreeTube version: 13.0.0 (latest on `development` actually)

**Additional context**

Issue #1387, opened 2021-06-02
Little (x) Button To Quickly Clear Search Bar Of Text


BrainstormSurge:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Usage. Less clicks means better UI. I've seen similar implementation where clicking the X button (whatever color such as red, black or white) will clear the search text and select the search bar.

One button and you're ready to type in your new search term.

Current way is to click the bar then use Backspace but this can feel annoying after a while compared to what can be done with one click of an X button.

**Describe the solution you'd like to see implemented**
Add an X button that will clear and select the search bar in one action/click. 

Best spot to put it for the current UI would be on the left side of the bar instead of next to the search arrow and filter buttons as that would start to make things cluttered there and would prevent accidental clicks that clear your current search.

--

Possible improvements would to be to add a couple hotkeys. One to clear text function while within the search bar or even while the FreeTube window is selected. The other to quickly navigate from anywhere to the search bar.

Possible clear function default hotkeys: [Ctrl]+[Backspace] or [Ctrl]+[Delete]

Possible hotkey for navigating to the search bar is: [Ctrl]+[Shift]+[F]

(I know hotkeys are not implement yet. These are future possibilities.)

**Describe alternatives you've considered**
Nothing that's as simple to implement or use. Anything more would be clunky.

**Screenshots**
![Searchbar Mockup](https://user-images.githubusercontent.com/26531295/120562327-6b3b0080-c3c3-11eb-8cad-2f09ae12bd93.JPG)

**Additional context**
I realize that #1392 was created as an improvement to mine. I'm no UI designer and only used MS Paint to create my mockup. Feel free to merge the issues if you feel it's applicable while adding anything that's not in the other together.

GilgusMaximus:
Short notice. Please use the template we provide. Otherwise I will start to close your issues without warning.

BrainstormSurge:
Okay. Sorry about that. I'll keep it in mind.

PikachuEXE:
See #1392

BrainstormSurge:
> Short notice. Please use the template we provide. Otherwise I will start to close your issues without warning.

I've updated the main post according to the feature request template.

BrainstormSurge:
> See #1392

Thank you for this. I'm glad it wasn't that difficult to test out.

Nice mustache by the way.

Issue #938, opened 2021-01-16
UI Search box


kgr0:
**Topic**
UI

**Question**
We're not sure if it is correct behavior of search box. A part of text is overlapped by arrow icon.
![image](https://user-images.githubusercontent.com/50838031/104810151-8d105e00-57fb-11eb-8fb7-6a7061203b5d.png)

While hovering on arrow:
![image](https://user-images.githubusercontent.com/50838031/104810158-9699c600-57fb-11eb-8a83-4957e2199e3a.png)

_Maybe the length of textbox should be shorter?_

****

Environment Information :
- OS: Windows
- OS Version: Windows 10
- FreeTube version: v0.11.1 Beta

@Vovany4, @abulenok

PikachuEXE:
Seems fixed already in #1392
https://github.com/FreeTubeApp/FreeTube/blame/development/src/renderer/components/ft-input/ft-input.css#L105-L112
![image](https://user-images.githubusercontent.com/1018543/133014739-c0f50408-fd7f-4b38-bcf3-e50caf010241.png)

efb4f5ff-1298-471a-8973-3d47447115dc:
@kgr0 is this fixed for u?

PR #1669, opened 2021-09-08
Focus in input after text clear

PikachuEXE:
**Pull Request Type**
- [x] Feature Implementation

**Related issue**
#1387
Missing change for #1392

**Description**
Update text input box to focus on text input element after "reset button" clicked

https://user-images.githubusercontent.com/1018543/132451601-ee44e74b-d7ff-4226-a7fd-42c7e4e239e0.mov

- Type something
- Press reset button

- OS: MacOS
- OS Version: 11.5.2
- FreeTube version: based on 4a871578bfb468ddae152147f3f40d1788c2a4ec

**Additional context**

### DIFF
--- a/src/renderer/components/ft-input/ft-input.css
+++ b/src/renderer/components/ft-input/ft-input.css
@@ -88,7 +88,7 @@
 
 .inputAction {
   position: absolute;
-  padding: 10px;
+  padding: 10px 8px;
   top: 5px;
   right: 0px;
   cursor: pointer;
--- a/src/renderer/components/ft-input/ft-input.css
+++ b/src/renderer/components/ft-input/ft-input.css
@@ -11,13 +11,19 @@
   cursor: pointer;
   border-radius: 200px 200px 200px 200px;
   color: var(--primary-text-color);
+
+  opacity: 0;
+
+  -moz-transition: background 0.2s ease-in, opacity 0.2s ease-in;
+  -o-transition: background 0.2s ease-in, opacity 0.2s ease-in;
+  transition: background 0.2s ease-in, opacity 0.2s ease-in;
 }
 
 .clearInputTextButton:hover {
   background-color: var(--side-nav-hover-color);
-  -moz-transition: background 0.2s ease-in;
-  -o-transition: background 0.2s ease-in;
-  transition: background 0.2s ease-in;
+}
+.clearInputTextButton.visible {
+  opacity: 1;
 }
 
 .forceTextColor .clearInputTextButton:hover {
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -74,6 +74,10 @@ export default Vue.extend({
 
     idDataList: function () {
       return `${this.id}_datalist`
+    },
+
+    inputDataPresent: function () {
+      return this.inputData.length > 0
     }
   },
   watch: {
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -104,6 +104,10 @@ export default Vue.extend({
     handleClearTextClick: function () {
       this.inputData = ''
       this.$emit('input', this.inputData)
+
+      // Focus on input element after text is clear for better UX
+      const inputElement = document.getElementById(this.id)
+      inputElement.focus()
     },
 
     addListener: function () {
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -61,7 +61,9 @@ export default Vue.extend({
         selectedOption: -1,
         isPointerInList: false
       },
-      clearTextButtonVisible: true
+      // This button should be invisible on app start
+      // As the text input box should be empty
+      clearTextButtonVisible: false
     }
   },
   computed: {
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -60,7 +60,8 @@ export default Vue.extend({
         showOptions: false,
         selectedOption: -1,
         isPointerInList: false
-      }
+      },
+      clearTextButtonVisible: true
     }
   },
   computed: {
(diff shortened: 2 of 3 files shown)


jasonhenriquez:
Screenshot_20210909_102040

PikachuEXE:
@jasonhenriquez 
Fixed
I still haven't get used to Vue yet 😅

jasonhenriquez:
It looks great visually! One more fix needed: the current solution removes the element's visibility, but the button is still present. As in, still keyboard-navigable, still changes the mouse cursor to "pointer", etc. This should not be the case.

PikachuEXE:
Right now it's using css transition + `opacity` to have the fade out effect
I got no idea how to make it disappear **only after** the animation

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE is it possible to also fix this issue https://github.com/FreeTubeApp/FreeTube/issues/938 because it is somewhat related. 

The only thing that needs to be fixed is when u input a long string and hover over the arrow icon the text behind it doesn't disappear.

peepopoggers:
@PikachuEXE you could use some hacky timer-based element hiding, but maybe you could also make it not selectable when it fades out? it does not necessarily have to disappear

jasonhenriquez:
> @PikachuEXE you could use some hacky timer-based element hiding, but maybe you could also make it not selectable when it fades out? it does not necessarily have to disappear

Yes, for this way you would have to set pointer-events to none, aria-hidden to true, and tabindex to -1.

PikachuEXE:
If `setTimeout` is used I think it simpler to just hide the button...

I will probably do that in a separate PR
Already got too many changes

peepopoggers:
@PikachuEXE It would make sense to add it in this PR as well, it seems unnecessary to open another PR IMO.

peepopoggers:
@PikachuEXE It would make sense to add it in this PR as well, it seems unnecessary to open another PR IMO.

PikachuEXE:
Seems fixed already in #1392
https://github.com/FreeTubeApp/FreeTube/blame/development/src/renderer/components/ft-input/ft-input.css#L105-L112
image

PikachuEXE:
Button now removed/hidden after 200ms via `setTimeout`

https://user-images.githubusercontent.com/1018543/133014975-6bb1ac85-85e5-48c9-90a6-48bd6f92eacb.mov

efb4f5ff-1298-471a-8973-3d47447115dc:
> Seems fixed already in #1392
> https://github.com/FreeTubeApp/FreeTube/blame/development/src/renderer/components/ft-input/ft-input.css#L105-L112
> image
> 

@PikachuEXE u need to input al long string, hover over the arrow icon and u'll see that the hover animation overlaps with the text.

PikachuEXE:
Overlap issue fixed
image

PikachuEXE:
For (3) that's behaviour of `input` for `text`
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_test

Not saying it can't be optimized but I wouldn't do it in this PR

https://user-images.githubusercontent.com/1018543/133060061-f3436eb3-3a38-4e7b-93c0-57f6ca4b3fd7.mov

efb4f5ff-1298-471a-8973-3d47447115dc:
So i just revisited the fact that the clear button is on the left, because it was bugging me a bit just like @peepopoggers. I checked how YT this does and i actually think it might be a solution worth talking about. My previous issue about putting the clear button at the right side of the bar was that it would cause miss clicks, after seeing this implementation i think it somthing similar will work out.

Capture

Also maybe its a good idea to change the search icon to this one.
https://fontawesome.com/v5.15/icons/search?style=solid

efb4f5ff-1298-471a-8973-3d47447115dc:
> Not saying it can't be optimized but I wouldn't do it in this PR
Ok, np. I'll open an issue for it.

PikachuEXE:
> Also maybe its a good idea to change the search icon to this one.
https://fontawesome.com/v5.15/icons/search?style=solid

I think one of the reasons we are using arrow instead of search icon on that button is
The button also makes the app jumps to video/playlist/channel directly.
So I am fine looking for another icon, but probably not the search icon.

PikachuEXE:
I would suggest merging this PR as there are improvements over current dev version.
The implementation of moving the clear button to the right, which requires style update of "action button" at the same time, can be done in a new PR.

PikachuEXE:
Will this be merged before 0.15?

efb4f5ff-1298-471a-8973-3d47447115dc:
I would like to see this also to be merged before release.

jasonhenriquez:
I only have two remaining suggestions for the search bar at the moment: 
1. Disappearing the search arrow when the text box is busy like how browsers handle their search bars.
2. Having the disappear animation of the Clear Text button also occur on appear/re-appear for visual consistency.

These do not have to be merged with this, of course, but if you think you can knock them out in time with it, I support it. If not, that's also fine, and I will gladly approve.

PikachuEXE:
When is the deadline?
If it's this weekend I will try to do 1 of the 2 suggestions. (Or both if really got time)

jasonhenriquez:
> When is the deadline?
> If it's this weekend I will try to do 1 of the 2 suggestions. (Or both if really got time)

No hard deadline, but we would like to release soon. Given that this PR is fine as is & we want to give more time for testing, we should probably approve this now, and you can have more time to work on those changes at your own pace.

jasonhenriquez (approved):
LGTM!

PikachuEXE:
> we want to give more time for testing

Sure!

ChunkyProgrammer (approved):
LGTM

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PikachuEXE:
@jasonhenriquez 
> Disappearing the search arrow when the text box is empty like how browsers do it in their search bars.

I am not sure I get what you mean here.
Does it mean "Make search arrow (action button I call it) disappear/appear with same logic as clear search button"?

jasonhenriquez:
> @jasonhenriquez
> 
> > Disappearing the search arrow when the text box is empty like how browsers do it in their search bars.
> 
> I am not sure I get what you mean here.
> Does it mean "Make search arrow (action button I call it) disappear/appear with same logic as clear search button"?

In the sense that it disappears when the text box is empty, yes (and also disabling the Enter key from triggering a search in this case). I say this because searching empty text has no purpose. When empty, the icon should either not be there or be non-interactive & possibly slightly grayed.

PikachuEXE:
@jasonhenriquez 
> When empty, the icon should either not be there or be non-interactive & possibly slightly grayed.

Included in https://github.com/FreeTubeApp/FreeTube/pull/1738

PR #1757, opened 2021-09-27
Date locale change

### LINKED ISSUE #725: Use format string for published Date.
Use format string for the date of the published date on the playback screen.
The current date format is American.
I would like to match the Japanese format, so could you use the format strings?
It would be nice to have a format string like C#.
### sample: 
0=year, 1=month, 2=day
Current "Oct 25, 2020" = "{1} {2}, {0}"
British "25th Oct 2020" = "{2}th {1} {0}"
Japanese & Chinese "2020 年 10 月 25 日" = "{0} 年 {1} 月 {2} 日"
simple "2020/10/25" = "{0}/{1}/{2}"

Airidasz:
---
Change publish date format based on locale
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
closes #725

**Description**
The PR makes a change to the "dateString" that is used to display a videos publish date and format it based on the currently selected locale.

**Screenshots**
Polish format
image

Finnish format
image

English format
image

**Testing**
Tested with multiple languages. When a locale is NOT set it will default to English so an error is unlikely to occur


src/renderer/components/watch-video-info/watch-video-info.js
```diff
@@ -227,9 +227,9 @@ export default Vue.extend({
 
     dateString() {
       const date = new Date(this.published)
-      const dateSplit = date.toDateString().split(' ')
-      const localeDateString = `Video.Published.${dateSplit[1]}`
-      return `${this.$t(localeDateString)} ${dateSplit[2]}, ${dateSplit[3]}`
+      const localeCode = this.$t('Locale Code')
```
ChunkyProgrammer, line 230: Instead of updating every yaml file, you could have just added a computed property like this https://github.com/FreeTubeApp/FreeTube/blob/f4e5537397c0b43e121c4e80f3d3c3cd386a5c6a/src/renderer/components/ft-video-player/ft-video-player.js#L143 to watch-video-info and replaced the underscores with dashes to get the locale code

Airidasz: Changed. Thanks

src/renderer/components/watch-video-info/watch-video-info.js
```diff
@@ -227,8 +231,7 @@ export default Vue.extend({
 
     dateString() {
       const date = new Date(this.published)
-      const localeCode = this.$t('Locale Code')
-      const localeDateString = new Intl.DateTimeFormat([localeCode, 'en'], { dateStyle: 'medium' }).format(date)
+      const localeDateString = new Intl.DateTimeFormat([this.currentLocale, 'en'], { dateStyle: 'medium' }).format(date)
```
ChunkyProgrammer, line 234:
```suggestion
      const localeDateString = new Intl.DateTimeFormat([this.currentLocale.replace('_', '-'), 'en'], { dateStyle: 'medium' }).format(date)
```

Airidasz:
DON'T APPROVE THIS YET. Something wrong with English-UK, US and others work fine :/

ChunkyProgrammer (commented):
A few locales return underscores instead of dashes (Ex. en_UK and es_AR) so the underscores will need to be replaced with dashes

ChunkyProgrammer (dismissed):
LGTM, thanks so much for taking the time to create this PR

jasonhenriquez:
Screenshot_20210928_083739

It looks good, but I'm still seeing the incorrect date behavior on playlist pages.

Airidasz:
> 
> 
> Screenshot_20210928_083739
> 
> It looks good, but I'm still seeing the incorrect date behavior on playlist pages.

OK I checked it out and it seems that this data is displayed by playlist-info.js. 
Playlist-info "last update" string is a part of a response that is received from ytdl.js. That means it is not formatted by the program, but by ytdl. For now I haven't found a way to change the locale of the request.

Hiers (dismissed):
Thank you for this PR. It's a small change, but it's nice to see proper dates in languages other than American English.

ChunkyProgrammer:
Haven't tested the playlist part yet but from what I can tell, YouTube uses underscores for their hl parameter.

"The example sets the hlparameter value to es_MX, indicating that text values in the API response should be provided in that language. That default value of that parameter is en_US." so it might be a better idea to replace dashes with underscores for that part (if it still works with underscores)

https://developers.google.com/youtube/v3/docs/i18nLanguages/list#common-use-cases

Airidasz:
> 
> 
> Haven't tested the playlist part yet but from what I can tell, YouTube uses underscores for their hl parameter.
> 
> "The example sets the hlparameter value to es_MX, indicating that text values in the API response should be provided in that language. That default value of that parameter is en_US." so it might be a better idea to replace dashes with underscores for that part (if it still works with underscores)
> 
> https://developers.google.com/youtube/v3/docs/i18nLanguages/list#common-use-cases

Tested. It works with both dashes and underscores.

ChunkyProgrammer (approved):

Hiers (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

livingentity:
Would be nice if this could default to the system language instead of english because the default language of FreeTube is system language, too.

PR #1777, opened 2021-10-01
Scroll for changing volume, like enhancer for youtube

anxiolyt:
---
Scroll for changing volume, like enhancer for youtube
---

I would like to request the feature of being able to scroll in order to change the volume by using the scroll wheel.
Ideally 1% by up/down.

There's a chrome externsion that supports this, for example.
https://chrome.google.com/webstore/detail/enhancer-for-youtube/ponfpcnoihfmfllpaingbgckeeldkhle?hl=de

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation


efb4f5ff-1298-471a-8973-3d47447115dc:
This is no place for feature requests. Create ur issue here https://github.com/FreeTubeApp/FreeTube/issues/new/choose and please fill in the template otherwise it will get removed.

anxiolyt:
Thank you. So it's fine to do feature requests in 'Issues'?

jasonhenriquez:
@anxiolyt That's correct, you can see those labeled as enhancements. Also, however, this feature is already implemented in FreeTube: see "Scroll Volume Over Video Player" under Player Settings.

PR #1783, opened 2021-10-02
Fix: Update thumbnail url and channel name for subscriptions when change detected

### LINKED ISSUE #1677: Channel icon in subscription list not updating with change
**Behavior of the program**
If you subscribe to a channel, it's icon appears in the subscription list, but, if the channel changes their icon then only the icon on the channel's page will change; icons in the subscription list don't get updated unless you manually unsubscribe and resubscribe.

**Expected behavior**
Checking every subscription for an icon change on every load (or even every couple of loads) would be inefficient from a network perspective, so instead I propose this:

The icon on the channel page changes as expected, so we can compare the icon loaded on the channel to the icon in the subscription list to see if the sub-list icon needs to be updated, the sub-list icon will only change when someone goes to the channel which is much nicer on the network then checking every channel at once.

**To Reproduce**
1. Subscribe to a channel.
2. Have that channel change it's icon.
3. See the channel's page display the new icon, but, not the subscription list.

**Environment Information**
 - OS: Arch Linux
 - FreeTube version: 0.15.1 Beta
 - Installation Method: FlatHub
 - Primary API Used: Local API [Happens with both]

ChunkyProgrammer:
---
Update thumbnail url for subscriptions when change detected
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation

**Related issue**
Closes #1677

**Description**
When navigating to a channel, video or playlist page: it will check if the user is subscribed to the author & if they are subscribed then it will check if the thumbnail needs updating and will update the thumbnail accordingly

I did the following for Invidious & local api:
I modified the thumbnail url to one of my subscriptions then:
- navigated to one of their videos (without navigating to channel)
- navigated to one of their playlists (without navigating to channel)
- navigated to their channel page.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.15.0 Beta

**Additional context**


ChunkyProgrammer:
the function to update thumbnails is pretty repetitive so I'm wondering if there's a file I can move it to so it only appears once (I'd update the function to take in thumbnail url & channel Id)

jasonhenriquez:
> the function to update thumbnails is pretty repetitive so I'm wondering if there's a file I can move it to so it only appears once (I'd update the function to take in thumbnail url & channel Id)

I use this implementation for global functions in the Accessibility Improvements PR: https://github.com/FreeTubeApp/FreeTube/blob/7391855a546d636fb05cfc21032f5ed4294fbf00/src/renderer/App.js#L477
Note that the function name begins with a `$` so as to explicitly denote that this function is global.

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer can u resolve the merge conflicts. i wanted to test but im unable to because of it :(

PrestonN:
> the function to update thumbnails is pretty repetitive so I'm wondering if there's a file I can move it to so it only appears once (I'd update the function to take in thumbnail url & channel Id)

I actually created the [Utils Store](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/store/modules/utils.js) for this exact reason.  The idea is that miscellaneous functions that are used across the app but don't have a good spot to go typically goes into utils.

I'm open to suggestions from you or @jasonhenriquez if this approach should be changed however, but that's how I've been handling it so far.

ChunkyProgrammer:
Sounds good, I'll move the method to utils when I get the chance

efb4f5ff-1298-471a-8973-3d47447115dc:
wait a second isnt https://github.com/FreeTubeApp/FreeTube/issues/2204 the same as #1677?

ChunkyProgrammer:
> wait a second isnt #2204 the same as #1677?

They are different. This PR should fix 1677. Another PR would have to fix 2204(lazyload/only load images when they are visible to user). This pr might fix some issues wirh 2204 though

src/renderer/views/Channel/Channel.js
```diff
@@ -364,7 +367,9 @@ export default Vue.extend({
         } else {
           this.subCount = response.subCount
         }
-        this.thumbnailUrl = response.authorThumbnails[3].url.replace('https://yt3.ggpht.com', `${this.currentInvidiousInstance}/ggpht/`)
+        const thumbnail = response.authorThumbnails[3].url
+        this.thumbnailUrl = thumbnail.replace('https://yt3.ggpht.com', `${this.currentInvidiousInstance}/ggpht/`)
+        this.updateChannelThumbnail(thumbnail, this.channelName, this.id)
```
hockerschwan, line 372: This part worked with
```suggestion
        this.updateChannelThumbnail({
          channelThumbnailUrl: this.thumbnailUrl,
          channelName: this.channelName,
          channelId: this.id
        })
```

ChunkyProgrammer: Did not realize that. I'll make that change when I get the chance. Thanks.

ChunkyProgrammer: I think I did it like this so it saves the youtube thumbnail url and not the invidious one. Will update it to use the named parameters though

absidue:
@ChunkyProgrammer please fix the merge conflicts.

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Followed the testcases u provided and tried some hacky things to break it. LGTM!

PikachuEXE:
How do you 
> I modified the thumbnail url to one of my subscriptions

ChunkyProgrammer:
> How do you
> 
> > I modified the thumbnail url to one of my subscriptions

- Export Subscriptions as FreeTube subscriptions 
- Modify the thumbnail url for one of the exported subscriptions
- Remove subscription data (delete data settings, i think it was called?)
- Import the subscriptions file

PikachuEXE (commented):
Tested locally
- navigated to one of their videos (without navigating to channel)
- navigated to one of their playlists (without navigating to channel)
- navigated to their channel page.

src/renderer/views/Playlist/Playlist.js
```diff
@@ -80,6 +80,10 @@ export default Vue.extend({
           channelId: result.author ? result.author.channelID : '',
           infoSource: 'local'
         }
+        const channelThumbnailUrl = this.infoData.channelThumbnail.replace('=s48', '=s176')
```
PikachuEXE, line 83: I totally can't understand `replace('=s48', '=s176')` part
Can you add code comment to explain it

Same for other places with same code

src/renderer/store/modules/profiles.js
```diff
@@ -89,6 +89,33 @@ const actions = {
     commit('setProfileList', profiles)
   },
 
+  async updateChannelThumbnail({ getters, dispatch }, { channelThumbnailUrl, channelName, channelId }) {
+    const profileList = getters.getProfileList
+    for (let i = 0; i < profileList.length; i++) {
```
PikachuEXE, line 94: Is it possible to use `forEach` to replace for loop?
I only see `i` being used in `profileList[i]`

src/renderer/store/modules/profiles.js
```diff
@@ -89,6 +89,33 @@ const actions = {
...
+      const currentProfile = JSON.parse(JSON.stringify(profileList[i]))
```
PikachuEXE, line 95: This is a copy of "current profile"
`currentProfileCopy` might not be a good name but it's not the original object
```suggestion
      const currentProfileCopy = JSON.parse(JSON.stringify(profileList[i]))
```

src/renderer/store/modules/profiles.js
```diff
@@ -89,6 +89,33 @@ const actions = {
...
+      const channelIndex = currentProfile.subscriptions.findIndex((channel) => {
+        return channel.id === channelId
+      })
+      if (channelIndex !== -1) {
+        console.log('channel found')
+        let updated = false
+        if (profileList[i].subscriptions[channelIndex].name !== channelName) {
```
PikachuEXE, line 102: Why do you need to read from `profileList[i].subscriptions[channelIndex]` when you read from `currentProfile.subscriptions[channelIndex]`?

PikachuEXE (changes requested):

PikachuEXE:
Feature review & code review are separate :)

efb4f5ff-1298-471a-8973-3d47447115dc:
> Feature review & code review are separate :)

Yeah that's my bad for not noticing that 😬

src/renderer/store/modules/profiles.js
ChunkyProgrammer: it uses a break statement so i can't use a foreach

absidue: you can use for-of in this case, if you want to:

```js
for (const profile of profileList) {
    // do stuff here with the profile variable
}
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -135,6 +141,10 @@ export default Vue.extend({
           channelId: result.authorId,
           infoSource: 'invidious'
         }
+        const channelThumbnailUrl = result.authorThumbnails[2].url
+        const channelName = this.infoData.channelName
+        const channelId = this.infoData.channelId
+        this.updateSubscriptionDetails({ channelThumbnailUrl, channelName, channelId })
```
absidue, line 147: Please inline these variables as you did in https://github.com/FreeTubeApp/FreeTube/pull/1783/commits/ba537cc1b8572ce20467634bd5dcf7baf0b7eb67.

src/renderer/views/Watch/Watch.js
```diff
@@ -276,6 +276,11 @@ export default Vue.extend({
             this.channelName = result.player_response.videoDetails.author
             this.channelThumbnail = result.player_response.embedPreview.thumbnailPreviewRenderer.videoDetails.embeddedPlayerOverlayVideoDetailsRenderer.channelThumbnail.thumbnails[0].url
           }
+          const channelName = this.channelName
+          const channelId = this.channelId
+          const channelThumbnailUrl = this.channelThumbnail
+          this.updateSubscriptionDetails({ channelThumbnailUrl, channelName, channelId })
```
absidue, line 282: Please inline these variables as well.

src/renderer/views/Watch/Watch.js
```diff
@@ -561,7 +566,13 @@ export default Vue.extend({
           }
           this.channelId = result.authorId
           this.channelName = result.author
-          this.channelThumbnail = result.authorThumbnails[1] ? result.authorThumbnails[1].url.replace('https://yt3.ggpht.com', `${this.currentInvidiousInstance}/ggpht/`) : ''
+          const channelThumb = result.authorThumbnails[1]
+          this.channelThumbnail = channelThumb ? channelThumb.url.replace('https://yt3.ggpht.com', `${this.currentInvidiousInstance}/ggpht/`) : ''
+          const channelThumbnailUrl = channelThumb?.url
+          const channelName = this.channelName
+          const channelId = this.channelId
+          this.updateSubscriptionDetails({ channelThumbnailUrl, channelName, channelId })
```
absidue, line 574: Please inline these variables as well.

absidue (changes requested):

src/renderer/views/Playlist/Playlist.js
ChunkyProgrammer: the s is just the thumbnail size, i think

src/renderer/store/modules/profiles.js
```diff
@@ -89,6 +89,33 @@ const actions = {
     commit('setProfileList', profiles)
   },
 
+  async updateSubscriptionDetails({ getters, dispatch }, { channelThumbnailUrl, channelName, channelId }) {
+    if (channelThumbnailUrl === null) {
```
PikachuEXE, line 93: What do we handle `undefined` value?

src/renderer/store/modules/profiles.js
```diff
@@ -89,6 +89,33 @@ const actions = {
...
+    const profileList = getters.getProfileList
+    for (const profile of profileList) {
+      const currentProfileCopy = JSON.parse(JSON.stringify(profile))
+      const channelIndex = currentProfileCopy.subscriptions.findIndex((channel) => {
+        return channel.id === channelId
+      })
+      if (channelIndex === 1) { continue }
+      let updated = false
+      const channel = currentProfileCopy.subscriptions[channelIndex]
+      if (channel.name !== channelName || (channel.thumbnail !== thumbnail && thumbnail !== null)) {
```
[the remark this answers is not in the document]
ChunkyProgrammer: Potentially with this change #2283  or maybe if there's an api error/update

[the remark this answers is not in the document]
PikachuEXE: I see `const thumbnail = channelThumbnailUrl.replace(/=s\d*/, '=s176')`
So I think this either will throw an error (when `channelThumbnailUrl` is not string) or it's returning a string
Am I missing anything? (Not sure)

src/renderer/store/modules/profiles.js
ChunkyProgrammer: fixed in latest commit (option chaining)

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

PR #1796, opened 2021-10-06
Parse like and dislike count to add comma separator

vallode:
---
Parse like and dislike count to add comma separator
---
**Pull Request Type**
- [*] Bugfix
- [ ] Feature Implementation

**Related issue**
Please link the issue your pull request is referring to. If this pull request fully resolves the relevant issue, put "closes" before the issue number. Example: "closes #123456".

Another pet peeve with no issue, sorry! :see_no_evil: 

**Description**

Uses toLocaleString to add a separator to larger like/dislike values. Eye candy.

Before:
Screenshot from 2021-10-06 19-09-03

After:
Screenshot from 2021-10-06 19-08-41

Go on any video above a thousand likes to see the difference. I browsed a bit to see if it caused issues.

 - OS: [e.g. iOS] Debian
 - OS Version: [e.g. 22] 10
 - FreeTube version: [e.g. 0.8] 0.15.0

**Additional context**

### DIFF
--- a/src/renderer/components/watch-video-info/watch-video-info.js
+++ b/src/renderer/components/watch-video-info/watch-video-info.js
@@ -198,6 +198,22 @@ export default Vue.extend({
       return this.likeCount + this.dislikeCount
     },
 
+    parsedLikeCount: function () {
+      if (this.hideVideoLikesAndDislikes) {
+        return null
+      }
+
+      return this.likeCount.toLocaleString()
+    },
+
+    parsedDislikeCount: function () {
+      if (this.hideVideoLikesAndDislikes) {
+        return null
+      }
+
+      return this.dislikeCount.toLocaleString()
+    },
+
     likePercentageRatio: function () {
       return parseInt(this.likeCount / this.totalLikeCount * 100)
     },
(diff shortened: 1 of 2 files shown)


vallode:
Interestingly enough the app does the same parsing to the view count but with a bespoke regex in: src/renderer/components/watch-video-info/watch-video-info.js:225.

Does any one know why the regex is used over the native toLocaleString? Only reason I prefer locale string is it actually uses the delimiter of the locale (comma for most places but some use spaces or decimals).

peepopoggers (approved):
LGTM

ChunkyProgrammer:
Maybe pass in the user's locale and "en" for toLocaleString()

Ex:
number.toLocaleString([userLocale, 'en-US'])

You can look at #1757 for reference

vallode:
> Maybe pass in the user's locale and "en" for toLocaleString()
> 
> Ex: number.toLocaleString([userLocale, 'en'])
> 
> You can look at #1757 for reference

Really good point, failed to notice this. Done :)

peepopoggers (approved):

Hiers (approved):
I'd also like this see a similar PR for views, as you suggested.

PR #1821, opened 2021-10-15
Feat: Verified account indicator

### LINKED ISSUE #944: Verified mark(Right mark) beside channel name

ChunkyProgrammer:
---
Verified account indicator
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
closes #944
supercedes: #1095

**Description**
This PR will show verified indicators beside channel names when the data is available or it will look into a cache of verified channel ids to figure it out (ex: subscriptions page, most popular, using invidious api)

Trending page:
image

Watch page:
image

Comments:
image

History:
image

Channel:
image

Playlist (List):
image

Playlist 
image

I navigated to different pages that could display veification check marks then navigated to pages that couldn't and added caching to support those

 - OS: Windows
 - OS Version: 10
 - FreeTube version: Latest nightly

**Additional context**
I don't know if something other than cache should be used instead for verified accounts because the cache clears on app close but I'm not sure what I'd do instead (a whole db just for verified channels?)


ChunkyProgrammer:
Verified Artists are considered "Verified" on some pages and "Official Artist" on others. Would be curious as to whether they should be considered verified on all pages or not be considered verified

ChunkyProgrammer:
> > Verified Artists are considered "Verified" on some pages and "Official Artist" on others. Would be curious as to whether they should be considered verified on all pages or not be considered verified.
> 
> Can you clarify how YT makes this distinction? Ideally we should replicate their behavior here, but if this is not possible I would like to know why.
> 
yt adds a music icon to Official Artists
image. I just looked at it again and it looks like it might actually be a bug causing the verified icon to be displayed for artists (it can display for non verified too). I don't think the "Official Artist" is parsed in enough places to make it worth while to include in this PR

> > Would be interested in other opinions of this PR as well like "Only add a channel to cache on specific pages and not all pages" or whether verified channel caching should be implemented as a setting
> 
> Is this particularly costly or disadvantageous? Is it a privacy concern? If no to either, then that should not be a setting.

I don't think so, I just saw this in a [comment](https://github.com/FreeTubeApp/FreeTube/pull/1095#issuecomment-803774279) on the PR that this replaces

efb4f5ff-1298-471a-8973-3d47447115dc:
wanted to test but merge conflicts :(

efb4f5ff-1298-471a-8973-3d47447115dc:
this is the only thing i could find that has no verified indicator

yt
ft

ChunkyProgrammer:
invidious now shows verified information. I don't want to put any more effort into this feature at the time so I will be closing this PR. This PR can likely help if someone else were to try to implement this. (The cache stuff probably wouldn't be needed)

PR #1888, opened 2021-11-12
Automatically compile to arm64 when running on apple silicon and add custom user agent option

Arctic26:
---
Automatically compile to arm64 when running on apple silicon and and add custom user agent option
---
**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Description**
- Automatically selects arm64 architecture when building on an apple chip
- Change the default user agent to hide that freetube is being used

 - OS: macOS
 - OS Version: 12.0.1
 - FreeTube version: 0.15.1


efb4f5ff-1298-471a-8973-3d47447115dc:
U need to change branch to development

src/main/index.js
```diff
@@ -205,6 +205,7 @@ function runApp() {
         commonBrowserWindowOptions
       )
     )
+    newWindow.webContents.userAgent = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/95.0.4638.69 Safari/537.36" // Prevent youtube from seeing the freetube header
```
PikachuEXE, line 208: I am not sure if we should set a custom user agent
Even if we do it should be an option (or at least not a constant value for all OS)

I think this should be reverted

PikachuEXE (changes requested):
User agent change should be reverted

_scripts/build.js
```diff
@@ -8,9 +8,17 @@ const args = process.argv
...
 
 if (platform == 'darwin') {
-  targets = Platform.MAC.createTarget()
+  let arch;
+
+  if (cpus[0].model.startsWith('Apple')) {
+    arch = Arch.arm64;
+  } else {
+    arch = Arch.x64;
+  }
```
PikachuEXE, line 20: Remove trailing commas, set default value...
```suggestion
  let arch = Arch.x64

  if (cpus[0].model.startsWith('Apple')) {
    arch = Arch.arm64
  }

  targets = Platform.MAC.createTarget(['dmg'], arch)
```

PikachuEXE (changes requested):

PikachuEXE (changes requested):
I have tested `_scripts/build.js` changes locally and the built binary seems running fine and reported as "Apple"

image

However we would only accept that change in this PR and not the custom user agent options
Please submit those changes in a separate PR

PR #2075, opened 2022-02-11
External player choice 'None' translated in Settings

### LINKED ISSUE #1670: Please allow `None` to be translatable
Please allow `None` to be translatable.

Settings > External Player Settings > External Player > Drop-down with: `None`.

kép

Thank you

FreeTube version 0.14.0 beta

9ekaitz:
---
External player choice 'None' translatable in Settings
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
closes #1670

**Description**
Enables the translation of the word 'None' in the external player selector in the Settings menu. In order to do that the name of the first player is changed from 'None' to a translatable string and the whole array of names is translated. (Only the 'none' one has a translation, the other ones remain the same)

Just tested if the word is substituted by the one on the locales file

 - OS: GNU/Linux
 - OS Version: Pop!_OS 21.04
 - FreeTube version: v0.16.0 Beta

**Additional context**
Is the first time I contribute to a project and also the first time using Vue, and haven't done a lot of things in JS, so if there is anything wrong with Vue or the format of the code tell me and I will try fix it

### DIFF
--- a/src/renderer/components/external-player-settings/external-player-settings.js
+++ b/src/renderer/components/external-player-settings/external-player-settings.js
@@ -5,6 +5,7 @@ import FtSelect from '../ft-select/ft-select.vue'
 import FtInput from '../ft-input/ft-input.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
 import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
+import i18n from '../../i18n/index.js'
 
 export default Vue.extend({
   name: 'ExternalPlayerSettings',
@@ -24,7 +25,7 @@ export default Vue.extend({
     },
 
     externalPlayerNames: function () {
-      return this.$store.getters.getExternalPlayerNames
+      return this.$store.getters.getExternalPlayerNames.map((value) => i18n.t(value))
     },
     externalPlayerValues: function () {
       return this.$store.getters.getExternalPlayerValues
--- a/src/renderer/components/external-player-settings/external-player-settings.js
+++ b/src/renderer/components/external-player-settings/external-player-settings.js
@@ -5,7 +5,6 @@ import FtSelect from '../ft-select/ft-select.vue'
 import FtInput from '../ft-input/ft-input.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
 import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
-import i18n from '../../i18n/index.js'
 
 export default Vue.extend({
   name: 'ExternalPlayerSettings',
@@ -25,7 +24,7 @@ export default Vue.extend({
     },
 
     externalPlayerNames: function () {
-      return this.$store.getters.getExternalPlayerNames.map((value) => i18n.t(value))
+      return this.$store.getters.getExternalPlayerNames.map((value) => this.$t(value))
     },
     externalPlayerValues: function () {
       return this.$store.getters.getExternalPlayerValues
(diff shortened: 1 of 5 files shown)


PikachuEXE (approved):

src/renderer/components/external-player-settings/external-player-settings.js
```diff
@@ -5,6 +5,7 @@ import FtSelect from '../ft-select/ft-select.vue'
 import FtInput from '../ft-input/ft-input.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
 import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
+import i18n from '../../i18n/index.js'
```
PrestonN, line 8: All Vue components have access to a global `i18n` object without the need to import. Please use that instead. It can be accessed with `this.$t`. Please check some other components for more information on how it's used.

PrestonN (changes requested):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
@PrestonN ?
The import of `i18n` is removed already

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
I think @PrestonN is blocking this merge

static/external-player-map.json
```diff
@@ -1,6 +1,6 @@
 [
     {
-        "name": "None",
+        "name": "Settings.External Player Settings.None",
```
ChunkyProgrammer, line 3: maybe a different key should be added called "translated" that is a bool value and the translating part of the code will only use the "$t" function if translated is true to avoid warnings from this PR

PikachuEXE: Now I notice that the other names are not translated

PikachuEXE: Another way I can think of would be 
```js
this.$te(value) ? this.$t(value) : this.$t(value)
```
[$te checks whether the translation key exists](https://kazupon.github.io/vue-i18n/api/#te)

ChunkyProgrammer: > Another way I can think of would be
> 
> ```js
> this.$te(value) ? this.$t(value) : value
> ```
> 
> [$te checks whether the translation key exists](https://kazupon.github.io/vue-i18n/api/#te)

that's probably the better way to do it. was not aware of the te function

PikachuEXE (changes requested):
Let's use `$te` then

9ekaitz:
> Let's use `$te` then
Should I name the null player "None" or keep the changes and name it with the translation key "Settings.External Player Settings.None"?

PikachuEXE:
I think the `"name": "Settings.External Player Settings.None"` can be kept
And expect it to keep either the name, or the translation key
If there is no translation entry for the value, we assume that's the name (without translations)
And to do that we update
`return this.$store.getters.getExternalPlayerNames` to
```js
return this.$store.getters.getExternalPlayerNames.map((value) => this.$te(value) ? this.$t(value) : value)
```
(Please test before committing, coz I haven't :P)

PikachuEXE:
I made some changes on my own branch
https://github.com/PikachuEXE/FreeTube/tree/feature/translated-player-name

Some screenshots:
When using zh-tw (with test translation `Nein`)
image

When using jp (without translation)
image

When using en (with translation)
image

Let me know what you think

PikachuEXE (approved):
Tested locally
Took me sometime to figure which one is `es`

ChunkyProgrammer (approved):

absidue (approved):

ChunkyProgrammer:
@PrestonN looks like you might need to approve this one for it to get merged

PrestonN (approved):

PR #2108, opened 2022-03-05
Prevent search input reset from overwriting clipboard on Linux

### LINKED ISSUE #2019: [Feature Request]: Reduce clipboard aggression of search field
### Guidelines

## Steps to Reproduce 

1. Use KDE/Plasma (Linux)
2. Use the default Plasma clipboard manager
3. In clipboard manager config, set "Synchronize contents of the clipboard and the selection" to on (checked)
1. Have some text in the search/URL field of Freetube already, e.g. `abc`
2. Go to another application, copy something, like a YouTube URL `https://youtube.com/foo`
3. Return to Freetube
4. Try to paste that URL in the search/URL field of Freetube

## Actual Behaviour

The previous text in the Freetube field is pasted (`abc`)

## Expected Behaviour

The text the user wants to paste is pasted (`https://youtube.com/foo`)

## Details

It seems that Freetube is very aggressive in re-copying the current field contents to the clipboard.  Even using the (X) button to clear the field copies the contents to the clipboard, which is kind of bizarre to me, because I don't know any other app in the history of the world which copies to clipboard when you use a UI element to clear a field.  I guess you guys have coded it to be a "select all" followed by "delete".

Maybe other people like the current behaviour, and that's fine, but please make a config setting so that Freetube isn't so aggressive with the clipboard.

(And making people change their desktop/OS settings is not a solution -- no, I do not want to change the configuration of my clipboard manager just to make it play nice with Freetube, while sacrificing the desired behaviour in every other application I use)

### Proposed Solution

Add a config setting to FreeTube so that clearing the field does not interact with the OS clipboard

### Alternatives Considered

* I could change the way I work with my OS, DE, and clipboard manager, but that is too much of an ask
* Clear the field in FreeTube first, then go copy the other thing

### Issue Labels

ease of use improvement, new optional setting

### Additional Information

_No response_

vallode:
---
Prevent search input reset from overwriting clipboard on Linux
---

**Pull Request Type**
- [x] Bugfix

**Related issue**
Closes #2019

**Description**
Removes the input event emitter in favour of a more standard way of clearing an input field to prevent clipboards being overwritten.

Test with any form of selection-based clipboards (middle mouse button on most Linux systems). Select any text, use middle mouse button to paste, type something else into FreeTube search, click the X button to clear it, paste again.

 - OS: [e.g. iOS] Debian
 - OS Version: [e.g. 22] 11
 - FreeTube version: [e.g. 0.8] v0.16.0 Beta

**Additional context**
https://github.com/FreeTubeApp/FreeTube/issues/2019#issuecomment-1019418085 user confirms that this is the expected behaviour for the X button. Cheers

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -139,10 +139,11 @@ export default Vue.extend({
       this.inputData = ''
       this.handleActionIconChange()
       this.updateVisibleDataList()
-      this.$emit('input', this.inputData)
+
+      const inputElement = document.getElementById(this.id)
+      inputElement.value = ''
 
       // Focus on input element after text is clear for better UX
-      const inputElement = document.getElementById(this.id)
       inputElement.focus()
     },
 


ChunkyProgrammer (approved):
Lgtm

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

Pistos:
Thank you so much for this. I've built (the `development` branch), and confirmed that the fix works for me.

(Thank you also for not just dismissing the issue as "works for me" or "user error".)

PikachuEXE:
There are so many platforms and no way we can be sure things work for all platforms well.
I only use Windows & MacOS so I would never dismiss Linux related issues.
We should always admit that there is **a possibility of making mistakes** and fix those mistakes when they appear (if worth fixing & can be fixed).

PR #2173, opened 2022-04-08
Restore subtitle player setting

### LINKED ISSUE #62: Enable subtitles by default
Please add an option to enable subtitles by default.

vallode:
---
Restore subtitle player setting
---

**Pull Request Type**
- [x] Feature Implementation

**Related issue**
Closes #62 

**Description**
Restores a setting in the player settings section for enabling subtitles/captions on by default

 - OS: [e.g. iOS] Debian
 - OS Version: [e.g. 22] 11
 - FreeTube version: [e.g. 0.8] 0.16.0

**Additional context**
I disabled the setting by default because I think it's better not to confuse users with subtitles always being on before going through the settings.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -367,7 +367,9 @@ export default Vue.extend({
 
             if (this.enableSubtitles) {
               const tracks = this.player.textTracks().tracks_
-              const userLocaleTrack = tracks.find(track => track.language == this.currentLocale)
+              const userLocaleTrack = tracks.find((track) => {
+                return track.language === this.currentLocale
+              })
 
               if (userLocaleTrack) {
                 userLocaleTrack.mode = 'showing'
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -165,6 +165,10 @@ export default Vue.extend({
       return parseInt(this.$store.getters.getDefaultQuality)
     },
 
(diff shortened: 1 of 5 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
1. Could u move the setting beneath Display Play Button In Video Player. I think that is a better place for it.

2. While testing on this (https://youtu.be/J6KeGIWZWJgIf) video i discovered that if it cant find the language it will default to the first one in the array and in this case that is Afrikaans. I dont think it is very useful to the user to display a subtitle that is not their language. We could do a few thing to address this. Check if there is auto generated captions available in that language, if that is not available dont turn on captions on that video. 

3. if we implement 2 then i think its fair to put a warning next to the toggle. Subtitles will only be shown when available in your language.

4. Testing on this video https://youtu.be/MkjPvfCSoWI. set ur language to deutsch > go to video > u only see one type of caption (german autogenerated) > change language to spanish > go to video > check captions > see that it generated a new spanish caption

vallode:
@absidue @PikachuEXE if you could see above and comment that'd be great :)

PikachuEXE:
I think the logic for caption options generation is outside of this PR (I think it's done in `transformAndInsertCaptions`)

efb4f5ff-1298-471a-8973-3d47447115dc:
Currently i'm trying to find captions on videos for all our locales to test out.

efb4f5ff-1298-471a-8973-3d47447115dc:
@vallode wanted to test this but keeps spitting out annoying errors. Could u merge your branch with latest dev branch

PR #2176, opened 2022-04-08
Bump marked and electron-context-menu

absidue:
---
Bump marked and electron-context-menu
---

**Pull Request Type**

- [x] Feature Implementation

**Description**

Bump `marked` and `electron-context-menu` to their latest versions using `yarn upgrade marked electron-context-menu`.

FreeTube has a lot of outdated dependencies (run `yarn outdated` in your terminal to see the full list), this pull request is a tiny start to the process of upgrading the dependencies to newer versions.

I tested this pull request for both of the dependencies:
- `electron-context-menu`: I checked all the context menus I could think of and they all worked fine.
- `marked`: I changed the version in the `package.json` file to `"0.15.0"` to force the changelog to show up and checked that it looked they way it did before upgrading `marked`.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 06a7298ded743c1db59f6b946ed038f996bd3340


efb4f5ff-1298-471a-8973-3d47447115dc:
isnt it weird that our dependencies bot didnt create a pr for this?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer:
> isnt it weird that our dependencies bot didnt create a pr for this?

I believe dependabot is configured only for certain vulnerabilities. Would have to check with Preston to verify though

PrestonN (approved):
In the past, I've been upgrading dependencies all in one go, but I've slowed down with doing this because of issues usually coming up because of it. The biggest packages I can think of that need extra care are Video.js and Electron as those likely need changes to accommodate for the newer releases.

The other one too is the core `Vue` package and it's associated packages as I believe they have finally updated to Vue 3. We'll need to see what needs to be converted if we decide to upgrade to Vue 3 and this could potentially be a lengthy process as well.

Feel free to continue upgrading some of the packages if you'd like. I could also setup a PR with most packages updated and we can use that for testing. I'm fine with whatever works best.

absidue:
I'll probably create separate pull requests instead of upgrading them all at once, that should hopefully make it easier to find out when things break and allow them to be rolled back if we don't want to fix the issues immediately. I'll group things like babel and eslint together. I agree that Vue, Electron and video.js need extra care as updates for those are likely to break stuff (e.g. latest video.js breaks the quality selector and a few other minor things that are styling related).

I won't be upgrading some dependencies like the proxy ones, as i don't have or use a proxy so i won't be able to test it. I may also have to leave things like upgrading Vue up to someone else, as I don't know anything about Vue and I'm expecting major things to break if that's upgraded.

PR #2208, opened 2022-04-17
Allow DASH formats for livestream replays

absidue:
---
Allow DASH formats for livestream replays
---

**Pull Request Type**

- [x] Bugfix

**Description**

Currently FreeTube enforces the legacy formats for all livestreams regardless of whether they are live or replays. This pull requests allows DASH formats to be used for livestream replays.

Before:
before

After:
after

I tested various videos to check that nothing broke. Here is an example of a livestream replay video that has DASH formats after this change: https://youtu.be/TEygSwHWhfA

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 7878e9e3a671830426bc0893a44c1baff5801cb9

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -310,7 +310,7 @@ export default Vue.extend({
             }
           }
 
-          if ((this.isLive || this.isLiveContent) && !this.isUpcoming) {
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):
LGTM

PikachuEXE (approved):

PR #2234, opened 2022-05-07
Always show general settings when visiting settings page

### LINKED ISSUE #1742: Automatically open general tab when visiting settings
It is nicer to see the general tab automatically appear because it contains settings that are commonly changed.

It would be cool for the general tab to open automatically so the user is brought to the most changed settings tab without needing to click.

Opening it manually.

**Screenshots**
image

laazyCmd:
---
Always show general settings when visiting settings page
---

**Pull Request Type**
- [ ] Bugfix
- [X] Feature Implementation

**Related issue**
closes #1742

**Description**
- I've set `general-settings.vue` to always open

(I added a gif instead, since I think it's more easier to see the result)
FreeTube 1742

 - OS: Windows 10
 - OS Version: 21H2
 - FreeTube version: 0.16.0


efb4f5ff-1298-471a-8973-3d47447115dc:
Q: @PikachuEXE @PrestonN @ChunkyProgrammer want this? This seems to be a worse UX then our current implementation. Lets say u almost never use the general settings. Now u have to close it every time when u go to the settings because its taking up screen space.

PikachuEXE:
Ya this is quite difficult to decide
Desktop & mobile would have different UX...
Different people want it closed/opened by default...

For me I can think of 2 things instead
- Remembering each section is opened/closed last time
  - If you are testing settings change this is useful
- Collapse/Expand all button
  - For people who wants to see all/see none quickly

PrestonN:
I think the better approach would be to redesign how we currently have our settings page setup. Right now, the settings layout works alright for both desktop and mobile, but it isn't great for either one. It'd likely be better to go towards a [GNOME Settings](https://apps.gnome.org/app/org.gnome.Settings/) sort of approach / layout for our settings page to better accommodate for both sides. The categories and content would still be the same, but the way that we present them would be slightly different.

I don't think that having a tab open by default (at least in this method) is the best approach, so I'll go ahead and close this PR. If someone wants to convert the current layout to something similar to GNOME Settings before I have a chance to do so then I'd be more open to merging that PR.

PR #2244, opened 2022-05-16
Format playlist dates with the selected locale

absidue:
---
Format playlist dates with the selected locale
---

**Pull Request Type**

- [x] Feature Implementation

**Description**

Currently FreeTube uses the dateFormat function provided by the [dateformat](https://github.com/felixge/node-dateformat) dependency to format the date for playlists when the Invidious API is selected as a backend. It does a good job of formatting the date however it doesn't respect the users selected locale, to fix this we can use the `Date.prototype.toLocaleDateString()` function that is built into JavaScript. The only downside is that we loose the ordinals (e.g. `9th` -> `9`).

This also fixes the date having 1 day added on to it (not sure why that was there in the first place).

I used German for the screenshots as the difference is clearly visible on the screenshots, it works with the other locales as well, I just didn't screenshot them.

before with the locale set to German:

before_german

after with the locale set to German:

after_german

after with the locale set to English US:

after_en_us

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 85bbffaf6c25b89f666df4cefee81bb456f03dc1

### DIFF
--- a/src/renderer/store/modules/ytdl.js
+++ b/src/renderer/store/modules/ytdl.js
@@ -288,10 +288,10 @@ const actions = {
             break
         }
       }
-      const locale = settings.currentLocale.replace('-', '_')
+      const locale = settings.currentLocale.replace('_', '-')
       ytpl(playlistId, {
         hl: locale,
-        limit: 'Infinity',
+        limit: Infinity,
         requestOptions: { agent }
       }).then((result) => {
         resolve(result)
--- a/src/renderer/store/modules/ytdl.js
+++ b/src/renderer/store/modules/ytdl.js
@@ -288,7 +288,12 @@ const actions = {
             break
         }
       }
-      const locale = settings.currentLocale.replace('_', '-')
+      let locale = settings.currentLocale.replace('_', '-')
+
+      if (locale === 'nn') {
+        locale = 'no'
+      }
+
       ytpl(playlistId, {
         hl: locale,
         limit: Infinity,
--- a/src/renderer/store/modules/ytdl.js
+++ b/src/renderer/store/modules/ytdl.js
@@ -7,6 +7,8 @@ import { SocksProxyAgent } from 'socks-proxy-agent'
 import { HttpsProxyAgent } from 'https-proxy-agent'
 import { HttpProxyAgent } from 'http-proxy-agent'
 
+import i18n from '../../i18n/index'
+
 const state = {
   isYtSearchRunning: false
 }
@@ -288,7 +290,7 @@ const actions = {
             break
         }
       }
-      let locale = settings.currentLocale.replace('_', '-')
+      let locale = i18n.locale.replace('_', '-')
 
       if (locale === 'nn') {
         locale = 'no'
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -5,6 +5,7 @@ import FtCard from '../../components/ft-card/ft-card.vue'
 import PlaylistInfo from '../../components/playlist-info/playlist-info.vue'
 import FtListVideo from '../../components/ft-list-video/ft-list-video.vue'
 import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
+import i18n from '../../i18n/index'
 
 export default Vue.extend({
   name: 'Playlist',
@@ -37,7 +38,7 @@ export default Vue.extend({
       return this.$store.getters.getCurrentInvidiousInstance
     },
     currentLocale: function () {
-      return this.$store.getters.getCurrentLocale.replace('_', '-')
+      return i18n.locale.replace('_', '-')
     }
   },
   watch: {
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -1,6 +1,5 @@
 import Vue from 'vue'
 import { mapActions } from 'vuex'
-import dateFormat from 'dateformat'
 import FtLoader from '../../components/ft-loader/ft-loader.vue'
 import FtCard from '../../components/ft-card/ft-card.vue'
 import PlaylistInfo from '../../components/playlist-info/playlist-info.vue'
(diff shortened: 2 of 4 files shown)


PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
cant test because i got rate limited on the Invidious API -.-

But i saw something interesting with the local api. the language of the date was shown based on the ip address of the country im in. So if my sever is in Germany the date in the description will be German instead of the language i set in the settings.

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc Select a different Invidious instance, most of them failed for me too while testing these changes but eventually (took more than 10 tries) I managed to get lucky enough that FreeTube picked an instance that was working.

The issue with the local API is unrelated but i'll try to fix it in the pull request as well. ytpl allows you to set the location and the language, FreeTube currently sets the location, i'll change it so sets the language instead.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

absidue:
Actually FreeTube does set the language but it does it wrongly, YouTube wants the locale to be `en-GB` but for some reason FreeTube replaces the `-` with an `_` which means FreeTube sends `en_GB` to YouTube, which YouTube sees as an invalid locale so YouTube chooses the locale based on the IP address.

While checking that I also noticed that FreeTube doesn't pass the chosen location on to YouTube (search, playlists and videos) so not sure why that setting is there unless it was just never full implemented.

efb4f5ff-1298-471a-8973-3d47447115dc:
Do u want me to create an issue for this so we can merge over this PR. Or would u rather fix it in this PR?

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc Commited it to this pull request, please review it again.

ChunkyProgrammer (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
Checked out all the languages just to be sure

Retrieves language based on IP if u switch to the following languages

- Norsk nynorsk
- system default

absidue:
That's because the closest YouTube has is "Norsk" which is `no`, however FreeTube has "Norsk nynorsk" (`nn`) and "Norsk bokmål" (`nb_NO`) so YouTube has no idea what `nn` is. I don't know how different "Norsk" and "Norsk nynorsk" are, if they are close enough I can add a special case to the code which sends `no` to YouTube if the language is set to `nn`.

efb4f5ff-1298-471a-8973-3d47447115dc:
> That's because the closest YouTube has is "Norsk" which is `no`, however FreeTube has "Norsk nynorsk" (`nn`) and "Norsk bokmål" (`nb_NO`) so YouTube has no idea what `nn` is. I don't know how different "Norsk" and "Norsk nynorsk" are, if they are close enough I can add a special case to the code which sends `no` to YouTube if the language is set to `nn`.

Quick search on the web leads me to believe that it is the same

absidue:
Added special handling for Norsk nynorsk.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Checked out all the languages just to be sure
> 
> Retrieves language based on IP if u switch to the following languages
> 
>     * Norsk nynorsk
> 
>     * system default

So if my IP is from Germany but the default language is English on my system. It will show me a German format instead of the System language. Which is understandable because YT doesnt have a langaguage thats called ```system default```

absidue:
FreeTube doesn't actually use the system language it picks the closest one to your system language and defaults to `en-US` if it doesn't have localisations for your system language ([`src/renderer/store/modules/settings.js` lines 232-245](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/store/modules/settings.js#L232-L245)). Maybe YouTube doesn't accept `en-US` as a valid language. If it doesn't like it then there are probably also a bunch of others it doesn't like, in which case I will need to add a list of which YouTube language should be used for each FreeTube language.

As I don't have access to your computer and your specific setup you might want to test it by adding `console.log(locale)` to [`src/renderer/store/modules/ytdl.js` line 296](https://github.com/absidue/FreeTube/blob/048379587d8464d88ec9aab464a5d7df87c20109/src/renderer/store/modules/ytdl.js#L296) and check what it logs in the devtools console for you.

PikachuEXE (approved):
Tested with Traditional Chinese

Should I test Norsk nynorsk too?

Playlist for testing:
- https://youtube.com/playlist?list=PLTariUrSxayszFKmTdIyty9MI28gFILvb

efb4f5ff-1298-471a-8973-3d47447115dc:
> Should I test Norsk nynorsk too?

@PikachuEXE just tested it and it works

ChunkyProgrammer (approved):
This issue with system happens elsewhere in the code so i am okay with approving the PR as is and having another PR fix the locales here & elsewhere

ChunkyProgrammer:
Maybe include 
`import i18n from '../../i18n/index'
`
at the top
and then get the current locale with:
`i18n.locale`

src/renderer/store/modules/ytdl.js
```diff
@@ -288,10 +288,15 @@ const actions = {
             break
         }
       }
-      const locale = settings.currentLocale.replace('-', '_')
+      let locale = settings.currentLocale.replace('_', '-')
```
ChunkyProgrammer, line 291: Maybe include
import i18n from '../../i18n/index'
at the top
and then get the current locale with:
i18n.locale

src/renderer/views/Playlist/Playlist.js
```diff
@@ -36,6 +35,9 @@ export default Vue.extend({
     },
     currentInvidiousInstance: function () {
       return this.$store.getters.getCurrentInvidiousInstance
+    },
+    currentLocale: function () {
+      return this.$store.getters.getCurrentLocale.replace('_', '-')
```
ChunkyProgrammer, line 40: Maybe include
import i18n from '../../i18n/index'
at the top
and then get the current locale with:
i18n.locale

ChunkyProgrammer (changes requested):

absidue:
I've updated the PR with @ChunkyProgrammer's requested changes

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #2276, opened 2022-05-30
Use the built-in subtle crypto library instead of node-forge

absidue:
---
Use the subtle crypto library built-in to Electron instead of node-forge
---

**Pull Request Type**

- [x] Feature Implementation

**Description**

Currently FreeTube uses the [node-forge](https://www.npmjs.com/package/node-forge) library to generate the hash prefix of the video id that it sends to the SponsorBlock API. This pull request removes the need for that dependency by using the [digest](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) function provided by the SubtleCrypto library which is built into all modern webrowsers (added to Chrome in v37, FreeTube uses Electron 16 -> Chrome 96).

I tested various different videos and it worked the same as it did before this pull request.

Example videos: https://youtu.be/qjw8ohwZ4nY and https://youtu.be/2ehSCWoaOqQ

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0ca42129341af43697418b6510e18207adb91f2c

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,6 @@
     "marked": "^4.0.15",
     "material-design-icons": "^3.0.1",
     "nedb-promises": "^5.0.1",
-    "node-forge": "^1.3.1",
     "opml-to-json": "^1.0.1",
     "rss-parser": "^3.12.0",
     "socks-proxy-agent": "^6.0.0",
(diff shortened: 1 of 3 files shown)


ChunkyProgrammer (approved):
Tested using SponsorBlock using this branch and it still worked

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested example videos in MacOS

PR #2284, opened 2022-06-01
Remove usage of fixed positioning for core UI elements

vallode:
---
Remove usage of fixed positioning for core UI elements
---
**Pull Request Type**
- [x] Bugfix

**Related issue**
N/A

**Description**
This PR attempts to remove the usage of margins for positioning the UI elements (routerView, top-nav, side-nav, etc.) and replace it with a more modern approach using flexbox and sticky positioning.

* Remove position: fixed from top-nav and side-nav
* Reduce z-index position of side-nav to make it not overlap the box-shadow on the top-nav
* Make video infoArea sticky so it follows users scroll (in the case there are many suggested videos)

Before:
image

After:
image

Before:
image

After:
image

I ran through most of the app, making sure that I checked as many views as I could.

 - OS: [e.g. iOS] 10
 - OS Version: [e.g. 22] Debian
 - FreeTube version: [e.g. 0.8] upstream/development

**Additional context**
Very open to discuss about this PR and how it should work. For now I just wanted to get rid of some of the CSS-smell that we had with the margin positioning that I feel were causing me headaches when working on the rest of the app.

### DIFF
--- a/src/renderer/components/side-nav/side-nav.css
+++ b/src/renderer/components/side-nav/side-nav.css
@@ -167,10 +167,11 @@
   }
 
   .sideNav {
-    bottom: 0;
-    display: flex;
-    left: 0;
     position: fixed;
+    left: 0;
+    bottom: 0;
+
+    display: flex;
   }
 
   .topNavOption {
--- a/src/renderer/components/side-nav/side-nav.css
+++ b/src/renderer/components/side-nav/side-nav.css
@@ -3,11 +3,10 @@
   height: calc(100vh - 60px);
   width: 200px;
   overflow-x: hidden;
-  position: fixed;
-  left: 0px;
-  top: 0px;
-  z-index: 4;
-  margin-top: 60px;
+  position: sticky;
+  left: 0;
+  top: 60px;
+  z-index: 3;
   box-shadow: 1px -1px 1px -1px var(--primary-shadow-color);
   background-color: var(--side-nav-color);
   transition-property: width;
@@ -168,7 +167,10 @@
   }
 
   .sideNav {
+    bottom: 0;
     display: flex;
+    left: 0;
+    position: fixed;
   }
 
   .topNavOption {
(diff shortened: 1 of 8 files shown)


src/renderer/views/Watch/Watch.sass
```diff
@@ -65,6 +65,8 @@
     grid-column: 1
 
   .infoArea
+    position: sticky
```
absidue, line 68: When the window is narrow (e.g. phones and tablets) the `.infoArea` and `.sidebarArea` are vertically stacked instead of being next to each other. This sticky should only be in effect when the `.infoArea` and `.sidebarArea` are next to each other as it causes weird overlay scrolling when those components are vertically stacked.

weird-scrolling

absidue (changes requested):

src/renderer/views/Watch/Watch.sass
vallode: Well spotted, fixed!

absidue (dismissed):
LGTM

PikachuEXE (commented):
Test results:

### Remove position: fixed from top-nav and side-nav
Not sure what to test I mean they look fine but I haven't change settings about these components

### Reduce z-index position of side-nav to make it not overlap the box-shadow on the top-nav
Compared the appearance of side-nav with screenshot, same as claimed

### Make video infoArea sticky so it follows users scroll (in the case there are many suggested videos)
Tested with wide & not so wide & narrow window size
No issue found

Still need to review the code

src/renderer/components/side-nav/side-nav.css
```diff
@@ -168,7 +167,10 @@
   }
 
   .sideNav {
+    bottom: 0;
     display: flex;
+    left: 0;
+    position: fixed;
```
PikachuEXE, line 173: Can we group the properties for positioning together?

```css
    position: fixed;
    left: 0;
    bottom: 0;

    display: flex;
```
The example code above is like talking to the renderer like this:
- For positioning, I want it to be fixed, 0 spacing to left and bottom
- For display behavior, I want flex

PikachuEXE (changes requested):

src/renderer/components/side-nav/side-nav.css
vallode: Personally I always alphabetically sort CSS properties (with very few exceptions), I find it to be the most reproducible method of property sorting. Implemented your change but I would argue for a future where we set up a linter for the CSS and not have to worry about this ;p

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

absidue:
@vallode Forcing the info area to be sticky has broken scrolling in the downloads list, please open a new PR to fix that issue.

PR #2286, opened 2022-06-01
Fix the redundant navigation error

absidue:
---
Fix the redundant navigation error
---

**Pull Request Type**

- [x] Bugfix

**Description**

Fixes the issue where spamming the nav buttons throws vue errors about redundant navigations.

duplicate-navigation-error

I spammed the nav buttons and the error didn't occur anymore.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 7287e0011ee21cb6d827c979b4cff2a44bc2ee6b

**Additional context**

### DIFF
--- a/src/renderer/router/index.js
+++ b/src/renderer/router/index.js
@@ -26,31 +26,30 @@ class CustomRouter extends Router {
   push(location) {
     // only navigates if the location is not identical to the current location
 
+    const currentQueryUSP = new URLSearchParams(router.currentRoute.query)
+    let newPath = ''
+    let newQueryUSP = new URLSearchParams()
+
     if (typeof location === 'string') {
       if (location.includes('?')) {
         const urlParts = location.split('?')
-
-        if (
-          urlParts[0] !== router.currentRoute.path ||
-          new URLSearchParams(urlParts[1]) !==
-            new URLSearchParams(router.currentRoute.query)
-        ) {
-          return super.push(location)
-        }
-      } else if (
-        location !== router.currentRoute.path &&
-        isQueryEmpty(router.currentRoute.query)
-      ) {
-        return super.push(location)
+        newPath = urlParts[0]
+        newQueryUSP = new URLSearchParams(urlParts[1])
+      } else {
+        newPath = location
+        // newQueryUSP already empty
       }
     } else {
-      if (
-        location.path !== router.currentRoute.path &&
-        (isQueryEmpty(location.query) ||
-          location.query !== router.currentRoute.query)
-      ) {
-        return super.push(location)
-      }
+      newPath = location.path
+      newQueryUSP = new URLSearchParams(location.query)
+    }
+
+    const pathsAreDiff = router.currentRoute.path !== newPath
+    // Comparing `URLSearchParams` objects directly will always be different
+    const queriesAreDiff = newQueryUSP.toString() !== currentQueryUSP.toString()
+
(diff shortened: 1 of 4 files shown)


ChunkyProgrammer:
Error still appears when clicking the profile setting icon
image

ChunkyProgrammer:
Was able to have this error happen again when searching (search the same thing twice)

vallode:
Potentially we need to hijack the prototype of the vue-router push function here? I mean is it feasible to expect every single next router interaction to always remember to put this if statement in place?

absidue:
I thought about that but I haven't looked into the source code for that yet (if you want to take a look remember to look at the source code for the version that FreeTube uses). Although I'll need to expand the checks to also check the query parameters etc.

absidue:
I've now removed the individual checks and wrapped the router push function in one that checks the path and query before navigating.

PikachuEXE:
Can we have a summary of what to be tested?

PikachuEXE (commented):
Tested
- spamming the nav buttons
- clicking the profile select's setting icon
- search the same thing twice
- Paste a YT Video URL to navbar, press arrow button multiple times

src/renderer/router/index.js
```diff
@@ -14,9 +14,50 @@ import Playlist from '../views/Playlist/Playlist.vue'
...
+      }
+    } else {
+      if (
+        location.path !== router.currentRoute.path &&
+        (isQueryEmpty(location.query) ||
+          location.query !== router.currentRoute.query)
+      ) {
+        return super.push(location)
+      }
+    }
```
PikachuEXE, line 54:
```suggestion
    const currentQueryUSP = new URLSearchParams(router.currentRoute.query)
    let newPath = ''
    let newQueryUSP = new URLSearchParams()

    if (typeof location === 'string') {
      if (location.includes('?')) {
        const urlParts = location.split('?')
        newPath = urlParts[0]
        newQueryUSP = new URLSearchParams(urlParts[1])
      } else {
        newPath = location
        // newQueryUSP already empty
      }
    } else {
      newPath = location.path
      newQueryUSP = new URLSearchParams(location.query)
    }

    const pathsAreDiff = router.currentRoute.path !== newPath
    // Comparing `URLSearchParams` objects directly will always be different
    const queriesAreDiff = newQueryUSP.toString() !== currentQueryUSP.toString()

    if (pathsAreDiff || queriesAreDiff) {
      return super.push(location)
    }
```

PikachuEXE (changes requested):
Cleanup duplicate statements (especially conditional)

`isQueryEmpty` can also be removed

PikachuEXE (approved):

src/renderer/router/index.js
PikachuEXE: Irrelevant now but I think
```js
location.path !== router.currentRoute.path &&
        (isQueryEmpty(location.query) ||
          location.query !== router.currentRoute.query)
```
is a wrong conditional statement
Upper conditional statements are like (new path != cur path **or** new query != cur query) (I think)
This one is (new path != cur path **and** new query != cur query) ?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer sorry for the spam :D

PR #2303, opened 2022-06-09
Prevent packaging unneeded dependency files

absidue:
---
Prevent packaging unneeded dependency files
---

**Pull Request Type**

- [x] Feature Implementation

**Description**
FreeTube's dependencies ship a lot of files that might be useful during development, these files aren't useful for users of FreeTube. As we don't have any control over the majority of the dependencies, we can exclude the files during the electron build and reduce the package size (`resources/app.asar`: 54.2MB -> ~49.5MB~ 35.2MB). This pull request also fixes the webpack builds no longer properly ignoring dotfiles.

I tested this by running `yarn run build` and then checking the files produced by the webpack builds in the `dist/` directory and the files produced by the electron build in the `build/` directory. I also clicked around FreeTube just to make sure that no required files were accidentally blacklisted.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 26e4433c999d2c32d925e2b2dae5d1da4886e6e3

### DIFF
--- a/_scripts/build.js
+++ b/_scripts/build.js
@@ -57,13 +57,22 @@ const config = {
     '_icons/iconColor.*',
     'icon.svg',
     './dist/**/*',
-    '!./dist/web/**/*',
+    '!dist/web/*',
+    '!**/node_modules/**/.*',
     '!**/{.github,Jenkinsfile}',
-    '!**/{CHANGES.md,CODE_OF_CONDUCT.md,CONTRIBUTING.md,DEVELOPMENT.md,docs,docs.md,examples,History.md,HISTORY.md,README.md,TODO.md,UPGRADE_GUIDE.md,UPGRADING.md}',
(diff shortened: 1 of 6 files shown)


PikachuEXE:
Does this affect `npm run dev/debug` too?
Anyway I probably just include this change in my custom build and use for a week as a test

absidue:
Managed to reduce the size even further :).

absidue:
Yes, this pull request does affect `npm/yarn run dev/debug`, only the changes to the webpack configuration though. The changes to `_scripts/build.js` only affect production builds triggered with `npm/yarn run build`.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Everything still works like a charm

GilgusMaximus (approved):
yup can confirm it working on my machine as well and code changes look good to me

ChunkyProgrammer (approved):
LGTM

PR #2359, opened 2022-06-25
Upgrade electron from 16 > 18

PikachuEXE:
**Pull Request Type**
Please select what type of pull request this is:
- [x] Feature Implementation

**Related issue**
Might or might not fix issue in #2113

Closes #2358

**Description**
16.x is not supported anymore
From https://www.electronjs.org/blog/electron-17-0
image

[Breaking change in 17.x](https://www.electronjs.org/blog/electron-17-0#breaking-changes) = 
- [`desktopCapturer.getSources`](https://www.electronjs.org/blog/electron-17-0#desktopcapturergetsources-in-the-renderer)
- The end

[Breaking change in 18.x](https://www.electronjs.org/blog/electron-18-0#breaking-changes) = 
- [Removed `nativeWindowOpen`](https://www.electronjs.org/blog/electron-18-0#removed-nativewindowopen)
- The end

Should be safe to upgrade if true

**Screenshots (if appropriate)**
N/A

**Testing (for code that is not small enough to be easily understandable)**
Open app, open new window via
- Keyboard shortcut
- New window button
- Middle click on links

**Desktop (please complete the following information):**
 - OS: MacOS
 - OS Version: 12.4
 - FreeTube version: 63442282a9a824ba77ed1c0d551213ff3e6abf99

**Additional context**

PR #2360, opened 2022-06-25
Use native web browser features instead of jquery

absidue:
---
Use native webbrowser features instead of jquery
---

**Pull Request Type**

- [x] Feature Implementation

**Related issue**
Please link the issue your pull request is referring to. If this pull request fully resolves the relevant issue, put "closes" before the issue number. Example: "closes #123456".

**Description**
Due to jquery's popularity most of jquery's features have now become web standards and are implemented into web browsers natively. This pull request replaces all uses of jquery in FreeTube with their web standards counter parts.

Here are some of the changes that I made:

replace `$.getJson()` with `fetch()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch))
```js
$.getJSON(requestUrl, (response) => {
  // ...
}).fail((xhr, textStatus, error) => {
  // ...
})

fetch(requestUrl)
  .then((response) => response.json())
  .then((json) => {
    // ...
  })
  .catch((error) => {
    // ...
  })
```

replace `$()` and `$().get(0)` with `document.getElementById()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById)), `document.querySelector()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector)) and `document.querySelectorAll()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll))
```js
$('#theId') /* -> */ document.getElementById('theId')

// selecting the first element matching a css query
$('.class').get(0) /* -> */ document.querySelector('.class')

// selecting all elements that match a css query
$('.class') /* -> */ document.querySelectorAll('.class')
```

adding event listeners:
replace `.on()` with `.addEventListener()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener))
```js
$('#profileList').on('focusout', (event) =>{
  // ...
})

document.getElementById('profileList').addEventListener('focusout', (event) => {
  // ...
})
```

replace `.addClass()` and `.removeClass()` with `.classList.add()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/add)) and `.classList.remove()` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/remove))
```js
$('#loopButton').removeClass('vjs-icon-loop-active')
$('#loopButton').removeClass('loop-black')
$('#loopButton').addClass('loop-white')

const loopButton = document.getElementById('loopButton')
loopButton.classList.remove('vjs-icon-loop-active')
loopButton.classList.remove('loop-black')
loopButton.classList.add('loop-white')
```

replace `.animate({ scrollTop: /* ... */})` with `.scrollTo({ top: /* ... */ })` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo))
```js
liveChatComments.animate({ scrollTop: liveChatComments.prop('scrollHeight') })
```

```js
liveChatComments.scrollTo({ top: liveChatComments.scrollHeight })
```
```css
@media (prefers-reduced-motion: no-preference) {
  .liveChatComments {
    scroll-behavior: smooth;
  }
}
```

**Testing**
Everything should work the same as it did before this PR. It's a good idea to have an instance of the nightly build open, so you can compare the before and after.

* checking for updates
    1. change the version to `0.15.0` in the `package.json` file
    2. check that the update available banner shows up
    3. check that the changelog shows up
* keyboard shortcuts (e.g. ALT+D to focus the search bar)
* clicking links in video descriptions, comments etc
* profile selector
    1. check that it opens correctly
    2. check that it closes when it looses focus
    3. check that it closes when a profile is selected
* video player
    * keyboard shortcuts
    * the highlighted quality is corrent and changes correctly in the quality selector
    * check that the loop, full window, theatre and screenshot buttons look and function as expected
    * captions
    * sponsor block segments
* Invidious API as backend
* history navigation arrows get updated correctly
* the search bar gets hidden/shown on mobile/in narrow windows
* proxy test button in settings
* live chat
    * set the backend to the local API (invidious API doesn't support live chat)
    * find a live stream (some are listed here https://www.youtube.com/live)
    * the live chat will auto scroll as new comments come it

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 63442282a9a824ba77ed1c0d551213ff3e6abf99

**Additional context**
The live chat stops smooth scrolling after a while, this seems to be a chromium/electron limitation, presumably some kind of power saving measure. Thankfully it keeps scrolling, it just jumps to the latest message instead of animating the scroll, manually scrolling still works smoothly, so this only affects scrolling programatically.


ChunkyProgrammer:
I think this is a good change. I think updating node to 18 might need to happen, however as that's when fetch is fully supported in node. I think updating the workflows to use node 18 & updating babel for node 18 would be a good idea. Whether included in this PR or a separate one. Note: I have not yet tested this PR but I looked at the code changes and the code looks good to me so far.
image
source: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

absidue:
@ChunkyProgrammer As far as I can tell all of these fetch calls are run in in the chromium part of electron not the node part, as they all work without updating node or electron. Electron 19 still uses node 16 according to the release notes, so 17 and 18 probably do too.

PikachuEXE:
I would prioritize this below electron upgrade (to [17](https://github.com/FreeTubeApp/FreeTube/pull/2358)/[18](https://github.com/FreeTubeApp/FreeTube/pull/2359))

I would still support this change but I think we don't have time to test so many changes at once

ChunkyProgrammer:
I would also like to see #1525 merged before this PR

absidue:
I'll wait until the accessibility PR is merged before fixing the merge conflicts.

PR #2372, opened 2022-07-02
Fix blank space in the subscription settings

absidue:
---
Title
---

**Pull Request Type**

- [x] Bugfix

**Description**
Fixes the blank space in the subscription settings.

before:
before

after:
after

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 3ccdf566998fc88350d55797c936b6014d65551c

### DIFF
--- a/src/renderer/components/subscription-settings/subscription-settings.js
+++ b/src/renderer/components/subscription-settings/subscription-settings.js
@@ -1,31 +1,17 @@
 import Vue from 'vue'
 import { mapActions } from 'vuex'
-import FtCard from '../ft-card/ft-card.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
-import FtButton from '../ft-button/ft-button.vue'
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
nice catch!

ChunkyProgrammer:
Why not remove the brs and remove the elements with v-if="false" as they won't be rendered?

absidue:
> Why not remove the brs and remove the elements with v-if="false" as they won't be rendered?

I wasn't sure if it would be needed again in the future, I can definitely get rid of them if that's better?

PikachuEXE:
Unless there is a value for keeping a "commented code" (well it's not commented but disabled)
Remove it

absidue:
I've removed the code for it now. It can always be added back in the future if it's needed then.

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #2425, opened 2022-07-30
Fix scrolling through the download dropdown

absidue:
---
Fix scrolling through the download dropdown (RC branch)
---

**Pull Request Type**

- [x] Bugfix

**Related issue**
works around issues introduced in #2284

**Description**
This pull request fixes the issue with not being able to scroll through the download dropdown. It toggles the sticky css position on the video info section when the downloads menu is open. It also includes some extra code to make the toggling feel seemless.

**Testing (for code that is not small enough to be easily understandable)**
I tested this PR both on wide and narrow screens and was able to scroll though the download dropdown without an issues.

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.0 RC (7ec1c1eb05738d9d45b1e2e5d3321cc086c0176f)

PR #2426, opened 2022-07-30
Fix scrolling through the download dropdown

absidue:
---
Fix scrolling through the download dropdown (development branch)
---

**Pull Request Type**

- [x] Bugfix

**Related issue**
works around issues introduced in #2284

**Description**
This pull request fixes the issue with not being able to scroll through the download dropdown. It toggles the sticky css position on the video info section when the downloads menu is open. It also includes some extra code to make the toggling feel seemless.

I tested this PR both on wide and narrow screens and was able to scroll though the download dropdown without an issues.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.0 RC/development (9fa4a5a36eb46cb7c2fb6d3d6a17749df5efbc97)

### DIFF
--- a/src/renderer/components/watch-video-info/watch-video-info.js
+++ b/src/renderer/components/watch-video-info/watch-video-info.js
@@ -300,6 +300,18 @@ export default Vue.extend({
           }
         ]
       })
+
+      this.$watch('$refs.downloadButton.dropdownShown', (dropdownShown) => {
+        this.$parent.infoAreaSticky = !dropdownShown
+
+        if (dropdownShown && window.innerWidth >= 901) {
+          // adds a slight delay so we know that the dropdown has shown up
+          // and won't mess up our scrolling
+          Promise.resolve().then(() => {
+            this.$parent.$refs.infoArea.scrollIntoView()
+          })
+        }
+      })
     }
   },
   methods: {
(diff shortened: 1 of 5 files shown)


PikachuEXE (approved):
https://github.com/FreeTubeApp/FreeTube/pull/2425#pullrequestreview-1056505581

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Only weird thing i could find is this.

https://user-images.githubusercontent.com/73130443/182027829-a815df01-13ce-4cde-82dc-37dfbc94619d.mp4

Up next videos should stay in same position when u click the download button

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc That's cause by this solution programatically scrolling up when you click the downloads button so that the info area and download menu are visible.

The sticky position that the info area has by default means that regardless of how much you scroll down it will stay in the same position on screen, which also means you can't scroll to see the options further down in the download menu.
To solve that issue, this solution makes the info area non sticky when the download menu is open, this results in the info area returning to it's original position in the page. If you have scrolled down quite a bit that means that you won't see the info area or the download menu, as they will be off screen further up the page.
I decided to do the scrolling up programatically, so that the user can use the download menu straight away, without being confused about the info area suddenly disappearing.

The ideal solution would be to redesign the download menu to not be a really long drop down, that is out of scope of this PR, as the goal was to provide a quick solution so we can release 0.17.0.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
in that case LGTM

ChunkyProgrammer (approved):

PR #2431, opened 2022-07-31
Fix banner width

absidue:
---
Fix banner width (RC branch)
---

**Pull Request Type**

- [x] Bugfix

**Description**
Makes the banners the same width as the cards below them.

before
fixed

I tested both the new blog and update banners.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.0 RC

### DIFF
--- a/src/renderer/App.css
+++ b/src/renderer/App.css
@@ -16,8 +16,12 @@
 }
 
 .banner {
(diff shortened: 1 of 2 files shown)


PrestonN (approved):
LGTM

PR #2452, opened 2022-08-04
Drastically improve loading speed of the settings page

absidue:
---
Drastically improve loading speed of the settings page
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

**Description**
This pull request fixes the slow loading of the settings page by only loading the language list once, instead of the 45 times it does it at the moment (once outside of the loop and once for each language inside the loop). As the list of language names is also regenerated when you choose a different language, this change also speeds up that as well.

<details><summary>Illustration of what it feels like to open the settings page now:</summary>

speed

</details>

Here is how long it takes to generate the list of language names with and without this change. The numbers are in milliseconds.

settings_locale_list_performance

Open the settings page and change language. You will definitely notice the difference, if you don't then your computer is too fast (I would definitely like to borrow it at some point).

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.0

### DIFF
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -122,8 +122,8 @@ export default Vue.extend({
         this.$t('Settings.General Settings.System Default')
       ]
 
-      Object.keys(this.$i18n.messages).forEach((locale) => {
-        const localeName = this.$i18n.messages[locale]['Locale Name']
+      Object.entries(this.$i18n.messages).forEach(([locale, localeData]) => {
+        const localeName = localeData['Locale Name']
         if (typeof localeName !== 'undefined') {
           names.push(localeName)
         } else {


efb4f5ff-1298-471a-8973-3d47447115dc:
@absidue your illustration really made laugh out loud hard. Definitely caught me off guard there.

You also made me chuckle when i read everything below testing.

PikachuEXE (approved):
Tested locally
Slight difference spotted
OK I got faster computer

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Before: 🏃💨
After: 🚀🚀🚀

Edit: noticeable performance improvement

ChunkyProgrammer (approved):

PR #2259, opened 2022-05-22
Fix: wrong info displayed when quickly navigate from a channel to another

hockerschwan:
**Pull Request Type**
- [x] Bugfix

**Related issue**
closes #1545

**Description**
Holds id in `getChannelInfoInvidious` etc. and compares to this.id, if they are different (route changed), do nothing.

**Screenshots (if appropriate)**
Before:
<video src="https://user-images.githubusercontent.com/80553357/169692829-4ff6cc7e-bb4a-4d21-9c9c-64141c790d00.mp4">

After:
<video src="https://user-images.githubusercontent.com/80553357/169692835-16c137be-daa9-4c17-b92f-bea410e2ae74.mp4">

**Desktop (please complete the following information):**
 - OS: Windows10
 - OS Version: 21H1
 - FreeTube version: 75d59a54990280c08488cb39ba136d2b38c9d0fe

PR #2457, opened 2022-08-06
Fix channel page ID handling

### LINKED ISSUE #2420: [Bug]: Custom channel URLs resolve by username which is incorrect
### Guidelines

### Describe the bug

1. Enter `https://www.youtube.com/c/RealEngineering` into the search bar and hit <kbd>Enter</kbd>
2. The channel goes to https://www.youtube.com/user/realengineering which is the wrong channel

### Expected Behavior

It should go to the correct channel

### Issue Labels

API issue, inconsistent behavior, usability issue

### FreeTube Version

v0.16.0 Beta

### Operating System Version

Linux 5.18.4, NixOS, 22.11 (Raccoon), 22.11.20220619.0d68d7c

### Installation Method

other

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

absidue:
---
Fix channel page ID handling
---

**Pull Request Type**
- [x] Bugfix

**Related issue**
closes #2420
fixes a race condition introduced by #2259

**Description**
This pull request resolves two issues relating to the handling of channel IDs on the channel page.

1. Treat `/channel/`, `/user/` and `/c/` channel URLs differently, as a `/user/` and `/c/` URLs with the same id might point to different channels.
2. Fixes the race condition introduced by #2259, that happens if fetching the info is faster than the video or playlist fetching. The issue is that after the info is fetched it overwrites the `id` property with the actual channel ID, so if you entered this URL `https://www.youtube.com/c/linustechtips`, the `id` property is set to `linustechtips` and when the info request returns it overwrites it with the channel ID `UCXuqSBlHAE6Xw-yeJA0Tunw`, as the videos and playlists functions still expect the `id` to be `linustechtips` they abort.

The fix for this was to introduce an `originalId` property that is only updated when the actual channel changes, which is then used for the abort checks.

I tested the following URLs:
- https://www.youtube.com/c/linustechtips
- https://www.youtube.com/c/RealEngineering
- https://www.youtube.com/user/realengineering

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.0

### DIFF
--- a/src/renderer/components/top-nav/top-nav.js
+++ b/src/renderer/components/top-nav/top-nav.js
@@ -174,10 +174,11 @@ export default Vue.extend({
           }
 
           case 'channel': {
-            const { channelId, subPath } = result
+            const { channelId, idType, subPath } = result
 
             this.$router.push({
-              path: `/channel/${channelId}/${subPath}`
+              path: `/channel/${channelId}/${subPath}`,
+              query: { idType }
             })
             break
           }
--- a/src/renderer/store/modules/utils.js
+++ b/src/renderer/store/modules/utils.js
@@ -621,7 +621,7 @@ const actions = {
     let urlType = 'unknown'
 
     const channelPattern =
-      /^\/(?:(c|channel|user)\/)?(?<channelId>[^/]+)(?:\/(join|featured|videos|playlists|about|community|channels))?\/?$/
+      /^\/(?:(?<type>channel|user|c)\/)?(?<channelId>[^/]+)(?:\/(join|featured|videos|playlists|about|community|channels))?\/?$/
 
     const typePatterns = new Map([
       ['playlist', /^\/playlist\/?$/],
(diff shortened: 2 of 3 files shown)


PikachuEXE (commented):
Do we have a place to stored these "magic values"?

I certainly won't be able to parse all the constant values for `idType` in the future.
Not to mention I can't parse them now.

absidue:
@PikachuEXE As those "magic values" come from yt-channel-info, if we create an enum it should definitely be in yt-channel-info and then imported into FreeTube.

PikachuEXE (approved):
Tested locally

All 3 provided URLs returned different channels

ChunkyProgrammer (approved):
Works well, also tested to make sure subpaths worked (ie c/name/playlsts)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2493, opened 2022-08-17
Eliminating a race condition with SponsorBlock markers

MarmadileManteater:
---
Eliminating a race condition with SponsorBlock markers
---

**Pull Request Type**
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

**Related issue**
closes FreeTubeApp#2492

**Description**
This modifies the SponsorBlock markers to use the `lengthSeconds` already provided by the API to calculate their size and position instead of using `this.player.duration()`. `this.player.duration()` is 0 when the view loads and only obtains a value some time after. `lengthSeconds` should never be 0 unless the API is giving a strange response. Because of this, changing `this.player.duration()` to `lengthSeconds` should eliminate the race condition where `this.player.duration()` is 0 at the time when `addSponsorBlockMarker` is called.

**Screenshots**
(These are with Invidious API selected as primary)
Before:
image
After:
image

 - OS: Windows 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -73,6 +73,10 @@ export default Vue.extend({
     videoId: {
       type: String,
       required: true
+    },
+    lengthSeconds: {
+      type: Number,
+      required: true
     }
   },
   data: function () {
@@ -566,6 +570,10 @@ export default Vue.extend({
             })
           })
         })
+      }).catch((error) => {
+        this.showToast({
+          message: error.message
+        })
       })
     },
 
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer:
Tested using a live video, a video with sponsors and a video with sponsors and it worked fine for me (tested with both invidious and local api)

ChunkyProgrammer (approved):

PikachuEXE (approved):
Tested locally but not sure where to find a live video to test

absidue (approved):
I was only able to review the code as it looks like the SponsorBlock API is ratelimiting me.

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -566,6 +570,10 @@ export default Vue.extend({
             })
           })
         })
+      }).catch((error) => {
+        this.showToast({
+          message: error.message
+        })
```
absidue, line 576: @MarmadileManteater This just shows and empty toast, I don't think that `error` object has a `message` property.

absidue:
@MarmadileManteater Could you please create a follow up PR that fixes the empty toast. If the SponsorBlock API doesn't have any segments for a video it 404s. The better option would be to show a toast that says something along the lines of no sponsorblock segments or even better don't show anything at all if there aren't any segments.

src/renderer/components/ft-video-player/ft-video-player.js
MarmadileManteater: You are correct. This was a mistake on my part.

MarmadileManteater:
I made a PR to remove the toast. It is just something I whipped up, but as stated, it would make more sense to opt for a solution which involves making the toast message meaningful.

PR #2497, opened 2022-08-20
PublishedText gets created using only one method. (toLocaleDateString…

### LINKED ISSUE #2439: [Bug]: il % in label of the video
### Guidelines

### Describe the bug

image
In the title of the French video, we get "il ya il %" instead of having the number of days.

### Expected Behavior

I would like to get the number of days since the video was uploaded

### Issue Labels

text/string issue

### FreeTube Version

v0.17.0 Bêta

### Operating System Version

Windows 11 Professionnel 21H2 22000.795

### Installation Method

.exe

### Primary API used

Invidious API

_No response_

### Additional Information

_No response_

### Nightly Build

MaxWienold:
…) (#2439)

---
The Published on String is now created in a (more) concise way
---

**Important note**

**Pull Request Type**
- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

**Related issue**
closes #2439

**Description**
There were two methods and two variables for creating the published ... min/hours/days string. One that was written in utils and the `Date.toLocaleString.` As well there were many conditionals that should produce all the same result. Should this solution be viable, the translation String and the utils method could also be purged.
**Edit:**
I realized that the real bug lay in that `toLocalePublicationString` relied on the api giving back a string in english. But apparently the locale is handled differently and is not affected by the language settings. So i think, it is more reliable to determine unit and number based on the published timestamp. 

I tried different language settings on different Pages.

 - OS: Debian
 - OS Version: 11
 - FreeTube version: latest

**Additional context**
I noticed that the recommended videos have no publication data associated.


src/renderer/views/Watch/Watch.js
```diff
@@ -591,7 +591,7 @@ export default Vue.extend({
             channelName: result.author,
             channelId: result.authorId
           })
-
+          console.log('recommended:', result.recommendedVideos)
```
PikachuEXE, line 594: Should we include these changes in the app...?
Not just this PR but all `console.log`

ChunkyProgrammer: I would be okay with getting rid of console.log. i also think some errors in the app currently use "console.log" instead of "console.error" and it would probably be best if those were changed

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -84,6 +88,8 @@ export default Vue.extend({
     },
 
     inHistory: function () {
+      console.log(JSON.stringify(this.$t('Video.Published')))
```
ChunkyProgrammer, line 91:
```suggestion
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -84,6 +88,8 @@ export default Vue.extend({
...
+
```
ChunkyProgrammer, line 92:
```suggestion
```

src/renderer/views/Watch/Watch.js
ChunkyProgrammer, line 594:
```suggestion
```

src/renderer/mixins/DateFormatter.js
```diff
@@ -0,0 +1,43 @@
+const DateFormatter = {
+  methods: {
+    toLocalePublicationString (videoData) {
```
ChunkyProgrammer, line 3: Named object parameters would probably be better here
Ex:
```suggestion
    toLocalePublicationString ({ publishedText, publishedDate, isLive = false, isUpcoming = false, isRSS = false }) {
```

ChunkyProgrammer (changes requested):

ChunkyProgrammer:
Just some minor changes

src/renderer/views/Watch/Watch.js
MaxWienold: Could maybe be  enforced by the linter in the future...

src/renderer/mixins/DateFormatter.js
MaxWienold: Yep, you're right. I'm not sure how to commit this suggestion, because it would mean that the references to `videoData` would have to be changed as well.

ChunkyProgrammer: You'd have to make the change in your code editor

ChunkyProgrammer (approved):

src/renderer/mixins/DateFormatter.js
```diff
@@ -0,0 +1,43 @@
...
+      const months = Math.floor(millis / 2629800000)
+      if (months > 0) return { unit: 'Month', num: months }
+      const weeks = Math.floor(millis / 604800000)
+      if (weeks > 0) return { unit: 'Week', num: weeks }
+      const days = Math.floor(millis / 86400000)
+      if (days > 0) return { unit: 'Day', num: days }
+      const hours = Math.floor(millis / 3600000)
+      if (hours > 0) return { unit: 'Hour', num: hours }
+      const minutes = Math.floor(millis / 60000)
+      if (minutes > 0) return { unit: 'Minute', num: minutes }
```
ChunkyProgrammer, line 38: after the if statements you should return seconds (even if seconds is 0)

ChunkyProgrammer (changes requested):

absidue (changes requested):
This PR changes the `ft-list-video` component that is used everywhere in FreeTube where there is a list of videos. Unfortunately the changes in this PR mean that the component now only works on the subscription and trending pages. All other pages are broken (search, channels page, playlists page, up next videos on the video page, playlist videos on the video page).

Basically what I'm trying to say is that the reason the existing logic is hard to understand, is because it needs to handle all of the different use cases. I'm sure it can be simplified slightly however not as much as this PR does it.

Issue #1837, opened 2021-10-18
Improve cached page loading


Winterhuman:
**Behavior of the program**
Even if a page is cached, as in the video details and all it's "Up Next" thumbnails *seem* to be restored when reloading the page, it still takes a substantial amount of time to load the page even though all the elements of the page appear to be fetched near instantly from local cache.

**Expected behavior**
When going to a video, navigating to another video from the "Up Next" section, and then going back to the first video, the page should load instantly assuming all elements on the page are fetched from cache instead of showing the loading screen.

**To Reproduce**
1. Go to any video.
2. Click on another video in the "Up Next" section.
3. Navigate back to the first video.
4. Observe the page take a long time to load, but, all the elements in the page seem to be fetched from cache instantly such as all the thumbnails.

**Environment Information:**
 - OS: Arch Linux
 - FreeTube version: v0.15.1 Beta
 - Installation Method: FlatHub
 - Primary API Used: Local API

Winterhuman:
Either this has been fixed, or it's not as noticeable anymore. Closing issue

PR #2498, opened 2022-08-20
Disable http disk cache and implempent in-memory image cache

absidue:
---
Disable http disk cache and implempent in-memory image cache
---

**Important note**

**Pull Request Type**

- [x] Bugfix
- [x] Feature Implementation

**Related issue**
#1837 could be related but I'm not sure

**Description**
Thank you to the Matrix user for bringing this issue to our attention.

To reduce RAM usage electron stores the http cache on disk. While that might be alright for other applications, in FreeTube that means that while you are watching a video, all of the buffered video data is actually being constantly streamed to and from your disk. In the long term this will likely have an inpact on disk health.

This pull request is made up of two parts:
1. Disabling the http cache, unfortunately there doesn't seem to be a way to get rid of the disk one without also getting rid of the in-memory one
2. To compensate for the now missing in-memory http cache, this pull request also implements an in-memory image cache.

_How the cache works:_
When FreeTube tries to load an image over HTTP(S), the request gets redirected to the `imagecache://` custom protocol handler. The protocol handler checks the cache for the image, if it finds it, the image data is returned. Otherwise it fetches the image, stores it to the cache and then returns the image data.

The amount of time an image stays in the cache is based on the `Cache-Control: max-age=<seconds>` and `Age: <seconds>` headers that are returned by the server when the image is fetched. The cache checks for and removes expired images every 5 minutes, this interval can altered in the future if needed.

_Performance:_
In my testing disabling the http cache reduced the disk usage while playing a 1080p DASH video from 1.4 - 2.2MB/s to 0MB/s, when you first start the video it will sit at about 0.1MB/s as the DASH manifest and storyboards are written to the disk and subsequently used.

I tried profiling the memory usage of the image cache using the `--inspect-brk` flag and connecting to the main process using the devtools in an external chromium browser. Unfortunately I'm not quite sure what the actual memory usage is, as the heap snapshot said it was using about 28.4MB for 1053 images, however the entire heap was only 19.1MB at the time of the snapshot, so something isn't adding up.

Please add before and after screenshots if there is a visible change.

Please test this PR with production/release builds as dev and debug builds use extra RAM because of the devtools.

To check the disk usage I used the Task Manager in Windows to monitor the disk usage during video playback (e.g. https://youtu.be/bafzQBSktwk)

RAM usage can be monitored using the Task Manager or by hooking up the devtools in an external chromium installation up to the main process:
1. Start FreeTube with the `--inspect-brk` (allows remote debugging of the main process and wait until you are ready before executing any javascript code).
2. Open an external chromium installation and open `chrome://inspect`
3. Depending on your browser their might be a link to open the dedicated node devtools
4. Connect to FreeTube using the connection details that FreeTube spat out in your terminal
5. In the Sources tab in the devtools, start code execution by pressing the play button in the top right corner
6. You can take heap snapshots in the Memory tab
7. As we are using a production build the `ImageCache` won't be called `ImageCache`, instead it will have a single letter name, in my case it was `q`. You can find the name by searching for `No image cache entry for` in the Sources tab, going back a bit until you find the `constructor` and then looking at the class name.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

**Additional context**
If the uptick in RAM usage is too significant we can always consider caching the images on disk instead of in-memory. This will be slower but will have the advantage of lower RAM usage. It will also allow us to use the cached images, that haven't expired yet, the next time you open FreeTube.

### DIFF
--- a/_scripts/dev-runner.js
+++ b/_scripts/dev-runner.js
@@ -53,10 +53,12 @@ async function restartElectron() {
 
   electronProcess = spawn(electron, [
     path.join(__dirname, '../dist/main.js'),
-    // '--enable-logging', Enable to show logs from all electron processes
+    // '--enable-logging', // Enable to show logs from all electron processes
     remoteDebugging ? '--inspect=9222' : '',
     remoteDebugging ? '--remote-debugging-port=9223' : '',
-  ])
+  ],
+    // { stdio: 'inherit' } // required for logs to actually appear in the stdout
+  )
 
   electronProcess.on('exit', (code, _) => {
     if (code === relaunchExitCode) {
--- a/_scripts/dev-runner.js
+++ b/_scripts/dev-runner.js
@@ -55,8 +55,7 @@ async function restartElectron() {
     path.join(__dirname, '../dist/main.js'),
     // '--enable-logging', // Enable to show logs from all electron processes
     remoteDebugging ? '--inspect=9222' : '',
-    remoteDebugging ? '--remote-debugging-port=9223' : '',
-    // '--experiments-disable-disk-cache'
+    remoteDebugging ? '--remote-debugging-port=9223' : ''
   ],
     // { stdio: 'inherit' } // required for logs to actually appear in the stdout
   )
--- a/_scripts/dev-runner.js
+++ b/_scripts/dev-runner.js
@@ -56,6 +56,7 @@ async function restartElectron() {
     // '--enable-logging', // Enable to show logs from all electron processes
     remoteDebugging ? '--inspect=9222' : '',
     remoteDebugging ? '--remote-debugging-port=9223' : '',
+    // '--experiments-disable-disk-cache'
   ],
     // { stdio: 'inherit' } // required for logs to actually appear in the stdout
   )
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -7,7 +7,7 @@ import cp from 'child_process'
 
 import { IpcChannels, DBActions, SyncEvents } from '../constants'
 import baseHandlers from '../datastores/handlers/base'
-import { ImageCache } from './ImageCache'
+import { extractExpiryTimestamp, ImageCache } from './ImageCache'
 
 if (process.argv.includes('--version')) {
   app.exit()
@@ -194,12 +194,10 @@ function runApp() {
         response.on('end', () => {
           const data = Buffer.concat(chunks)
 
-          const age = parseInt(response.headers.age ?? 0)
-          const maxAge = parseInt(response.headers['cache-control'].match(/max-age=([0-9]+)/)[1])
-
+          const expiryTimestamp = extractExpiryTimestamp(response.headers)
           const mimeType = response.headers['content-type']
 
-          imageCache.add(url, mimeType, data, maxAge - age)
+          imageCache.add(url, mimeType, data, expiryTimestamp)
 
           // eslint-disable-next-line node/no-callback-literal
           callback({
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -219,12 +219,32 @@ function runApp() {
           })
 
           response.on('error', (error) => {
-            console.log('error', error)
+            console.error('image cache error', error)
+
+            // error objects don't get serialised properly
+            // https://stackoverflow.com/a/53624454
+
+            const errorJson = JSON.stringify(error, (key, value) => {
+              if (value instanceof Error) {
+                return {
+                  // Pull all enumerable properties, supporting properties on custom Errors
+                  ...value,
+                  // Explicitly pull Error's non-enumerable properties
+                  name: value.name,
+                  message: value.message,
+                  stack: value.stack
+                }
+              }
+
+              return value
+            })
+
             // eslint-disable-next-line node/no-callback-literal
             callback({
-              statusCode: response.statusCode ?? 400
+              statusCode: response.statusCode ?? 400,
+              mimeType: 'application/json',
+              data: Buffer.from(errorJson)
             })
-            throw error
           })
         })
 
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1,12 +1,13 @@
 import {
   app, BrowserWindow, dialog, Menu, ipcMain,
-  powerSaveBlocker, screen, session, shell, nativeTheme
+  powerSaveBlocker, screen, session, shell, nativeTheme, net, protocol
 } from 'electron'
 import path from 'path'
 import cp from 'child_process'
 
 import { IpcChannels, DBActions, SyncEvents } from '../constants'
 import baseHandlers from '../datastores/handlers/base'
+import { ImageCache } from './ImageCache'
 
 if (process.argv.includes('--version')) {
   app.exit()
@@ -44,10 +45,15 @@ function runApp() {
 
   let mainWindow
   let startupUrl
+  const imageCache = new ImageCache()
 
   app.commandLine.appendSwitch('enable-accelerated-video-decode')
   app.commandLine.appendSwitch('enable-file-cookies')
   app.commandLine.appendSwitch('ignore-gpu-blacklist')
+
+  // the http cache causes excessive disk usage during video playback
+  // we've got a custom image cache to make up for disabling the http cache
+  app.commandLine.appendSwitch('disable-http-cache')
 
   // See: https://stackoverflow.com/questions/45570589/electron-protocol-handler-not-working-on-windows
   // remove so we can register each time as we run the app.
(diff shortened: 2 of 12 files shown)


_scripts/dev-runner.js
```diff
@@ -51,10 +51,12 @@ async function restartElectron() {
...
   electronProcess = spawn(electron, [
     path.join(__dirname, '../dist/main.js'),
-    // '--enable-logging', Enable to show logs from all electron processes
+    // '--enable-logging', // Enable to show logs from all electron processes
     remoteDebugging ? '--inspect=9222' : '',
     remoteDebugging ? '--remote-debugging-port=9223' : '',
-  ])
+  ],
+    // { stdio: 'inherit' } // required for logs to actually appear in the stdout
+  )
```
PikachuEXE, line 59: What are these changes?

src/main/index.js
```diff
@@ -48,6 +50,10 @@ function runApp() {
   app.commandLine.appendSwitch('enable-file-cookies')
   app.commandLine.appendSwitch('ignore-gpu-blacklist')
 
+  // the http cache causes excessive disk usage during video playback
+  // we've got a custom image cache to make up for disabling the http cache
+  app.commandLine.appendSwitch('disable-http-cache')
```
PikachuEXE, line 55: Has [`--disk-cache-size=0`](https://www.electronjs.org/docs/latest/api/command-line-switches/#--disk-cache-sizesize) been tried?

src/main/index.js
```diff
@@ -152,6 +158,85 @@ function runApp() {
       })
     })
 
+    protocol.registerBufferProtocol('imagecache', (request, callback) => {
+      const url = decodeURIComponent(request.url.substring(13))
```
PikachuEXE, line 162:
```suggestion
      Remove `imagecache://` prefix
      const url = decodeURIComponent(request.url.substring(13))
```

src/main/index.js
```diff
@@ -152,6 +158,85 @@ function runApp() {
...
+      if (imageCache.has(url)) {
+        const cached = imageCache.get(url)
+
+        // eslint-disable-next-line node/no-callback-literal
+        callback({
+          mimeType: cached.mimeType,
+          data: cached.data
+        })
+      } else {
```
PikachuEXE, line 171: Perform an early return here to reduce one level of nesting afterward?

PikachuEXE:
I confirmed on MacOS 12.5.1 with Activity Monitor that the disk usage (bytes written) has not increased during video playback after switching to this branch
image

PikachuEXE:
This error occurs on my custom build
image

_scripts/dev-runner.js
absidue: `{ stdio: 'inherit' }` makes `spawn()` pipe the stdout and stderr of the spawned command to the stdout and stderr of the parent process, in this case `_scripts/dev-runner.js`. Without that change you won't see any logs from the `main` process, I've left it in there commented out so that if someone needs to log something from the `main` process in the future they don't need to spend hours trying to work out why they can't see the logs, like I did.

Adding the extra `//` to the enable logging line, makes it easier to use your IDE's keyboard shortcuts to uncomment the line without needing to go and add the slashes in manually after you have uncommented it. As `Enable to show logs from all electron processes` is not valid javascript code and will need to stay commented even when you uncomment that line.

src/main/index.js
```diff
@@ -152,6 +158,85 @@ function runApp() {
...
+          const chunks = []
+          response.on('data', (chunk) => {
+            chunks.push(chunk)
+          })
+
+          response.on('end', () => {
+            const data = Buffer.concat(chunks)
+
+            const age = parseInt(response.headers.age ?? 0)
+            const maxAge = parseInt(response.headers['cache-control'].match(/max-age=([0-9]+)/)[1])
```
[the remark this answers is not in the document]
absidue: I've added fallbacks now.
1. It will try to use the max-age in the cache-control header if it is available, alternatively it will use the expires header, otherwise it will fallback to a hardcoded expiry time (currently 2 hours).

src/main/index.js
absidue: Yep and it doesn't make a difference to disk usage during video playback. I also tried setting it to 1 which makes it only do 0.1MB/s which is already a lot better than it was, but it doesn't eliminate the problem completely. I suspect that setting it to 0 tells it to just use their default limit.

PikachuEXE:
Using it locally (MacOS)
Will use it in Windows after 1 week of testing if no issue found

absidue:
I've added fallback so for the expiry timestamp so that it should never fail, if none of the headers are available it defaults to a hardcoded 2 hour expiry time (can be changed with the `const` at the top of the ImageClass file).

As discussed I've also moved these changes behind a CLI argument so that this behaviour is opt in for now. (If you want to enable it while using `yarn dev`, you can uncomment the flag in the `_scripts/dev-runner.js` file.)

PikachuEXE:
I wonder how this can be enabled when released...

absidue:
On Windows you can either run the .exe from the CLI e.g. `C:\path\to\FreeTube.exe --experiments-disable-disk-cache` or just add the arg to the target field.

On Linux you can either run FreeTube from the CLI or add it to the `.desktop` file.

macOS is annoying though, you can only pass the flag when you launch FreeTube from the terminal: `open -a /Applications/FreeTube --args '--experiments-disable-disk-cache'`.

PikachuEXE:
Really need a new settings to enable/disable this... (and be effective on restart)

absidue:
Finally figured out a way to add a GUI setting for this. Also removed the CLI flag.

absidue:
copy of @PikachuEXE's comment here https://github.com/FreeTubeApp/FreeTube/pull/2511#issuecomment-1232319410

> Style suggestion on settings UI
>
> image
image
>
> Code changes:
image
image

PikachuEXE:
An error encountered on Windows 10
Disabling via settings to see if it happens again

image

PikachuEXE:
I still see error popups on windows, disabled on windows for now
No issue on MacOS though
But I am having VPN enabled all the time on Windows while no VPN on Mac
I can try to turn on VPN on Mac later to see if the error would be reproduced

manifest-9:
Working for me on arch linux 5.19.10 playing 1080p dash. 
Disk writes went from 1.0MB/s to 0MB/s.
Screenshot_20220923_100000

PikachuEXE:
This can *almost* be merged except those random JS errors
Is there anything can be done?

PikachuEXE:
Also just found out the code is outdated due to settings panel update
I got a fix https://github.com/PikachuEXE/FreeTube/commit/ae9fad898aa35e227953ba2411077a49d0497147 for my custom build
Use it ~

src/main/index.js
```diff
@@ -152,6 +165,93 @@ function runApp() {
...
+            })
+          })
+
+          response.on('error', (error) => {
+            console.log('error', error)
+            // eslint-disable-next-line node/no-callback-literal
+            callback({
+              statusCode: response.statusCode ?? 400
+            })
+            throw error
```
absidue, line 230: @PikachuEXE I suspect the error you are seeing is happening here. Instead of throwing it I could probably return the error in the response and then it would be visible in the devtools instead of throwing up a random error popup?

PikachuEXE: As long as it behaves the same as when it's disabled

absidue: Okay so instead of throwing the error it will return it as JSON, so you should be able to see it in the network tab in the devtools. I still don't understand why the error is happening in the first place, but this should at least make errors less intrusive and should make it easier for you to debug it. Of course the images that caused the errors still won't load but you shouldn't get the popup now.

Please let me know when you have some error JSON so I can try and work out what the problem is.

PikachuEXE (dismissed):
I think this is OK to merge as an experimental setting

ChunkyProgrammer:
Got this error on Windows 
image

PikachuEXE:
Issue reproduced
image

static/locales/en-US.yaml
```diff
@@ -405,6 +405,10 @@ Settings:
     Download Behavior: Download Behavior
     Download in app: Download in app
     Open in web browser: Open in web browser
+  Experimental Settings:
+    Experimental Settings: Experimental Settings
+    Warning: These settings are experimental, enabling them may result in a worse FreeTube experience. Use at your own risk!
```
ChunkyProgrammer, line 410: If you say it has a potential for crashes and recommend the data is backed up then I would be fine with approving as is. It is an experimental setting after all

absidue: Updated the message now.

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2528, opened 2022-09-01
Only generate the web assets directory during web builds

absidue:
---
Only generate the web assets directory during web builds
---

**Pull Request Type**
- [x] Feature Implementation

**Description**
This pull request stops the web assets directory being created during webpack builds that are only used for Electron packages. This doesn't make a big difference to the build duration, but it does stop webpack doing unnecessary work. While I was at it I also removed the references to the `LoaderOptionsPlugin` as that doesn't do anything in webpack 5, webpack 5 optimises during productions builds by default.

I decided to keep the asset copying in the main webpack config instead of moving it to the renderer one, as the renderer build is already a lot slower than the main one, so we might as well have the main build do it so that the renderer one doesn't take even longer (yes, I know that my other webpack PR makes the renderer build take even longer).

I tested `yarn dev`, `yarn run pack` (`yarn pack` packages FreeTube as a node module in a gzipped tar archive) and `yarn build` and didn't see any issues.

The only difference in output is that `yarn run pack` doesn't copy create the `web` directory in the `dist` directory and copy any files to it, as the Electron build doesn't package those anyway. Those files and directories will now only be created during the web build.

To compare the output you can run `yarn run pack` on the development branch with commit 717f23e0dd48a543ac2aac429e33615131747558, rename the `dist` directory to something else e.g. `dist_old` and then switch to this branch and run `yarn run pack`.

After that you can compare the two directories like this: `git diff --no-index dist_old dist` (`--no-index` makes `git diff` behave like a normal diffing tool, comparing just those two directories without caring about any git history)

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

### DIFF
--- a/_scripts/webpack.main.config.js
+++ b/_scripts/webpack.main.config.js
@@ -98,10 +98,7 @@ if (isDevMode) {
           },
         ]
       }
-    ),
-    new webpack.LoaderOptionsPlugin({
-      minimize: true,
-    })
+    )
   )
 }
 
--- a/_scripts/webpack.main.config.js
+++ b/_scripts/webpack.main.config.js
@@ -69,12 +69,8 @@ if (isDevMode) {
     new CopyWebpackPlugin({
         patterns: [
           {
-            from: path.join(__dirname, '../static/pwabuilder-sw.js'),
-            to: path.join(__dirname, '../dist/web/pwabuilder-sw.js'),
-          },
-          {
             from: path.join(__dirname, '../static'),
-            to: path.join(__dirname, '../dist/web/static'),
+            to: path.join(__dirname, '../dist/static'),
             globOptions: {
               dot: true,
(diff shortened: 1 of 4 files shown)


PikachuEXE:
Creating another custom build...

absidue:
Creating a custom build for this is not necessary as it's purely a clean up of the config and stopping webpack generating files that we have already blacklisted in the electron build.

Testing a custom build for this is the same as testing a build of the development branch without this branch, because non of the changes that this PR makes are noticable in the final electron build.

Creating a custom build for this is definitely not necessary but I can't stop you from making one if you really want to.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #2535, opened 2022-09-03
Localization inconsistency fix for Local API

MarmadileManteater:
---
Localization inconsistency fix for Local API
---

**Pull Request Type**
- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

**Related issue**
closes FreeTubeApp#2530

**Description**
This applies a workaround for the inconsistency in the responses from `ytdl` and `ytpl`. The result of this change is that the video title and description are localized to `en-US` by `ytdl` (as it was expected to be based on the following code snippet).
```js
// line 353 of src\renderer\store\modules\ytdl.js
ytdl.getInfo(videoId, {
        lang: 'en-US',// this locale was not being honored before this change
        requestOptions: { agent }
      }).then((result) => {
        resolve(result)
      }).catch((err) => {
        reject(err)
      })
```

Before this change:
image
image
image

After this change:
image
image
image

 - OS: Windows 10
 - FreeTube version: 0.17.1

**Additional context**
The inconsistency described in FreeTubeApp#2530 still occurs when using the Invidious API; however, as far as I can tell, this is an issue with invidious. Even if we aren't sending a locale, it should give back titles and descriptions that reflect a consistent locale, but I don't believe there is a way to fix that inconsistency without digging into the invidious source, and at that point, it isn't an issue with FreeTube.

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -309,7 +309,8 @@ export default Vue.extend({
             // workaround for description localization
             const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description.runs
             this.videoDescription = descriptionLines?.join('\n') ?? ''
-          } catch {
+          } catch (err) {
+            console.error('Failed to extract localised video description, falling back to the standard one.', err)
             // if the workaround for localization fails, this sets the description to the potentially non-localized value
             this.videoDescription = result.player_response.videoDetails.shortDescription
           }
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -308,7 +308,7 @@ export default Vue.extend({
           try {
             // workaround for description localization
             const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description.runs
-            this.videoDescription = descriptionLines?.join('\n') ?? ''
+            this.videoDescription = descriptionLines?.map(line => line.text).join('\n') ?? ''
           } catch (err) {
             console.error('Failed to extract localised video description, falling back to the standard one.', err)
             // if the workaround for localization fails, this sets the description to the potentially non-localized value
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -307,7 +307,7 @@ export default Vue.extend({
           this.videoPublished = new Date(result.videoDetails.publishDate.replace('-', '/')).getTime()
           try {
             // workaround for description localization
-            const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description.runs
+            const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description?.runs
             this.videoDescription = descriptionLines?.map(line => line.text).join('\n') ?? ''
           } catch (err) {
             console.error('Failed to extract localised video description, falling back to the standard one.', err)
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -277,7 +277,8 @@ export default Vue.extend({
           try {
             // workaround for title localization
             this.videoTitle = result.response.contents.twoColumnWatchNextResults.results.results.contents[0].videoPrimaryInfoRenderer.title.runs[0].text
-          } catch {
+          } catch (err) {
+            console.error('Failed to extract localised video title, falling back to the standard one.', err)
             // if the workaround for localization fails, this sets the title to the potentially non-localized value
             this.videoTitle = result.videoDetails.title
           }
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -308,7 +308,7 @@ export default Vue.extend({
           try {
             // workaround for description localization
             const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description?.runs
-            this.videoDescription = descriptionLines?.map(line => line.text).join('\n') ?? ''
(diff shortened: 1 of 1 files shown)


MarmadileManteater:
I accidentally pressed enter. I am still filling this out.

src/renderer/views/Watch/Watch.js
```diff
@@ -274,8 +274,13 @@ export default Vue.extend({
 
             throw new Error(`${reason}: ${subReason}`)
           }
-
-          this.videoTitle = result.videoDetails.title
+          try {
+            // workaround for title localization
+            this.videoTitle = result.response.contents.twoColumnWatchNextResults.results.results.contents[0].videoPrimaryInfoRenderer.title.runs[0].text
+          } catch {
```
absidue, line 280: It's a good idea to catch and log the error, that way it can be fixed if it breaks in the future.
```suggestion
          } catch (err) {
            console.error('Failed to extract localised title, falling back to the standard one.', err)
```

src/renderer/views/Watch/Watch.js
```diff
@@ -299,7 +304,17 @@ export default Vue.extend({
...
 
           this.videoPublished = new Date(result.videoDetails.publishDate.replace('-', '/')).getTime()
-          this.videoDescription = result.player_response.videoDetails.shortDescription
+          try {
+            // workaround for description localization
+            const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description.runs
+            this.videoDescription = ''
+            for (let i = 0; i < descriptionLines.length; i++) {
+              this.videoDescription += `${descriptionLines[i].text}\n`
+            }
```
absidue, line 313: this can be simplified by using [Array#join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join)
```suggestion
            this.videoDescription = descriptionLines?.join('\n') ?? ''
```

src/renderer/views/Watch/Watch.js
```diff
@@ -299,7 +304,17 @@ export default Vue.extend({
...
+          } catch {
```
absidue, line 314: It's a good idea to catch and log the error, that way it can be fixed if it breaks in the future.
```suggestion
          } catch (err) {
            console.error('Failed to extract localised title, falling back to the standard one.', err)
```

absidue (changes requested):

MarmadileManteater (commented):
I just approved this, but I walked away from my desk, and the more I thought about it, the more I realized it won't work because the array contains objects with a field named 'text', not strings.
image

src/renderer/views/Watch/Watch.js
```diff
@@ -299,7 +305,15 @@ export default Vue.extend({
           })
 
           this.videoPublished = new Date(result.videoDetails.publishDate.replace('-', '/')).getTime()
-          this.videoDescription = result.player_response.videoDetails.shortDescription
+          try {
+            // workaround for description localization
+            const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description.runs
+            this.videoDescription = descriptionLines?.join('\n') ?? ''
```
absidue, line 311: Oh whoops, sorry missed that part. To keep it as a one liner you could change it to this:
```suggestion
            this.videoDescription = descriptionLines?.map(line => line.text).join('\n') ?? ''
```

Or go back to your old solution, as that will definitely work.

MarmadileManteater:
I feel like these were all very good changes. I see the merging is blocked because 1 review is requesting changes, but I believe that I reviewed all of the requested changes. I can make more changes if it is necessary. I am just checking in on where we left off on this.

absidue (commented):
Hi @MarmadileManteater

Sorry for the delay, at the time I had finished the code review but hadn't tested the code yet.

While testing I noticed that it spits out the fallback error if the video is missing a description, one easy way to fix that would be to check if there is a non translated description (empty string) before doing the whole try catch stuff. Ideally we only want the error being logged if it legitimately failed to extract the translated description (not being able to do it when there is no description is expected behaviour).

MarmadileManteater:
Hi @absidue,

You are exactly correct.
image
Looking at the error, I believe it might just be easy enough to use optional chaining on the description property. Then, `descriptionLines` would be assigned the value of `[]`, and this would make `videoDescription` a blank string.

```javascript
// Adding a ? to line 310 of src\renderer\views\Watch\Watch.js
const descriptionLines = result.response.contents.twoColumnWatchNextResults.results.results.contents[1].videoSecondaryInfoRenderer.description?.runs
this.videoDescription = descriptionLines?.map(line => line.text).join('\n') ?? ''
```

absidue (approved):
Looks good to me now. Thanks!

PikachuEXE (commented):
Code looks fine but how do I test the behaviour locally?

MarmadileManteater:
These are some examples of videos I have collected of where this code changes the behavior:
- [uGQWZCVFJe4](https://youtu.be/uGQWZCVFJe4), [QtLMR7K69Mc](https://youtu.be/QtLMR7K69Mc), [9mz_HwlAurw](https://youtu.be/9mz_HwlAurw)
- [HtBqK6xsQ9k](https://youtu.be/HtBqK6xsQ9k), [NkBq2bYuvJI](https://youtu.be/NkBq2bYuvJI), [KushW6zvazM](https://youtu.be/KushW6zvazM), [hxSg2Ioz3LM](https://youtu.be/hxSg2Ioz3LM) (these are all from searching for "deco*27 feat hatsune miku")
- [CAnHY-9juA0](https://youtu.be/CAnHY-9juA0) (the video locale is `NL` but it has an `en-US` localization available; i also created this one explicitly as an example)

It would also be worth testing some videos which have a default locale of `en-US` because nothing should change for those videos . . . wait

I just noticed a problem. There are extra new lines that shouldn't be there.
image

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):
Hopefully we won't have to hardcode en-us in the future and will be able to respect user's locale preferences

PR #2536, opened 2022-09-03
Update babel and specify exact node version target

absidue:
---
Update babel and specify exact node version target
---

**Pull Request Type**

- [x] Other - Update dependencies

**Description**
Updates babel to the latest versions:

| Package | Old version | New version |
| :--- | --- | --- |
| `@babel/core` | `^7.17.10` | `^7.18.13` |
| `@babel/plugin-proposal-class-properties` | `^7.16.7` | `^7.18.6` |
| `@babel/preset-env` | `^7.17.10` | `^7.18.10` |

The babel changelogs are a bit annoying as they don't always list changes for some packages, however I can't spot anything in them that would prevent us from upgrading: https://github.com/babel/babel/releases.

I also specified the exact node version we are using in the babel config. That shrunk the main.js file by about 100 bytes and the renderer.js file by about 400 bytes, which means we are using less polyfills. Using less polyfills is a good thing as the native implementations of things are usually faster. Yay minor performance improvement.

I tested both `yarn dev` and `yarn build` and FreeTube worked in both cases.

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1


.babelrc
```diff
@@ -5,7 +5,7 @@
       {
         "targets": {
           "chrome": "104",
-          "node": 16
+          "node": "16.15.0"
```
PikachuEXE, line 8: I don't understand why this change is necessary

ChunkyProgrammer:
Could also upgrade babel-eslint "babel-eslint is now @babel/eslint-parser. This package will no longer receive updates." (3 years ago)

.babelrc
absidue: babel treats `"node": 16` as if it were `"node": "16.0.0"`. As there have been a lot of changes between node 16.0 and node 16.15.0 (the version used by the version of electron we are using), it makes sense to tell babel the exact version of node that we are targetting so that it can generate less pollyfills.

It isn't strictly necessary, though as their are only positive results of this change, it makes sense to do it.

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
OK that's minimal version
https://babeljs.io/docs/en/options#targets

Issue #1429, opened 2021-06-18
Convert "Permieres on" UTC time to local time


dennis1248:
Currently "Premieres on" seems to be using UTC time instead of local time. I live in UTC+02:00 so a stream or video premier starting at 19:00:00 local time will be reported as starting on 17:00:00 UTC.

Could someone knowledgeable with the codebase confirm that this is indeed what FreeTube is doing? I am just guessing here as to what this behavior is caused by.

Winterhuman:
@PrestonN Any update on this issue? It seems to of been incorrectly labelled as an enhancement instead of a bug

Aiz0:
I use Freetube v0.16.0 on Arch Linux and currently use CEST.
I don't experience this issue, premiere time is reported as correct local time for me.
@Winterhuman in the issue you opened you said you use FreeTube v0.13.2 Beta, have you tried any later versions of freetube?

While i say premiere time is reported correctly, it's displayed in the mm/dd/yyyy format with a 12 hour clock, I don't use this format at all.
I thought this is might be related to some locale setting on my end but It's still formatted like that despite changing LC_TIME.
Should I open an issue for this is, or is it something on my end.

Winterhuman:
@Aiz0 The date format is probably hard-coded, probably best to open an issue for supporting `LC_TIME`

Aiz0:
After setting ALL locale variables it seems to be using correct date format, yyyy-mm-dd with 24 hour clock.

Freetube uses the toLocaleString() to return the premier date as a string with locale settings.

Firefox seems to behave the same.
Launching Firefox with `LC_TIME=en_US.UTF-8` does not change date formatting.
Launching it with `LC_ALL=en_US.UTF-8` does.

It's very weird.

Winterhuman:
Yep, this is still an issue in FreeTube v0.16. `timedatectl` shows:

```
               Local time: Fri 2022-07-01 22:03:10 BST
           Universal time: Fri 2022-07-01 21:03:10 UTC
                 RTC time: Fri 2022-07-01 21:03:10
                Time zone: Europe/London (BST, +0100)
```

But, https://youtu.be/_BDClzHu-ho says it starts at `01/07/2022, 22:00:00`, and it's 22:04 now; the video should say it starts at `23:00:00` and use BST time.

Aiz0:
After Installing FreeTube via flatpak I can confirm that premiere time will display in UTC.
[...]

PR #2557, opened 2022-09-09
Properly localize numbers

ChunkyProgrammer:
---
Properly localize numbers
---

**Pull Request Type**
Please select what type of pull request this is:
- [x] Bugfix

**Related issue**
closes #2547

**Description**
Number seperators (",", "." and " ") were not chosen based on locale and were always shown as ","

**Screenshots (if appropriate)**
image

**Testing (for code that is not small enough to be easily understandable)**
Used search with different locales (had to clear search to test)

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: 11
 - FreeTube version: 0.17.1

PR #2559, opened 2022-09-09
Improve premiere timestamp to be easier to read

Aiz0:
---
Improve premiere timestamp to be easier to read
---

**Important note**

**Pull Request Type**
- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

**Related issue**
no related issue that i know of

**Description**
Changes the premiere timestamp to be easier to read.
Removes year and seconds from string because they are not important.
Displays month as it's "long name" (eg., March)

also displays timestamp in the same locale set in freetube.

before:
my locale
2022-09-10_T00:28:02

after:
my locale / system default
2022-09-10_T00:10:35
english locale
2022-09-10_T00:10:55
french locale
2022-09-10_T01:11:38

I have tested some different locales in freetube.
Open this in browser and you should be able to find some upcoming live streams.
https://www.youtube.com/live

 - OS Version: [e.g. 22]
 - FreeTube version: [e.g. 0.8]

**Additional context**
~~I noticed that the locale set in freetube was not used when using `date.toLocaleString()`.
So I searched, found, and copied the code from `watch-video-info.js` to solve this problem.
This obviously however caused 'en' to be used when i would have preferred my system locale.
I changed 'en' to 'default' to solve that.
Should the same be done to the other functions that do this?~~

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -419,6 +419,11 @@ export default Vue.extend({
                 hour: 'numeric',
                 minute: '2-digit'
               }
+              if (new Date().getFullYear() < upcomingTimestamp.getFullYear()) {
+                Object.defineProperty(timestampOptions, 'year', {
+                  value: 'numeric'
+                })
+              }
               this.upcomingTimestamp = Intl.DateTimeFormat(this.currentLocale, timestampOptions).format(upcomingTimestamp)
 
               let upcomingTimeLeft = upcomingTimestamp - new Date()
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -419,7 +419,7 @@ export default Vue.extend({
                 hour: 'numeric',
                 minute: '2-digit'
               }
-              this.upcomingTimestamp = upcomingTimestamp.toLocaleString(this.currentLocale, timestampOptions)
+              this.upcomingTimestamp = Intl.DateTimeFormat(this.currentLocale, timestampOptions).format(upcomingTimestamp)
 
               let upcomingTimeLeft = upcomingTimestamp - new Date()
 
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -419,7 +419,7 @@ export default Vue.extend({
                 hour: 'numeric',
                 minute: '2-digit'
               }
-              this.upcomingTimestamp = upcomingTimestamp.toLocaleString([locale, 'en'], timestampOptions)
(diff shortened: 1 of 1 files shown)


src/renderer/views/Watch/Watch.js
```diff
@@ -163,6 +163,9 @@ export default Vue.extend({
     },
     theatrePossible: function () {
       return !this.hideRecommendedVideos || (!this.hideLiveChat && this.isLive) || this.watchingPlaylist
+    },
+    currentLocale: function () {
+      return this.$store.getters.getCurrentLocale
```
absidue, line 168: Please don't use `getCurrentLocale`, instead do it the same way that is done in https://github.com/FreeTubeApp/FreeTube/pull/2557.

Aiz0: This should be fixed now.
You were right that I didn't need to specify 'default' when doing it this way.

efb4f5ff-1298-471a-8973-3d47447115dc:
this closes https://github.com/FreeTubeApp/FreeTube/issues/1429 right?

Aiz0:
> this closes #1429 right?

I don't think so.
That issue has to do with the flatpak build using UTC instead of the users timezone.
Locale was the same for both, i posted an image in there comparing two builds and only the timezone differed in how the timestamp was displayed.

Aiz0:
I found out about [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) and am using it to simplify the logic used to get `upcomingTimeLeft`.
Should I add these changes to this pull request or would you rather I open a new one later?

src/renderer/views/Watch/Watch.js
```diff
@@ -409,7 +413,13 @@ export default Vue.extend({
 
             if (typeof startTimestamp !== 'undefined') {
               const upcomingTimestamp = new Date(result.videoDetails.liveBroadcastDetails.startTimestamp)
-              this.upcomingTimestamp = upcomingTimestamp.toLocaleString()
+              const timestampOptions = {
+                month: 'long',
+                day: 'numeric',
+                hour: 'numeric',
+                minute: '2-digit'
```
ChunkyProgrammer, line 420: I think you should include year as well. Example: https://www.youtube.com/watch?v=O3cCYok_ukk (this video is to be published in 2024)

Aiz0: Thanks for the video.

I've added a conditional so year is displayed if the video isn't published in current year

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #2573, opened 2022-09-12
Workflow: Use slim instead of full super linter

ChunkyProgrammer:
---
Use slim linter option for super linter
---

**Pull Request Type**
- [x] Other

**Related issue**
Speeds up Super Linter workflow

**Description**
The slim github/super-linter:slim-v4 comes with all supported linters but removes the following:
- rust linters
- dotenv linters
- armttk linters
- pwsh linters
- c# linters

"By removing these linters, we were able to bring the image size down by 2gb and drastically speed up the build and download time. The behavior will be the same for non-supported languages, and will skip languages at run time."

I also removed some comments that were not needed

 - OS: Windows
 - OS Version: 11
 - FreeTube version: 0.17.1

### DIFF
--- a/.github/workflows/super-linter.yml
+++ b/.github/workflows/super-linter.yml
@@ -1,53 +1,24 @@
-#################################
-#################################
-## Super Linter GitHub Actions ##
-#################################
-#################################
 name: Lint Code Base
-
-#
 # Documentation:
 # https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions
-#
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer:
Before:
image
After:
image

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #2574, opened 2022-09-12
Electron variable at build time instead of at runtime

absidue:
---
Electron variable at build time instead of at runtime
---

**Pull Request Type**

- [x] Bundle optimisation

**Description**
This PR replaces the current runtime electron detection with build time variables, this allows webpack to optimise out (dead code elimination) web stuff in the electron build and vice versa. Unfortunately the define plugin doesn't seem to affect `.vue` files so for the Settings page i still needed to use a computed property but I was able to change it to use the variable, so it's still an improvement. 80KB drop doesn't seem too bad to me :).

@MarmadileManteater This should also be useful to you for the web build. With some adjustments you will be able to achieve even better results than the Electron build (e.g. excluding local API stuff).

`renderer.js`:
renderer_before
renderer_after

resources:
resources_before
resources_after

`yarn dev` and `yarn build`

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

### DIFF
--- a/_scripts/webpack.renderer.config.js
+++ b/_scripts/webpack.renderer.config.js
@@ -105,7 +105,10 @@ const config = {
     global: isDevMode,
   },
   plugins: [
-    // new WriteFilePlugin(),
+    new webpack.DefinePlugin({
+      'process.env.PRODUCT_NAME': JSON.stringify(productName),
+      'process.env.IS_ELECTRON': true
+    }),
     new HtmlWebpackPlugin({
       excludeChunks: ['processTaskWorker'],
       filename: 'index.html',
@@ -115,9 +118,6 @@ const config = {
         : false,
     }),
     new VueLoaderPlugin(),
-    new webpack.DefinePlugin({
-      'process.env.PRODUCT_NAME': JSON.stringify(productName),
-    }),
     new MiniCssExtractPlugin({
       filename: isDevMode ? '[name].css' : '[name].[contenthash].css',
       chunkFilename: isDevMode ? '[id].css' : '[id].[contenthash].css',
--- a/_scripts/webpack.web.config.js
+++ b/_scripts/webpack.web.config.js
@@ -109,7 +109,10 @@ const config = {
     dns: 'empty'
   },
   plugins: [
-    // new WriteFilePlugin(),
+    new webpack.DefinePlugin({
+      'process.env.PRODUCT_NAME': JSON.stringify(productName),
+      'process.env.IS_ELECTRON': false
+    }),
     new HtmlWebpackPlugin({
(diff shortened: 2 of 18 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Super linter detected code duplication ._.

ChunkyProgrammer (approved):

PikachuEXE (approved):
Tested `yarn dev` `yarn build` & built binary

MarmadileManteater:
I feel like it is worth noting that this flag feels, sort of, vestigial in the current code base, and there are some issues caused by setting it to `false`. For instance, in the `Channel` and `Trending` views, there are calls to a non-existent function that only occur when this flag is set to `false`.
```javascript
// line 184 and line 240 of src/renderer/views/Channel/Channel.js
 if (!process.env.IS_ELECTRON) {
      this.getVideoInformationInvidious()// if IS_ELECTRON is false, this always throws 
      // 'this.getVideoInformationInvidious' is not a function
```
I'm pretty sure this is should be:
```javascript
 if (!process.env.IS_ELECTRON) {
    this.getChannelInfoInvidious()
    this.getPlaylistsInvidious()
```
That technically would fix the problem, but this brings up something else. I see this same code structure repeated throughout several of the views.
```javascript
  if (!process.env.IS_ELECTRON) {
    // calls to invidious
    // (often this is just `getVideoInformationInvidious` even if the function is undefined)
  } else {
    switch (this.backendPreference) {
      case 'local':
        // calls to local api
        break
      case 'invidious':
        // calls to invidious
        break
    }
  }
```
Unless I am missing something, this could be reduced into a single condition because the same code should execute whether IS_ELECTRON is `false` or the backend preference is set to invidious.
```javascript
  if (!process.env.IS_ELECTRON || this.backendPreference === 'invidious') {
    // calls to invidious
  } else if (this.backendPreference === 'local') {
    // calls to local api
  }
```
This would eliminate the issue with `Channel` and `Trending` entirely, but it is possible there is a good reason the switch and conditional were used which I do not understand.

I like all of the changes that this PR makes. This change will definitely be helpful for the web build, but I was reminded of these existing problems related to the flag because these changes touch all of the problem areas. I wanted to say something in case others were not aware of these problems because it feels like this information might be relevant to testing this PR.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue:
Thanks for mentioning those problems, we haven't looked into the web build very much, so we hadn't noticed them yet. Would be great if we can get those issues fixed. I can look into some of them although I would appreciate you testing the web side of things when I PR it, as you've got quite a few web specific changes in your fork, like the locale handling?

PR #2585, opened 2022-09-17
Fully replace electron-debug with our own implementation

absidue:
---
Fully replace electron-debug with our own implementation
---

**Pull Request Type**

- [x] Bugfix
- [x] Feature Implementation

**Description**
This pull removes the `electron-debug` dependency and adds our own implementation for the features that we didn't already have our own implementations for. There are various advantages to getting rid of that dependency and replacing it with our own implementations:
1. Our keyboard shortcuts will work on release builds, which is fine as this isn't a proprietary app, so being able to debug easily on release builds is quite useful.
2. electron-debug was causing issues with our release webpack builds, getting rid of the dependency, means we don't need to mark it as external in the webpack config.
3. this also gets rid of 6 development dependencies which is nice, also our implementation uses a lot less code than electron-debug and it's dependencies :).

**Screenshots**
The new visible menu item:
new-menu-element

**Testing**
`yarn dev` and `yarn build`

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -80,10 +80,6 @@ function runApp() {
         }
       }
     })
-  } else {
-    require('electron-debug')({
-      showDevTools: !(process.env.RENDERER_REMOTE_DEBUGGING === 'true')
-    })
   }
 
   app.on('ready', async (_, __) => {
@@ -303,10 +299,20 @@ function runApp() {
 
     // Show when loaded
     newWindow.once('ready-to-show', () => {
-      if (newWindow.isVisible()) { return }
+      if (newWindow.isVisible()) {
+        // only open the dev tools if they aren't already open
+        if (isDev && !newWindow.webContents.isDevToolsOpened()) {
+          newWindow.webContents.openDevTools({ activate: false })
+        }
+        return
+      }
 
       newWindow.show()
       newWindow.focus()
+
+      if (isDev) {
+        newWindow.webContents.openDevTools({ activate: false })
+      }
     })
 
     newWindow.once('close', async () => {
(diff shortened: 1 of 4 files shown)


src/main/index.js
```diff
@@ -816,6 +822,21 @@ function runApp() {
...
+            click: (_, window) => {
+              if (window.webContents.isDevToolsOpened()) {
+                window.devToolsWebContents.executeJavaScript('DevToolsAPI.enterInspectElementMode()')
+              } else {
+                window.webContents.once('devtools-opened', () => {
+                  window.devToolsWebContents.executeJavaScript('DevToolsAPI.enterInspectElementMode()')
+                })
+                window.webContents.openDevTools()
+              }
+            }
```
absidue, line 838: I don't know if we really need this keyboard shortcut in the FreeTube windows, as you can use this keyboard shortcut in the devtools windows themselves already. I just added this as it was in electron-debug and I wasn't sure if anyone was actively using it (I wasn't).

PikachuEXE: I am not

PikachuEXE (commented):
Let me test this on my custom build

src/main/index.js
efb4f5ff-1298-471a-8973-3d47447115dc: Also not me

PikachuEXE (approved):
Tested on MacOS & Win 10 with custom build

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #2591, opened 2022-09-18
Replace jquery getJSON and get with native fetch

absidue:
---
Replace jquery getJSON and get with native fetch
---

**Pull Request Type**

- [x] Feature Implementation

**Related issue**
Original giant PR #2360

**Description**
This PR replaces the calls to jquery's networking functions (`getJSON` and `get`) with the `fetch` API that is available in all browsers these days. It doesn't remove the jquery dependency as it's still used in quite a few other places in FreeTube, I'll change those in future PRs.

The idea is to split up my giant jquery PR into smaller more manageable ones, as that makes them easier to test and less likely that bugs sneak into them.

1. Update check
    Set version in the `package.json` file to `0.16.0` and check that the "Update available..." banner shows up
2. Proxy test
    Click the proxy test button, if you have a proxy it will succeed. If you don't have a proxy you can check that the error handling works by looking in the devtools console
3. Switch to the Invidious API and just check that stuff still works
4. Enable SponsorBlock
    Video with segments: https://youtu.be/nFDBxBUfE74
    Video without segments: https://youtu.be/0bvZ70wgZVM
5. Enable captions on a video and see if they show up

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/renderer/components/proxy-settings/proxy-settings.js
+++ b/src/renderer/components/proxy-settings/proxy-settings.js
@@ -1,5 +1,4 @@
 import Vue from 'vue'
-import $ from 'jquery'
 import { mapActions } from 'vuex'
 import FtCard from '../ft-card/ft-card.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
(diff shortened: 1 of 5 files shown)


src/renderer/components/proxy-settings/proxy-settings.js
```diff
@@ -124,27 +123,29 @@ export default Vue.extend({
...
-      }).fail((xhr, textStatus, error) => {
-        console.log(xhr)
-        console.log(textStatus)
-        console.log(error)
-        this.showToast({
-          message: this.$t('Settings.Proxy Settings["Error getting network information. Is your proxy configured properly?"]')
+      fetch(this.proxyTestUrl)
+        .then((response) => response.json())
+        .then((json) => {
+          console.log(json)
```
ChunkyProgrammer, line 129: I know you added this just because it was in the previous code but i think we should remove this console.log

absidue: Removed it now.

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2597, opened 2022-09-19
Cleaning up calls to the Clipboard API

MarmadileManteater:
---
Cleaning up calls to the Clipboard API
---

**Pull Request Type**
- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

**Description**
I have found some issues with the current usage of the clipboard API in FreeTube. `navigator.clipboard.writeText` returns a promise, but most of the code that uses it in FreeTube doesn't seem to acknowledge that; this can lead to unexpected behavior.
```javascript
// line 337 of src/renderer/components/ft-list-video/ft-list-video.js
navigator.clipboard.writeText(this.invidiousChannelUrl)
this.showToast({// we don't necessarily know if writeText is successful when this executes
  message: this.$t('Share.Invidious Channel URL copied to clipboard')
})
```
```javascript
// line 115 of src/renderer/components/ft-share-button/ft-share-button.js
this.showToast({
  message: this.$t('Share.Invidious URL copied to clipboard')
})
this.copy(this.getFinalUrl(this.invidiousURL))// even if this is a synchronous function, it won't prevent the success message from being displayed
```
There are also some instances where the content being copied to the clipboard is not ideal.
```javascript
// line 171 of src/renderer/views/Subscriptions/Subscriptions.js
message: `${errorMessage}: ${err.responseText}`,
time: 10000,
action: () => {
  navigator.clipboard.writeText(err)// err is an object, so this is copying `[object Object]` to the clipboard
}
```
This PR aims to clean-up and centralize these calls to the clipboard API into a utility function called `copyToClipboard`.
```javascript
/**
 * This writes to the clipboard. If an error occurs during the copy,
 * a toast with the error is shown. If the copy is successful and
 * there is a success message, a toast with that message is shown.
 * @param {string} content - the content to be copied to the clipboard
 * @param {string} messageOnSuccess - the message to be displayed as a toast when the copy succeeds (optional)
 * @param {string} messageOnError - the message to be displayed as a toast when the copy fails (optional)
 */
async copyToClipboard ({ dispatch }, { content, messageOnSuccess, messageOnError }) {
  const locale = i18n._vm.locale
  const translations = i18n._vm.messages[locale]
  if (navigator.clipboard !== undefined && window.isSecureContext) {
    try {
      await navigator.clipboard.writeText(content)
      if (messageOnSuccess !== undefined) {
        dispatch('showToast', {
          message: messageOnSuccess
        })
      }
    } catch (error) {
      if (messageOnError !== undefined) {
        dispatch('showToast', {
          message: `${messageOnError}: ${error}`,
          time: 5000
        })
      } else {
        dispatch('showToast', {
          message: `${translations.Clipboard['Copy failed']}: ${error}`,
          time: 5000
        })
      }
    }
  } else {
    dispatch('showToast', {
      message: translations.Clipboard['Can not access without a secure connection'],
      time: 5000
    })
  }
}
```

This change effects all views and components that copy to the clipboard. The easiest ones to test would be the `ft-share-button` or the `ft-list-video` components. The easiest way to test the copy to clipboard events which happen as a result of clicking on toast messages is to set your invidious server to something ridiculous like `abasdcassadas` or some other fake domain.

I have tested these components and views with the clipboard API both failing and succeeding, but the only way I know to reliably cause the clipboard API to fail in electron is to pause on a break point during the function that triggers the `copyToClipboard` function and continue with the electron window unfocused. This is the error it should display:
image

 - OS: Windows 10
 - FreeTube version: 0.17.1

**Additional context**
Errors do not usually occur with the clipboard API in electron, but they are much more likely in a web browser. My main motivation for making this PR is to make it easier to extend the 'copy to clipboard' functionality in non-electron environments.

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -298,46 +298,31 @@ export default Vue.extend({
           }
           break
         case 'copyYoutube':
-          navigator.clipboard.writeText(this.youtubeShareUrl)
-          this.showToast({
-            message: this.$t('Share.YouTube URL copied to clipboard')
-          })
+          this.copyToClipboard({ content: this.youtubeShareUrl, name: 'Share.YouTube URL', t: this.$t.bind(this) })
           break
         case 'openYoutube':
           this.openExternalLink(this.youtubeUrl)
           break
         case 'copyYoutubeEmbed':
-          navigator.clipboard.writeText(this.youtubeEmbedUrl)
-          this.showToast({
-            message: this.$t('Share.YouTube Embed URL copied to clipboard')
-          })
+          this.copyToClipboard({ content: this.youtubeEmbedUrl, name: 'Share.YouTube Embed URL', t: this.$t.bind(this) })
           break
         case 'openYoutubeEmbed':
           this.openExternalLink(this.youtubeEmbedUrl)
           break
         case 'copyInvidious':
-          navigator.clipboard.writeText(this.invidiousUrl)
-          this.showToast({
-            message: this.$t('Share.Invidious URL copied to clipboard')
-          })
+          this.copyToClipboard({ content: this.invidiousUrl, name: 'Share.Invidious URL', t: this.$t.bind(this) })
           break
         case 'openInvidious':
           this.openExternalLink(this.invidiousUrl)
           break
         case 'copyYoutubeChannel':
-          navigator.clipboard.writeText(this.youtubeChannelUrl)
-          this.showToast({
-            message: this.$t('Share.YouTube Channel URL copied to clipboard')
-          })
+          this.copyToClipboard({ content: this.youtubeChannelUrl, name: 'Share.YouTube Channel URL', t: this.$t.bind(this) })
           break
         case 'openYoutubeChannel':
           this.openExternalLink(this.youtubeChannelUrl)
           break
         case 'copyInvidiousChannel':
-          navigator.clipboard.writeText(this.invidiousChannelUrl)
-          this.showToast({
-            message: this.$t('Share.Invidious Channel URL copied to clipboard')
-          })
+          this.copyToClipboard({ content: this.invidiousChannelUrl, name: 'Share.Invidious Channel URL', t: this.$t.bind(this) })
           break
         case 'openInvidiousChannel':
           this.openExternalLink(this.invidiousChannelUrl)
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -298,31 +298,31 @@ export default Vue.extend({
           }
           break
         case 'copyYoutube':
-          this.copyToClipboard({ content: this.youtubeShareUrl, name: 'Share.YouTube URL', t: this.$t.bind(this) })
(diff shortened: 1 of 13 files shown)


src/renderer/store/modules/utils.js
```diff
@@ -246,6 +246,46 @@ const actions = {
...
+        }
+      } catch (error) {
+        if (messageOnError !== undefined) {
+          dispatch('showToast', {
+            message: `${messageOnError}: ${error}`,
+            time: 5000
+          })
+        } else {
+          dispatch('showToast', {
+            message: `${translations.Clipboard['Copy failed']}: ${error}`,
```
absidue, line 276: Please use `i18n.t()` (works the same as `$t()` in vue files) here instead of needing to handle the translations yourself.

src/renderer/store/modules/utils.js
```diff
@@ -246,6 +246,46 @@ const actions = {
...
+    const translations = i18n._vm.messages[locale]
+    if (navigator.clipboard !== undefined && window.isSecureContext) {
+      try {
+        await navigator.clipboard.writeText(content)
+        if (messageOnSuccess !== undefined) {
+          dispatch('showToast', {
+            message: messageOnSuccess
+          })
+        }
+      } catch (error) {
```
absidue, line 268: Please log the error to the console as well.
```suggestion
      } catch (error) {
          console.error(`Failed to copy ${content} to clipboard`, error)
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -298,46 +298,31 @@ export default Vue.extend({
...
           break
         case 'openYoutubeChannel':
           this.openExternalLink(this.youtubeChannelUrl)
           break
         case 'copyInvidiousChannel':
-          navigator.clipboard.writeText(this.invidiousChannelUrl)
-          this.showToast({
-            message: this.$t('Share.Invidious Channel URL copied to clipboard')
-          })
+          this.copyToClipboard({ content: this.invidiousChannelUrl, messageOnSuccess: this.$t('Share.YouTube Channel URL copied to clipboard') })
```
ChunkyProgrammer, line 325: Looks like this should be should be "this.$t('Share.Invidious Channel URL copied to clipboard')"

static/locales/en-US.yaml
```diff
@@ -688,6 +688,9 @@ Share:
   YouTube URL copied to clipboard: YouTube URL copied to clipboard
   YouTube Embed URL copied to clipboard: YouTube Embed URL copied to clipboard
   YouTube Channel URL copied to clipboard: YouTube Channel URL copied to clipboard
+Clipboard:
+  Copy failed: Copy to clipboard failed
+  Can not access without a secure connection: Can not access clipboard without a secure connection
```
ChunkyProgrammer, line 693:
```suggestion
  Cannot access clipboard without a secure connection: Cannot access clipboard without a secure connection
```

src/renderer/store/modules/utils.js
MarmadileManteater: I did not realize `$t()` was accessible. I had been trying to pass it through. The reason I switched to handling the translation manually is because that is how it is handled in the `downloadMedia` function, but this is so much cleaner.

ChunkyProgrammer (approved):
LGTM

absidue (approved):

PikachuEXE (approved):
Tested copying various stuff on a video

PR #2601, opened 2022-09-20
Fixing issue with visibleDataList holding onto stale data

MarmadileManteater:
---
Fixing issue with visibleDataList holding onto stale data
---

**Pull Request Type**
- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

**Description**
This PR intends to fix an unintentional bug introduced by FreeTubeApp#2567. 

When selecting an invidious instance in the settings, the dropdown which displays possible instances will only update when the view is mounted. The result is that you can not get the dropdown to populate with invidious instances other than the current instance at the time the view was mounted:
before-change

The core issue is that the call to `updateVisibleDataList` was moved out of `handleInput` and into the watch method for `dataList`; however, there was no watch method added `inputData`, This means that `visibleDataList` sometimes retains stale data when it should be updating. This PR aims to resolve this by adding `updateVisibleDataList` to a watch method for `inputData`.

**Screenshots**
after-change

I have tested the search suggestions and the invidious instance selector, but there may be other relevant `ft-input`s to test.

 - OS: Windows 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -92,6 +92,11 @@ export default Vue.extend({
       if (val !== oldVal) {
         this.updateVisibleDataList()
       }
+    },
+    inputData(val, oldVal) {
+      if (val !== oldVal) {
+        this.updateVisibleDataList()
+      }
     }
   },
   mounted: function () {


ChunkyProgrammer (approved):

ChunkyProgrammer:
Good catch

PikachuEXE (approved):

absidue (approved):

PR #2603, opened 2022-09-21
Compress and lazy load locales

absidue:
---
Compress and lazy load locales
---

**Pull Request Type**

- [x] Feature Implementation

**Description**
This pull request adds a custom webpack plugin to compress the language files with brotli compression for the electron build. I tested gzip, brotli, zip and deflate, as those are the compression algorithms supported by node's built-in zlib library and brotli resulted the best compression ratios. While testing I also noticed that I could get the files even smaller by converting them from YAML to JSON, so the build process takes care of that as well. I also made the relevant changes to decompress and parse the JSON files at runtime.

I also implemented lazy loading of the languages which means that only the ones that are needed are actually loaded. At startup the language that the user selected and `en-US` are loaded, as `en-US` is the fallback if other languages are missing translations. When you change language in the FreeTube settings, the newly selected language is only loaded into memory if it wasn't already loaded.

As that language names are all inside the language files the build step also extracts the language names, which are then passed on to the define plugin to replace the `process.env.LOCALE_NAMES` variable.

`yarn dev` will still behave the same as it did before, so that the development process isn't slowed down by preprocessing the language files.

The language files are intentionally not compressed for the web build, as the webserver or cdn will usually automatically work out what compression algorithms the user's web browser supports and compress them for transit to the user's web browser. GitHub pages for example automatically serves static files from a cdn.

@MarmadileManteater 
For the web build the languages are loaded using `fetch`, this should mean that you don't need to include then in your zip file for the in-memory filesystem anymore. You probably can't get rid of the filesytem shim fully though as the country files are still loaded using the filesytem APIs.

For some reason webpack doesn't optimise out the `js-yaml` import in the release build, I'll look into this separately later on, as it probably happens with other stuff as well.

before:
before

after:
after

`yarn dev` and `yarn build`

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1


PikachuEXE (commented):
My brain hurts reading the code change
Include in test build first = =

src/renderer/i18n/index.js
```diff
@@ -1,31 +1,84 @@
...
+          const data = JSON.parse(brotliDecompressSync(compressed).toString())
+          this.setLocaleMessage(locale, data)
+        } catch (err) {
+          console.error(err)
+        }
+      } else {
+        const url = new URL(window.location.href)
+        url.hash = ''
+        if (url.pathname) {
+          url.pathname += `/static/locales/${locale}.json`
```
MarmadileManteater, line 59: I noticed that this uses `./static/locales` as the directory, but in `webpack.web.config.js`, the output directory is defined as `./web/static/locales`. 
```javascript
// line 162 of _scripts/webpack.web.config.js
const processLocalesPlugin = new ProcessLocalesPlugin({
  compress: false,
  inputDir: path.join(__dirname, '../static/locales'),
  outputDir: 'web/static/locales',
})
```
I believe this would cause a 404 when fetching for those resources. This could be fixed by making this point to `./web/static/locales` or by making `webpack.web.config.js` point to `./static/locales`.

absidue: The reason for having a separate subfolder for the web build is so that you can theoretically build both the electron and web builds on the same machine without needing to `yarn clean` between the builds. When deploying the web build you can just copy the contents of the web directory onto the web server, so when it's deployed the "web" directory shouldn't exist anymore.

Also as the index.html file will already be in the web directory adding `/web` to the path in the code would then result in the url to the file being wrong, as it will then be looking for a web directory inside the web directory.

MarmadileManteater: I see. I was confused because I was thinking about this wrong. I hadn't realized that index.html would be in `web/`, but that makes complete sense.

_scripts/webpack.web.config.js
```diff
@@ -160,7 +163,17 @@ if (isDevMode) {
     })
   )
 } else {
+  const processLocalesPlugin = new ProcessLocalesPlugin({
+    compress: false,
+    inputDir: path.join(__dirname, '../static/locales'),
+    outputDir: 'web/static/locales',
```
MarmadileManteater, line 169: I believe this will generate an output at `./dist/web/web/static/locales` because `web` is already defined as part of the `output` path above.
```javascript
// line 22 of _scripts\webpack.web.config.js
  output: {
    publicPath: '',
    path: path.join(__dirname, '../dist/web'),
```
image

absidue: Sorry about that, fixed it now.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #2611, opened 2022-09-23
Improve styling of premiere date box

Aiz0:
---
Improve styling of premiere date box
---

**Pull Request Type**
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

**Related issue**
no related issue

**Description**
Fixes some issues I had with the styling of the premiere date box and tries to bring it up to par with how it looks on YouTube.

- Border radius is set with `px` instead of `%`
This is good because it makes the border radius circular instead of oval-shaped.

- Adds some additional offset so the box isn't hugging the bottom-left of the video
- Use flex-box to align text vertically instead of relying on margins
The text is now more centered than ever.

- use min-width instead of a set width. 
Looks better because there's less empty space normally and the "Please refresh soon" message is put on 1 line

- increase opacity of the box from 0.7 to 0.8.
- Make timestamp smaller and bold.
looks similar to YouTube.
- adds more margin between text, icon, and box edges

Before:
2022-09-23_T22:03:50
2022-09-23_T22:04:32

After:
2022-09-23_T21:54:53
2022-09-23_T21:48:40

https://www.youtube.com/watch?v=O3cCYok_ukk

  - OS: Arch Linux
  - OS Version: 5.19.8-arch1-1
  - FreeTube version: 0.17

**Additional context**
I added a span around Premiere time left but ended up not using it

~~Might need to wait for #2610 or the other way around since both change the same line in Watch.vue~~


Aiz0:
Merge conflicts have been fixed now

src/renderer/views/Watch/Watch.sass
```diff
@@ -48,23 +48,28 @@
...
+          margin: 0 12px
 
         .premiereText
-          margin-left: 50px
-          margin-top: 10px
+          min-width: 200px
+          margin: 0 12px
+
+          .premiereTextTimestamp
+              font-size:0.85em
```
absidue, line 71: Could you please fix the formatting on this line, the rest of the code looks good.
```suggestion
              font-size: 0.85em
```

Aiz0: sure thanks

absidue (approved):
looks good

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #2612, opened 2022-09-24
Use native addEventListener instead of jquery's .on

absidue:
---
Use native addEventListener instead of jquery's .on
---

**Pull Request Type**

- [x] Feature Implementation

**Description**
This pull request uses the native `.addEventListener` method instead of jquery's `.on`. The `event.key` property unfortunately will contain an upper case or lower case letter in it depending on whether the shift or caps lock key are enabled but it's the only property that is not deprecated and also respects the user's keyboard layout, so it's the one I've used for this PR.

**Testing**
They should all behave the same as they did before:
- Global keyboard events like focusing the search bar
- Player keyboard events like seeking with the numbers, pausing, skipping frames, taking a screenshot etc
- Right, middle and left clicking on links in video descriptions

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

**Additional context**
More jquery removal PRs will follow.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -21,6 +21,7 @@ export default Vue.extend({
     'ft-card': FtCard
   },
   beforeRouteLeave: function () {
+    document.removeEventListener('keydown', this.keyboardShortcutHandler)
     if (this.player !== null) {
       this.exitFullWindow()
     }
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -1850,99 +1833,56 @@ export default Vue.extend({
...
-          case 51:
-            // 3 Key
-            // Jump to 30% in the video
-            event.preventDefault()
-            this.changeDurationByPercentage(0.3)
-            break
-          case 52:
-            // 4 Key
-            // Jump to 40% in the video
+          case '0':
```
ChunkyProgrammer, line 1836: I guess if we allow customizing keybinds in the future then we can exclude using different shortcuts for this

ChunkyProgrammer (approved):

PikachuEXE (approved):

Issue #2530, opened 2022-09-02
[Bug]: Title inconsistency between watch and ft-list-playlist


MarmadileManteater:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a bug report that matches the one I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.

### Describe the bug

1. Navigate to a playlist which uses Japanese in the video titles. EX: [PLB6Mfu4v-kdUKpNVGykSxx5CpnTP2_rcB](https://youtube.com/playlist?list=PLB6Mfu4v-kdUKpNVGykSxx5CpnTP2_rcB)
2. Click play on one of the videos

The title under the video is different than the title displayed in the playlist.
![image](https://user-images.githubusercontent.com/106682128/188131058-abe1496d-8323-4bf0-b274-10f0cc15e985.png)

### Expected Behavior

The expected behavior is just that these titles match, not that they are specifically one version. This might involve using the user's locale preference to determine which to display.

This is what I expect it to look like:
![image](https://user-images.githubusercontent.com/106682128/188131168-99f34e81-cce0-4580-a688-73fece57f809.png)

I could also expect it to look like this:
![image](https://user-images.githubusercontent.com/106682128/188139226-308227bb-ae0d-42d7-b023-88b4806ac977.png)

### Issue Labels

API issue, text/string issue

### FreeTube Version

v0.17.1

### Operating System Version

Windows 10 Pro Version 21H2 Installed on ‎4/‎3/‎2022 OS build  19044.1889 Experience	 Windows Feature Experience Pack 120.2212.4180.0

### Installation Method

.exe

### Primary API used

Local API

### Additional Information

This issue has to do with the response from the APIs. It seems that they always hand back a romanized version of the title when requesting a playlist, but they hand back a non-romanized version of the title when requesting a video directly.

[Response from the Invidious API for a single video](https://invidious.sethforprivacy.com/api/v1/videos/iLp1wjlNSm8?fields=title&pretty=1)
```json
{
  "title": "私、能力は平均値でって言ったよね! OST (CD 1) ~ 01.ティルス王国の王都!"
}
```
[...]

PR #2614, opened 2022-09-24
Fix empty description box showing for the Invidious API

absidue:
---
Fix empty description box showing for the Invidious API
---

**Pull Request Type**
- [x] Bugfix
- [x] Feature Implementation

**Description**
This pull request fixes an empty description box showing for videos that don't have a description when the Invidious API is selected. I noticed this problem while removing jquery from the description component so I've killed (fixed) two birds (issues) with one stone (PR).

before:
before

after:
after

https://youtu.be/Dmu-Yt4Nzzs (from #2530)

 - OS: Windows
 - OS Version: 10
 - FreeTube version: 0.17.1

**Additional context**
jquery is used in a lot of places

### DIFF
--- a/src/renderer/components/watch-video-description/watch-video-description.js
+++ b/src/renderer/components/watch-video-description/watch-video-description.js
@@ -2,7 +2,6 @@ import Vue from 'vue'
 import FtCard from '../ft-card/ft-card.vue'
 import FtTimestampCatcher from '../ft-timestamp-catcher/ft-timestamp-catcher.vue'
 import autolinker from 'autolinker'
-import $ from 'jquery'
 
 export default Vue.extend({
   name: 'WatchVideoDescription',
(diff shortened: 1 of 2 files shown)


src/renderer/components/watch-video-description/watch-video-description.js
```diff
@@ -31,13 +30,21 @@ export default Vue.extend({
...
-      this.shownDescription = autolinker.link(this.description)
-    }
+      const parsed = this.parseDescriptionHtml(this.descriptionHtml)
+
+      // the invidious API returns emtpy html elements when the description is empty
+      // so we need to check for that here
 
-    if (/^\s*$/.test(this.shownDescription)) {
-      $('.videoDescription')[0].style.display = 'none'
+      const testDiv = document.createElement('div')
```
ChunkyProgrammer, line 38: I'm guessing this should be removed?

absidue: No, the purpose of the test div is to check if there is any text in the descriptionHtml other than just HTML elements. I could just check for `<p></p>` but as Invidious might change in the future I wanted a more generic solution.
We are letting the browser parse the HTML by assigning it to the `innerHTML` property of the testDiv. When descriptionHtml is just `<p></p>` the `innerText` property of the testDiv is an empty string but when there is text in descriptionHtml e.g. `<p>text</p>` then the `innerText` property of the testDiv is `text`. So by checking the length of the `innerText` property of the testDiv we know whether the description is empty or not.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/watch-video-description/watch-video-description.js
absidue: Added a comment about the HTML parsing to the code now.

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/watch-video-description/watch-video-description.js
SamantazFox: If that's needed, we could make the API return an empty string.
I'd prefer for us to return clean and consistent data, rather than requiring hacky workarounds downstream!

absidue: That would be very much appreciated, as long is it doesn't break stuff for other people using the API.

Issue #539, opened 2020-10-01
Collection of all feature requests so far


GilgusMaximus:
So in this issue I summarize all feature requests that are marked as "rewrite-featurerequests" from the old FreeTubeApp.
This way everyone can take a look inside here before creating issues and might find the right issue to look into.

### Video playback

**Multiple Tabs**
#96 and #333 suggested the implementation of multiple Tabs, where each represents it's own instance of FreeTube and can play a different video than the other tabs. This is also planned for the future, once a good way of implementing is found. Also refer to the other section and the second window.

**Removing of the play button on video stop**
#347 noted that it is hard to take screenshots of a video when it is paused. Mainly because the play button in the middle of the video player cannot be removed. So an option to remove it could help in this case.

**Dynamic and temporary playlist**
#413 and #519 proposed the idea to add the ability to add videos to a temporary playlists so that they are played one after another. This way the user does not have to go back and click each video individually or create a permanent playlist, once they are implemented.

**External players**
#418 and #459 noted that the ability to play videos in external players like VLC would be very useful for quite some people. For the old FreeTube a PR was made. However this is not implemented in the rewrite at this point in time.

**Streaming video to other devices**
#437 proposed the idea to be able to stream the video to other devices, like a smart TV. Especially with a mobile application this is extra useful.

**Sponsor block**
#471 proposed the idea to integrate sponsor block into the playback of the video in order to skip the in-video ads content creators do for companies.

### Subscriptions

**Dynamic subscription sorting**
#17 mentioned a setting to sort the subscriptions by different ways. For example the most recently updated channel is on the top.

**Subscribe to playlists**
#312 suggested the idea to be able to subscribe to playlists instead of channels. So instead of fetching all new videos of a channel, only the new videos in the playlist are fetched.

**Option to not reload subscriptions automatically at start**
#330 proposed the idea to not reload all subscriptions when the program is started. Especially with many subscriptions this can take a while and therefore block the usage of the application. This could be implemented as a setting. The reload button is already available in the rewrite.

### Search

**In-Video frame instead of thumbnails**
#374 proposed the idea for an option to show a frame from the video instead of the original thumbnail.

**Select search region**
#484 noted that at the moment it is not possible to change the location of where the trending page comes from. So changing that would allow different pages to appear.

### History

**Mark video as watched**
[...]

PR #2632, opened 2022-09-28
Add setting to disable automatic fetching of subscription feed

### LINKED ISSUE #2515: [Feature Request]: Optionally disable automatic subscription fetching
### Guidelines

### Problem Description

For users with slow connections / many subscriptions, the behavior of automatically refreshing the subscriptions feed without input from the user can be frustrating and undesirable, as this process can take quite a long time and prevents usage of the app.  This is doubly true because a broad variety of seemingly unrelated behaviors can trigger an auto-refresh to occur, and it cannot be cancelled after it has begun.

### Proposed Solution

I seem to recall an option to refresh the feed exclusively manually in a previous version of the application, but I cannot locate it in the current build.  Adding an option to disable the auto-refresh behavior would greatly improve my experience as a user.

### Alternatives Considered

Increasing the speed of subscription fetching without encountering rate limiting is arguably the "correct" solution, but it's a much larger task, one that this project has tried to tackle before.  A simple toggle to only refresh the feed upon the user's request seems a suitable workaround for people with my usecase.

### Issue Labels

ease of use improvement, new optional setting

### Additional Information

v0.17.1 beta installed via AUR

Aiz0:
# Add setting to disable automatic fetching of subscription feed

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #2515
part of #539

### Main changes
Adds a new setting in subscription settings to disable automatic fetching of the subscription feed.
Prevents calls to the method `getSubscriptions()` in` Subscriptions.js` in to places.

1. When the subscriptions component is mounted
2. In `getProfileSubscriptions` method, when the users switches profile.

When the users switches profile and has disabled auto fetching 2 things can happen
1. the video list is emptied.
2. if the video list is saved in `profileSubscriptions` then it will be loaded instead.

### other stuff

In `Subscriptions.vue` I made some changes to the messages that show when the video list is empty
1. The regular "You have not subscribed to any channel" message will now only show when you have in fact not subscribed to any channel.
2. Added a new message that only shows if you have disabled automatic fetching.
3. Added a new message that will show when the user is subscribed to channels that have not uploaded any videos.

~~message 2 will block message 3 from being displayed even if they have attempted two fetch subscriptions.
This will probably not be a huge deal since I imagine it rarely happens.
The alternatives is to check if the user has attempted to refresh subscriptions.~~

## Screenshots 

This is the message that shows when the setting is disabled and the user starts the app or navigates to a new profile.
2022-09-28_T02:27:41

## Testing 

I created 2 new profiles
one without any channels to test if message 1 is displayed
the second one included only a channel without any videos to test if message 3 is displayed.

with the new setting disabled
I tested to make sure subscriptions where not fetched when freetube is started.
I switched between profiles to make sure subscriptions where not fetched automatically
I fetched subscriptions from one profile switched to another and the switched back to see if the video list was kept.

chosen api or rss should not affect anything since this check is done before that part of the code.

## Desktop

- **OS:** Arch Linux
- **OS Version:** 5.19.8-arch1-1

## Additional context

It's likely that I've missed something or implemented this weirdly since I am unfamiliar with this part of the code.

I had to remove the newline for the `Your Subscription list is currently empty` string in the locale file to get the new strings to work.

If disabling fetching when switching profile is unwanted I could make a new setting for that or just remove it.

### DIFF
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -14,6 +14,12 @@
         :tooltip="$t('Tooltips.Subscription Settings.Fetch Feeds from RSS')"
         @change="updateUseRssFeeds"
       />
+      <ft-toggle-switch
+        :label="$t('Settings.Subscription Settings.Load Subscriptions Automatically')"
+        :default-value="loadSubscriptionsAutomatically"
+        :tooltip="$t('Tooltips.Subscription Settings.Load Subscriptions Automatically')"
+        @change="updateLoadSubscriptionsAutomatically"
+      />
     </ft-flex-box>
   </ft-settings-section>
 </template>
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -8,12 +8,14 @@
           :label="$t('Settings.Subscription Settings.Fetch Automatically')"
           :default-value="fetchSubscriptionsAutomatically"
           :tooltip="$t('Tooltips.Subscription Settings.Fetch Automatically')"
+          :compact="true"
           @change="updateFetchSubscriptionsAutomatically"
         />
         <ft-toggle-switch
           :label="$t('Settings.Subscription Settings.Fetch Feeds from RSS')"
           :default-value="useRssFeeds"
           :tooltip="$t('Tooltips.Subscription Settings.Fetch Feeds from RSS')"
+          :compact="true"
           @change="updateUseRssFeeds"
         />
       </div>
@@ -21,6 +23,7 @@
         <ft-toggle-switch
           :label="$t('Settings.Subscription Settings.Hide Videos on Watch')"
           :default-value="hideWatchedSubs"
+          :compact="true"
           @change="updateHideWatchedSubs"
         />
       </div>
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -2,25 +2,29 @@
   <ft-settings-section
     :title="$t('Settings.Subscription Settings.Subscription Settings')"
   >
-    <ft-flex-box class="settingsFlexStart500px">
-      <ft-toggle-switch
-        :label="$t('Settings.Subscription Settings.Hide Videos on Watch')"
-        :default-value="hideWatchedSubs"
-        @change="updateHideWatchedSubs"
-      />
-      <ft-toggle-switch
-        :label="$t('Settings.Subscription Settings.Fetch Feeds from RSS')"
-        :default-value="useRssFeeds"
-        :tooltip="$t('Tooltips.Subscription Settings.Fetch Feeds from RSS')"
-        @change="updateUseRssFeeds"
-      />
-      <ft-toggle-switch
-        :label="$t('Settings.Subscription Settings.Fetch Automatically')"
-        :default-value="fetchSubscriptionsAutomatically"
-        :tooltip="$t('Tooltips.Subscription Settings.Fetch Automatically')"
-        @change="updateFetchSubscriptionsAutomatically"
-      />
-    </ft-flex-box>
+    <div class="switchColumnGrid">
+      <div class="switchColumn">
+        <ft-toggle-switch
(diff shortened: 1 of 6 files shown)


Aiz0:
Accidentally opened this before filling in the template... It's filled in now so I'll reopen it.

PikachuEXE (commented):
Code changes look fine
Still need to test locally tomorrow (21:08 here)

PikachuEXE (approved):
Tested
- Main changes
- Profile with no channel for message 1
- Switching between profiles to see fetched video list kept

message 3 not tested since no channel with no video provided

src/renderer/components/subscription-settings/subscription-settings.vue
```diff
@@ -14,6 +14,12 @@
         :tooltip="$t('Tooltips.Subscription Settings.Fetch Feeds from RSS')"
         @change="updateUseRssFeeds"
       />
+      <ft-toggle-switch
```
ChunkyProgrammer, line 17: I think you should add a switch column grid and two switch columns so it matches the layout of other setting components (parental control settings, distraction free settings, privacy settings, player settings)

Aiz0:
> message 3 not tested since no channel with no video provided

Sorry forgot to include a one. Didn't want to pick a random users channel.
Youtube music channel works though since it's a special channel without any videos uploaded.
https://www.youtube.com/channel/UC-9-kyTW8ZkZNDHQJ6FgpwQ

src/renderer/components/subscription-settings/subscription-settings.vue
Aiz0: I've added one now, It does look nicer. 
2022-09-29_T18:49:07
Should I change their order?

src/renderer/components/subscription-settings/subscription-settings.vue
```diff
@@ -2,25 +2,29 @@
   <ft-settings-section
     :title="$t('Settings.Subscription Settings.Subscription Settings')"
   >
-    <ft-flex-box class="settingsFlexStart500px">
```
absidue, line 5: Can you please remove the ft-flex-box component and import from the `.js` file as well please.

Aiz0: should be removed now unless i missed something

Aiz0:
I was over complicating this in my mind...
The correct message should now display in all cases.

2022-09-29_T19:15:14

absidue (approved):

ChunkyProgrammer (dismissed):

static/locales/en-US.yaml
```diff
@@ -768,6 +770,8 @@ Tooltips:
     Fetch Feeds from RSS: When enabled, FreeTube will use RSS instead of its default
       method for grabbing your subscription feed. RSS is faster and prevents IP blocking,
       but doesn't provide certain information like video duration or live status
+    Fetch Automatically: When enabled, FreeTube will automatically fetch
+      your subscription feed on start and when switching profile.
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 774: Not sure if important to mention but feed will also be fetched when opening a new window/tab

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Some observations

static/locales/en-US.yaml
absidue: The visual space between them can be fixed by adding the `:compact="true"` attribute to all 3 switches.

Aiz0: > Not sure if important to mention but feed will also be fetched when opening a new window/tab

I don't think _"on start"_ is the best wording.
I'll see if I can think of something better or if there's any suggestions.
Maybe _"when a window is opened"_ or something.

Aiz0: > The visual space between them can be fixed by adding the `:compact="true"` attribute to all 3 switches.

Thanks, I'll add that.

Aiz0: I've made the switches compact and reworded the tooltip.

PikachuEXE (approved):
All tested locally

ChunkyProgrammer (approved):

PR #2646, opened 2022-09-30
Fix the development only js-yaml dependency being included in the release builds

absidue:
# Fix the development only js-yaml dependency being included in the release builds

## Pull Request Type
- [x] Bugfix

## Related issue
follow up to #2603

## Description
`js-yaml` is only used to load the languages for the development builds, as the language files are converted to JSON and compressed during the release build process. Unfortunately webpack doesn't notice that `js-yaml` isn't needed in the release builds and still includes it in the renderer.js file. This pull request fixes that by explicitly telling webpack to ignore the `js-yaml` import for the release builds.

## Screenshots

before:
before

after:
after

## Testing
`yarn build` and then change that language to make sure that locales are still loaded properly.

## Desktop
- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/webpack.renderer.config.js
+++ b/_scripts/webpack.renderer.config.js
@@ -163,6 +163,12 @@ if (isDevMode) {
     processLocalesPlugin,
     new webpack.DefinePlugin({
       'process.env.LOCALE_NAMES': JSON.stringify(processLocalesPlugin.localeNames)
+    }),
+    // webpack doesn't get rid of js-yaml even though it isn't used in the production builds
+    // so we need to manually tell it to ignore any imports for `js-yaml`
+    new webpack.IgnorePlugin({
+      resourceRegExp: /^js-yaml$/,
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
Is the issue caused by
https://github.com/FreeTubeApp/FreeTube/blob/development/_scripts/ProcessLocalesPlugin.js#L3

absidue:
> Is the issue caused by https://github.com/FreeTubeApp/FreeTube/blob/development/_scripts/ProcessLocalesPlugin.js#L3

No, packages used by plugins are ignored unless they are used in the actual code. That plugin file is in the _scripts folder which webpack doesn't look at, so that import shouldn't make a difference.

ChunkyProgrammer (approved):

PikachuEXE (approved):

PR #2653, opened 2022-09-30
Label new PR's

efb4f5ff-1298-471a-8973-3d47447115dc:
# Label new PR's

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other - Automation 

## Description

Add the waiting for review label when pr is created


absidue (commented):
Do you need to specifically specify that list of files or would `'*'` be enough already?

Might make sense to also add the `PR: dependencies` label for PRs that change the `package.json` or `yarn.lock` files.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Do you need to specifically specify that list of files or would `'*'` be enough already?

I found out that '*' is only looking into the root files and the files in the root that have a .example1 before them are considered a seperate path.

> Might make sense to also add the `PR: dependencies` label for PRs that change the `package.json` or `yarn.lock` files.

Will do!

absidue (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):

PR #2665, opened 2022-10-02
Prevent adding extra `/` when concatenating locale paths

MarmadileManteater:
# Prevent adding extra `/` when concatenating locale paths

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
This PR formats the URLs used to fetch the locale json (in web builds) to be more consistent using `path.join`. Currently, the URLs that are being fetched can end up looking like this: `path/to/freetube//static/locales/${locale}.json` because of a trailing `/` inside of the path. This typically works fine because it often gets interpreted as `path/to/freetube/static/locales/${locale}.json` which is where the files are actually stored, but if a web build is stored on a server which is more particular about paths or if FreeTube is in Cordova, this causes an error.

## Screenshots 
*Before* (on Android in Cordova in my fork)
image
image
*After* (on Android in Cordova in my fork)
image

## Testing 
This should only effect locales fetching when `IS_ELECTRON` is `false`. 

 - OS: Windows 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/renderer/i18n/index.js
+++ b/src/renderer/i18n/index.js
@@ -1,6 +1,7 @@
 import Vue from 'vue'
 import VueI18n from 'vue-i18n'
 import fs from 'fs'
+import path from 'path'
 
 // List of locales approved for use
 import activeLocales from '../../../static/locales/activeLocales.json'
@@ -60,7 +61,7 @@ class CustomVueI18n extends VueI18n {
         }
 
         if (url.pathname) {
-          url.pathname += `/static/locales/${locale}.json`
+          url.pathname = path.join(url.pathname, 'static', 'locales', `${locale}.json`)
         } else {
           url.pathname = `/static/locales/${locale}.json`
         }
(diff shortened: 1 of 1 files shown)


absidue:
Would it be better to fix the trailing slash? Using path.join means that you'll need to include a shim/pollyfill for the path library in the web build.

absidue:
Ah looks like you are using browserify to generate the shim for you.

MarmadileManteater:
You are correct about it needing the polyfill. I used `path` because I didn't want to handle edge cases where the `pathname` does not end with a `/`, but I had forgotten that `path` is coming from the browserify polyfill on my end. Honestly, this should probably be done in a way which doesn't require that polyfill. I am trying out a version of this change which just looks at the last character of the path name and uses that to decide what to do.

absidue (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2666, opened 2022-10-02
Replace jquery in live chat and improve the message styling

absidue:
# Replace jquery in live chat and improve the message styling

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description
This pull request removes jqeury from the live chat component, some cleanup and fixes the emojis messing up the layout of the messages.

## Screenshots 
There is a visible change but i didn't want to screenshot and post live chat comments here

## Testing 
- Add my updated fork of youtube-chat (includes the open PR) `yarn add https://github.com/absidue/youtube-chat`
- Pick a live stream from https://www.youtube.com/live/
- Check that the live chat scrolls and that the scroll to bottom button shows up when you scroll and also works

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/watch-video-live-chat/watch-video-live-chat.css
+++ b/src/renderer/components/watch-video-live-chat/watch-video-live-chat.css
@@ -137,12 +137,6 @@
   overflow-y: auto;
 }
 
-@media (prefers-reduced-motion: no-preference) {
-  .liveChatComments {
-    scroll-behavior: smooth;
-  }
-}
-
 .comment .superChatMessage {
   padding: 5px;
 }
--- a/src/renderer/components/watch-video-live-chat/watch-video-live-chat.css
+++ b/src/renderer/components/watch-video-live-chat/watch-video-live-chat.css
@@ -19,6 +19,11 @@
   margin: 0;
 }
 
+:deep(.liveChatEmoji) {
+  vertical-align: middle;
+  margin: 0 2px;
+}
+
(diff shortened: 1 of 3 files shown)


PikachuEXE (commented):
It can be scrolled but not auto scrolled to bottom (Or do you mean it can be scrolled)
Scroll to bottom button shows up when you scroll and also works

Which open PR is related? Should that be merged and released before this is merged?

absidue:
This youtube-chat PR is related: https://github.com/FreeTubeApp/youtube-chat/pull/6 (currently has changes requested that need to be addressed)
Without that PR you just end up in an endless loop of live chat errors in the console, the moment that anyone sends an emoji, but to anyone that doesn't have the console open it just looks like nothing is happening in the live chat. The only way to fix it is to close FreeTube entirely.

Hmm okay so the scrolling seems to be an issue with the native scrollTo function. If you can scrollTo again before it has finished smooth scrolling to the position specified in the original scrollTo call it just does nothing. So the current implementation works fine for a slow chat but I might need to implement my own scrolling with setInterval to fix that problem.

absidue:
Fixed the scrolling, I had forgot about vue not making the changes to the DOM immediately.

PikachuEXE (approved):
Let the (conditional) console spam begin

ChunkyProgrammer (approved):

PR #2667, opened 2022-10-03
Prevent search from closing when using a virtual keyboard

MarmadileManteater:
# Prevent search from closing when using a virtual keyboard

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
This PR adds a condition to the window `resize` event listener which prevents the search bar from closing unless the width of the page has changed. On mobile, the screen height may change when opening a virtual keyboard, but the screen width does not typically change. This change should prevent issues with the search bar closing when opening a virtual keyboard.

## Screenshots 
*before* (on Android in my fork)

*after* (on Android in my fork)
 

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional context
This would also effect Electron builds. I am open to modifying this PR to only change the existing behavior when `IS_ELECTRON` is `false` if that would make more sense.

### DIFF
--- a/src/renderer/components/top-nav/top-nav.js
+++ b/src/renderer/components/top-nav/top-nav.js
@@ -76,6 +76,7 @@ export default Vue.extend({
     }
   },
   mounted: function () {
+    let previousWidth = window.innerWidth
     if (window.innerWidth <= 680) {
       this.showSearchContainer = false
     }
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer:
This would probably also be benefitial for the pinephone which is supported by FreeTube as well as some laptops so I don't think this would need to check if IS_ELECTRON is false

ChunkyProgrammer (approved):
LGTM

PikachuEXE (approved):

PR #2678, opened 2022-10-05
Making `pack:web` work by adding polyfill

MarmadileManteater:
# Making `pack:web` work by adding polyfill

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
This PR makes the `pack:web` script work by adding the polyfill from browserify to webpack. This should effectively be the same thing as running `browserify` on the result of webpack; except that, it is way faster to build, and it won't bugger up the nice clean directory structure that webpack puts in `Inspect Element` in development builds. These things alone make this web build infinitely easier to debug and run than my existing web build. 

## Screenshots 

Before:
image
After:
image
image
image

image
## Testing 
`yarn pack:web` will create all of the files needed for the web build, but since this web app needs to be hosted on a server and cannot be served directly off the filesystem, you will probably want to break out your favorite one liner like `http-server` and navigate to `http://127.0.0.1:8080/dist/web/index.html`.

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional context
It may be worth noting, but there is code currently in the codebase which assumes that the application is in Electron if the environment is development. That code may have to be changed in the future if it would make sense to run the web build in development mode.

One major caveat of this solution is that it depends on bringing in two new packages `browserify@^17.0.0` and `process@^0.11.10`. I don't see a way to achieve this without bringing in at least one new package or manually writing all of our own polyfill. Both of those solutions have their advantages and disadvantages.


ChunkyProgrammer:
I think you still need to run `yarn install` to update the lock file

ChunkyProgrammer (dismissed):

PikachuEXE (approved):

absidue (approved):

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2683, opened 2022-10-06
Replacing `setImmediate` with `setTimeout`

MarmadileManteater:
# Replacing `setImmediate` with `setTimeout`

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
`setImmediate` is a global in node, but it is technically also part of the `timers` module, and it is not a global in web. This causes an error to occur when `showToast` is called in a browser.

## Screenshots 
_Before: (in a web build)_
image
_After:_
image

## Testing 
This change can be tested by checking if any toast messages display in both electron and web environments.

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional context
This error did not occur in my original web build because `setImmediate` was being globally defined, but I believe it makes more sense to import it explicitly as to not pollute global scope. Technically, this same behavior could be achieved by modifying the `webpack.web.config.js` to globally define `setImmediate` only in web builds.

### DIFF
--- a/src/renderer/components/ft-toast/ft-toast.js
+++ b/src/renderer/components/ft-toast/ft-toast.js
@@ -1,6 +1,5 @@
 import Vue from 'vue'
 import FtToastEvents from './ft-toast-events.js'
-import { setImmediate } from 'timers'
 
 export default Vue.extend({
   name: 'FtToast',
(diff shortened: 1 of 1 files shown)


absidue:
You should be able to replace `setImmediate(() => { toast.isOpen = true })` with `setTimeout(() => { toast.isOpen = true })` or `Promise.resolve().then(() => { toast.isOpen = true })`. The only reason it even works in electron is because we have the node integration enabled but we should probably avoid using node specific stuff if we don't need to.

ChunkyProgrammer (approved):

PR #2684, opened 2022-10-07
Move fetching of the sponsorblock segments out of the store

absidue:
# Move fetching of the sponsorblock segments out of the store

## Pull Request Type

- [x] Feature Implementation

## Description
As the `sponsorBlockSkipSegments` function only needs to read from the store, without adding extra properties or updating values in the store, it can easily be moved out of the store. This allows it to be imported with a normal ES6 import, instead of needing to use vuex's `mapActions()` to import it. Importing the function with an ES6 import also gives you autocomplete and allows webpack to optimise a bit better (webpack doesn't understand vuex's mapActions magic so it can't do a lot of optimisation).

Although it looks like there are a lot of changes to `ft-video-player.js`, most of them are just indentation changes.

## Testing
Open https://youtu.be/6T9d9LM1TwY with SponsorBlock enabled.

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (approved):
Tested locally

Also you can review changes without whitespace changes
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #2686, opened 2022-10-07
Remove unused property from the invidious store

absidue:
# Remove unused property from the invidious store

## Pull Request Type

- [x] Cleanup

## Description
This pull request removes the unused `isGetChannelInfoRunning` property from the invidious store. I couldn't find any other references to that property so it doesn't seem worth keeping it around.

Technically in vuex 3 it's all one big store, the concept of separate stores only exists in Pinia.

## Testing 
Channel pages and fetching subscriptions still work fine.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/store/modules/invidious.js
+++ b/src/renderer/store/modules/invidious.js
@@ -2,15 +2,10 @@ import fs from 'fs'
 
 const state = {
   currentInvidiousInstance: '',
-  invidiousInstancesList: null,
-  isGetChannelInfoRunning: false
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):
Speaking of Pinia, it'd probably be a good idea to migrate to that sometime

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #2689, opened 2022-10-08
Add shortcuts for refresh buttons on Subscription, Trending, and Popular views

### LINKED ISSUE #2631: [Feature Request]: Shortcut to refresh subscriptions (R)
### Guidelines

### Problem Description

I often refresh my subscriptions with `CTRL + R` but this reloads the entire view and is therefore slower than just refreshing the subscriptions.

### Proposed Solution

I would like a shortcut key for the `refresh subscriptions` button.
`R` would be my recommendation, Since it's already used with modifiers to reload the app.

### Alternatives Considered

Refreshing just the subscriptions with  `CTRL + R` might be better, i'm not sure.
Since the plan is to replace all other refresh buttons with https://github.com/FreeTubeApp/FreeTube/pull/1578 a shortcut for that would be preferable. There hasn't been any updates to that for about a year now since the rework was planned so I don't know how it would reload the subscription page.

### Issue Labels

ease of use improvement, new keyboard shortcut

### Additional Information

Other relevant issues.
https://github.com/FreeTubeApp/FreeTube/issues/1966

Aiz0:
# Add shortcuts for refresh buttons on Subscription, Trending, and Popular views

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #2631

## Description

Adds shortcuts to the 3 views with refresh buttons.
I stole the keyboard shortcut handler function from ft-video-player.js
Just a quick implementation.

## Screenshots 

No visible Difference

## Testing 

Testing should be done on all three views

- Pressing `R` When the view isn't loading should perform the same action as the refresh button.
- Nothing should happen if `R` is pressed while the view is loading
- Leaving the view should destroy the event handler and pressing `R` should no longer do anything
- Typing `R` into a input element. such as the search box should not trigger a refresh.

## Desktop

- **OS:** Arch Linux
- **OS Version:** 5.19.13-arch1-1

## Additional context

Since everything uses the same implementation with only differences being which function to call there ends up being a lot of code duplication. I'm open to other ways to implement it (if you could point me to the right direction).

### DIFF
--- a/src/renderer/views/Popular/Popular.js
+++ b/src/renderer/views/Popular/Popular.js
@@ -69,7 +69,9 @@ export default Vue.extend({
       switch (event.key) {
         case 'r':
         case 'R':
-          this.fetchPopularInfo()
+          if (!this.isLoading) {
+            this.fetchPopularInfo()
+          }
           break
       }
     },
--- a/src/renderer/views/Popular/Popular.js
+++ b/src/renderer/views/Popular/Popular.js
@@ -25,10 +25,15 @@ export default Vue.extend({
     }
   },
   mounted: function () {
+    document.addEventListener('keydown', this.keyboardShortcutHandler)
+
     this.shownResults = this.popularCache
     if (!this.shownResults || this.shownResults.length < 1) {
       this.fetchPopularInfo()
     }
+  },
+  beforeDestroy: function () {
+    document.removeEventListener('keydown', this.keyboardShortcutHandler)
   },
   methods: {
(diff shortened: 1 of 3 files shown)


absidue (approved):

PikachuEXE (approved):

PikachuEXE:
forgot to add "Tested locally"

ChunkyProgrammer (approved):

PR #2696, opened 2022-10-09
Prevent default event for shortcut

Aiz0:
# Prevent default event for shortcut

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

no related issue

## Description

Adds `event.preventDefault()` to the shortcuts added in https://github.com/FreeTubeApp/FreeTube/pull/2689.
This fixes an issue where pressing shortcut also caused sidebar to expand/close when hamburger button was last element clicked.

## Screenshots 

no visible difference
## Testing 

1. Go to Subscription page
2. Press the hamburger menu to expand/close the sidebar.
3. Press the any alpha character key such as the `A` key 
4. See that the menu expands or closes. (Is this intended?)
5. Press the `R` key 
6. See that nothing happens to the menu.

## Desktop

- **OS:** Arch Linux
- **OS Version:** 5.19.13-arch1-1

## Additional context

I don't really know why the sidebar behaves that way or if it's something to be fixed.

### DIFF
--- a/src/renderer/views/Popular/Popular.js
+++ b/src/renderer/views/Popular/Popular.js
@@ -69,6 +69,7 @@ export default Vue.extend({
       switch (event.key) {
         case 'r':
         case 'R':
+          event.preventDefault()
           if (!this.isLoading) {
             this.fetchPopularInfo()
           }
--- a/src/renderer/views/Subscriptions/Subscriptions.js
+++ b/src/renderer/views/Subscriptions/Subscriptions.js
@@ -484,6 +484,7 @@ export default Vue.extend({
       switch (event.key) {
         case 'r':
         case 'R':
+          event.preventDefault()
           if (!this.isLoading) {
             this.getSubscriptions()
           }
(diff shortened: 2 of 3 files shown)


Aiz0:
`event.preventDefault()` seems to be missing from some video player shortcuts causing the same behavior.

Should I add it to those as well or should we consider some other solution?

Aiz0:
This line seems to be the issue
https://github.com/FreeTubeApp/FreeTube/blob/7ca6440a888008401caeb5ebdc892afa0f318e37/src/renderer/components/top-nav/top-nav.vue#L13
It seems to read all keypressess.
Using `@keydown.enter.prevent` would probably be preferable.

I'll open a new PR to address that.
preventing default events for these keys will no longer be necessary as far as I know.

Aiz0:
Closing this because i opened the other one.
I can reopen this if we do want to add event.preventDefault() to the keybinds.

Issue #1580, opened 2021-08-12
Adding eslint-plugin-vuejs-accessibility to linter rules


kommunarr:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Currently, FreeTube's linter has little to no rules necessitating basic accessible design. Without accessibility rules in place, devs unknowingly utilize harmful anti-patterns and push out features that disabled users cannot benefit from. Irregularly addressing these built-up accessibility issues all at once is an extremely comprehensive, daunting, and time-consuming task. Instituting thorough linter accessibility rules will help keep devs accountable for their own code, lower sum development time, and improve the overall code quality of FreeTube.

**Describe the solution you'd like to see implemented**
[`eslint-plugin-vuejs-accessibility`](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility) is an open source `eslint` plugin whose gamut of accessibility rules will massively improve the accessibility & code quality of FreeTube.

I strongly recommend that all of the recommended accessibility rules from `eslint-plugin-vuejs-accessibility` be added to FreeTube's `.eslintrc.js` as such:
```
{
  "rules": {
    "vuejs-accessibility/rule-name": "error"
  }
}
```
For newly required changes to the existing project that cannot be automatically applied, I would be glad to implement them all concomitantly in #1525. Alternatively, a separate PR can be made if needed.

**Describe alternatives you've considered**
1. [Less favorable] Implement most (but not all) of the accessibility rules. If there are certain pre-existing features that would require a difficult rework, implementing only most of the accessibility rules is a possible workaround (albeit not favorable).
2. Use a different open source project's accessibility linter rules.

**Additional context**
I would categorize this as a relatively high priority. Every new PR that does not follow accessibility guidelines is more work needed down the line. This is increasingly unsustainable, and at some point the duty of basic accessibility needs to shift onto devs' own code.

PR #2709, opened 2022-10-11
Improve accessibility of FreeTube input elements

ChunkyProgrammer:
# Improve accessibility of FreeTube input elements

## Pull Request Type
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Implements part of #1525 (removes jquery)

## Description
Makes generic ft-(input) items more accessible. Future PRs will tackle more components.

## Testing 
FtDropdown:
- Find a Playlist
- Tab to the "Share" bar (clicking on the description then clicking tab might be easier)
- Click enter on the "Share"
- Click enter on copy link

FtNotificationBanner:
- modify version in package.json to a previous version (0.16.0)(verify that check for app updates is enabled in the setting as well)
- Tab to notification banner
- Click enter
- (Repeat but instead of clicking enter, tab to the close button then click enter)

FtPrompt:
- go to data settings
- click "Export Subscriptions"
- click tab then click enter

FtIconButon
- navigate to a video
- tab to the share button
- click enter
- tab to different options in list
- click enter

FtSelect
- go to settings
- go to download settings
- tab to "Download Behavior"
- click arrow keys up/down
- see that value changes
- click enter
- list of values opens up

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** 0.17.2

## Additional context
Co-Authored-By: Jason <84899178+jasonhenriquez@users.noreply.github.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "autolinker": "^4.0.0",
     "browserify": "^17.0.0",
     "electron-context-menu": "^3.5.0",
+    "eslint-plugin-vuejs-accessibility": "^1.2.0",
     "http-proxy-agent": "^4.0.1",
     "https-proxy-agent": "^5.0.0",
     "lodash.debounce": "^4.0.8",
--- a/src/renderer/components/ft-icon-button/ft-icon-button.vue
+++ b/src/renderer/components/ft-icon-button/ft-icon-button.vue
@@ -12,8 +12,12 @@
         padding: padding + 'px',
         fontSize: size + 'px'
       }"
+      tabindex="0"
+      role="button"
       @click="handleIconClick"
       @mousedown="handleIconMouseDown"
+      @keydown.enter.prevent="handleIconClick"
+      @keydown.space.prevent="handleIconClick"
     />
     <div
       v-show="dropdownShown"
@@ -33,12 +37,19 @@
         <ul
           v-if="dropdownOptions.length > 0"
           class="list"
+          role="listbox"
+          aria-expanded="false"
         >
           <li
             v-for="(option, index) in dropdownOptions"
+            :id="removeWhitespace(title + '-' + index)"
             :key="index"
+            role="option"
+            aria-selected="false"
+            tabindex="-1"
             :class="option.type === 'divider' ? 'listItemDivider' : 'listItem'"
-            @click="handleDropdownClick({url: option.value, index: index})"
+            @click="handleDropdownClick({url: option.value, index: index}, $event)"
+            @keydown="handleDropdownClick({url: option.value, index: index}, $event)"
           >
             {{ option.type === 'divider' ? '' : option.label }}
           </li>
--- a/src/renderer/components/ft-list-dropdown/ft-list-dropdown.js
+++ b/src/renderer/components/ft-list-dropdown/ft-list-dropdown.js
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-
+import { removeWhitespace, handleDropdownKeyboardEvent } from '../../helpers/accessibility'
 export default Vue.extend({
   name: 'FtListDropdown',
   props: {
@@ -30,6 +30,37 @@ export default Vue.extend({
   computed: {
     listType: function () {
       return this.$store.getters.getListType
+    }
+  },
+  methods: {
+    removeWhitespace,
+    handleIconKeyPress(event) {
+      if (event instanceof KeyboardEvent) {
+        if (event.key === 'Tab') {
+          return
+        }
+
+        event.preventDefault()
+
+        if (event.key !== 'Enter' && event.key !== ' ' && event.key !== 'ArrowDown') {
+          return
+        }
+      }
+
+      const firstOption = document.getElementById('buttonOption0')
+      firstOption.setAttribute('tabIndex', 1)
+      firstOption.focus()
+    },
+    handleDropdownClick: function(index, event) {
+      if (!handleDropdownKeyboardEvent(event, event?.target)) {
+        return
+      }
+
+      const unspacedTitle = removeWhitespace(this.title)
+      const allOptions = document.querySelector(`#${unspacedTitle} + ul`)
+      allOptions.setAttribute('tabindex', '-1')
+      event.target.setAttribute('tabindex', '0')
+      this.$emit('click', this.labelValues[index])
     }
   }
 })
--- a/src/renderer/components/ft-list-dropdown/ft-list-dropdown.vue
+++ b/src/renderer/components/ft-list-dropdown/ft-list-dropdown.vue
@@ -1,6 +1,12 @@
 <template>
-  <div class="dropDown">
-    <div class="buttonTitle">
+  <div class="dropdown">
+    <div
+      :id="removeWhitespace(title)"
+      class="buttonTitle"
+      tabindex="0"
+      role="listbox"
+      @keypress="handleIconKeyPress($event)"
+    >
       {{ title }}
       <font-awesome-icon
         class="angleDownIcon"
@@ -10,9 +16,14 @@
     <ul>
       <li
         v-for="(label, index) in labelNames"
+        :id="'buttonOption' + index"
         :key="index"
         class="buttonOption"
-        @click="$emit('click', labelValues[index])"
+        role="option"
+        tabindex="-1"
+        aria-selected="false"
+        @click="handleDropdownClick(index, $event)"
+        @keydown="handleDropdownClick(index, $event)"
       >
         {{ label }}
       </li>
--- a/src/renderer/components/ft-prompt/ft-prompt.js
+++ b/src/renderer/components/ft-prompt/ft-prompt.js
@@ -2,6 +2,7 @@ import Vue from 'vue'
 import FtCard from '../../components/ft-card/ft-card.vue'
 import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
 import FtButton from '../../components/ft-button/ft-button.vue'
+import { removeWhitespace } from '../../helpers/accessibility'
 
 export default Vue.extend({
   name: 'FtPrompt',
@@ -26,11 +27,20 @@ export default Vue.extend({
     optionValues: {
       type: Array,
       default: () => { return [] }
+
     }
   },
+  mounted: function () {
+    setTimeout(() => {
+      document.querySelector('.prompt')
+        .querySelectorAll('button')[0]
+        .focus()
+    })
+  },
   methods: {
+    removeWhitespace,
     handleHide: function (event) {
-      if (event.target.className === 'prompt') {
+      if (event.target.getAttribute('role') === 'button' || event.target.className === 'prompt') {
         this.$emit('click', null)
       }
     }
--- a/src/renderer/components/ft-prompt/ft-prompt.vue
+++ b/src/renderer/components/ft-prompt/ft-prompt.vue
@@ -1,11 +1,19 @@
 <template>
   <div
     class="prompt"
+    tabindex="-1"
     @click="handleHide"
+    @keydown.enter="handleHide"
   >
-    <ft-card class="promptCard">
+    <ft-card
+      class="promptCard"
+      :aria-labelledby="'dialog-' + removeWhitespace(label)"
+    >
       <slot>
-        <h2 class="center">
+        <h2
+          :id="'dialog-' + removeWhitespace(label)"
+          class="center"
+        >
           {{ label }}
         </h2>
         <p
--- a/src/renderer/components/ft-select/ft-select.vue
+++ b/src/renderer/components/ft-select/ft-select.vue
@@ -1,9 +1,11 @@
 <template>
   <div class="select">
     <select
+      :id="removeWhitespace(placeholder)"
       class="select-text"
       :class="{disabled: disabled}"
       :value="value"
+      :name="removeWhitespace(placeholder)"
       :disabled="disabled"
       @change="$emit('change', $event.target.value)"
     >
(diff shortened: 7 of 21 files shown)


src/renderer/components/ft-list-dropdown/ft-list-dropdown.vue
```diff
@@ -1,6 +1,12 @@
 <template>
-  <div class="dropDown">
-    <div class="buttonTitle">
+  <div class="dropdown">
+    <div
+      :id="removeWhitespace(title)"
+      class="buttonTitle"
+      tabindex="0"
+      role="listbox"
+      @keypress="handleIconKeyPress($event)"
```
ChunkyProgrammer, line 8:
```suggestion
      @keydown="handleIconKeyPress($event)"
```

src/renderer/components/ft-prompt/ft-prompt.js
```diff
@@ -26,11 +27,20 @@ export default Vue.extend({
     optionValues: {
       type: Array,
       default: () => { return [] }
+
```
ChunkyProgrammer, line 30:
```suggestion
```

ChunkyProgrammer (commented):
Some small changes i need to make

PikachuEXE:
Can I wave white flag so many things to test ._.

PikachuEXE (approved):
Tested
- FtDropdown
- FtPrompt - however if I press tab "export YT(csv)" is highlighted, is it expected
- FtIconButon
- FtSelect

Not tested (lazy ._.)
- FtNotificationBanner

PikachuEXE:
Tested updated FtPrompt

How do I escape from `Select Export Type` with keyboard only?

ChunkyProgrammer:
> Tested updated FtPrompt
> 
> How do I escape from `Select Export Type` with keyboard only?

Just added a close button

PikachuEXE (approved):

src/renderer/components/ft-prompt/ft-prompt.js
```diff
@@ -30,14 +30,21 @@ export default Vue.extend({
...
+      this.$emit('click', null)
+    },
     handleHide: function (event) {
       if (event.target.getAttribute('role') === 'button' || event.target.className === 'prompt') {
-        this.$emit('click', null)
+        this.hide()
       }
+    },
+    focusItem: function (index) {
+      document.querySelector(`#prompt-${removeWhitespace(this.label)}-${index}`)
```
absidue, line 46: This seems to be broken for the restart prompt for the smooth scrolling toggle:
smooth_scrolling_prompt_error

src/renderer/components/ft-prompt/ft-prompt.vue
```diff
@@ -28,9 +28,19 @@
         <ft-flex-box>
           <ft-button
             v-for="(option, index) in optionNames"
+            :id="'prompt-' + removeWhitespace(label) + '-' + index"
             :key="index"
             :label="option"
             @click="$emit('click', optionValues[index])"
+            @keydown.left.prevent="focusItem(index-1)"
+            @keydown.right.prevent="focusItem(index+1)"
```
absidue, line 36: This probably needs handling (wasn't able to test my theory due to the error above) for when the left key is pressed on the first button and the right key is pressed on the last button in the prompt.

src/renderer/components/ft-icon-button/ft-icon-button.vue
```diff
@@ -33,12 +37,19 @@
         <ul
           v-if="dropdownOptions.length > 0"
           class="list"
+          role="listbox"
+          aria-expanded="false"
```
absidue, line 41: The expanded property should be updated when the drop down is shown and hidden.
```suggestion
          :aria-expanded="dropdownShown"
```

src/renderer/components/ft-icon-button/ft-icon-button.vue
```diff
@@ -33,12 +37,19 @@
...
             v-for="(option, index) in dropdownOptions"
+            :id="removeWhitespace(title + '-' + index)"
             :key="index"
+            role="option"
+            aria-selected="false"
+            tabindex="-1"
             :class="option.type === 'divider' ? 'listItemDivider' : 'listItem'"
-            @click="handleDropdownClick({url: option.value, index: index})"
+            @click="handleDropdownClick({url: option.value, index: index}, $event)"
+            @keydown="handleDropdownClick({url: option.value, index: index}, $event)"
```
absidue, line 52: Should every keyboard event really trigger the selection or should only specific ones do it?

src/renderer/components/ft-list-dropdown/ft-list-dropdown.js
```diff
@@ -31,5 +31,36 @@ export default Vue.extend({
...
+
+        if (event.key !== 'Enter' && event.key !== ' ' && event.key !== 'ArrowDown') {
+          return
+        }
+      }
+
+      const firstOption = document.getElementById('buttonOption0')
+      firstOption.setAttribute('tabIndex', 1)
+      firstOption.focus()
+    },
```
absidue, line 53: This can be handled a lot cleaner and with less code by filtering the keys on the vue side.

absidue:
> > Tested updated FtPrompt
> > How do I escape from `Select Export Type` with keyboard only?
> 
> Just added a close button

Would probably be good to make the <kbd>Escape</kbd> key close it too.

src/renderer/components/ft-prompt/ft-prompt.js
ChunkyProgrammer: I'll update the replace function to remove special characters as well

absidue: Might be easier to use a regex whitelist e.g. `/^[a-Za-z0-9]$/`, that way you don't have to worry about adding more characters to the blacklist in the future.

ChunkyProgrammer:
@absidue i believe that I addressed your concerns

PikachuEXE (dismissed):
Escape!!!

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @RastislavKish Is it possible for u to help out with testing this PR? This PR only implements a part of https://github.com/FreeTubeApp/FreeTube/pull/1525 but we want to make sure the things implemented in this PR behaves the same for u as #1525.

efb4f5ff-1298-471a-8973-3d47447115dc:
i think this closes https://github.com/FreeTubeApp/FreeTube/issues/1580

package.json
```diff
@@ -56,6 +56,7 @@
     "autolinker": "^4.0.0",
     "browserify": "^17.0.0",
     "electron-context-menu": "^3.5.0",
+    "eslint-plugin-vuejs-accessibility": "^1.2.0",
```
absidue, line 59: This eslint plugin should be listed as a development dependency, as it is only needed while developing, not while you are actually using FreeTube. I'm alright with browserify being in the dependencies because the polyfills it provides are included in the web builds, but eslint stuff is really only used during development.

PikachuEXE (dismissed):

RastislavKish:
@efb4f5ff-1298-471a-8973-3d47447115dc thanks for notifying me!
Sure, I'll test these out.

> FtDropdown:
* Find a Playlist
* Tab to the "Share" bar (clicking on the description then clicking tab might be easier)
* Press enter on the "Share"
* Press enter on copy link

To be honest, this is the first time I see this menu, it's present, but it's so inconspicuous for a screenreader I almost overlooked it even when searching for it.

I wonder, what is it exactly? An expandable menu?

Anyway, the menu is basically working, the arrow navigation works as it should, just like pressing the return key. I don't have a share item in the menu, but the one for copying playlist link is present and works.

One problem is, the screenreader for some reason tends to forget the menu is a menu after moving in either direction and switches into browsing mode intended for webpage reading, which doesn't send key commands to the elements.
The SR's focus mode has to be turned on manually again to make the SR keep sending key presses to it.

Though, this behavior appears also in #1525, and I don't remember seeing it anywhere else inn the app.

Are there other places where this component is being used?

absidue:
That component is only be used in that place, we should probably replace it with a better share menu in the future.

efb4f5ff-1298-471a-8973-3d47447115dc:
When u hover over that component with your cursor it will expand the menu. Indeed a bit weird to only have that menu there. I agree we should probably replace it

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
@RastislavKish i'll wait for u to give the green light on this one before i approve.

ChunkyProgrammer:
I won't be updating the playlist share menu as we'll likely replace it with the same menu on the video page for sharing

efb4f5ff-1298-471a-8973-3d47447115dc:
@RastislavKish are there other things regarding this pr u like changes for except for the playlist share menu?

RastislavKish:
Here are my test results, I apologize for the delay

FtNotificationBanner

The banner works as expected, I can both open more details and close the banner.

One peculiarity that appears and is not present in #1525 is the behavior of the more details page, which sort of disappears when I arrow navigate in the subscriptions section, which stays present. I'm not sure whether this is also a part of the component, so I'm mentioning it just in case.

FtPrompt

The component is basically accessible, there is a slight diversion from the behavior of #1525. After opening the menu, tabbing out of it makes it disappear, as well as simply navigating out by arrow keys.
In #1525, the menu simply exposes its content via the buttons and doesn't explicitly respond to keyboard actions except the return key. This way, the user can explore the whole menu without accidentally closing it when they cross the border by either tabbing or using the arrow keys.

FtIconButton

This element behaves identically to #1525.
One thing I just noticed is the way it's composed, as far as I understand, there are two panes, Youtube and Invidious, with each having buttons for copying and opening the link/embed.
One problem is, when the menu is supposed to be used by tabbing, the information about the pane of a particular button is not spoken. Seems like something that could be easily resolved by aria tags, like [aria-describedby,](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) depending on the structure of the menu.

FtSelect

Works as expected.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @RastislavKish, changes have been implemented. Please test them if u can :)

PikachuEXE (dismissed):

src/renderer/components/ft-prompt/ft-prompt.js
```diff
@@ -28,10 +29,61 @@ export default Vue.extend({
...
+      } else if (index >= this.promptButtons.length) {
+        index = 0
+      }
+      if (index >= 0 && index < this.promptButtons.length) {
+        this.promptButtons[index].focus({ focusVisible: true })
+      }
+    },
+    // close on escape key and unfocus
+    closeEventFunction: function(event) {
+      const targ = event.target
```
ChunkyProgrammer, line 73:
```suggestion
```

src/renderer/components/ft-prompt/ft-prompt.js
```diff
@@ -28,10 +29,61 @@ export default Vue.extend({
...
-      if (event.target.className === 'prompt') {
-        this.$emit('click', null)
+      if (event.target.getAttribute('role') === 'button' || event.target.className === 'prompt') {
+        this.hide()
+      }
+    },
+    focusItem: function (value) {
+      let index = value
+      if (index < 0) {
+        index = promptButtons.length
```
ChunkyProgrammer, line 63:
```suggestion
        index = this.promptButtons.length
```

PikachuEXE (approved):

src/renderer/helpers/accessibility.js
```diff
@@ -0,0 +1,45 @@
...
+    nextElement.setAttribute('tabindex', '0')
+    nextElement.focus()
+  }
+
+  return event.key === 'Enter' || event.key === ' '
+}
+
+export function sanitizeForHtmlId(attribute) {
+  // html doesn't like spaces in ids, queryselector/css wants symbols encoded
+  return attribute.replace(/[^A-Za-z0-9]*/g, '')
```
absidue, line 44: Does this return an empty string for languages that don't use those characters? e.g. Logographic systems?

ChunkyProgrammer: Good point, all ascii characters are valid for html ids (besides whitespace) so  we could include them but the issue from that would be with querySelector requiring some escaping. For example, `const allOptions = document.querySelector(`#${unspacedTitle} + ul`)` wouldn't work if it had those characters in it without some sort of escaping.

absidue: using `CSS.escape` might work

PikachuEXE (dismissed):
FtDropdown re-tested with zh-tw

PikachuEXE (approved):

src/renderer/components/ft-select/ft-select.vue
```diff
@@ -1,9 +1,11 @@
 <template>
   <div class="select">
     <select
+      :id="sanitizeForHtmlId(placeholder)"
```
absidue, line 4: you could make this a computer property so the function only has to run once instead of 3 times

src/renderer/components/ft-prompt/ft-prompt.vue
```diff
@@ -1,11 +1,19 @@
...
   <div
     class="prompt"
+    tabindex="-1"
     @click="handleHide"
+    @keydown.enter="handleHide"
   >
-    <ft-card class="promptCard">
+    <ft-card
+      class="promptCard"
+      :aria-labelledby="'dialog-' + sanitizeForHtmlId(label)"
```
absidue, line 10: you could make `sanitizeForHtmlId(label)` a computed property as it's used 4 times in this component

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2710, opened 2022-10-11
Move colours from the store to the utils helper

absidue:
# Move colours from the store to the utils helper

## Pull Request Type

- [x] Refactor

## Description
This pull request move the colorNames and colorValues arrays out of the store, as they don't change at runtime. I also moved the getRandomColor and getRandomColorClass functions out of the store, as they just need access to the colorNames and colorValues arrays.

## Testing 
- Colours are listed in the theming and sponsorblock settings
- The colourful cirlces are shown when editing a profile
- A random colour is auto selected when you create a new profile

Very unlikely that you'll be able to find a livestream with superchats
- Superchats get a random colour as the background.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -11,7 +11,7 @@ import { MAIN_PROFILE_ID } from '../../../constants'
 import fs from 'fs'
 import { opmlToJSON } from 'opml-to-json'
 import ytch from 'yt-channel-info'
-import { calculateColorLuminance } from '../../helpers/utils'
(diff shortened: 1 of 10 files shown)


ChunkyProgrammer:
I think to make the code more maintainable, we should probably have a list of color objects that have a "colorName" and a "colorValue" value

absidue:
Switched it over to a single array now.

ChunkyProgrammer (approved):

PikachuEXE (approved):
Tested locally except Superchats

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2712, opened 2022-10-12
Make the save as and open file dialogs blocking

absidue:
# Make the save as and open file dialogs blocking

## Pull Request Type

- [x] Bugfix

## Description
Currently the Save as... and Open file... dialogs are non-blocking, which means you can switch back to the FreeTube window and open as many of those dialogs as you want or navigate to a different page, causing unexpected behaviour when you select a file.
As we aren't Discord, I decided to fix that bug, so that when one of those dialogs is open, you have to deal with it before you can return to FreeTube.

## Testing 
The easiest place to test this is in the data import and export settings, by opening on of those dialogs and then attempting to click into the FreeTube window. On Windows the dialog window will flash for a short moment. It should stop you from going back to the FreeTube window until you cancel or select a file to open/save to.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -433,21 +433,27 @@ function runApp() {
     return app.getPath('pictures')
   })
 
-  ipcMain.handle(IpcChannels.SHOW_OPEN_DIALOG, async (_, options) => {
+  ipcMain.handle(IpcChannels.SHOW_OPEN_DIALOG, async ({ sender }, options) => {
+    const senderWindow = findSenderWindow(sender)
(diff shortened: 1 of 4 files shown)


PikachuEXE (approved):
Tested on MacOS & compared to current unfixed version

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2714, opened 2022-10-12
Filtering out invidious instances that don't support the API

MarmadileManteater:
# Filtering out invidious instances that don't support the API

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Currently, when filtering invidious instances, there isn't any check for if an invidious instance supports API calls. The result is that, occasionally, the selected invidious instance is one that doesn't support the API whatsoever.  For instance, if the selected instance is set to `https://yewtu.be/` or `https://invidious.snopyta.org/`, all invidious API calls will fail, and that won't change until the user selects another instance or the app is restarted.

This PR aims to address this issue by adding two extra conditions to the invidious instances filter:
- Does this instance support the API?
- Does this instance support CORS? (This is only checked when `IS_ELECTRON` is `false`.)

## Screenshots 
_Before / After:_
 

 - OS: Windows 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/renderer/store/modules/invidious.js
+++ b/src/renderer/store/modules/invidious.js
@@ -24,7 +24,7 @@ const actions = {
       const response = await fetch(requestUrl)
       const json = await response.json()
       instances = json.filter((instance) => {
-        if (instance[0].includes('.onion') || instance[0].includes('.i2p')) {
+        if (instance[0].includes('.onion') || instance[0].includes('.i2p') || !instance[1].api || (!process.env.IS_ELECTRON && !instance[1].cors)) {
           return false
         } else {
           return true


ChunkyProgrammer:
Might be a good idea to update https://github.com/FreeTubeApp/FreeTube/blob/development/static/invidious-instances.json as well.

absidue:
These two hardcoded fallback ones probably need changing then as well: https://github.com/FreeTubeApp/FreeTube/pull/2714/files#diff-afab392b004dec9b55a80e6f19b78dff91e7d211e2b2ae5fd54bcdbc08c36d25R52-R56

ChunkyProgrammer (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #2487, opened 2022-08-16
[Bug]: Share Playlist drop down on mobile layout causes scrolling on hover


ChunkyProgrammer:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a bug report that matches the one I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.

### Describe the bug

1. Go to any playlist (ex: https://youtube.com/playlist?list=PLBcZM94cb2UxtAqsfIWY1R4OrlkD323qd)
2. Change window size until mobile layout appears
3. Hover on "Share Playlist" dropdown
4. See issue

### Expected Behavior

You shouldn't have to scroll to copy links

### Issue Labels

visual bug

### FreeTube Version

v0.17.1 (latest nightly)

### Operating System Version

Windows 11

### Installation Method

.exe

### Primary API used

Local API

### Additional Information

Share menu (not hovering):
<img width="457" alt="image" src="https://user-images.githubusercontent.com/78101139/184874377-c994f067-333d-4e33-a748-8b021a089ed5.png">
Share menu (hovering)
<img width="460" alt="image" src="https://user-images.githubusercontent.com/78101139/184874419-7cd6400e-0df9-4739-8d87-fadf09623152.png">
Share menu (hovering, scrolled down)
<img width="452" alt="image" src="https://user-images.githubusercontent.com/78101139/184874507-5671a017-9a4b-470d-8792-1303bb33a7c9.png">

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Issue #1865, opened 2021-10-30
Fullscreen button and quality selection don't fit on mobile screen


pkrasicki:
**Behavior of the program**
On mobile the fullscreen button and quality selection are not shown in portrait mode. In landscape mode quality selection is visible, but fullscreen button is only partially visible.

**Expected behavior**
Fullscreen button and quality selection should be visible on mobile screen.

**To Reproduce**
1. Use a phone or set your window dimensions to 360x650px on desktop.
2. Go to any video page.

**Screenshots**
![freetube-fullscreen-mobile](https://user-images.githubusercontent.com/49721209/139536216-2b0a9dbb-d1d2-4911-9261-11dd0a09bc93.png)

![freetube-fullscreen-mobile2](https://user-images.githubusercontent.com/49721209/139536242-50f51331-a3b1-4ca9-ba0e-d4d85a5e619b.png)

**Environment Information (please complete the following information):**
 - OS: Mobian
 - OS Version: Bookworm
 - FreeTube version: 0.15.0
 - Installation Method (If applicable): .deb arm64 package
 - Primary API Used: Invidious API

**Additional context**
I can always double tap to toggle fullscreen, so the app is still usable.

absidue:
Completed in #2688

Issue #337, opened 2019-08-14
Discussion | GNU/Linux Mobile


ggg27:
**Description**: Supporting GNU/Linux for Mobile Devices
(i.e. [Ubuntu Touch](https://www.wikipedia.org/wiki/Ubuntu_Touch), [Purism tablet/phone](https://www.wikipedia.org/wiki/Librem#Librem_11_tablet), [PinePhone](https://www.wikipedia.org/wiki/Pine_Microsystems#Smartphone), [PineTab](https://www.pine64.org/pinetab/), e.c.t)
**How?**: In theory most of it should still work, the key piece is verifying the UI works on a much smaller touch screen (i.e [adaptive UI](https://www.wikipedia.org/wiki/Adaptive_user_interface)).

__________________

Want to support this issue? Post a [bounty on it](https://www.bountysource.com/issues/78761559-discussion-gnu-linux-mobile)! We accept bounties via [BountySource](https://www.wikipedia.org/wiki/Bountysource).

PrestonN:
This is something I've been thinking about for quite a while now.  Personally I'd love to see it happen.

At this moment, FreeTube already provides ARM builds which should theoretically work on the Librem 5 as-is with no further modifications.  Having said that, there's a good chance that it may not be a great experience with how FreeTube is at it's current state.

Most of the UI should resize properly to the small display size, though there are definitely some tweaks to be make in order for it to be a better experience.  It also wouldn't incorporate any sort of touch gestures (such as sliding from the side to bring up the side menu).

There's also the potential problem on performance.  With FreeTube being an Electron app, I'm not sure how well performance will be on a device like the Librem 5, especially when resources and battery life are at a premium.

I will likely be purchasing a Librem 5 whenever it comes out, so I will be able to do some actual tests with it.  I will attempt to get a VM up and running with their OS to get an idea on how well it runs though.

So yes, it should work, but there will likely be a lot of changes that need to be made in order for it to work well on the device.  I'm prepared to make it work as well as I can when the device is released.

PrestonN:
I'll have more news about this in the future, though I have successfully been able to get FreeTube running on a Pinephone.  I don't have too much more to say on it though as there are still quite a few issues with it and won't be a priority for now, but some progress has been made at least.

okias:
If needed, I can give a shot on Librem 5 devkit (== final phone at least for screen & handling).

PrestonN:
[...]

PR #2719, opened 2022-10-13
Improving touch input by bringing in `videojs-mobile-ui`

MarmadileManteater:
# Improving touch input by bringing in `videojs-mobile-ui`

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
MarmadileManteater#56

## Description
This PR aims to improve the touch input on the video player by adding in the [videojs-mobile-ui](https://github.com/mister-ben/videojs-mobile-ui) plugin. This plugin adds the ability to double tap on the left or right side of the video in order to rewind or fast forward. It also adds in its' own `pause / play` button which responds more naturally to touch input, and it puts the window into fullscreen mode when it detects a rotation from portait to landscape. 

## Testing 
This can be tested in the web build by using the touch simulation available in Chrome and Firefox.
Theoretically, this implementation should also work in Electron, but the touch simulation doesn't seem to work in Electron (afaik), so I have no idea how to test this in Electron without a touch screen device that can run Electron apps.

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional context
I added `v0.8.0` of `videojs-mobile-ui`. This is technically the beta version. My reason for not using the latest `v0.7.0` is because it contains a bug. `v0.7.0` references `videojs` as a dependency, instead of a dev-dependency. The result is that `v0.7.0` loads a second version of `videojs`  which is not the one FreeTube uses, and this causes the plugin to fail to load. This bug is fixed in `v0.8.0`.

The `videojs-mobile-ui` plugin, by default, only works if videojs detects it is running in Android or iOS. This is why I enabled the `forceForTesting` flag. Instead of letting `videojs-mobile-ui` try to detect if touch input should be used, I added logic to show the touch input when a touch event occurs and to hide the touch input as soon as a mouseover event occurs. The result is that the touch controls should be able to load on any device that has a touch screen even if it isn't Android or iOS, and they won't load show up at all if the user doesn't use any touch input.

A lot of the options available in the `.mobileUi` method seem like good candidates for new settings: `fullscreen.enterOnRotate`, `fullscreen.exitOnRotate`,  and `fullscreen.lockOnRotate`. I could understand wanting them enabled on some devices and disabled on others.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -111,6 +111,7 @@ export default Vue.extend({
       statsModal: null,
       showStatsModal: false,
       statsModalEventName: 'updateStats',
+      usingTouch: false,
       dataSetup: {
         fluid: true,
         nativeTextTracks: false,
@@ -484,6 +485,7 @@ export default Vue.extend({
             this.transformAndInsertCaptions()
           }
           this.toggleScreenshotButton()
+          this.usingTouch = this.$refs.video.parentNode.querySelector('.vjs-touch-overlay') !== null
         })
 
         this.player.on('ended', () => {
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -11,7 +11,8 @@ import 'videojs-overlay/dist/videojs-overlay.css'
 import 'videojs-vtt-thumbnails-freetube'
 import 'videojs-contrib-quality-levels'
 import 'videojs-http-source-selector'
-
+import 'videojs-mobile-ui'
+import 'videojs-mobile-ui/dist/videojs-mobile-ui.css'
 import { IpcChannels } from '../../../constants'
 import { sponsorBlockSkipSegments } from '../../helpers/sponsorblock'
 import { calculateColorLuminance } from '../../helpers/utils'
@@ -362,6 +363,21 @@ export default Vue.extend({
               allowSeeksWithinUnsafeLiveWindow: true,
               handlePartialData: true
             }
+          }
+        })
+        // TODO : map actual settings to this object
+        this.player.mobileUi({
+          fullscreen: {
+            enterOnRotate: true,
+            exitOnRotate: true,
+            lockOnRotate: true,
+            iOS: false
+          },
+          touchControls: {
+            seekSeconds: 10,
+            tapTimeout: 300,
+            disableOnEnd: false,
+            disabled: false
           }
         })
 
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -370,7 +370,7 @@ export default Vue.extend({
             enterOnRotate: true,
             exitOnRotate: true,
             lockOnRotate: true,
-            forceForTesting: this.hasUserTouched
+            forceForTesting: true
           },
           touchControls: {
             seekSeconds: this.defaultSkipInterval,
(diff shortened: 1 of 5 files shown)


ChunkyProgrammer (approved):

absidue:
We've got an option to hide the play button, should this PR respect that settings?

MarmadileManteater:
I believe it should respect that setting because it acts as a replacement for the existing play button on mobile.

src/renderer/components/ft-video-player/ft-video-player.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="relative">
+  <div
+    :class="wrapperClass"
```
absidue, line 3:
```suggestion
    class="relative"
    :class="{
      'vjs-using-touch': usingTouch,
      'vjs-hide-play-button': !displayVideoPlayButton
    }"
```

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -291,6 +292,10 @@ export default Vue.extend({
 
     screenshotFolder: function() {
       return this.$store.getters.getScreenshotFolderPath
+    },
+
+    wrapperClass: function () {
+      return `relative ${this.usingTouch ? 'vjs-using-touch' : ''} ${!this.displayVideoPlayButton ? 'vjs-hide-play-button' : ''}`
```
absidue, line 298:
```suggestion
```

absidue (commented):
Instead of having a long string with the classes, you can let Vue programatically add and remove individual classes when the properties change. You'll probably want to batch both of these suggestions together into a single commit, which you can do from the "Files changed" tab.

absidue (dismissed):
Thanks

ChunkyProgrammer (dismissed):

src/renderer/videoJS.css
```diff
@@ -2201,4 +2201,37 @@ video::-webkit-media-text-track-display {
...
+}
+
+.vjs-hide-play-button .video-js .vjs-touch-overlay .vjs-play-control .vjs-icon-placeholder:before {
+  content: " " !important;
+}
+
+.video-js .vjs-touch-overlay.skip {
+  -webkit-filter: drop-shadow(0px 0px 4px black);
+  filter: drop-shadow(0px 0px 4px black);
 }
```
PikachuEXE, line 2237: Can we have final newline like what's set inside [`.editorconfig`](https://github.com/FreeTubeApp/FreeTube/blob/development/.editorconfig#L9)?

PikachuEXE (changes requested):
Also I have no idea how to test web build ._.

absidue:
> Also I have no idea how to test web build ._.

Build the web build with `yarn pack:web` then host it on localhost with something like `node-static` https://github.com/cloudhead/node-static#command-line-interface

ChunkyProgrammer (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Cant test rn but maybe this closes one of these issues https://github.com/FreeTubeApp/FreeTube/issues/337, https://github.com/FreeTubeApp/FreeTube/issues/1865, https://github.com/FreeTubeApp/FreeTube/issues/2487

absidue:
#337: Don't think so, the main complaints there seem to be about RAM and CPU usage (linux phones have very limited hardware as I understand it), which we might be able to do something about it but probably not too much as it is a chromium browser after all.
#1865: looks like that was fixed in #2688, so it can probably be closed as completed now
#2487: nope this PR is for the video player, that issue is about the playlist page

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Powers granted

efb4f5ff-1298-471a-8973-3d47447115dc:
Could somebody dismiss pikachu's review. I cant on mobile

PR #2726, opened 2022-10-15
Replace rss-parser with native the fetch and DomParser

absidue:
# Replace rss-parser with native the fetch and DomParser

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description
This pull request replaces the rss-parser dependency with the fetch function and the native DomParser. This change has quite a few advantages over the old implementation with rss-parser:
* Using native functionality over rss-parser with it's XML parsing dependencies, shaves 120KB off both the Electron renderer and web builds
* Fetching the feeds now respects the user's proxy settings, as rss-parser uses node's http and https libraries directly, bypassing the proxy settings
* Fetching the RSS subscriptions used to timeout while using a VPN (see screenshot below), which doesn't happen with the new implementation
* It feels faster, especially on a VPN as you don't need to wait for the timeouts to happen (24 seconds vs 2 seconds with my subscriptions list and VPN, might differ for you)
* Video views show in the subscriptions now (would probably have been possible to implement with rss-parser too with extra configuration)

## Screenshots 

Timing out with VPN enabled:
errors_with_vpn

Size before:
before

Size after:
after

## Testing 
**Subscriptions:**
1. Enable the "Fetch Feeds from RSS" setting in the subscription settings
2. Fetch your subscriptions

**Blog posts:**
1. Enable the "Check for latest blog posts" setting in the general settings
2. In the devtools Application tab delete the `lastAppWasRunning` key and value from the local storage, if it's there for you.
3. Close and reopen FreeTube
4. Wait for the banner to show up, it might take a while, the network request to write.as seems to be quite slow

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (approved):
Tested both

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2737, opened 2022-10-17
Fix for `Region for Trending` dropdown not populating

MarmadileManteater:
# Fix for `Region for Trending` dropdown not populating

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
FreetubeApp#1644

## Description
This PR aims to fix an issue with the `Region for Trending` dropdown not populating with data before the user changes the `currentLocale`.

## Screenshots 

_Before:_

_After:_

## Testing 
1. Delete your `settings.db` file _(I believe it is located in the `Electron` roaming directory when debugging)_
2. Start FreeTube
3. Check the `Region for Trending` dropdown in settings

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional Information
An alternative solution might be to add an additional call to `getRegionData` somewhere; however, it may be difficult to tell if `getRegionData` has already been called recently without introducing a race condition. That could cause `getRegionData` to be called twice in a row. The reason I went with this solution was because this issue goes away when all of the side effects are evaluated on app mounted.

### DIFF
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -387,9 +387,14 @@ Object.assign(customGetters, {
 const customActions = {
   grabUserSettings: async ({ commit, dispatch, getters }) => {
     try {
-      const userSettings = await DBSettingHandlers.find()
+      // Assigning default settings for settings that have side effects
+      const userSettings = Object.entries(Object.assign({},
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):

PikachuEXE (approved):
Tested locally

absidue (approved):

PR #2743, opened 2022-10-18
Implement `getRegionData` in web

MarmadileManteater:
# Implement `getRegionData` in web

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
The `Region for Trending` dropdown doesn't currently populate on web builds, so this PR adds logic to `getRegionData` in order to correct this.

## Screenshots 
_Before / After:_
 

## Testing 
1. `yarn dev:web`
2. Navigate to general settings
3. Click on the `Region for Trending` dropdown to see if it is populated with data

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional context
This PR also replaces the currently provided `__dirname` in web builds with one which is more accurate. Previously, it was referencing the location of the currently executing file in the source (ex: `src/renderer/components/general-settings` etc.), and now, it should be the path that the web app is currently located on; if the app is hosted at `/freetube/`, the `__dirname` would be `/freetube`.

### DIFF
--- a/_scripts/webpack.web.config.js
+++ b/_scripts/webpack.web.config.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const fs = require('fs')
 const webpack = require('webpack')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const VueLoaderPlugin = require('vue-loader/lib/plugin')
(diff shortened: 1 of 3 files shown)


_scripts/webpack.web.config.js
```diff
@@ -108,13 +109,14 @@ const config = {
...
-    __dirname: true,
+    __dirname: false,
     __filename: isDevMode,
   },
   plugins: [
     new webpack.DefinePlugin({
       'process.env.PRODUCT_NAME': JSON.stringify(productName),
-      'process.env.IS_ELECTRON': false
+      'process.env.IS_ELECTRON': false,
+      '__dirname': 'window.location.pathname.endsWith("/")?window.location.pathname.substring(0, window.location.pathname.length - 1):window.location.pathname'
```
absidue, line 119: I would prefer that you create a helper function (it can go straight into helpers/utils.js instead of store/utils.js, that way I don't have to move it over in a later PR) for this, based on https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/i18n/index.js#L57-L67. That makes the code much more maintainable and less cryptic, personally I wouldn't look at the `__dirname` variable and assume that it is being reassinged in a webpack config. It will also handle `index.html` being in the URL path, which this implementation doesn't at the moment.

The function could be named something along the lines of createWebURL(path).

PikachuEXE (approved):

ChunkyProgrammer (approved):

absidue (approved):

PR #2749, opened 2022-10-19
Fix double focus on the sidebar channels

absidue:
# Fix double focus on the sidebar channels

## Pull Request Type

- [x] Bugfix

## Description
Currently there are two focusable and clickable elements which could be confusing for people tabbing through them. This pull request removes the div with the event handlers, instead just relying on the router-link element.

You might want to make the github diff not show whitespace differences while reviewing this PR.

## Screenshots 
**Before with the sidebar compacted:**
before-collapsed-outer before-collapsed-inner

**Before with the sidebar expanded:**
before-expanded-outer before-expanded-inner

**After:**
after-collapsed after-expanded

## Testing 
Clicking on the channels
Tabbing through them
Selecting them by pressing enter

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/side-nav/side-nav.css
+++ b/src/renderer/components/side-nav/side-nav.css
@@ -69,7 +69,6 @@
 .navOption .navLabel {
   margin-left: 40px;
   overflow: hidden;
-  margin-left: 40px;
   word-break: break-word;
 }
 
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Tested in cloud😁

PR #2750, opened 2022-10-19
Switch workflows from npm to yarn

absidue:
# Switch workflows from npm to yarn

## Pull Request Type
- [x] Feature Implementation

## Description
This pull request replaces the remaining uses of npm in the workflows, with yarn, which is quicker. I also changed the package.json file to use yarn, I had to add the `--silent` argument to prevent the inner yarn command from outputting it's duration (could be confusing if you don't know why there are two durations). Got rid of the `rebuild:node` target as that isn't used anymore and doesn't work.

## Screenshots 
before and after:
yarn-install

## Testing 
All of the commands that i changed work locally, so they should work on GitHub actions too.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,12 +38,11 @@
     "pack:renderer": "webpack --mode=production --node-env=production --config _scripts/webpack.renderer.config.js",
     "pack:web": "webpack --mode=production --node-env=production --config _scripts/webpack.web.config.js",
     "pack:workers": "webpack --mode=production --node-env=production --config _scripts/webpack.workers.config.js",
-    "postinstall": "npm run rebuild:electron",
+    "postinstall": "yarn run --silent rebuild:electron",
     "prettier": "prettier --write \"{src,_scripts}/**/*.{js,vue}\"",
     "rebuild:electron": "electron-builder install-app-deps",
-    "rebuild:node": "npm rebuild",
     "release": "run-s test build",
-    "ci": "yarn install --frozen-lockfile"
+    "ci": "yarn install --silent --frozen-lockfile"
   },
   "dependencies": {
     "@fortawesome/fontawesome-svg-core": "^6.2.0",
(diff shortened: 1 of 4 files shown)


ChunkyProgrammer (approved):

package.json
```diff
@@ -38,12 +38,11 @@
     "pack:renderer": "webpack --mode=production --node-env=production --config _scripts/webpack.renderer.config.js",
     "pack:web": "webpack --mode=production --node-env=production --config _scripts/webpack.web.config.js",
     "pack:workers": "webpack --mode=production --node-env=production --config _scripts/webpack.workers.config.js",
-    "postinstall": "npm run rebuild:electron",
+    "postinstall": "yarn run --silent rebuild:electron",
```
PikachuEXE, line 41: Why `--silent` is needed? (I have never used that flag)

package.json
```diff
@@ -38,12 +38,11 @@
...
     "prettier": "prettier --write \"{src,_scripts}/**/*.{js,vue}\"",
     "rebuild:electron": "electron-builder install-app-deps",
-    "rebuild:node": "npm rebuild",
     "release": "run-s test build",
-    "ci": "yarn install --frozen-lockfile"
+    "ci": "yarn install --silent --frozen-lockfile"
```
PikachuEXE, line 45: Why `--silent` is needed? (I have never used that flag)

PikachuEXE (commented):
Only question is on `--silent`
Otherwise good to go

absidue:
Without `--silent` the inner yarn command, the one in the postinstall script, outputs it's time as well.

without-silent

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #2752, opened 2022-10-20
Move `Dialog` funtions to `utils/helpers`

MarmadileManteater:
# Move `Dialog` funtions to `utils/helpers`

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [X] Other

## Description
The functions pertaining to file pickers don't touch the state.

```javascript
  // line 346 of src/store/modules/utils.js
  async showSaveDialog (context, options) {
    const webCbk = async () => {
         // ...
    }
    // These functions technically pass the context into this function.
    return await invokeIRC(context, IpcChannels.SHOW_SAVE_DIALOG, webCbk, options)
  }
```
```javascript
// line 107 of src/store/modules/utils.js
async function invokeIRC(context, IRCtype, webCbk, payload = null) {
  // However, context is completely unused here.
  let response = null
  if (process.env.IS_ELECTRON) {
    const { ipcRenderer } = require('electron')
    response = await ipcRenderer.invoke(IRCtype, payload)
  } else if (webCbk) {
    response = await webCbk()
  }

  return response
}
```
I believe it makes sense to move them out of `src/store/modules/utils.js` and into `src/helpers/utils.js`.

 - OS: Windows 10
 - FreeTube version: 0.17.1

## Additional context
If there is a scenario which could occur in the future which would involve the `invokeIRC` needing to modify state; it might make  sense to only move the `writeFileFromDialog` and `readFileFromDialog` functions.

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -10,7 +10,7 @@ import { MAIN_PROFILE_ID } from '../../../constants'
 
 import { opmlToJSON } from 'opml-to-json'
 import ytch from 'yt-channel-info'
-import { calculateColorLuminance, copyToClipboard, getRandomColor, showToast } from '../../helpers/utils'
+import { calculateColorLuminance, copyToClipboard, getRandomColor, readFileFromDialog, showToast, writeFileFromDialog } from '../../helpers/utils'
 
 export default Vue.extend({
   name: 'DataSettings',
@@ -103,7 +103,7 @@ export default Vue.extend({
       }
       let textDecode
       try {
-        textDecode = await this.readFileFromDialog({ response })
+        textDecode = await readFileFromDialog(response)
       } catch (err) {
         const message = this.$t('Settings.Data Settings.Unable to read file')
         showToast(`${message}: ${err}`)
@@ -509,7 +509,7 @@ export default Vue.extend({
         return
       }
       try {
-        await this.writeFileFromDialog({ response, content: subscriptionsDb })
+        await writeFileFromDialog(response, subscriptionsDb)
       } catch (writeErr) {
         const message = this.$t('Settings.Data Settings.Unable to read file')
         showToast(`${message}: ${writeErr}`)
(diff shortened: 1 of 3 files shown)


src/renderer/helpers/utils.js
```diff
@@ -196,3 +197,130 @@ export function openExternalLink(url) {
...
+    const { ipcRenderer } = require('electron')
+    return await ipcRenderer.invoke(IpcChannels.SHOW_SAVE_DIALOG, options)
+  } else {
+    // If the native filesystem api is available
+    if ('showSaveFilePicker' in window) {
+      return {
+        canceled: false,
+        handle: await window.showSaveFilePicker({
+          suggestedName: options.defaultPath.split('/').at(-1),
+          types: options?.filters[0]?.extensions?.map((extension) => {
```
absidue, line 277: As options is used on the line above without any checks, that means that it will always have a value, so one of the optional chaining operators can be removed here.
```suggestion
          types: options.filters[0]?.extensions?.map((extension) => {
```

absidue (commented):
`showSaveDialog` is also used in the ft-player component, please make the changes in there too.

absidue (dismissed):

ChunkyProgrammer (approved):

src/renderer/components/data-settings/data-settings.js
```diff
@@ -10,7 +10,7 @@ import { MAIN_PROFILE_ID } from '../../../constants'
 
 import { opmlToJSON } from 'opml-to-json'
 import ytch from 'yt-channel-info'
-import { calculateColorLuminance, copyToClipboard, getRandomColor, showToast } from '../../helpers/utils'
+import { calculateColorLuminance, copyToClipboard, getRandomColor, readFileFromDialog, showOpenDialog, showSaveDialog, showToast, writeFileFromDialog } from '../../helpers/utils'
```
PikachuEXE, line 13: Longggg
```suggestion
import {
  calculateColorLuminance,
  copyToClipboard,
  getRandomColor,
  readFileFromDialog,
  showOpenDialog,
  showSaveDialog,
  showToast,
  writeFileFromDialog
} from '../../helpers/utils'
```

PikachuEXE (changes requested):

ChunkyProgrammer (approved):

PikachuEXE (approved):

absidue (approved):

PR #2769, opened 2022-10-25
Use native DomParser instead of opml-to-json dependency

absidue:
# Use native DomParser instead of opml-to-json dependency

## Pull Request Type

- [x] Feature Implementation

## Description
This pull request replaces the opml-to-json dependency with the native DomParser. I did a similar thing for parsing the RSS feeds, which went down well, so I thought I would do it for the OPML import too.

## Screenshots 
before:
before

after:
after

## Testing 

1. `yarn dev`
2. Download the 3 test files (Just a list of all the LTT channels)
3. Before importing each of the the files, click "Remove All Subscriptions / Profiles" in the privacy settings.
4. Import the subscription files, the broken one is intended to fail, this is to test that the error handling works.

[opml-testfiles.zip](https://github.com/FreeTubeApp/FreeTube/files/9863346/opml-testfiles.zip)

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (dismissed):
Tested locally

src/renderer/components/data-settings/data-settings.js
```diff
@@ -335,79 +334,80 @@ export default Vue.extend({
...
-          })
-          if (subExists === -1) {
-            let channelInfo
-            if (this.backendPreference === 'invidious') {
-              channelInfo = await this.getChannelInfoInvidious(channelId)
-            } else {
-              channelInfo = await this.getChannelInfoLocal(channelId)
-            }
+      feedData.forEach(async (channel) => {
+        const channelId = channel.getAttribute('xmlUrl').replace('https://www.youtube.com/feeds/videos.xml?channel_id=', '')
```
ChunkyProgrammer, line 371: Invidious can also export the subscriptions like this: `https://yewtu.be/feed/channel/{CHANNELID}` (for yewtu.be as the instance) so it might be worth doing a replace for everything before 'channel/' as well to support the invidious one. This could be done in a different PR if you'd prefer

absidue: This should be fixed now.

absidue:
@ChunkyProgrammer You should now be able to import your subscriptions. XML is very strict but as HTML isn't, I've made it fallback to the HTML parser. That will allow you to import subscriptions that are invalid XML but still valid HTML.

PikachuEXE (dismissed):

src/renderer/components/data-settings/data-settings.js
```diff
@@ -335,79 +334,98 @@ export default Vue.extend({
...
-        if (typeof feedData === 'undefined') {
-          if (json.title.includes('gPodder')) {
-            feedData = json.children
-          } else {
-            const message = this.$t('Settings.Data Settings.Invalid subscriptions file')
-            showToast(message)
-            return
-          }
+          xmlDom = htmlDom
+        } catch {
```
ChunkyProgrammer, line 353: Would you need to throw the errornode again to get to this catch block?

absidue: Not sure, the docs don't mention what happens if it fails to parse HTML https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString#error_handling

The HTML parser is too lenient, for the broken file it produced this:
broken_file

Maybe I should check the error message and only fall back to the HTML parser if the HTML entity error occurs?

absidue: Okay, only doing it for a specific error won't work, as Firefox always outputs the same error if the XML is problematic. So checking the error, wouldn't work for the web build.

ChunkyProgrammer: Might be good to log the error with the xml and then try to do it with the html dom parser

absidue: It'll log the error now before trying to parse it with the HTML parser.

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2780, opened 2022-10-28
Centralize map used for parsing invidious comments

MarmadileManteater:
# Centralize map used for parsing invidious comments

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
When I added `toLocalePublicationString` to the invidious comments, I missed that this function should also probably be called on comment replies; however, this brought my attention to the fact that there are two maps which do virtually the same thing which could be simplified into one `parseInvidiousCommentData` function similar to how the local API comments work.

This PR aims to fix the locale date string inconsistency between replies and top level comments that come from the Invidious API.

## Screenshots 
_Before:_
before
_After:_
after

## Testing 
1. `yarn dev`
2. Make sure your locale preference is set to something that is not English
3. Check a comment reply to see if the date string is in your selected locale

## Desktop
 - OS: Windows 10
 - FreeTube version: 0.17.1

### DIFF
--- a/src/renderer/components/watch-video-comments/watch-video-comments.js
+++ b/src/renderer/components/watch-video-comments/watch-video-comments.js
@@ -240,6 +240,37 @@ export default Vue.extend({
       }
     },
 
+    parseInvidiousCommentData: function (response) {
+      return response.comments.map((comment) => {
+        comment.showReplies = false
+        comment.authorLink = comment.authorId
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
Tested locally
image

However the provided test steps do not include changing preferred API to invidious?

MarmadileManteater:
I forgot to mention that, but that is important; Thank you for catching that. I just updated the test steps above to add the missing step.

absidue (approved):

ChunkyProgrammer (approved):

PR #2812, opened 2022-11-03
exempt some labels from stale workflow

ChunkyProgrammer:
# exempt some labels from stale workflow

## Pull Request Type
- [x] Other - workflow

## Description
I don't feel that the stale workflow should apply to feature requests or PRs that are waiting for review

## Testing 
Tested in the FreeTube test repo

## Desktop

- **OS:** Windows
- **OS Version:** 11

## Additional context
Let me know if you agree/disagree with the changes 😃

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -19,3 +19,5 @@ jobs:
           days-before-pr-close: 14
           stale-issue-label: 'U: stale'
           stale-pr-label: 'PR: stale'
+          exempt-pr-labels: 'PR: waiting for review'
+          exempt-issue-labels: 'enhancement'


efb4f5ff-1298-471a-8973-3d47447115dc:
I agree because it is logical to review a PR when we have time for it but i also disagree because if a PR with that label gets stale after a month it means that we are slacking and should really review it. 

I disagree with the issue because i already found many duplicate issues thanks to the stale bot saying, hi this is stale. If something like that is stale i often investigate if its a valid request and if this could be a duplicate of another issue

PikachuEXE:
Agree with @efb4f5ff-1298-471a-8973-3d47447115dc 
The bot also agrees /s

efb4f5ff-1298-471a-8973-3d47447115dc:
If needed i will summon all the bots for this battle

ChunkyProgrammer:
> I agree because it is logical to review a PR when we have time for it but i also disagree because if a PR with that label gets stale after a month it means that we are slacking and should really review it.
> 
> I disagree with the issue because i already found many duplicate issues thanks to the stale bot saying, hi this is stale. If something like that is stale i often investigate if its a valid request and if this could be a duplicate of another issue

Do you think we should have a seperate label we can assign to issues to prevent it from going stale?

PikachuEXE:
That would be useful

efb4f5ff-1298-471a-8973-3d47447115dc:
> Do you think we should have a seperate label we can assign to issues to prevent it from going stale?

I think that is a good idea. Maybe we can also use a label for our own PR's that have been laying around for a while?

PR #2835, opened 2022-11-06
Only show Select All context menu item in text fields

absidue:
# Only show Select All context menu item in text fields

## Pull Request Type

- [x] Feature Implementation

## Description
Currently the Select All context menu item is shown all the time. This pull request makes it so that the option is only shown for text fields. If the text field is empty the Select All item will be greyed out and enabled if the text field contains text.

## Screenshots 
Before:
before_1
before_2

After:
after_1
after_2 after_3

## Testing 
Invoke the context menu in random places around FreeTube
Invoke the context menu on an empty text field
Invoke the context menu on a non empty text field

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -21,6 +21,7 @@ function runApp() {
     showSearchWithGoogle: false,
     showSaveImageAs: true,
     showCopyImageAddress: true,
+    showSelectAll: false,
     prepend: (defaultActions, parameters, browserWindow) => [
       {
         label: 'Show / Hide Video Statistics',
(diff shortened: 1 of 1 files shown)


PikachuEXE:
I don't have "select all" on MacOS ._.
image

absidue:
@PikachuEXE is that before or after this PR? electron-context-menu's default Select All option is hidden on macOS, but the implementation in this PR should show up.
https://github.com/sindresorhus/electron-context-menu/blob/main/index.js#L241

absidue:
Should I hide it on macOS like the default one? Do other macOS apps have a Select All context menu action?

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2844, opened 2022-11-08
Move colours and the colour functions to their own file

absidue:
# Move colours and the colour functions to their own file

## Pull Request Type

- [x] Feature Implementation

## Description
This PR makes the utils helper file less crowded by moving the colours and the colour functions to their own helper file. It doesn't change any functionality, all it does is move that code and change the imports.

## Testing 
`yarn dev` would complain if the imports were wrong and as there is no actual functional change that should be fine

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (approved):
Changed theme to different colors a few times

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2845, opened 2022-11-08
Reduce duplicate code by introducing a function to create a proxy agent

absidue:
# Reduce duplicate code by introducing a function to create a proxy agent

## Pull Request Type
- [x] Feature Implementation

## Description
Currently the code to create the proxy agents is copied inside all 4 functions in the ytdl store module. As that code is identical, this PR moves that code to a single function, that way it's easier to maintain, as you only need to make changes in a single place. 

Noticed this while upgrading the socks proxy agent.

## Testing 
Setup a proxy and use the local API to search, view playlists and watch videos.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/store/modules/ytdl.js
+++ b/src/renderer/store/modules/ytdl.js
@@ -15,6 +15,33 @@ const state = {
 
 const getters = {}
 
+function createProxyAgent(protocol, hostname, port) {
+  switch (protocol) {
+    case 'http':
+      return new HttpProxyAgent({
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
I have no proxy to test this...

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2846, opened 2022-11-08
Cleanup webpack aliases

absidue:
# Cleanup webpack aliases

## Pull Request Type

- [x] Feature Implementation

## Description
Most of the resolve aliases we have setup in the webpack configuration files go unused, this PR aims to get rid of all of the unused ones. The `image` alias was used in a single place, so that was easy to fix. I've kept the vue one around so that we still import the specific vue framework files that we want.

## Testing 
`yarn dev` and `yarn build` check that there are no build errors and then check that you don't see any missing images while FreeTube is running

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/webpack.main.config.js
+++ b/_scripts/webpack.main.config.js
@@ -34,8 +34,7 @@ const config = {
   },
   node: {
     __dirname: isDevMode,
-    __filename: isDevMode,
-    global: isDevMode,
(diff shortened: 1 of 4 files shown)


PikachuEXE (approved):
Tested locally

ChunkyProgrammer (approved):

PR #2849, opened 2022-11-09
Option to hide videos from certain channels

petaded:
# Option to hide videos from certain channels

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

partial implementation of #95

## Description
This PR adds a new setting under distraction free settings to hide specific channels.
You can write either the channel name or the channel id into the input box - multiple channels allowed via separating with semi-colon ;

Videos and playlists from this channel ( and the channel itself ) will no longer show up in a search, recommended or trending.
Note: the box is case sensitive and must be an exact match to the channel name.

## Screenshots 

The new setting ( see bottom right ):
image
Option box with multiple channels entered
image

Trending with empty box
image
Trending with "shane" in hide channel box
image
Tending after having "shane;A24" in hide channels box
image

## Testing 

Tested the following:
 - since this touched the livestream filtering code
   - Search "music live" with hide livestreams as false, validate live streams still show
   - Search "music live" with hide livestreams as true, validate live streams no longer show
 - Search "ltt" with empty Hide channels, ltt videos show
 - Search "ltt" with "Linus Tech Tips" in hide channel box, ltt videos hidden + channel also hidden
 - Search "ltt" with "Linus Tech Tips;LMG Clips" in hide channel box, ltt and lmg clips videos hidden
 - Search "ltt" with "UCXuqSBlHAE6Xw-yeJA0Tunw" in hide channel box ( ltt channel id), ltt videos hidden

## Desktop

- **OS:** Manjaro 
- **OS Version:** 22
- **FreeTube version:** 0.18.0

Issue #2755, opened 2022-10-21
[Feature Request]: Option to block Shorts and Upcoming videos


MaxGripe:
### Guidelines

- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for open and closed issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.

### Problem Description

It would be really great to have an option to hide "Shorts" and "Upcoming" YouTube videos from the Subscriptions panel.

In FireFox I can do that using some plugins, for example:
- BlockTube plugin has "Block Youtube Shorts videos" option
- uBlock Origin plugin has advanced filter that allows to remove Upcoming videos with the following code:

```
www.youtube.com##ytd-browse ytd-grid-video-renderer:has(ytd-thumbnail-overlay-time-status-renderer[overlay-style="UPCOMING"])

www.youtube.com##ytd-browse ytd-rich-item-renderer:has(ytd-thumbnail-overlay-time-status-renderer[overlay-style="UPCOMING"])

www.youtube.com##ytd-browse[page-subtype="subscriptions"] ytd-video-renderer ytd-thumbnail-overlay-time-status-renderer[overlay-style="UPCOMING"]:upward(ytd-item-section-renderer)
```

### Proposed Solution

Additional options in the "Settings" > "Subscription settings" section:

- Hide Youtube Shorts videos
- Hide Youtube Upcoming videos

### Alternatives Considered

1) the above options available in the Search Filters
2) new feature - advanced filters that allow custom filtering / scripts

### Issue Labels

new optional setting

efb4f5ff-1298-471a-8973-3d47447115dc:
Duplicate of #2008

MaxGripe:
> Duplicate of #2008

#2008 is considered duplicate of #1070 but that request is about title filtering only.
I don't understand how my request is a duplicate of that?

ExInferi:
This was one of my favourite features in Vanced, the ability to hide shorts in subscription feed. Upcoming videos don't bug me, but this feature isn't used often by the creators I'm subscribed to. I can see why this would be a wanted feature. 

> #2008 is considered duplicate of #1070 but that request is about title filtering only. I don't understand how my request is a duplicate of that?

I suppose your suggested alternatives are a duplicate?

Flashwalker:
The feed is flooded with shorts,
[...]

Issue #1070, opened 2021-02-27
Video Title Filter / Blacklist


Industrial:
Hi.

I would like to have a Filter or I guess the better word for it is Blacklist functionality for video titles. I want video's with titles partly matching any entry in this list to be hidden from search results.

The blacklist would work for Subscriptions, Trending, Most Popular and searches.

Is this possible?

PrestonN:
This is planned, however it's low priority at the moment. I'd be happy to accept a pull request if someone wanted to work on it ahead of me.

dbogdanov:
Another related feature request: hide videos from a configurable blacklist of YouTube channels.

Industrial:
I'm now using BlockTube for this: https://chrome.google.com/webstore/detail/blocktube/bbeaicapbccfllodepmimpkgecanonai

Rabcor:
I second this, I would also like the blacklist to affect the 'up next' lists, and the ability to blacklist specific videos and channels.

crisp5:
Based on the fact that #2008 and #2655 (both proposing to hide Shorts) were closed as duplicates of this issue, has the scope of this issue been expanded? The title implies filtering based on video titles, but the discussion has also considered filtering specific videos and channels, shorts, premieres, scheduled live streams.

MaxGripe:
I have requested option to block Shorts and Upcoming videos via #2755 and my request has been closed as duplicate of this issue. I'm requesting option to block Shorts and Upcoming specifically, not just based on title.

EvilGremlin:
+1, option to put all shorts to separate tab, but more than that option to simply hide them is sorely needed. Well, maybe show shorts on per channel basis. But in my feed literally every single short is trash i don't want to see.

BarbzYHOOL:
> Based on the fact that #2008 and #2655 (both proposing to hide Shorts) were closed as duplicates of this issue, has the scope of this issue been expanded? The title implies filtering based on video titles, but the discussion has also considered filtering specific videos and channels, shorts, premieres, scheduled live streams.

yeah seems pretty unrelated, two different issues

Shadow8472:
+1 I have a family member who relies on channel/keyword blocking. I'd use it myself, but it's not a deal breaker for me.

classicQ:
Please be so kind as to also consider filtering "Upcoming Premieres" and "Shorts" as in some cases it might be super annoying as well

johnno66:
[...]

PR #2853, opened 2022-11-10
Add Distraction Free Settings -> Hide Upcoming Premieres

### LINKED ISSUE #507: [Feature Request] Settings to hide video premieres
Can we have a "Hide premiere videos" inside "Subscription Settings" please? Too many people abusing of video premieres and filling my subscription page with videos that can't be watched yet

miangraham:
# Add Distraction Free Settings → Hide Upcoming Premieres

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #507
see also #1070, #2755

## Description
Add a new boolean setting to Distraction Free: "Hide Upcoming Premieres."

When enabled, removes non-live future premieres from subscription and channel pages.

YT-side data unpredictably flips some streams that get "waiting" viewers from premieres into live streams and back. The live streams introduced can be removed by also enabling the "Hide Live Streams" setting.

Seeking feedback: Due to the weird shapeshifting streams issue above, and in order to avoid adding yet another setting, IMO there's a good argument for combining this new option with "Hide Live Streams" and having one switch do both. If you want one on, you probably want the other. I definitely need both myself. Made it a separate switch for now but happy to make the change if maintainers think combining makes sense.

### New setting
prem_new_setting

### An affected video
prem_vid_info

### Setting OFF

The first item is our target.

prem_before

### Setting ON

And it's gone.

prem_after

## Testing

I've tested the new setting using `npm run dev` against Local API (no fallback) due to Invidious API not cooperating at present, with Fetch Feeds from RSS both OFF and ON.

For a quick test:
1. Find a channel with a non-live upcoming premiere video. Subscribe to the channel.
2. With the setting OFF, verify that the target video appears in Subscriptions.
3. Turn the setting ON.
4. Return to Subscriptions and reload.
5. Verify that the target video no longer appears.

I've tested Subscriptions, Channel and Search pages. If you see areas where I'm not removing videos as appropriate please let me know.

## Desktop
- **OS:** NixOS
- **OS Version:** 22.05
- **FreeTube version:** 3f7a2cf

## Additional context
This PR conflicts with #2849 and the later merge will need rebasing/fixing. I'm borrowing their excellent showResult() addition to ft-list-lazy-wrapper.js since it's easier to work with than the previous v-if one-liner.

### DIFF
--- a/src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
+++ b/src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
@@ -36,11 +36,38 @@ export default Vue.extend({
   computed: {
     hideLiveStreams: function() {
       return this.$store.getters.getHideLiveStreams
+    },
+    hideUpcomingPremieres: function () {
+      return this.$store.getters.getHideUpcomingPremieres
     }
   },
   methods: {
     onVisibilityChanged: function (visible) {
       this.visible = visible
+    },
+
+    showResult: function (data) {
+      if (!data.type) {
+        return false
+      }
+      if (data.type === 'video') {
+        if (this.hideLiveStreams && (data.liveNow || data.lengthSeconds == null)) {
+          // hide livestreams
+          return false
+        }
+
+        if (this.hideUpcomingPremieres &&
+            // Observed for premieres in Local API Channels.
+            (data.durationText === 'PREMIERE' ||
+             // viewCount is our only method of detecting premieres in RSS
+             // data without sending an additional request.
+             // If we ever get a better flag, use it here instead.
+             (data.isRSS && data.viewCount === '0'))) {
+          // hide upcoming
+          return false
+        }
+      }
+      return true
     }
   }
 })
--- a/src/renderer/views/Subscriptions/Subscriptions.js
+++ b/src/renderer/views/Subscriptions/Subscriptions.js
@@ -88,6 +88,11 @@ export default Vue.extend({
     hideLiveStreams: function() {
       return this.$store.getters.getHideLiveStreams
     },
+
+    hideUpcomingPremieres: function () {
+      return this.$store.getters.getHideUpcomingPremieres
+    },
+
     fetchSubscriptionsAutomatically: function() {
       return this.$store.getters.getFetchSubscriptionsAutomatically
     }
@@ -193,6 +198,18 @@ export default Vue.extend({
           if (this.hideLiveStreams) {
             videoList = videoList.filter(item => {
               return (!item.liveNow && !item.isUpcoming)
+            })
+          }
+          if (this.hideUpcomingPremieres) {
+            videoList = videoList.filter(item => {
+              if (item.isRSS) {
+                // viewCount is our only method of detecting premieres in RSS
+                // data without sending an additional request.
+                // If we ever get a better flag, use it here instead.
+                return item.viewCount !== '0'
+              }
+              // Observed for premieres in Local API Subscriptions.
+              return item.durationText !== 'PREMIERE'
             })
           }
           const profileSubscriptions = {
--- a/src/renderer/components/distraction-settings/distraction-settings.js
+++ b/src/renderer/components/distraction-settings/distraction-settings.js
@@ -49,6 +49,9 @@ export default Vue.extend({
     hideLiveStreams: function() {
       return this.$store.getters.getHideLiveStreams
     },
+    hideUpcomingPremieres: function () {
+      return this.$store.getters.getHideUpcomingPremieres
+    },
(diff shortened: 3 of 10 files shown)


PikachuEXE:
#2849 might require some time to be merged
Is it possible to exclude the logic (but not the code pattern, like `showResult` can be keep but without `splitChannels`)

miangraham:
Yes, if #2849 looks like it should come after I can definitely clean this one up with the aim of landing first and having no stubs in the meantime.

If that looks likely then I'm also considering moving the new option higher in the list to sit alongside Hide Live Streams, both because it makes sense for them to have adjacent positions and because it should help avoid many-but-not-all merge conflicts when #2849 needs to rebase on top.

miangraham:
I've done some more testing and I think this is ready for a look.

First time contributor playing it by ear on style since it looks like prettier hasn't been used in awhile. Glad to make tweaks as needed--let me know!

PikachuEXE (approved):
Tested locally
Thanks for your work!

src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
```diff
@@ -36,11 +36,33 @@ export default Vue.extend({
...
+        return false
+      }
+      if (data.type === 'video') {
+        if (this.hideLiveStreams && (data.liveNow || data.lengthSeconds == null)) {
+          // hide livestreams
+          return false
+        }
+        if (this.hideUpcomingPremieres &&
+            (data.isUpcoming || data.premiere || data.durationText === 'PREMIERE' ||
+             (data.isRSS && data.viewCount === '0'))) {
```
ChunkyProgrammer, line 60: Not sure if we should include the rss check as there could be videos that aren't premiers that will be hidden

PikachuEXE: I guess this is `(data.isRSS && data.viewCount === '0')`
Meaning if using RSS feed and view count is zero, **assume** it's premiere

src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
```diff
@@ -36,11 +36,33 @@ export default Vue.extend({
...
+      if (!data.type) {
+        return false
+      }
+      if (data.type === 'video') {
+        if (this.hideLiveStreams && (data.liveNow || data.lengthSeconds == null)) {
+          // hide livestreams
+          return false
+        }
+        if (this.hideUpcomingPremieres &&
+            (data.isUpcoming || data.premiere || data.durationText === 'PREMIERE' ||
```
PikachuEXE, line 59: I think we do need code comment to explain 
`data.premiere || data.durationText === 'PREMIERE'`

Why need both ...?

src/renderer/views/Subscriptions/Subscriptions.js
```diff
@@ -195,6 +200,15 @@ export default Vue.extend({
               return (!item.liveNow && !item.isUpcoming)
             })
           }
+          if (this.hideUpcomingPremieres) {
+            videoList = videoList.filter(item => {
+              if (item.isRSS) {
+                // Nothing else to pivot on in RSS data :(
+                return item.viewCount !== '0'
+              }
+              return (!item.isUpcoming && !item.premiere && item.durationText !== 'PREMIERE')
```
PikachuEXE, line 209: Same as above

PikachuEXE (changes requested):

src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
miangraham: Pika is correct. This is the only feasible way to remove them in the case of RSS data--no other flags to pivot on.

src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
miangraham: Added comments explaining cases I've observed personally and removed cases I haven't observed.

src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
miangraham: I've added some additional comments alongside this to clarify.

PikachuEXE (approved):

ChunkyProgrammer (approved):

absidue (approved):

PR #2855, opened 2022-11-11
Migrate search suggestions to youtubei.js

absidue:
# Migrate search suggestions to youtubei.js

## Pull Request Type

- [x] Feature Implementation

## Description
This is the first step in migrating over to youtubei.js (https://github.com/LuanRT/YouTube.js), as discussed I decided to start with the search suggestions as those are the easiest to do.

The full Innertube instance creation is only needed when you want to decode streaming URLs, so for us that is only needed on the watch page, all other parts of FreeTube don't need it, so we can just create new sessions without needing to fetch the player, for most of FreeTube. This approach should keep FreeTube quick.

**I had to do some shenanigans in the webpack config to get the library to work:**

1. The youtubei.js package is distributed in 3 different versions, only the last one works for our use case:
    - `bundle/browser.js`: browser ready bundle with all of it's dependencies already bundled in (webpack likes to do it's own bundling of dependencies, which is why it errored), this is what webpack picks by default as it's listed in the browser field in their package.json
    - `dist/index.js`: node version, this is the one that overwrites the global fetch function, as most versions of node don't have fetch anyway so it's fine in 99% of situations. It's bad for us though, as that means that none of the fetch calls that FreeTube makes (update check, RSS subscriptions, SponsorBlock) show up in the devtools anymore.
    - `dist/browser.js`: browser version but not prebundled, so it doesn't make webpack complain and it doesn't overwrite the fetch function, so I can make the library use the browser fetch function, which means that it will respect the proxy settings (no need for the {http,https,socks}-proxy-agent dependencies once we fully migrate, as we configure the proxy in electron's settings at startup, the browser fetch function routes everything through the proxy)

2. One of youtubei.js' dependencies linkedom, has an optional dependency on `canvas`, which it tries to import in a try-catch. Webpack doesn't care about the optional nature of the dependency and complains about it being missing. I decided to use the IgnorePlugin to ignore it instead of adding it, as it is a native dependency and failed to build on my machine, additionally youtubei.js only uses linkedom to generate DASH manifests, so it definitely doesn't need the canvas dependency.

Sadly due to webpack not being able to tree shake classes and only having partial support for tree shaking commonjs, this PR blows up the `renderer.js` file by 0.87MB and increases the release build time by 3-6 seconds on my machine. In the future we might be able to improve this by switching to a more modern bundler such as vite. For the electron build that's fine, as electron itself is still massive and the web build isn't affected. The bundle size will go down slightly again once we migrate over other parts of FreeTube, getting rid of the dependencies that they currently use.

## Screenshots 
before
after

## Testing 
Use the search field as usually and check that the search suggestions show up.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -1,19 +1,19 @@
 import { Innertube } from 'youtubei.js'
 
-let searchSuggestionsInnerTube = null
+let searchSuggestionsSession = null
 
-export async function getYTSearchSuggestions(query) {
+export async function getLocalSearchSuggestions(query) {
   // reuse innertube instance to keep the search suggestions snappy
-  if (searchSuggestionsInnerTube === null) {
-    searchSuggestionsInnerTube = await Innertube.create({
+  if (searchSuggestionsSession === null) {
+    searchSuggestionsSession = await Innertube.create({
       // use browser fetch
       fetch: (input, init) => fetch(input, init)
     })
   }
 
-  return await searchSuggestionsInnerTube.getSearchSuggestions(query)
+  return await searchSuggestionsSession.getSearchSuggestions(query)
 }
 
-export function clearYTSearchSuggestionsSession() {
-  searchSuggestionsInnerTube = null
+export function clearLocalSearchSuggestionsSession() {
+  searchSuggestionsSession = null
 }
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -1,0 +1,19 @@
+import { Innertube } from 'youtubei.js'
+
+let searchSuggestionsInnerTube = null
+
+export async function getYTSearchSuggestions(query) {
+  // reuse innertube instance to keep the search suggestions snappy
+  if (searchSuggestionsInnerTube === null) {
+    searchSuggestionsInnerTube = await Innertube.create({
+      // use browser fetch
+      fetch: (input, init) => fetch(input, init)
+    })
+  }
+
+  return await searchSuggestionsInnerTube.getSearchSuggestions(query)
+}
+
+export function clearYTSearchSuggestionsSession() {
+  searchSuggestionsInnerTube = null
+}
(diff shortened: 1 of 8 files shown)


src/renderer/helpers/api/local.js
```diff
@@ -0,0 +1,19 @@
+import { Innertube } from 'youtubei.js'
+
+let searchSuggestionsInnerTube = null
+
+export async function getYTSearchSuggestions(query) {
```
ChunkyProgrammer, line 5: What are you planning on naming the invidious function once it moves to a helper? Just want to make sure the names don't collide. Other than that, code looks good to me (haven't tested yet)

ChunkyProgrammer (commented):
If performance lost from creating new InnerTube clients is noticable, maybe there could be a setting for whether to clear the client or keep it?

src/renderer/helpers/api/local.js
absidue: All the Invidious ones will have Invidious in the name. It probably makes sense to use local instead of YouTube in the name, so I'll rename this one as well.

ChunkyProgrammer: Sounds good

absidue:
The initialisation is certainly expensive, it makes 3 requests during it.

PikachuEXE:
> The initialisation is certainly expensive

What does this mean?

ChunkyProgrammer (dismissed):

ChunkyProgrammer:
What still needs to be changed, exactly?

absidue:
Currently waiting for a release upstream, so we can create new InnerTube instances/sessions without fetching the player javascript. The player is only needed for the watch page to decode the streaming URLs, so for the majority of the functionality in FreeTube we don't need it. Being able to just create new sessions means that it will only require one request instead of three for each new innertube instance, will still need 3 on the watch page of course.

absidue:
This is ready for review now.

absidue:
Closing and reopening to trigger the auto merge workflow

ChunkyProgrammer (dismissed):
Lgtm

PikachuEXE (approved):
Tested locally

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2859, opened 2022-11-12
#1414 share channels

### LINKED ISSUE #1414: Add button to copy channel link on channel pages
When I visit a channel page, I sometimes need to copy the channel link. The only method (i know of) to copy the channel link in either invidious format or youtube format is by clicking the three dots on a random video and copying there. It would make mich more sense to add a new button to the page with either a share icon or the words "share" which could share to invidious or youtube. Picture is below.

Button added to channel pages.

**Screenshots**
image
Adding something like this would be cool. I think a share icon might look nicer here. Also it would be better not colored so the unsubscribe stands out

Alhuin:
# Title

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
 closes #1414

## Description
Add a new share icon in the Channel view.

At first, I implemented this as a dropdown button but this wasn't pretty in terms of UI.

In a second step, I wrote a component ft-share-channel-button, but it was clearly a duplicate of the ft-share-button with less content.

I finally edited the ft-share-button so that it takes a new optionnal parameter: mediaType:
- Not providing it keeps the original behaviour ( by default the mediaType is 'Video')
- Setting it as 'Channel' computes the urls as: {{ provider }}/channel/{{ id }} and renders only Copy and Open links for Youtube and Invidious, as there is no way (as far as I know) to provide an embed link to a Channel.

#### Before
image
image

#### After
image
image

image
image
Screenshot from 2022-11-12 18-24-32
Screenshot from 2022-11-12 18-28-27

## Testing
Click on the shared icon on any Channel view.

## Desktop

- **OS: Fedora**
- **OS Version: Fedora Linux 37 (Workstation Edition)**
- **FreeTube version: 0.18.0**

## Additional context
This is my first contribution to an open source project and first time using VueJs and Electron, any feedback on the worflow or the code style will be greatly appreciated !

### DIFF
--- a/src/renderer/components/ft-share-button/ft-share-button.js
+++ b/src/renderer/components/ft-share-button/ft-share-button.js
@@ -72,9 +72,13 @@ export default Vue.extend({
       return `${this.currentInvidiousInstance}/embed/${this.id}`
     },
 
+    youtubeChannelUrl() {
+      return `https://www.youtube.com/channel/${this.id}`
+    },
+
     youtubeURL() {
       if (this.isChannel) {
-        return `https://www.youtube.com/channel/${this.id}`
+        return this.youtubeChannelUrl
       }
       let videoUrl = `https://www.youtube.com/watch?v=${this.id}`
       // `playlistId` can be undefined
@@ -86,6 +90,9 @@ export default Vue.extend({
     },
 
     youtubeShareURL() {
+      if (this.isChannel) {
+        return this.youtubeChannelUrl
+      }
       // `playlistId` can be undefined
       if (this.playlistId && this.playlistId.length !== 0) {
         // `index` seems can be ignored
--- a/src/renderer/components/ft-share-button/ft-share-button.js
+++ b/src/renderer/components/ft-share-button/ft-share-button.js
@@ -15,6 +15,15 @@ export default Vue.extend({
     'ft-toggle-switch': FtToggleSwitch
   },
   props: {
+    mediaType: {
+      /**
+       * Allows to render the dropdown conditionally
+       * 'Channel' will exclude embed links
+       * 'Video' (default) keeps the original behaviour
+       **/
+      type: String,
+      default: 'Video'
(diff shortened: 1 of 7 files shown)


src/renderer/components/ft-share-button/ft-share-button.vue
```diff
@@ -1,14 +1,15 @@
...
-    theme="secondary"
+    :title="$t(`Share.Share ${mediaType}`)"
+    :theme="isVideo?'secondary':'base-no-default'"
     :icon="['fas', 'share-alt']"
     dropdown-position-x="left"
     :force-dropdown="true"
   >
     <ft-flex-box>
       <ft-toggle-switch
+        v-if="!isChannel"
```
PikachuEXE, line 12: Same for other places
```suggestion
        v-if="isVideo"
```

src/renderer/components/ft-share-button/ft-share-button.js
```diff
@@ -15,6 +15,15 @@ export default Vue.extend({
     'ft-toggle-switch': FtToggleSwitch
   },
   props: {
+    mediaType: {
```
PikachuEXE, line 18: `shareTargetType`?
Channel doesn't seem to be a media type to me...

PikachuEXE (changes requested):

src/renderer/components/ft-share-button/ft-share-button.vue
Alhuin: Done !

src/renderer/components/ft-share-button/ft-share-button.js
Alhuin: You're right !  It's updated

PikachuEXE (approved):
Tested locally

ChunkyProgrammer (approved):

PR #2868, opened 2022-11-14
Themes css deduplication

absidue:
# Themes css deduplication

## Pull Request Type

- [x] Other - Cleanup

## Description
Just deduplicates some of the CSS rules, that way the output is 6KB smaller. You probably want to review this commit by commit, instead of looking at it all in one go.

## Screenshots 
before
after

## Testing 
Check that the colours still work, you don't need to test all of them if you don't want to.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/themes.css
+++ b/src/renderer/themes.css
@@ -343,130 +343,107 @@
   --logo-text-bar-color: url("../../_icons/textDraculaDarkSmall.png");
 }
 
+.mainCatppuccinMochaRosewater,
+.mainCatppuccinMochaFlamingo,
+.mainCatppuccinMochaPink,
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
Is it possible to have tools to do this for us semi-auto/full-auto in the future?

ChunkyProgrammer (approved):

absidue:
I can try some of the other tools that the CssMinimizerPlugin supports, currently we are using cssnano which is the default.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #1869, opened 2021-11-03
[MacOS] Extended shortcuts


lothar-cell:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
More shortcuts would be great, because shortcuts are a good thing.

**Describe the solution you'd like to see implemented**
Would be great if commands like:
- Switch Channel 
- Show Previous/Next Tab
- ~New Page (⌘N)~
- ~New Tab (⌘T)~
- ~Home~
- ~Settings (⌘,)~ 
- ~History (⌘Y)~
- ~Back, Forward~
- ~Subscriptions~
- ~Playlists~

and so on would be implemented, they would be menu titles, so they could be customised in OS settings as well. 

**Screenshots**
![Screenshot 2021-11-03 at 7 53 18](https://user-images.githubusercontent.com/2356354/140019587-c06435e4-663f-4700-903d-2ed9afd27230.png)

Shortcut customisation in OS, needs a menu title:
![Screenshot 2021-11-03 at 7 55 17](https://user-images.githubusercontent.com/2356354/140019763-582fbdf1-929b-4ec6-bad5-b046caadbbb1.png)

**Additional context**
no additional context.

efb4f5ff-1298-471a-8973-3d47447115dc:
Duplicate of https://github.com/FreeTubeApp/FreeTube/issues/251

lothar-cell:
Actually it's not a duplicate of #251, but a completely different issue. This one is about extending the menu items, which can use the native MacOS shortcut system. See screenshots.

efb4f5ff-1298-471a-8973-3d47447115dc:
I'll let @PikachuEXE decide if this is worth reopening.

PikachuEXE:
I think this is a valid issue and not a dupe to #251. 
However I have no idea how "menu titles" can be implemented.

lothar-cell:
Template updated, I think it was completely unnecessary, but hopefully @efb4f5ff-1298-471a-8973-3d47447115dc will be satisfied and U: doesn't follow template can be removed.

lothar-cell:
> However I have no idea how "menu titles" can be implemented.

Maybe this can be of help? https://stackoverflow.com/questions/38208469/electron-osx-application-menu-how-to-add-custom-file-menu

PikachuEXE:
Create first PR for this: https://github.com/FreeTubeApp/FreeTube/pull/2306

lothar-cell:
[...]

PR #2870, opened 2022-11-16
Update app menu to add item for history

PikachuEXE:
# Update app menu to add item for history

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/1869 Partially

## Description

Update app menu to add item for history
with keyboard shortcut

## Screenshots 

image

## Testing 

- Use new menu item to navigate to history view
- Use new keyboard shortcut (Ctrl/Cmd + `Y`) to navigate to history view

## Desktop

- **OS:** MacOS
- **OS Version:** 12.6
- **FreeTube version:** 2cc5c2e1ddd7c9d54d508725fc1fdf861fc7527d

## Additional context
The namespace for translation entries used in history view needs to be updated to avoid key conflict


absidue:
Pretty sure you didn't need to update the translation keys, as the menu items don't get translated anyway at the moment.
I think this should be two separate shortcuts:
- cmd + Y on macOS as that's what the issue requestor wants
- ctrl + H on Windows and Linux, as CTRL + Y is the keyboard shortcut for redo (the opposite of CTRL + Z to undo), CTRL + H is also the shortcut for opening the history on Firefox and Chromium browsers on Windows

PikachuEXE:
Too complicated to have different keywords, I will just update it to Ctrl/Cmd + H

src/main/index.js
```diff
@@ -997,7 +997,21 @@ function runApp() {
           { role: 'zoomout' },
           { role: 'zoomout', accelerator: 'CmdOrCtrl+numsub', visible: false },
           { type: 'separator' },
-          { role: 'togglefullscreen' }
+          { role: 'togglefullscreen' },
+          { type: 'separator' },
+          {
+            label: 'History',
+            accelerator: 'CmdOrCtrl+U',
```
absidue, line 1004: It wouldn't be too hard to have a separate shortcut on macOS and have a meaningful one on Windows and Linux
```suggestion
            accelerator: process.platform === 'darwin' ? 'Cmd+Z' : 'Ctrl+H',
```

PikachuEXE: `Cmd+Z`? `Cmd+Y`?

absidue: What's the standard shortcut for history in apps on macOS?

PikachuEXE: `Cmd+Y` I guess
- [Safari](https://www.wikihow.com/View-Your-Browser-History-on-a-Mac)
- [Chrome](https://support.google.com/chrome/answer/157179?hl=en)

PikachuEXE: Updated but still need to be tested on windows

yarn.lock
```diff
@@ -7133,11 +7133,6 @@ sax@^1.1.3, sax@^1.2.4:
   resolved "https://registry.yarnpkg.com/sax/-/sax-1.2.4.tgz#2816234e2378bddc4e5354fab5caa895df7100d9"
   integrity sha512-NqVDv9TpANUjFm0N8uM5GxL36UgKi9/atZw+x7YFnQ8ckwFGKrl4xX4yWtrey3UJm5nP1kUbnYgLopqWNSRhWw==
 
-sax@~0.6.0:
```
PikachuEXE, line 7136: I keep getting this on MacOS after running `yarn install`

.eslintrc.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
 
   rules: {
     'space-before-function-paren': 'off',
-    'comma-dangle': ['error', 'never'],
+    'comma-dangle': ['error', 'only-multiline'],
```
PikachuEXE, line 36: I think multiline should **allow** (but not **enforce**) last comma

src/main/index.js
```diff
@@ -997,7 +997,23 @@ function runApp() {
           { role: 'zoomout' },
           { role: 'zoomout', accelerator: 'CmdOrCtrl+numsub', visible: false },
           { type: 'separator' },
-          { role: 'togglefullscreen' }
+          { role: 'togglefullscreen' },
```
PikachuEXE, line 1000: Allowing last comma on multiline makes duplicating last code and modifying it much easier, also no more diff on N-1th line

src/main/index.js
lothar-cell: Thank you for implementing shortcuts! Once a menu item, a shortcut can be customised in macOS, so anything is good with me.

PikachuEXE: > Once a menu item, a shortcut can be customised in macOS, so anything is good with me.

I didn't know that~

lothar-cell: Yep, like this: 
Screenshot 2022-11-21 at 8 00 25

static/locales/en-US.yaml
```diff
@@ -119,7 +120,7 @@ User Playlists:
     it listed here
   Empty Search Message: There are no videos in this playlist that matches your search
   Search bar placeholder: Search in Playlist
-History:
+History Page:
```
ChunkyProgrammer, line 123: Why rename this?
`$t("History.History")` would have gotten the history value before the update.

Also all locales that already have the History translated will need to update to "History Page" in order to not use the english translations

PikachuEXE: According to the comment `# Webkit Menu Bar` those entries are for `Webkit Menu Bar` and seems that they can only be placed on top level
I have no idea if that's true or not
Let me know?

The commit for that comment and many other entries are from [a commit 2 years ago](https://github.com/FreeTubeApp/FreeTube/commit/8aa0dcd243cf09312a25559b6be59ab71ee5895b)

absidue: At the moment the menu doesn't get translated at all, so the translation change is not necessary yet, until someone decides to setup translations for the menu.

PikachuEXE: OK I will revert the change
Confusing!

PikachuEXE:
translation namespace changes reverted

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #2832, opened 2022-11-05
[Feature Request]: Open video/channel in new window option


ghost:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

Trying to visit multiple profiles or videos can get cluttered:
If you want to view a video you have to open it in the same window, and if you want multiple video windows up you have to open new windows every time and find where you where.

Although the middle mouse click can fix this, it can be really helpful to have this in a visible choice, encase of using touch screen, touch pad (anything other than a mouse), or simply for ease to the average person.

### Proposed Solution

Add "Open in New Window" in the ellipsis options ⋮ of videos, channels, playlists and such.

### Alternatives Considered

Or using the "Open New Window" button next to the logo, a second option asking which place to visit (Previous page, Playlists... etc)

### Issue Labels

ease of use improvement

efb4f5ff-1298-471a-8973-3d47447115dc:
Videos and channels can already be opened in new window by using CTRL + CLICK

eightyy8:
Updated my request, read the additional info.

Further, if it is a basic feature it needs to be accessible in sight.

eightyy8:
Also you cannot open History, Favorites or other side options other then the channels.

efb4f5ff-1298-471a-8973-3d47447115dc:
Okay so this is somewhat related to https://github.com/FreeTubeApp/FreeTube/issues/2094 in the sense that the user always needs the option to open something in a new window/tab.

Could u reword your issue title and body a bit to match this description better.

eightyy8:
> Please reword your issue title and body a bit to match this description better.

Done, hopefully it's clear to everyone.

eightyy8:
All this feature request is to allow interface support for the open in new window feature, which is otherwise only accessible through mouse or keyboard...

Just a button on videos, channels and playlists would really help your accessibility, whilst using touch screen, or anything other than a mouse.

absidue:
[...]

PR #2873, opened 2022-11-18
Fix "Open in new window" context menu action in release builds

absidue:
# Fix "Open in new window" context menu action in release builds

## Pull Request Type

- [x] Bugfix

## Related issue
Partially addresses #2832

## Description
In the release builds, all files are accessed using the `file:` protocol, including the index.html file. Interestingly how the `new URL()` class handles file URLs differs depending on the implementation. In node the `origin` field is `null`, in Firefox it's a string with null in it `"null"` and in Chromium it's `file:///`. In release builds, the current implementation was checking if the linkURL contained `null`, which it never did, so it always hid the context menu entry.

## Testing 
1. `yarn build`
2. run the built release build
3. right click on links and check that the Open in New Window context menu entry shows up

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -31,7 +31,7 @@ function runApp() {
       },
       {
         label: 'Open in a New Window',
-        visible: parameters.linkURL.includes((new URL(browserWindow.webContents.getURL())).origin),
+        visible: parameters.linkURL.startsWith(browserWindow.webContents.getURL()),
         click: () => {
           createWindow({ replaceMainWindow: false, windowStartupUrl: parameters.linkURL, showWindowNow: true })
         }


src/main/index.js
```diff
@@ -31,7 +31,7 @@ function runApp() {
       },
       {
         label: 'Open in a New Window',
-        visible: parameters.linkURL.includes((new URL(browserWindow.webContents.getURL())).origin),
+        visible: parameters.linkURL.startsWith(browserWindow.webContents.getURL()),
```
PikachuEXE, line 34: Probably should add code comment to document this pitfall

PikachuEXE (changes requested):

src/main/index.js
absidue: Added a comment now.

PikachuEXE (approved):
Tested locally

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2879, opened 2022-11-20
Add ft-input enter listener the vue way

absidue:
# Add ft-input enter listener the vue way

## Pull Request Type

- [x] Other - Cleanup

## Description
The current implementation uses setTimeout to add the listener and has a theoretical race condition, in which it would not add the listener at all. This PR moves away from the setTimeout approach and instead makes vue manage the event handler, which avoids the race condition and also makes our code simpler.

I also replaced the getElementById calls with vue refs (the ref was already there).

## Testing 
Enter some search queries into the search bar and press enter, it should still do the search.
Check that the clear button still works.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -103,8 +103,6 @@ export default Vue.extend({
     this.id = this._uid
     this.inputData = this.value
     this.updateVisibleDataList()
-
-    setTimeout(this.addListener, 200)
   },
   methods: {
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer (approved):

PikachuEXE (approved):
I can still search Pikachu thx
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2904, opened 2022-12-01
Make replaceFilenameForbiddenChars a helper

absidue:
# Make replaceFilenameForbiddenChars a helper

## Pull Request Type

- [x] Other - Refactor

## Description
The benefits are that we get rid of one unnecessary use of await, the function can now be dead code eliminated in web builds and it also gets rid of one place where a polyfill would be needed in the web build. We will need to replace more local API modules with YouTube.js and check other places in FreeTube, before we can get rid of polyfills entirely, this is a start though.

## Testing 
Set the download behaviour to "Download in app" and download this video https://youtu.be/O16lQlK7bkw, the slash in the title should be replaced by a similar looking unicode character, the same as before.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -422,3 +422,36 @@ export function searchFiltersMatch(filtersA, filtersB) {
     filtersA?.type === filtersB?.type &&
     filtersA?.duration === filtersB?.duration
 }
+
+export function replaceFilenameForbiddenChars(filenameOriginal) {
+  let filenameNew = filenameOriginal
+  let forbiddenChars = {}
+  switch (process.platform) {
+    case 'win32':
+      forbiddenChars = {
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer (commented):
Should we replace this with that function too?  (Instead of doing an if for different environments) https://github.com/FreeTubeApp/FreeTube/blob/40ff0fad7ece01b9adfe240d051259d3f00eda21/src/renderer/store/modules/utils.js#L256

ChunkyProgrammer (approved):

PikachuEXE (approved):
Tested locally

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2918, opened 2022-12-05
Increase dependebot PR limit to 10

absidue:
# Increase dependebot PR limit to 10

## Pull Request Type

- [x] Other - Automation

## Description
Even after the recent dependency updates we still have a lot of outdated dependencies (see screenshot below), increasing the PR limit from the default of 5 to 10, should hopefully make a bigger dent in that list, in the future.

## Screenshots 
dependencies

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,7 @@ updates:
     labels:
       - "PR: waiting for review"
       - "PR: dependencies"
+    open-pull-requests-limit: 10
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:


ChunkyProgrammer:
We could also make it daily so it'll check daily and only have a max of 10 prs open

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> We could also make it daily so it'll check daily and only have a max of 10 prs open

Lets try this first and check daily if this doesnt work as expected

PR #2956, opened 2022-12-13
Implement playlist pagination for the local API

absidue:
# Implement playlist pagination for the local API

## Pull Request Type

- [x] Feature Implementation

## Description
This pull request implement pagination on the playlist page when the local API is selected. I also got rid of the unused methods and variables that were copied from the search page.

I didn't implement pagination for Invidious in this PR, as their pagination system is confusing, weird and will likely require talking to the Invidious devs to understand it or ask them to fix their system (see the discussion on Matrix for more details).

## Testing 

playlist with 100 videos, only a single page (check that load more button is hidden)
https://www.youtube.com/playlist?list=PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i

playlist with 872 videos, multiple pages
https://www.youtube.com/playlist?list=PLbMjU_TVMIFvmkQ871iUvE5gy5ezXc6hE

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -4,6 +4,8 @@ import FtLoader from '../../components/ft-loader/ft-loader.vue'
 import FtCard from '../../components/ft-card/ft-card.vue'
 import PlaylistInfo from '../../components/playlist-info/playlist-info.vue'
 import FtListVideo from '../../components/ft-list-video/ft-list-video.vue'
+import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
+import FtButton from '../../components/ft-button/ft-button.vue'
 import i18n from '../../i18n/index'
 import { getLocalPlaylist } from '../../helpers/api/local'
 import { extractNumberFromString } from '../../helpers/utils'
@@ -14,17 +16,18 @@ export default Vue.extend({
     'ft-loader': FtLoader,
     'ft-card': FtCard,
     'playlist-info': PlaylistInfo,
-    'ft-list-video': FtListVideo
+    'ft-list-video': FtListVideo,
+    'ft-flex-box': FtFlexBox,
+    'ft-button': FtButton
   },
   data: function () {
     return {
       isLoading: false,
       playlistId: null,
-      nextPageRef: '',
-      lastSearchQuery: '',
-      playlistPage: 1,
       infoData: {},
-      playlistItems: []
+      playlistItems: [],
+      continuationData: null,
+      isLoadingMore: false
     }
   },
   computed: {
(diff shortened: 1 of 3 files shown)


ChunkyProgrammer:
For invidious, it does give you a `videoCount` so we could look at that to see if there's any more pages

absidue:
I haven't looked at Invidious' code but they are probably just sending the estimate that YouTube has on their page, I say estimate because that number also counts the videos that can't be played (privated, copyright striked or deleted) and are not shown in the playlist. So that number might be the same as the total number of videos but it lots of playlists, especially big ones, that number will always be larger than the total number of videos.

ytpl and ytch both check for continuation data, Invidious doesn't seem to have that mechanism, or at least not exposed through the API

absidue:
eg.
with the 100 video playlist from above that only has a single page

with page=1 (the same as not specify the page parameter) you get all 100 videos
https://inv.riverside.rocks/api/v1/playlists/PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i?page=1

with page=2, you get the last 50 videos in the playlist, despite the video only having one playlist
https://inv.riverside.rocks/api/v1/playlists/PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i?page=2

with page=3, you finally get an empty array of videos
https://inv.riverside.rocks/api/v1/playlists/PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i?page=3

ChunkyProgrammer:
We should definitely open an issue on the invidious repo then

src/renderer/views/Playlist/Playlist.js
```diff
@@ -155,22 +164,32 @@ export default Vue.extend({
...
 
-    nextPage: function () {
-      const payload = {
-        query: this.query,
-        options: {
-          nextpageRef: this.nextPageRef
-        },
-        nextPage: true
+    getNextPage: function () {
+      switch (this.backendPreference) {
```
ChunkyProgrammer, line 168: Not sure if we should have this at the moment. If the show more button is only visible when local api is used then wouldn't this only error if someone has invidious as preferred backend and it had a fallback of local api for the playlist request?

absidue: Changed it to check `this.infoData.infoSource` instead, that way it should work for the backend fallback too.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested with

playlist with 100 videos, only a single page (check that load more button is hidden)
https://www.youtube.com/playlist?list=PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i

playlist with 872 videos, multiple pages
https://www.youtube.com/playlist?list=PLbMjU_TVMIFvmkQ871iUvE5gy5ezXc6hE

ChunkyProgrammer:
It appears to be a known invidious issue https://github.com/iv-org/invidious/issues/2694

PR #2943, opened 2022-12-12
Fix ft-input keyboard naviagation

ChunkyProgrammer:
# Fix keyboard navigation for input lists (search + setting invidious instance)
## Pull Request Typ
- [x] Bugfix

## Related issue
closes #2708

## Description
No longer set the input value on arrow keys (this is an accessibility enhancement)

## Testing 
- Use arrow keys to navigate search suggestions + invidious instance
- Click enter to set instance

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** 0.18.0

PR #2960, opened 2022-12-13
fix error when searching

ChunkyProgrammer:
# Fix error when searching

## Pull Request Type
- [x] Bugfix

## Description
Error introduced in #2943 
If you click enter without selecting an option in the list then an error will occur

## Testing 
Write something in search bar, click enter without hovering over any of the options

## Desktop
- **OS:** Windws
- **OS Version:** 11

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -197,14 +197,13 @@ export default Vue.extend({
     handleKeyDown: function (event) {
       if (this.visibleDataList.length === 0) { return }
       if (event.key === 'Enter') {
-        this.inputData = this.visibleDataList[this.searchState.selectedOption]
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

absidue:
Even with this PR, if there are no search suggestions, searching by pressing enter doesn't work, it does work if you click the search button. If you enter a URL e.g. `https://youtube.com/c/linustechtips` you won't get any search suggestions and then it doesn't work with enter.

PikachuEXE:
searching by pressing enter does work for me BUT only when search suggestion appears
enter with URL does not (coz no suggestion)

PikachuEXE:
Opened #2966

PR #2961, opened 2022-12-14
Add support for lefthook local config for setting up option `rc`

PikachuEXE:
Add support for lefthook local config for setting up option `rc`

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue
Improvement for lefthook added in #1788

## Description
I use SourceTree
And it's impossible to run lefthook in SourceTree due to ENV values in SourceTree are different from those in CLI
lefthook `1.2.4` introduced [`rc` option](https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#rc)
Which allows GUI apps to also run lefthook with extra setup steps

This PR makes git ignores local config for lefthook `lefthook-local.yml` and an example file for it
Though due to [an issue (still present in `1.2.5`)](https://github.com/evilmartians/lefthook/pull/392#issuecomment-1350252074) the value must be absolute path, i.e. cannot use prefix `~/`

## Screenshots 
What do you expect to see ._.

## Testing 
- Install & Open SourceTree
- Switch to this branch
- Try committing a JS file, confirms it failed due to missing binary (might succeed for some setup but not for mine)
- Create `lefthook-local.yml` (just copy from `lefthook-local.yml.example`)
- Commit again, should work now

## Desktop

- **OS:** MacOS
- **OS Version:** 12.6.1
- **FreeTube version:** 29f5e38623d02f2e19a0d396aa615fbebf3d1238


lefthook-local.yml.example
```diff
@@ -0,0 +1,7 @@
+# See following doc for details
+# https://github.com/evilmartians/lefthook/blob/master/docs/configuration.md#rc
+
+# You can choose whatever name you want.
```
ChunkyProgrammer, line 4: Maybe we should have a message that says something like `we recommend you use 'lefthook-local.yml' as the name to avoid accidentally commiting your lefthook file but feel free to use any filename`

PikachuEXE: `You can choose whatever name you want` means the `~/.lefthookrc` file
I have updated comment to be clearer

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2965, opened 2022-12-15
Update app menu to add items for side nav items

PikachuEXE:
# Update app menu to add items for side nav items

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
For #1869

## Description
Update app menu to add items for side nav items
- Subscriptions
- Channels
- Trending
- Most Popular
- Playlists

## Screenshots 
image

## Testing 
- Open app menu `View`
- Click x5

## Desktop

- **OS:** MacOS
- **OS Version:** 12.6.1
- **FreeTube version:** 083ae0e622d4fd90d42f23969a5b8ed612229383


absidue:
The View menu is getting quite busy, it might make sense to have a separate one for these (history could be moved to that menu too.

PikachuEXE:
I need a name ._.

absidue:
Navigation maybe?

absidue:
also not sure if you want to do it, but it might be good to hide options depending on the users settings (e.g. if they have the hide trending distraction free setting enabled, then it would probably be good to hide the menu entry too)

PikachuEXE:
However it can only work on boot right?
i.e. no way to hide it on the fly after setting updated
That's the limitation I can think of

Gonna use `Navigate`
image

absidue:
I think if you regenerate the whole menu it should work, well at least native apps are allowed to change their menus when they want to, so i suspect you could do similar stuff with electron. You could also try adjusting the `visible` property on the menu items, remember to set `acceleratorWorksWhenHidden` to false on the hidden menu items, as it defaults to true

PikachuEXE:
Do it later ._.

ChunkyProgrammer (approved):

PikachuEXE:
@absidue Implemented app menu update! Time to review!
image

ChunkyProgrammer (dismissed):

absidue (commented):
Was wondering why the navigate menu wasn't showing up for me, turns out if any of the hide trending, hide popular or hide playlists settings have never been set, they don't exist in the database, so an error is thrown while getting the values. This makes Electron show the default menu instead.

navigate-menu-missing

after uncommenting the `{ stdio: 'inherit' }` line in dev-runner i was able to see this error:
error

PikachuEXE:
@absidue Fixed and tested by updating keys to read to something absent in db
image

src/main/index.js
```diff
@@ -1043,7 +1054,21 @@ function runApp() {
...
+      hideTrendingVideos = (await sidenavSettings.hideTrendingVideos).value
+    } catch {}
+    let hidePopularVideos = false
+    try {
+      hidePopularVideos = (await sidenavSettings.hidePopularVideos).value
+    } catch {}
+    let hidePlaylists = false
+    try {
+      hidePlaylists = (await sidenavSettings.hidePlaylists).value
+    } catch {}
```
absidue, line 1070: This can be simplified to this, as the error was occuring when it was trying to read the `value` property and `null` is falsey.
```suggestion
    const hideTrendingVideos = (await sidenavSettings.hideTrendingVideos)?.value
    const hidePopularVideos = (await sidenavSettings.hidePopularVideos)?.value
    const hidePlaylists = (await sidenavSettings.hidePlaylists)?.value
```

absidue: I tested the change before suggesting it.

PikachuEXE:
Done!

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

lothar-cell:
Great to see these implemented as menu items! Any chance for Profiles as well?

PikachuEXE:
This one? 
image

lothar-cell:
Yes. Probably it's not possible yet, but the best would be to have slots for profile 0-9, so individual shortcuts could be assigned to each, like opt+cmd 0-9.

PikachuEXE:
I doubt it's acceptable to generate menu items for individual profiles (the number can be unlimited)
But just for profile list is fine

lothar-cell:
The idea came from Firefox and Safari, where Container Tabs / Profiles have the same option.

PR #2970, opened 2022-12-17
Pass playlist

absidue:
# Pass playlist data from playlist page to watch page

## Pull Request Type

- [x] Feature Implementation

## Description
Currently when you start playling a video in a playlist from the playlist page, the watch page fetches the playlist information again. With this PR the watch page can reuse the playlist information that was already fetched on the playlist page, this prevents unnecessary network activity and also speeds up the loading of the playlist component on the watch page, as it has less to do. If the playlist has extra pages that haven't been fetched yet, the watch page will fetch those.

The watch page will of course still need to fetch everything itself if you come from a different page.

I also got rid of some unused properties, as I noticed them while implementing the changes mentioned above.

## Testing 

**Single page playlist**
https://www.youtube.com/playlist?list=PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i

**Playlist with lots of pages**
1. Don't load anymore pages on the playlist page, should be slow on the watch page but slightly faster than before
2. Click "Load more" a few times on the playlist page before playing a video, the playlist view on the watch page should load significantly faster
https://www.youtube.com/playlist?list=PLbMjU_TVMIFvmkQ871iUvE5gy5ezXc6hE

**Jump directly to the watch page, without visiting the playlist page first**
https://youtu.be/Uq9gPaIzbe8?list=PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/watch-video-playlist/watch-video-playlist.js
+++ b/src/renderer/components/watch-video-playlist/watch-video-playlist.js
@@ -31,7 +31,6 @@ export default Vue.extend({
       reversePlaylist: false,
       channelName: '',
       channelId: '',
-      channelThumbnail: '',
       playlistTitle: '',
       playlistItems: [],
       randomizedPlaylistItems: []
@@ -251,9 +250,7 @@ export default Vue.extend({
       this.setCachedPlaylist(null)
 
       this.playlistTitle = cachedPlaylist.title
-      this.videoCount = cachedPlaylist.videoCount
       this.channelName = cachedPlaylist.channelName
-      this.channelThumbnail = cachedPlaylist.channelThumbnail
       this.channelId = cachedPlaylist.channelId
 
       if (!process.env.IS_ELECTRON || this.backendPreference === 'invidious' || cachedPlaylist.continuationData === null) {
@@ -286,9 +283,7 @@ export default Vue.extend({
         let playlist = await getLocalPlaylist(this.playlistId)
 
         this.playlistTitle = playlist.info.title
-        this.videoCount = playlist.info.total_items
         this.channelName = playlist.info.author?.name
-        this.channelThumbnail = playlist.info.author?.best_thumbnail?.url
         this.channelId = playlist.info.author?.id
 
         const videos = playlist.items.map(parseLocalPlaylistVideo)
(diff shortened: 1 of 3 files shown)


src/renderer/components/watch-video-playlist/watch-video-playlist.js
```diff
@@ -242,16 +245,45 @@ export default Vue.extend({
       }
     },
 
+    loadCachedPlaylistInformation: async function (cachedPlaylist) {
+      this.isLoading = true
+      this.setCachedPlaylist(null)
```
ChunkyProgrammer, line 250: Would this clear the cache when the next video in the playlist loads?

absidue: Nope, the playlist view just gets hidden with CSS, while changing to a new video in the same playlist ([src/renderer/views/Watch/Watch.vue#L162-L163](https://github.com/FreeTubeApp/FreeTube/blob/176716483df81c8ea138d0d4d568652fa973f55e/src/renderer/views/Watch/Watch.vue#L162-L163)), it doesn't get unmouted and mounted again with new data, like the description or up next views.

This is a good idea, as the playlist view is one of the most expensive components to process and render inside FreeTube, as each ft-list-video element has quite a lot of checks (the amount of checks has increased a lot with #2849 and #2853) to see if it should actually be shown and then Vue also has to create all of those DOM elements (remember each ft-list-video element also has a dropdown menu inside of it), so doing it 872 times, for the large playlist above, is extremely processing intensive (you can notice it if you try to pause or start the video just as vue is inserting all those DOM elements, it will lag for a moment before resuming/pausing the video).

Most video lists in FreeTube have pagination, the playlist view on the watch page on the other hand has to do all of that work in one go.

ChunkyProgrammer (approved):

PikachuEXE (approved):
Tested locally with long list
Loaded up all pages than switch to a video, loaded instantly

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2982, opened 2022-12-19
Cleanup and simplify the settings password code

absidue:
# Cleanup and simplying the settings password code

## Pull Request Type

- [x] Other - Refactor

## Description
A lot of the changes in this PR are likely from when the original PR was altered to split the settings password into two separate parts. The goal of this PR is to clean it up the code, without affecting any functionality, so it's easier to maintain in the future.

You may want to enable hide whitespace while reviewing this PR, that'll make it easier to see the actual changes and not just the indentation that changed due to removing code.

List of changes:
1. Removed unused code (e.g. the wrong password prompt, which is never displayed)
2. Add some CSS to stop stretching the set password button
3. Remove unneeded divs and ft-flex-box in the templates, as we don't need to make Vue insert elements into the DOM that don't make a difference to the appearance or layout
4. Make the Settings view check for the existence of a password itself, instead relying on both the password-dialog and the password-settings section to emit locked and unlocked events
5. Change the names of some of the variables

## Screenshots 
There is only one visual change, which is that the set password button is no longer vertically stretched.

before

after

## Testing 
Try various combinations of setting, clearing and entering the password, while leaving the settings inbetween.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/password-dialog/password-dialog.css
+++ b/src/renderer/components/password-dialog/password-dialog.css
@@ -1,10 +1,9 @@
 .card {
-    position: relative;
-    width: 85%;
-    height: 25%;
-    margin: auto;
-    box-sizing: border-box;
+  width: 85%;
+  margin: auto;
+  box-sizing: border-box;
 }
+
 .passwordInput {
-    width: 100%;
+  width: 100%;
 }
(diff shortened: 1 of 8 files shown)


PikachuEXE (approved):
Tested original behaviour

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2293, opened 2022-06-05
Improve channel info bar

vallode:
---
Improve channel info bar
---

**Pull Request Type**
Please select what type of pull request this is:
- [x] Feature Implementation

**Related issue**
N/A

**Description**
Slight refactor to the way we render the channel info bar fixing some issues and improving on others:

* Fixes issue where tabs would "jump around" when you navigate over to the about tab
* Fixes cramped space on mobile and desktop view

**Screenshots (if appropriate)**
Before:
image

After:
image

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] 10
 - OS Version: [e.g. 22] Debian
 - FreeTube version: [e.g. 0.8] upstream/development

PR #2984, opened 2022-12-20
Improve accessibility of Channel View

ChunkyProgrammer:
# Improve Channel View accessibility

## Pull Request Type
- [x] Bugfix

## Related issue
Includes modified changes from #2293 and #1525

Screenshots:
Before:
image

After:
image

## Description
Improves navigation of channel page using only a keyboard. 
coauthors: @vallode @jasonhenriquez 

## Testing 
Here's a good account to use for testing (LinusTechTips) https://www.youtube.com/channel/UCXuqSBlHAE6Xw-yeJA0Tunw
### Channel Tabs
- using arrow keys should change focus of the tab
- hitting tab while on the channel tabs should go to the search input
- hitting enter or space should change the tab panel that is shown

### About
- you should be able to tab to featured channels
- clicking enter or space will navigate to the focused channel

### Fetch More
- you should be able to tab to the `Fetch More` control.
- Clicking enter or space on the  Fetch More control should fetch more 

## Desktop

- **OS:** Windows
- **OS Version:** 11


src/renderer/components/ft-channel-bubble/ft-channel-bubble.js
```diff
@@ -21,8 +21,20 @@ export default Vue.extend({
...
+  computed: {
+    sanitizedId: function() {
+      return 'channelBubble' + sanitizeForHtmlId(this.channelName)
+    }
+  },
   methods: {
-    handleClick: function () {
+    handleClick: function (event) {
+      if (event instanceof KeyboardEvent) {
+        if (event.target.getAttribute('role') === 'link' && event.key !== 'Enter') {
```
absidue, line 32: What is this check for? the ft-channel-bubble component doesn't contain any elements with the link role

src/renderer/views/Channel/Channel.vue
```diff
@@ -151,10 +160,29 @@
             :channel-name="channel.author || channel.channelName"
             :channel-id="channel.channelId"
             :channel-thumbnail="channel.authorThumbnails[channel.authorThumbnails.length - 1].url"
+            role="link"
```
ChunkyProgrammer, line 163: @absidue i believe this is why link is checked in the ft-channel-bubble component

src/renderer/components/ft-channel-bubble/ft-channel-bubble.js
absidue: Ah that makes sense.

src/renderer/components/ft-channel-bubble/ft-channel-bubble.js
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue'
-
+import { sanitizeForHtmlId } from '../../helpers/accessibility'
```
PikachuEXE, line 2: Spacing!!
```suggestion
import { sanitizeForHtmlId } from '../../helpers/accessibility'

```

PikachuEXE (changes requested):
Tested

PikachuEXE:
Found an issue while testing but probably not caused by this PR:
Cannot fetch more in channel (local API)/keep fetching first page (Invidious)

Local API
image

Invidious
image

ChunkyProgrammer:
> Found an issue while testing but probably not caused by this PR:
> 
> * Cannot fetch more in channel (local API) (Update: might need to fetch twice to reproduce)
> * Keep fetching first page (Invidious)
> 
> Local API image
> 
> Invidious image

I don't think that's caused by this PR.
Another issue I noticed (not entirely caused by this PR) is that you can spam enter on the fetch more and it will just add duplicates to the videolist if you spam quick enough.

ChunkyProgrammer:
Would also like to add that this change is pretty necessary before we add livestreams + shorts to channel page (as space is not handled that well before this PR)

absidue (approved):
tested locally

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #1115, opened 2021-03-14
WIP: refactor

komachi:
refactor: remove vendor prefixes from css/scss
refactor: get subscriptions in parallel
refactor: us fs/promises instead of sync functions from fs
refactor: replace some setTimeout calls in wrong places with better callbaack
refactor: better promise usage, cleaner code
refactor: remove unneded babel plugins

---
This is a little refactoring I am in progress of
---

**Description**

I removed jquery dependency. Jquery should not be used to manipulate any content vue is responsible to render, `$.get` and `$.getJSON` is replaced with moden ky, `$.param` replaced with small module `jquery-param` which implements the same algo. There is still some cases where we need to manipulate DOM (on video.js side mainly), but this could be easy solved with `document.querySelector` and some small helpers from `shorter-js`.

In some places `Promises` were used incorrectly, I tried to solve those cases with more cleaner code. Also I replaced sync calls to fs module with async ones.

Ideally this shouldn't broke anything, only makes codebase cleaner and an app a little bit faster

**Testing (for code that is not small enough to be easily understandable)**

There is still some bugs with search menu and profile selector I plan to fix later, but I want to know what do you think about this PR.

PR #2991, opened 2022-12-24
Switch from sync fs APIs to fs/promises

absidue:
# Switch from sync fs APIs to fs/promises

## Pull Request Type

- [x] Other - Refactor

## Related issue
related to #1115

## Description
Currently we use the synchronous file system APIs for the majority of the filesystem operations, these are blocking, which is problematic as JavaScript is single threaded, so they prevent any other code from running while the operations are happening. Migrating to the promises file system APIs should alleviate that issue.

## Testing 
1. Enable and disable the experimental replace HTTP cache setting and check that the file is created and removed properly in the userdata directory
2. Change the language, check that the new one is loaded and that the region names for trending show up
3. Watch a video through the local API and check that the storyboard and DASH files are created and removed correctly when you change to a different page
4. Export something in the data settings, checking that the file is written properly to your chosen export location
5. Import something in the data settings
6. Download a video inside FreeTube (downloading an audio track will be the fastest to test), check that the file is correctly written to your download location

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (dismissed):
all tested

PikachuEXE (dismissed):

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #2994, opened 2022-12-25
Improve responsiveness of `Hide Videos From Channels`

MarmadileManteater:
# Improve responsiveness of `Hide Videos From Channels` section in Settings

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
This PR aims to improve the mobile responsiveness of the `Hide Videos From Channels` input by adjusting the alignment of the action button and adjusting the whitespace surrounding the section.

Currently, the action icon is absolutely positioned within the relative `ft-input-component`. The `ft-input-component` contains a `label` which is displayed before the `input` element, and this can cause the icon to become misaligned if the label is taller than the `top` position of the icon. This PR aims to fix this by adding a relatively positioned wrapper around just the `input` and action icon. This should ensure that the action icon remains in line with the input even if the height of the label changes due to text wrapping.

Additionally, there is a lot of space between the `Hide Videos From Channel` section and the above section on small displays, so this PR aims to address this by hiding the line break in-between these sections with a media query. It may make more sense to use a margin/padding which changes based on a media query instead of hiding and showing the existing line-break.

## Screenshots 
_(the below screenshots are from a 412px wide window)_
| _before:_ | _after:_ |
| :---: | :---: |
|||

## Testing 
1. `yarn dev`
2. Open the `Settings` view
3. Navigate to `Distraction Free Settings`
4. Reduce the width of the window to something below `576px`
5. Make sure that the `Hide Videos From Channels` section doesn't contain any overflow

_It would also make sense to do a run through of all other `ft-input`s which display an action icon._

## Desktop
 - OS: Windows 10
 - FreeTube version: 0.18.0

## Additional context
This can also be tested in a web build; however, there is currently an issue with the web build which is resolved by FreeTubeApp/FreeTube#2991. I went ahead and tested this PR in the web build by adding the additional fallback for `fs/promises` to the `web` webpack config which is present in that PR, and it seems to work fine. 

I am unsure if the break-point I used is the best one to use. I tried to pick a pretty standard break-point, but I am aware that FreeTube uses a lot of non-standard breakpoints, so it might make more sense for this to use one of the already existing breakpoints in FT.

### DIFF
--- a/src/renderer/components/distraction-settings/distraction-settings.vue
+++ b/src/renderer/components/distraction-settings/distraction-settings.vue
@@ -104,7 +104,7 @@
         />
       </div>
     </div>
-    <br>
+    <br class="hide-on-mobile" />
     <ft-flex-box>
       <ft-input-tags
         :label="$t('Settings.Distraction Free Settings.Hide Channels')"
@@ -119,3 +119,4 @@
 </template>
 
 <script src="./distraction-settings.js" />
+<style src="./distraction-settings.css" />
--- a/src/renderer/components/distraction-settings/distraction-settings.css
+++ b/src/renderer/components/distraction-settings/distraction-settings.css
@@ -1,0 +1,5 @@
+@media only screen and (max-width: 800px) {
+  br.hide-on-mobile {
+    display: none;
+  }
+}
--- a/src/renderer/components/ft-input-tags/ft-input-tags.css
+++ b/src/renderer/components/ft-input-tags/ft-input-tags.css
@@ -48,3 +48,8 @@
   margin-top: 10px;
 }
 
+@media(max-width: 532px) {
+  .ft-input-tags-component {
+    width: 100%;
+  }
+}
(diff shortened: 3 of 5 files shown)


ChunkyProgrammer:
There's a github lint issue btw

src/renderer/components/distraction-settings/distraction-settings.vue
```diff
@@ -104,7 +104,7 @@
         />
       </div>
     </div>
-    <br>
+    <br class="hide-on-mobile" />
```
PikachuEXE, line 107: Lint Warning

PikachuEXE (changes requested):
Tested but the lint warning need to be fixed
image

ChunkyProgrammer (approved):

PikachuEXE (approved):

absidue (approved):

absidue:
The inputWrapper span introduced in this PR is overlayed on top of the clear button in the main search bar, so you can no longer click it.

absidue:
Sorry it's my fault for not testing that specific part of FreeTube while testing this PR.

PikachuEXE:
@absidue https://github.com/FreeTubeApp/FreeTube/pull/3016

PR #3005, opened 2022-12-26
Migrate the trending page to YouTube.js

absidue:
# Migrate the trending page to YouTube.js

## Pull Request Type

- [x] Feature Implementation

## Description
This pull request changes the local API backend of the trending page from @freetube/yt-trending-scraper to YouTube.js.

This removes the need for the @freetube/yt-trending-scraper dependency. Yay! 😄

## Testing 
Check that the different trending tabs work

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/views/Trending/Trending.js
+++ b/src/renderer/views/Trending/Trending.js
@@ -6,8 +6,8 @@ import FtElementList from '../../components/ft-element-list/ft-element-list.vue'
 import FtIconButton from '../../components/ft-icon-button/ft-icon-button.vue'
 import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
 
-import { scrapeTrendingPage } from '@freetube/yt-trending-scraper'
 import { copyToClipboard, showToast } from '../../helpers/utils'
+import { getLocalTrending, parseLocalListVideo } from '../../helpers/api/local'
 
 export default Vue.extend({
   name: 'Trending',
@@ -22,7 +22,8 @@ export default Vue.extend({
     return {
       isLoading: false,
       shownResults: [],
-      currentTab: 'default'
+      currentTab: 'default',
+      trendingInstance: null
     }
   },
   computed: {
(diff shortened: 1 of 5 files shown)


ChunkyProgrammer:
We should probably archive the yt-trending project then so we no longer have to maintain it

src/renderer/views/Trending/Trending.js
```diff
@@ -77,27 +78,35 @@ export default Vue.extend({
       }
     },
 
-    getTrendingInfoLocal: function () {
+    getTrendingInfoLocal: async function () {
```
ChunkyProgrammer, line 81: Should most of this method be moved to the api file as well? I imagine a function in the local api file could return the trending instance and the shown results. This could also help with decoupling local and invidious api from our components which would be useful if in the future we ever want to add support for other services (ex. PeerTube or Piped) or if we decide to switch to a different library in the future.

Let me know your thoughts on this 🙂 It can also be done in a future PR if you agree with this suggestion.

absidue: Done the changes now, most of the logic is now in the api/local.js file. Please let me know if I didn't do it the way you wanted it.

absidue:
Archiving the repo wiht a message in the README sounds good, although we should wait until this is merged and I've removed yt-trending-scraper from the docs.

src/renderer/views/Trending/Trending.js
ChunkyProgrammer: Yeah, the code lgtm. 🙂 Haven't tested it yet though.

PikachuEXE:
Do that after next FT release?

PikachuEXE (approved):
Tested locally

Off topic: The Internet is full of weird things
image

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3019, opened 2023-01-01
Fix no sources error on the watch page

absidue:
# Fix no sources error on the watch page

## Pull Request Type

- [x] Bugfix

## Description
When you open the watch page, an error pops up in the console `No sources available`, but it appears harmless as the video clearly starts playing when it's loaded. Still I thought I would track it down.

The ft-player and other components on the watch page have `v-if="!isLoading"`, which means they are created when `isLoading` is `false` and destroyed when it is `true`. `isLoading` is set to `true` inside the invidious and local API functions, before that it is `false` just long enough for the ft-player component to start doing it's thing, as at that point the data hasn't been fetched there aren't any formats available so the error pops up.

By setting the default value of `isLoading` to `true` we make sure that the ft-player and other components are only created once the data is available and `isLoading` is set to `false`.

## Screenshots 
no-sources-available

## Testing 
Play a video and check that the error doesn't show up anymore

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -43,7 +43,7 @@ export default Vue.extend({
   },
   data: function () {
     return {
-      isLoading: false,
+      isLoading: true,
       firstLoad: true,
       useTheatreMode: false,
       showDashPlayer: true,


ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
I thought that error was expected

PR #3023, opened 2023-01-02
update linters, add stylelint, switch from sass to scss

ChunkyProgrammer:
# Update Linters, Add StyleLint, Swithc from SASS to SCSS

## Pull Request Type
- [x] Development chores

## Description
Changes look a lot bigger than they actually are
#### `eslintrc` file updated to:
-  support json 
- migrate to new babel parser
- add `unicorn` rules to improve code quality

#### `sass` files converted to `scss` notation.
This allows stylelint to work on them & is more compatible with css files. (all css files are valid `scss` files)

#### `package.json` 
- eslint modules updated
- new lint modules added
- new scripts for npm run

#### Everything else
- changes made by running lint-fix
- updated references from `eslint-plugin-node` to `eslint-plugin-n`

## Testing 
- FreeTube can load scss files
- Linting works

## Desktop
- **OS:** Windows
- **OS Version:** 11


src/renderer/components/data-settings/data-settings.js
```diff
@@ -228,7 +228,7 @@
       let count = 0
 
       const ytsubs = youtubeSubscriptions.slice(1).map(yt => {
-        const splitCSVRegex = /(?:,|\n|^)("(?:(?:"")*[^"]*)*"|[^",\n]*|(?:\n|$))/g
+        const splitCSVRegex = /(?:,|\n|^)("(?:(?:"")*[^"]*)*"|[^\n",]*|(?:\n|$))/g
```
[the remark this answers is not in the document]
ChunkyProgrammer: This is not a new alert see: https://github.com/FreeTubeApp/FreeTube/security/code-scanning/44

src/renderer/components/player-settings/player-settings.scss
```diff
@@ -0,0 +1,18 @@
...
+  margin: 0 auto;
+  align-items: center;
+  column-gap: 1rem;
+
+  .screenshotFolderLabel,
+  .screenshotFolderButton,
+  .screenshotFilenamePatternTitle {
+    flex-grow: 0;
+  }
+  .screenshotFolderPath,
```
absidue, line 12: Could you please make the spacing consistent, either with empty lines between all of the rules (easier to read) or remove the empty lines. Currently you've got a mix between both in different and even with the same files.

ChunkyProgrammer: Done

package.json
```diff
@@ -31,8 +31,15 @@
     "dev": "run-s rebuild:electron dev-runner",
     "dev:web": "node _scripts/dev-runner.js --web",
     "dev-runner": "node _scripts/dev-runner.js",
+    "lint-all": "npm run lint & npm run lint-json & npm run lint-style",
     "lint-fix": "eslint --fix --ext .js,.vue ./",
     "lint": "eslint --ext .js,.vue ./",
+    "lint-json": "eslint --ext .json ./",
+    "lint-style": "stylelint \"**/*.scss\" & stylelint \"**/*.css\"",
```
absidue, line 38: Please avoid using npm, everything else uses yarn. Also the behaviour of `&` is different on Windows compared to in shells on Linux and macOS
On Windows in cmd: `&` makes it run the commands one after the other, regardless of whether the previous one failed or not
in other shells: `&` makes it run the previous commands in the background in parallel and the last one in the foreground

As we already have the `npm-run-all` dependency for other scripts in this `package.json` file, the better solution would be to use either `run-s` to run them in series or `run-p` to run them in parallel (`npm-run-all` will also correctly use yarn if the script was invoked with yarn)

PikachuEXE:
Why mixed so many changes in 1 PR...
I hate reviewing big changes -,-...

PikachuEXE (dismissed):
Reviewed except sass/scss files

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #2385, opened 2022-07-07
[Bug]: search `#shorts` in searchbar results in `TypeError: Cannot read properties of undefined (reading 'url')`


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a bug report that matches the one I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.

### Describe the bug

1. Search #shorts
2. see error

### Expected Behavior

no error

### Issue Labels

API issue, text/string issue

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/04c276ca73a62db8c9ecdeddc38b9acbf0c469e9

### Operating System Version

win10 21h2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
looks like @ChunkyProgrammer opened a PR https://github.com/TimeForANinja/node-ytsr/pull/164 to fix that in the ytsr repo a while ago, it might make sense to make FreeTube use the branch in that PR as upstream hasn't merged the PR yet.

efb4f5ff-1298-471a-8973-3d47447115dc:
Noticed that now anything with a # in the search query will give u this error

Try searching 

1. #anything 
2. #ltt

ChunkyProgrammer:
We could switch to my branch for the search module (my pr got closed due to inactivity)

PR #3028, opened 2023-01-03
Migrate search to YouTube.js

absidue:
# Migrate search to YouTube.js

## Pull Request Type

- [x] Feature Implementation
- [x] Other - Refactor

## Description
This PR migrates the search functionality over to YouTube.js, this also fixes an issue with the number of channel subscribers being displayed as the number of videos. I also added support for showing the channel handles in the search results. Searching with search filters should be as fast as searching without them now, ytsr requires additional requests, whereas YouTube.js encodes them as protobuf and sends them along in a single request (like YouTube does).

Side note:
If you see any base64 string being sent to YouTube or being returned in a response, it's likely to be a base64 encoded protobuf, you can paste them into this website and see what is inside of it https://protobuf-decoder.netlify.app/.

## Screenshots 
before:
channel-before
after:
channel-after

## Testing 
Try various search queries with and without filters. Also search for some live and upcoming videos to make sure they are displayed correctly (you can copy video titles from https://www.youtube.com/@live)

## Desktop

- **OS:** Windows
- **OS Version:** 10


ChunkyProgrammer:
🎉 code looks great, i'll try it out later tonight

ChunkyProgrammer:
Haven't tested yet but might be a good idea to make sure https://github.com/FreeTubeApp/FreeTube/issues/2385 isn't an issue  (we can filter out hashtags if we need to at the moment)

absidue:
Searching `#shorts` makes these warnings show up in the console, however search results do show up, so it's probably something to report upstream so the warnings go away:
warnings

_scripts/build.js
```diff
@@ -62,13 +62,6 @@ const config = {
     './dist/**/*',
     '!dist/web/*',
     '!node_modules/**/*',
-
-    // renderer
-    'node_modules/{miniget,ytsr}/**/*',
```
ChunkyProgrammer, line 67: I think miniget is still a ytdl-core dependency, would that cause any issues here?

absidue: That shouldn't be a problem, ytdl-core and it's dependencies get bundled into the `renderer.js` file. Webpack leaves any dependency marked as external untouched or replaces it with your replacement. So with ytsr, the `renderer.js` file still had a `require('ytsr')` call in it, even after bundling, which also means that any dependencies of ytsr will need to be available in the `node_modules` folder at runtime, hence the miniget dependency needing to be listed here while we were using ytsr.
That also meant that we were shipping miniget twice, once inside the `node_modules` and in the `renderer.js` file.

TL;DR ytdl-core will still work as miniget is also bundled into the renderer.js file when webpack bundles ytdl-core.

ChunkyProgrammer (approved):
Lgtm

PikachuEXE (approved):
teste locally

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Could someone open a PR in the docs repo for this?

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc https://github.com/FreeTubeApp/FreeTube-Docs/pull/78

PR #3053, opened 2023-01-12
Decipher format URLs with the local API

absidue:
# Decipher format URLs with the local API

## Pull Request Type

- [x] Bugfix

## Description
So it turns out YouTube.js doesn't automatically decipher the format URLs, not sure how I didn't realise this earlier 🙈.

Anyway this should fix the audio and legacy formats not working sometimes (DASH worked, as YouTube.js's toDash function deciphers them while generating the DASH manifest).

## Testing 
Try a few videos and switch between audio, DASH and legacy formats, to make sure that they all work.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -115,14 +115,42 @@ export async function getLocalSearchContinuation(continuationData) {
 }
 
 export async function getLocalVideoInfo(id, attemptBypass = false) {
+  let info
+  let player
+
   if (attemptBypass) {
     const innertube = await createInnertube({ withPlayer: true, clientType: ClientType.TV_EMBEDDED })
+    player = innertube.actions.session.player
+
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):

PikachuEXE:
Anyone got any example video(s) which breaks before this fix?

absidue:
For most videos the audio formats seemed to be broken at random, e.g. if you played it 4 times, one of those times would be broken.
The audio formats seem to be always broken with this video: https://youtu.be/FWf6L-wKvOA and it fallsback to DASH when it tries to play the audio formats.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3055, opened 2023-01-12
Add a formatNumber helper and other i18n cleanup

absidue:
# Add a formatNumber helper and other i18n cleanup

## Pull Request Type

- [x] Feature Implementation

## Description
Add a `formatNumber` function as `Intl.NumberFormat().format()` is used in quite a few places. I also replaced the i18n import in components with `this.$i18n` as vue-i18n makes itself available through that variable inside components.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-list-channel/ft-list-channel.js
+++ b/src/renderer/components/ft-list-channel/ft-list-channel.js
@@ -1,6 +1,6 @@
 import Vue from 'vue'
-import i18n from '../../i18n/index'
 import { youtubeImageUrlToInvidious } from '../../helpers/api/invidious'
+import { formatNumber } from '../../helpers/utils'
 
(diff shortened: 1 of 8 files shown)


PikachuEXE (commented):
Code looks fine but how to test

absidue:
I'm guessing that number should be displayed the same as before is not an acceptable answer for testing?

absidue:
Okay so one way to test it is the subscription count on watch page, the K or M should change depending on the selected language.

PikachuEXE (approved):
Tested en-US & zh-TW

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3056, opened 2023-01-12
Add an IS_ELECTRON_MAIN define instead of runtime detection

absidue:
# Add an IS_ELECTRON_MAIN define instead of runtime detection

## Pull Request Type

- [x] Other - Refactor

## Description
Instead of detecting whether we are in the main process with a runtime check, we can do it at build time, allowing the other part of the if-else to be optimised out.

## Testing 
Check that the database still gets loaded, so just open FreeTube and see if your settings are loaded.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/webpack.main.config.js
+++ b/_scripts/webpack.main.config.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const webpack = require('webpack')
 const CopyWebpackPlugin = require('copy-webpack-plugin')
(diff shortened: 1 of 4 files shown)


PikachuEXE (approved):
Tested

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3060, opened 2023-01-13
Use ES6 classes for video.js components instead of videojs.extend

absidue:
# Use ES6 classes for video.js components instead of videojs.extend

## Pull Request Type

- [x] Feature Implementation

## Description
Video.js 8 removes the `videojs.extend` function and recommends switching to ES6 classes as they provide the same functionality. Even though we haven't upgraded to video.js 8 just yet (7 is still tagged as latest on npmjs.org, 8 is tagged as next), we can do this change now already, as it doesn't break anything and will make the upgrade easier when we want to upgrade (we will need to check that all plugins work).

## Testing 
* Loop button
* Full window button
* Theatre mode button
* Screenshot button
* DASH quality selector

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -1196,27 +1196,28 @@ export default Vue.extend({
     },
 
     createLoopButton: function () {
+      const toggleVideoLoop = this.toggleVideoLoop
+
       const VjsButton = videojs.getComponent('Button')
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
buttons work

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3064, opened 2023-01-14
Import store modules with ES6 imports instead of dynamically

absidue:
# Import store modules with ES6 imports instead of dynamically

## Pull Request Type

- [x] Feature Implementation

## Description
This changes the way that the store modules are imported. While there was nothing wrong with the previous approach, doing it dynamically also meant that it was done dynamically at runtime too (the for loop and the replaceAll stuck around in the output JavaScript files). ES6 imports are static and therefore statically analysable, allowing webpack to produce a smaller and more efficient output.

The main goal of this change is to reduce the output size and as it doesn't worsen the development experience, I think it's an alright change. This will probably also result in a very minor performance increase, not that anyone will notice it.

renderer.js: -7375 bytes
web.js: -8017 bytes

_Yes, I edited that file that said it shouldn't need to be edited._

## Screenshots 
renderer:
renderer_before renderer_after

web:
web_before web_after

## Testing 
Check that all your settings get loaded correctly.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/store/modules/index.js
+++ b/src/renderer/store/modules/index.js
@@ -1,14 +1,22 @@
 /**
  * The file enables `@/store/index.js` to import all vuex modules
- * in a one-shot manner. There should not be any reason to edit this file.
+ * in a one-shot manner.
  */
 
-const files = require.context('.', false, /\.js$/)
-const modules = {}
+import history from './history'
+import invidious from './invidious'
+import playlists from './playlists'
+import profiles from './profiles'
+import settings from './settings'
+import subscriptions from './subscriptions'
+import utils from './utils'
 
-files.keys().forEach(key => {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
> This will probably also result in a very minor performance increase, not that anyone will notice it.

All minor performance increases u made do add up to a significant performance increase

ChunkyProgrammer (approved):

PikachuEXE (approved):
Here is a Pikachu for FT becoming slightly faster
image

efb4f5ff-1298-471a-8973-3d47447115dc:
U really give me the feeling to rewatch pokemon again

PR #3062, opened 2023-01-14
Fix searching in channel view broken

PikachuEXE:
# Fix searching in channel view broken

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Bug might be introduced by https://github.com/FreeTubeApp/FreeTube/pull/2984

## Description
Searching in channel currently completely broken due to code error

## Screenshots 
Expected to see search results
image

## Testing 
- View random channel
- Search via search input box
- Should see results (not getting result via Invidious API should be a separate issue)

## Desktop

- **OS:** MacOS
- **OS Version:** 12.6.2
- **FreeTube version:** ae50ec72056d2ba56d9ea4188acf7a0fa5cf87f5

PR #3073, opened 2023-01-16
Fix searching (2nd time onward) in channel view broken

PikachuEXE:
# Fix searching (2nd time onward) in channel view broken

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
#3062 only fixes 1st search

## Description
Since there is no search tab
Code being run when already in channel search result view would cause code error (missing node)
This PR fixes that issue

## Screenshots 
Test yourself~

## Testing 
- View random channel
- Search something via search input box
- Search something else via search input box (2nd search)
- Should see results (not getting result via Invidious API should be a separate issue)

## Desktop

- **OS:** MacOS
- **OS Version:** 12.6.2
- **FreeTube version:** 8db8ca2992711285309d83d3df30b9953ae2ebb9

### DIFF
--- a/src/renderer/views/Channel/Channel.js
+++ b/src/renderer/views/Channel/Channel.js
@@ -676,12 +676,13 @@ export default Vue.extend({
         }
       }
 
+      // `currentTabNode` can be `null` on 2nd+ search
       const currentTabNode = document.querySelector('.tabs > .tab[aria-selected="true"]')
       // `newTabNode` can be `null` when `tab` === "search"
       const newTabNode = document.getElementById(`${tab}Tab`)
-      document.querySelector('.tabs > .tab[tabindex="0"]').setAttribute('tabindex', '-1')
+      document.querySelector('.tabs > .tab[tabindex="0"]')?.setAttribute('tabindex', '-1')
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):

absidue (commented):
Instead of the many `?.` would it be better to wrap them in an if that doesn't trigger them for the search tab?

PikachuEXE:
The code is written based on behaviour

For `currentTabNode` it's
`If there is current tab, update stuff, else do nothing`
Same for `newTabNode` and `document.querySelector('.tabs > .tab[tabindex="0"]')` (probably should assign a name for the latter)

In case we have something else (I know we don't) that also have "tabless" behaviour this code would also work without any change

Also `currentTabNode` is absent when **previous** `tab` is `'search'`
i.e. Search first > switching to videos/playlist tab also causes error
In that case code like `tab === 'search'` doesn't even work

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3072, opened 2023-01-15
Migrate comments to YouTube.js

absidue:
# Migrate comments to YouTube.js

## Pull Request Type

- [x] Other - Migration

## Description
Migrated the comments to YouTube.js, this allows us to get rid of the `@freetube/yt-comment-scraper` dependency.

Nearly there 😁

## Testing 
Try the comments under various videos.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.18.0

PR #3075, opened 2023-01-16
Fix comment reply author link style

PikachuEXE:
# Fix comment reply author link style

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Style broken since class name applied on wrapper `span` instead of the link itself

## Screenshots 
After fix
image

Before fix
image

## Testing 
- View a video with comment and reply (e.g. https://youtu.be/FPF7Z7TLdsk)
- Open reply
- Reply author names' style should be same as comment author (and also still work as links)

## Desktop

- **OS:** macOS
- **OS Version:** 12.6.2
- **FreeTube version:** 8db8ca2992711285309d83d3df30b9953ae2ebb9

### DIFF
--- a/src/renderer/components/watch-video-comments/watch-video-comments.vue
+++ b/src/renderer/components/watch-video-comments/watch-video-comments.vue
@@ -168,18 +168,15 @@
               >
             </router-link>
             <p class="commentAuthorWrapper">
-              <span
+              <router-link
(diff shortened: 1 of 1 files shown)


PikachuEXE:
#3072 already fixes this

PR #3097, opened 2023-01-21
Adjust `side-nav-more-options` to account for anchor style

MarmadileManteater:
# Adjust `side-nav-more-options` to account for difference in style between `router-link` and `div`

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Somewhat recently, the `side-nav-more-options` icons were changed to be `router-link`s instead of `div`s, but the style difference between `router-link` and `div` still needs to be accounted for because `router-link`s carry with them all of the styles of anchors. Due to this, the text in the more options menu is the primary link color _(instead of white)_ and the background color is not visible on hover _(due to the display being inline)_. This PR aims to address this by adding some of the styles that `div`s have to the `.navOption` class in the `side-nav-more-options` component.

## Screenshots 

|Before|After|
|--|--|
|before|after|

## Testing 
1. Make sure the window is narrow enough that the `More` navbar item appears: _(width < 678px)_.
2. Click the `More` navbar item.
3. Ensure that the navbar options under `More` have white text and a background hover effect.

## Desktop
 - OS: Windows 10
 - FreeTube version: 0.18.0

### DIFF
--- a/src/renderer/components/side-nav-more-options/side-nav-more-options.css
+++ b/src/renderer/components/side-nav-more-options/side-nav-more-options.css
@@ -6,6 +6,9 @@
   position: relative;
   padding: 5px;
   cursor: pointer;
+  color: white;
+  text-decoration: none;
+  display: block;
 }
 
 .navOption:hover {


PikachuEXE (approved):
image

ChunkyProgrammer (approved):

absidue (approved):

PR #3101, opened 2023-01-21
Set the Vue compiler whitespace option to condense

absidue:
# Set the Vue compiler whitespace option to condense

## Pull Request Type

- [x] Feature Implementation

## Description
In Vue 3 the whitespace option defaults to `condense`. This is a small step towards migrating to Vue 3 that we can already do now. I fixed the only spot I noticed that was broken by that option (the spacing in the ft-list-video component), I'm opening this as a draft, so we can test the rest of FreeTube with this option, to check if we need to make any other changes.

## Testing 
The goal is for FreeTube to be visually the same before and after changing the value of that option, so please let me know if you notice any missing whitespace after this change.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/webpack.renderer.config.js
+++ b/_scripts/webpack.renderer.config.js
@@ -46,6 +46,11 @@ const config = {
       {
         test: /\.vue$/,
         loader: 'vue-loader',
+        options: {
+          compilerOptions: {
+            whitespace: 'condense',
+          }
+        }
       },
       {
         test: /\.scss$/,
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Will be testing it with my [latest custom build](https://github.com/PikachuEXE/FreeTube/actions/runs/3977233751)

PikachuEXE (approved):
Nothing noticed for far

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #3105, opened 2023-01-22
Filter out WebM DASH formats

absidue:
# Filter out WebM DASH formats

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description
video.js only supports MP4 DASH, so when the player catches wind of the WebM DASH formats it breaks down. This pull request filters out the WebM DASH formats. It also adds a setting to choose between using the MP4 AV1 formats if they are available or always use the MP4 H.264 formats.

Only allowing the quality selector to see the MP4 H.264 formats, seems to make it work sometimes, unfortunately switching between qualities with AV1 formats is still broken.

AV1 formats retain more quality than H.264 formats, so they are definitely preferred, unfortunately they are also more demanding on your hardware, especially if you don't have dedicated hardware AV1 decoders.

This also fixes the duplicate qualities in the selector in almost all situations, HDR qualities still show two, they seem identical except the itag and almost double as high bitrate on one format compared to the other. My guess is that they might be different HDR implementations or maybe one is SDR and the other is HDR but they are both labelled as HDR.

LTT video about AV1: https://www.youtube.com/watch?v=WVjtK71qqXU
Tom Scott video about why YouTube's compression can't cope with snow and confetti (even though it's a 6 year old video it has AV1 formats, I wonder if YouTube used it for testing purposes): https://youtu.be/r6Rp-uo6HmI

## Testing 
Animals https://youtu.be/Ac07Qt84WDw
Confetti https://youtu.be/4JLLokZWDqE (this looks a lot better with AV1 compared to H264)

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -495,3 +495,33 @@ export function parseLocalComment(comment, commentThread = undefined) {
     replies: []
   }
 }
+
+/**
+ * video.js only supports MP4 DASH not WebM DASH
+ * so we filter out the WebM DASH formats
+ * @param {Format[]} formats
+ * @param {boolean} allowAv1 Use the AV1 formats if they are available
+ */
+export function filterFormats(formats, allowAv1 = false) {
+  const audioFormats = []
+  const h264Formats = []
+  const av1Formats = []
+
+  for (const format of formats) {
+    const mimeType = format.mime_type
+
+    if (mimeType.startsWith('audio/mp4')) {
+      audioFormats.push(format)
+    } else if (allowAv1 && mimeType.startsWith('video/mp4; codecs="av01')) {
+      av1Formats.push(format)
+    } else if (mimeType.startsWith('video/mp4; codecs="avc')) {
+      h264Formats.push(format)
+    }
+  }
+
+  if (allowAv1 && av1Formats.length > 0) {
+    return [...audioFormats, ...av1Formats]
+  } else {
+    return [...audioFormats, ...h264Formats]
+  }
+}
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -507,7 +507,7 @@ export function filterFormats(formats, allowAv1 = false) {
   const h264Formats = []
   const av1Formats = []
 
-  for (const format of formats) {
+  formats.forEach(format => {
     const mimeType = format.mime_type
 
     if (mimeType.startsWith('audio/mp4')) {
@@ -517,7 +517,7 @@ export function filterFormats(formats, allowAv1 = false) {
     } else if (mimeType.startsWith('video/mp4; codecs="avc')) {
       h264Formats.push(format)
     }
-  }
+  })
 
   if (allowAv1 && av1Formats.length > 0) {
     return [...audioFormats, ...av1Formats]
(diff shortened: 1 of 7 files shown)


KarateCowboy:
> Tom Scott video about why YouTube's compression can't cope with snow and confetti (even though it's a 6 year old video it has AV1 formats, I wonder if YouTube used it for testing purposes): https://youtu.be/r6Rp-uo6HmI

That was very interesting. Also, very concise.

src/renderer/helpers/api/local.js
```diff
@@ -437,3 +437,33 @@ export function mapLocalFormat(format) {
...
+ * so we filter out the WebM DASH formats
+ * @param {Format[]} formats
+ * @param {boolean} allowAv1 Use the AV1 formats if they are available
+ */
+export function filterFormats(formats, allowAv1 = false) {
+  const audioFormats = []
+  const h264Formats = []
+  const av1Formats = []
+
+  for (const format of formats) {
```
PikachuEXE, line 452: Why use `for of` instead of [`forEach`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)?

src/renderer/helpers/api/local.js
```diff
@@ -437,3 +437,33 @@ export function mapLocalFormat(format) {
...
+    const mimeType = format.mime_type
+
+    if (mimeType.startsWith('audio/mp4')) {
+      audioFormats.push(format)
+    } else if (allowAv1 && mimeType.startsWith('video/mp4; codecs="av01')) {
+      av1Formats.push(format)
+    } else if (mimeType.startsWith('video/mp4; codecs="avc')) {
```
PikachuEXE, line 459: How do we get these values? (In case there are new ones in the future)

src/renderer/store/modules/settings.js
```diff
@@ -278,7 +278,8 @@ const state = {
   screenshotFolderPath: '',
   screenshotFilenamePattern: '%Y%M%D-%H%N%S',
   fetchSubscriptionsAutomatically: true,
-  settingsPassword: ''
+  settingsPassword: '',
+  allowDashAv1Formats: false
```
PikachuEXE, line 282: Add trailing commas?
```suggestion
  allowDashAv1Formats: false,
```

ChunkyProgrammer:
https://github.com/LuanRT/YouTube.js/pull/293

"I'm also considering adding an option to filter out unwanted formats from the DASH manifest and download function, which is why this is a WIP."

absidue:
I considered opening a feature request for it but then realised that it's probably a pretty niche to us thing, so I just implemented it in FreeTube, however if he thinks it's going to be useful enough, then we can definitely use it.

We'll just have to hope that the filtering for the toDash function gives us a lot more control than the filtering/options that the download function uses. It's probably fine for most users of the library that just want to download the highest quality stream without touching the formats, but we want want that fine grained control.

Even with the filtering in the toDash function, we'll still need to do our own filtering, it'll just mean that we won't need line 620 in the Watch.js file.

src/renderer/helpers/api/local.js
absidue: `console.log(formats)` then look at the `mime_type` property.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):
lgtm

Issue #1380, opened 2021-06-01
SponsorBlock - Auto-skip as an option


ScriptForAll:
I've started using the SponsorBlock feature for a while and it has been working quite well as it should. However, I found the fact that you can't choose to see an ad without having to disable the feature in the settings quite bothersome.

My suggestion to tackle this issue would be to make the auto-skip an option in the SponsorBlock's settings.  Thus, when chosen, the program would function that same as currently does and when not, the user would manually skips the ad via a "skip this ad" button (in similar fashion of YouTube's ads).

I believe this to be the best option since it covers for the users that don't care the slightest for ads and those that sometimes do.

GilgusMaximus:
A good idea, but as we are not actively supporting SponsorBlock as a team (see [disclaimer](https://docs.freetubeapp.io/usage/sponsorblock/)) , we have to ask someone from the community to implement this.

ScriptForAll:
Okay, I understand. However, I have some comments regarding it:

Since the FreeTube team encourages content creators to use sponsors inside their videos, as it's stated in the aforementioned disclaimer, I reckon that the feature I've requested complies with said position since it's a better implementation that allows to skip ad segments only and only if the user wishes to do rather than doing it automatically. 
That being said, I believe that the team itself could support it if the decision is based on the idea of supporting content creators.

Hope this serves as food for thought and thanks for reading.

GilgusMaximus:
I agree with you. But still, we do not want to work on Sponsorblock. We only merged this because @starsep put in the effort to make it. 
So he might be able to help or provide some guidance if needed.

efb4f5ff-1298-471a-8973-3d47447115dc:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
At the moment u always get forced to skip the sponsor. It would be nice that it doesn't force u to skip but still shows u the color of the sponsorblock segment. So the user has more control over what sponser he/she like to see or not.

**Describe the solution you'd like to see implemented**
Maybe this could done with a toggle in the settings?

**Describe alternatives you've considered**
This comment was made by another user but i cant ping him/her :( 

The way I've seen it done in most other implementations is that for each category, each type of section (like sponsor or self promotion) you have several settings. You can disable it completely, you can show the bar, you can show the bar and skip, and in some implementations you can show the bar and a button to skip manually. Probably the first 3 are what the vast majority of people use.

[...]

Issue #251, opened 2019-04-04
Request: Option to change the shortcut keys


PunnyBoi:
Is it possible to add an option for changing the shortcut keys? I started using FreeTube just a few days back, and I miss the {shift+.} and {shift+,} keys (as in youtube) to change the playback speed. Also, flashing the playback speed on the screen (when it is altered), would be very helpful.

BloodRaven0:
Big yes to this suggestion! Sad it's been more than 4 years since it's been opened.

pseudomonas:
This would be really helpful - or at least having:
→skip forwards
←skip backwards
↑ volume up
↓ volume down 
[spacebar] pause/play
as youtube does, and Netflix, and VLC, and I think a fair number of other players.

HelmicNewciv:
In particular, having the option to bind navigation in the app to keys would be really useful.  I try to avoid using my mouse too much to mitigate pain and I'm able to find vim-like bindings in applications for most things, but there isn't really one for watching YouTube other than navigating their website through qutebrowser or similar.  It would be nice to, by default, have the arrow keys select videos or options, and have the option to configure vim-ish bindings (namely hjkl, shift-H and shift-L being back and forward in most keyboard-driven browsers and extensions, / to search).

It would also help with making gamepad navigation easier, like on the Steam Deck, as it's a lot easier to bind the left stick to act as arrow keys and pick a video than to have to use the trackpads to click something on a small screen or having to let go of the device with one hand to tap something.

nemya9066:
Same request here, just have a menu and allow the user to choose the keybinding that he wants.
Big thanks to anyone already helping on implementing this.

cybergazer:
Really hope this is added soon. This can be done extremely easily if FreeTube is bundled with the Vimium chrome extension, and would work extremely well because it's the same keybindings as vimium in the browser, so the end user won't have to learn anything new if they learn vimium, and it would take almost 0 extra development time since you just bundle the unpacked extension with it.

Walter-o:
I always use `j` and `k` to browse up and down
`J` and `K` to browse up and down quickly
`m` and `,` to go forwards or backwards.

If vimium was added as a drop-in, i could easliy do this as it already supports custom keybinds.

Going the custom route is probably better in the long run but will cause a ton of issues / bugs that have to be fixed.

PietJanHein:
Hotkeys for actions changing profiles and selecting videos would be a great addition to FreeTube.

[...]

PR #3116, opened 2023-01-25
Allows sponsor segments to be watched even if Auto Skip is enabled

lodwhet:
# Allow sponsor segments to be watched even if Auto Skip is enabled

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Partial fix for #1380.

## Description
Currently, when SponsorBlock is enabled with the Auto Skip option turned on, sponsor segments are automatically skipped without letting the user watch them. They have to go into the settings, disable SponsorBlock Auto Skip and then return on the video.

This PR allows the user to watch sponsor segments (even if Auto Skip is enabled) by manually seeking into them (by clicking on the progress bar or with arrow keys). A keyboard shortcut has also been added: 'b' or 'B' (may be changed). When it is pressed, the playback is moved at the beginning of the previous sponsor segment (if there is one), thus allowing the user to watch it completely if it has been auto-skipped.

## Testing
I manually tested the changes in `dev:web` on videos with multiple sponsor segments. I also traced the execution to see if the code works as intended.

## Desktop
- **OS:** Arch Linux
- **OS Version:** Rolling
- **FreeTube version:** development branch

## Additional context
There is an annoying behavior in videojs that can happen while seeking. The corresponding `seeking` event can be called *after* the first `timeupdate` event on the new time. Therefore, sponsor segments can be skipped even before the `seeking` event has been triggered. To deal with this problem, I used a countdown (`skipCountdown`) that only skip sponsor segments after 2 `timeupdate` events. Note that I have never seen 2 `timeupdate` events triggered before the `seeking` event. If it happens sometimes, then we have to increase the countdown value, or find another (better) workaround.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -109,6 +109,12 @@ export default defineComponent({
       showStatsModal: false,
       statsModalEventName: 'updateStats',
       usingTouch: false,
+      // whether or not sponsor segments should be skipped
+      skipSponsors: true,
+      // countdown before actually skipping sponsor segments
+      skipCountdown: 1,
+      // sponsor segments
+      skipSegments: null,
       dataSetup: {
         fluid: true,
         nativeTextTracks: false,
@@ -580,10 +586,16 @@ export default defineComponent({
           if (skipSegments.length === 0) {
             return
           }
+          this.skipSegments = skipSegments
 
           this.player.ready(() => {
             this.player.on('timeupdate', () => {
-              this.skipSponsorBlocks(skipSegments)
+              this.skipSponsorBlocks()
+            })
+
+            this.player.on('seeking', () => {
+              // disabling sponsors auto skipping when the user manually seeks
+              this.skipSponsors = false
             })
 
             skipSegments.forEach(({
@@ -601,24 +613,33 @@ export default defineComponent({
         })
     },
 
-    skipSponsorBlocks(skipSegments) {
+    skipSponsorBlocks() {
       const currentTime = this.player.currentTime()
       const duration = this.player.duration()
       let newTime = null
       let skippedCategory = null
-      skipSegments.forEach(({ category, segment: [startTime, endTime] }) => {
+      this.skipSegments.forEach(({ category, segment: [startTime, endTime] }) => {
         if (startTime <= currentTime && currentTime < endTime) {
           newTime = endTime
           skippedCategory = category
         }
       })
-      if (newTime !== null && Math.abs(duration - currentTime) > 0.500) {
+      if (this.skipSponsors && newTime !== null && Math.abs(duration - currentTime) > 0.500) {
         if (this.sponsorSkips.autoSkip[skippedCategory]) {
-          if (this.sponsorBlockShowSkippedToast) {
-            this.showSkippedSponsorSegmentInformation(skippedCategory)
+          if (this.skipCountdown === 0) {
+            if (this.sponsorBlockShowSkippedToast) {
+              this.showSkippedSponsorSegmentInformation(skippedCategory)
+            }
+            this.player.currentTime(newTime)
+          } else {
+            this.skipCountdown--
           }
-          this.player.currentTime(newTime)
         }
+      }
+      // restoring sponsors skipping default values
+      if (newTime === null && !this.skipSponsors) {
+        this.skipSponsors = true
+        this.skipCountdown = 1
       }
     },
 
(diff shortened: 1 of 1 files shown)


PikachuEXE:
I agree adding `Allow sponsor segments to be watched` but not the keyboard shortcut

lodwhet:
> but not the keyboard shortcut

I agree that 'b' is probably not the best key choice. It can be changed. But why do you want to remove it completely?

PikachuEXE:
That feature description is confusing to me
`the playback is moved at the beginning of the previous sponsor segment (if there is one)`
Looking at the implementation it seems moving to the closest segment in the past
But not just for sponsor type, any type that has an action set
Since the `skipSegments` comes from `sponsorBlockSkipSegments`, and the FT passes categories with action not set as `doNothing`
https://github.com/FreeTubeApp/FreeTube/blob/2cd8f0f2d3e949c46bd5d87eb4ed507a194c8a60/src/renderer/components/ft-video-player/ft-video-player.js#L248-L250

For the solution itself (as to make disabling auto skip easier), what about having a keyboard shortcut to disable Auto Skip for this video only?
Allowing the player to auto skip it **and then** pressing a keyboard shortcut to jump back seems like an odd solution/workaround to me.
Remember there are segments which are quite long, jumping forth and back might make video loading even slower.

Let me know what you think

lodwhet:
Yes, it would probably be better to only jump back to the previous auto-skipped segment. I'm going to fix that.

Disabling Auto Skip for the video would require the user to know in advance whether they want to watch the sponsor segments or not, otherwise they would have to press the keyboard shortcut and then go back to the segment manually. Sometimes we can start watching a video without thinking about sponsors, and when FreeTube auto-skips a sponsor segment, we'd still like to watch this skipped segment. This has happened to me several times, which is why I've implemented this keyboard shortcut.

Obviously I can also add a keyboard shortcut to disable Auto Skip temporarily but perhaps a visual feedback should be added.

PikachuEXE:
> Sometimes we can start watching a video without thinking about sponsors, and when FreeTube auto-skips a sponsor segment, we'd still like to watch this skipped segment.

I need some examples to understand this

lodwhet:
> > Sometimes we can start watching a video without thinking about sponsors, and when FreeTube auto-skips a sponsor segment, we'd still like to watch this skipped segment.
> 
> I need some examples to understand this

For example, you can start watching a video immersively, not wondering if you want to watch sponsor segments or not, or even without thinking about sponsors at all. And when FreeTube skips the first sponsor segment, you notice there's one and you might want to watch it. In this situation, I find the keyboard shortcut relevant. 

But maybe I'm the only one in this situation. If you really don't want it, I can remove the keyboard shortcut.

PikachuEXE:
I am certain I wouldn't want this new keyboard shortcut to be `b`

Existing shortcuts: https://docs.freetubeapp.io/usage/keyboard-shortcuts/

`\` (Backslash)?

ChunkyProgrammer:
Personally, I think that the keyboard shortcut shouldn't be included until https://github.com/FreeTubeApp/FreeTube/issues/251 is implemented

lodwhet:
> I am certain I wouldn't want this new keyboard shortcut to be `b`
> 
> Existing shortcuts: https://docs.freetubeapp.io/usage/keyboard-shortcuts/
> 
> `\` (Backslash)?

Yes I think `Backslash`, `Backspace`, or `Ctrl+left arrow` are better shortcuts.

> Personally, I think that the keyboard shortcut shouldn't be included until #251 is implemented but if the other reviewers agree with the shortcut then i'm okay with it being implemented

OK, should I open a different PR for the keyboard shortcut?

efb4f5ff-1298-471a-8973-3d47447115dc:
> Personally, I think that the keyboard shortcut shouldn't be included until https://github.com/FreeTubeApp/FreeTube/issues/251 is implemented

I agree.

lodwhet:
I just removed the keyboard shortcut.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

ChunkyProgrammer:
Thanks for opening this PR 😄

ApurvR20:
Hi, does this feature still work? While watching on firefox using sponsorblock, I use the enter key to unskip the segment to check whether the segment is worth my time, and enter again if it isn't. Often sponsors are useless and generic, and something I wouldn't ever spend money on. Some implementation like this would be extremely helpful.

absidue:
@ApurvR20 If you are using a nightly build you can seek back into the sponsor and watch it, e.g. withe arrow click of clicking on the seek bar with a mouse.

PR #3117, opened 2023-01-25
Pinia migration

ChunkyProgrammer:
# Pinia migration

## Pull Request Typen
- [x] Development Chore

## Description
Pinia is now the preferred state management for vue and vuex is no longer recommended.

## Desktop
- **OS:** Windows
- **OS Version:** 11


ChunkyProgrammer:
@absidue @PikachuEXE any suggestions on how the settings module should be done?

PikachuEXE:
I got too little pinia experience to even know what to look at...
image

absidue:
Me neither but I can take a look and see if I can come up with something.

Also sidenote:
> The directory for Pinia is generally called stores instead of store. This is to emphasize that Pinia uses multiple stores, instead of a single store in Vuex.

https://pinia.vuejs.org/cookbook/migration-vuex.html#restructuring-modules-to-stores

absidue (commented):
As Pinia allows actions to have whatever arguments they want (as it no longer needs to pass the store in as the first one), it probably makes sense to change away from passing in objects payloads when there are multiple arguments and instead define them as their own arguments, unless of course some of them are optional.

ChunkyProgrammer:
TODO: use subscribe/watcher to update database when changes are made

ChunkyProgrammer:
I think this change is really good.  it adds javascript autocomplete when accessing stores and automatically adds getters/setters. But I don't think now is the right time to implement this and I don't want to be maintaining this PR for a year.

Issue #373, opened 2019-09-05
FreeTube doesn't handle its protocol properly


maricn:
When using https://github.com/FreeTubeApp/freetube-redirect extension or when simply issuing command (on OSX terminal):
```
open 'freetube://https://www.youtube.com/watch?v=whatever'
```
The FreeTube app opens if closed or gets in focus if it's already open, but nothing else happens. Video doesn't load. Since the extension properly passes the URL to the FreeTube app, and since opening the protocol address outside of extension should also pass the URL to the app properly, I conclude the problem is the app doesn't handle received URL.

Originally observed at https://github.com/FreeTubeApp/freetube-redirect/issues/6.
Documentation about protocol handlers in electron: https://github.com/electron/electron/blob/master/docs/api/app.md#appsetasdefaultprotocolclientprotocol-path-args.

PrestonN:
The problem with this is that I simply don't own a Mac to do testing with.  It's something that I'd like to do soon but I need to be in the right place financially before I get one as getting one would be bought solely for FreeTube testing.  This is also why I don't currently provide `.dmg` builds of FreeTube.

This problem doesn't occur on Windows or Linux, so I am unable to reproduce the issue with the hardware that's available to me.  Unfortunately I don't know when I will be able to fix this, but it's something that I'd like to do when I am able to.

maricn:
@PrestonN Thanks for your reply and your effort on the FreeTube in general.
I'm limited with my knowledge and time to contribute, otherwise would look deeper into it.

But knowing this doesn't happen on Linux/Windows and the link/protocol handling works fine there is an indication it might not be related to electron in the trivial way I imagined (misuse of their API).

Anyways, I think it's good to track this issue here, as opposed to `freetube-redirect` because it belongs here.

Mech0z:
I have a similar problem on Win10, if the app is closed, clicking a youtube link in firefox and having the Freetube redirect addon simply opens Freetube, it does not load the specific video, if I then click the link again when the app is open it will play the video correctly

stevenfang1997:
I have the same issue on Manjaro Linux.

PrestonN:
This should be fixed now in the latest dev builds.  Full release should be coming out sometime this week.

fabianski7:
As I understand it this problem happens with the portable or third party versions of freetube, like AUR? 
I keep having the same problem: freetube opens but the video does not load.

I am using the Privacy Redirect extension, as recommended.

Any way to solve this?

[...]

PR #3119, opened 2023-01-26
Add QT Url support for freetube protocol

ChunkyProgrammer:
# Add QT Url support for FreeTube protocol

## Pull Request Type
- [x] Bugfix

## Related issue
https://github.com/libredirect/libredirect/issues/212
https://github.com/FreeTubeApp/FreeTube/issues/373#issuecomment-1116352210

## Description
The way QT parses the url, it removes the : after https as it recognizes "freetube://" as the protocol.
This PR makes it so that "freetube:" can be used as the protocol instead and will readd the missing colon after the 'https'

## Testing 
None, I don't use QT (I use Windows)

## Desktop
- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1026,7 +1026,15 @@ function runApp() {
   }
 
   function baseUrl(arg) {
-    return arg.replace('freetube://', '')
+    let newArg = arg.replace('freetube://', '')
+    // add support for authority free url
+      .replace('freetube:', '')
+
+    // fix for QT Url
+    if (newArg.startsWith('https') && !newArg.charAt(5) !== ':') {
+      newArg = 'https:' + newArg.substring(5)
+    }
+    return newArg
   }
 
   function getLinkUrl(argv) {


src/main/index.js
```diff
@@ -1026,7 +1026,15 @@ function runApp() {
   }
 
   function baseUrl(arg) {
-    return arg.replace('freetube://', '')
+    let newArg = arg.replace('freetube://', '')
+    // add support for authority free url
+      .replace('freetube:', '')
+
+    // fix for QT Url
```
PikachuEXE, line 1033:
```suggestion
    // fix for Qt URL, like `freetube://https//www.youtube.com/watch?v=...`
    // For details see https://github.com/FreeTubeApp/FreeTube/pull/3119
```

PikachuEXE (changes requested):

PikachuEXE (dismissed):

src/main/index.js
```diff
@@ -1026,7 +1026,16 @@ function runApp() {
...
 
   function baseUrl(arg) {
-    return arg.replace('freetube://', '')
+    let newArg = arg.replace('freetube://', '')
+    // add support for authority free url
+      .replace('freetube:', '')
+
+    // fix for Qt URL, like `freetube://https//www.youtube.com/watch?v=...`
+    // For details see https://github.com/FreeTubeApp/FreeTube/pull/3119
+    if (newArg.startsWith('https') && !newArg.charAt(5) !== ':') {
```
PikachuEXE, line 1035: `!newArg.charAt(5) !== ':'`
Is there something wrong here?

ChunkyProgrammer: Good catch

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ssnor:
I have tested this solution and can confirm that it solves the problem.
I'm using openSUSE Tumbleweed with KDE.

ChunkyProgrammer:
> I have tested this solution and can confirm that it solves the problem. I'm using openSUSE Tumbleweed with KDE.

Thanka for testing, I'm glad it fixed the issue for you 🙂

PR #3093, opened 2023-01-18
Shorts tab on channel page

KarateCowboy:
# Add 'Shorts' tab to the Channel page

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Adds the Shorts tab for a channel to the Channel Page. Allows for playing via direct or invidious api

### Before:
before

#### After:

after

## Testing

1) Find a channel which has Shorts and Subscribe to it.
2) View the Channels page, then click on that channel
3) The new "Shorts" tab should be visible. Click it
4) There should be a grid with Shorts in it. 
5) Click any of the shorts. You should be taken to the Watch page and be able to watch the Short

PR #3128, opened 2023-01-29
Support storyboards on very short shorts

absidue:
# Support storyboards on very short shorts

## Pull Request Type

- [x] Bugfix

## Related issue
Issue noticed in #3093 

## Description
For long videos, YouTube returns storyboards with an image for every 10 seconds, as you are unlikely to notice it as the progress bar can only be so wide, so making them more granular would be pointless. However for very short videos like the one in the test case below, which is only 13 seconds long, it's definitely worth having more granular thumbnails, so YouTube returns them in a sub one seconds intervals (for that video 0.14s), so that means that the interval value in the storyboards, which is an integer is 0. The current code relies on that interval always being larger than 0.

This pull request adds support for those 0 interval storyboards by dividing the length of the video by the number of thumbnails on the storyboards. Unfortunately YouTube also only returns very low quality thumbnails for those videos, so we can either leave them being tiny like they are now, or stretch them to be larger like YouTube does, which looks bad too.

## Testing 
https://www.youtube.com/shorts/Jj9PmISOzCQ

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -89,13 +89,18 @@ export function toLocalePublicationString ({ publishText, isLive = false, isUpco
   return i18n.t('Video.Publicationtemplate', { number: strings[0], unit })
 }
 
-export function buildVTTFileLocally(storyboard) {
+export function buildVTTFileLocally(storyboard, videoLengthSeconds) {
   let vttString = 'WEBVTT\n\n'
   // how many images are in one image
   const numberOfSubImagesPerImage = storyboard.columns * storyboard.rows
   // the number of storyboard images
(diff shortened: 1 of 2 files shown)


PikachuEXE (approved):
image

absidue:
@ChunkyProgrammer @efb4f5ff-1298-471a-8973-3d47447115dc please review 🥺

PikachuEXE:
Let's get 3 ~pokemons to drink water~ approvals
FlPNYZMaYAEoGL1

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #3129, opened 2023-01-29
Add support for symlinked database files

### LINKED ISSUE #3114: [Feature Request]: Do not overwrite symlinked db files, or offer a db location setting
### Guidelines

### Problem Description

I saw this issue [here](https://github.com/FreeTubeApp/FreeTube/issues/477). You said you would rather specify the db folder than change the behaviour. Though I do not agree, because:
1-  the files themselves are here, they're just accessible through a symlink
2- I'd need to modify settings instead of having it work "out of the box" when I install my dotfiles

At least a setting would make it manageable.

As of now, I don't see anyway to backup my history the same way I manage all my other apps since only on your app the files are overwritten on each start. Did I miss the setting to change the location, or is it not yet implemented?

PS: Thanks for all your work on the app, I know I'm asking a lot! ;)

### Proposed Solution

Check if the files exists, or if a symlink of the same name exists. If either, append and do not overwrite the db files.

### Alternatives Considered

Alternatively if the previous is too complex to implement, do like you proposed and offer a setting to provide an alternative folder for the db files. 

### Issue Labels

new optional setting

### Additional Information

If not implemented, do you need a hand in implementing the feature? I'm not familiar with your code base but I'm a fullstack dev (and did some low-level work when I was working in electronics), I'm sure I could try my hand at helping for this.

absidue:
# Add support for symlinked database files

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #3114 

## Description
Linux users have asked many a time for the ability to symlink their database files from the location they wish to store them in, to the location that FreeTube expects them to be in. Historically we took the stance that it would need to be supported in `nedb`, the database dependency that we use, however I decided to throw this together and according to the Linux users, this seems to work, so I'm opening this pull request for it.

## Testing 
https://docs.freetubeapp.io/usage/data-location/
**Backup your database files!!!**

1. Close FreeTube
2. Move your database files to another location on the disk
3. Create symlinks at the original location pointing to the new location
4. Open FreeTube and check that your settings, profiles, history, playlists are still there

Now to check that the old behaviour still works
5. Close FreeTube
6. Move your database files back to their proper location
7. Open FreeTube and check that your settings, profiles, history, playlists are still there

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/datastores/index.js
+++ b/src/datastores/index.js
@@ -5,8 +5,18 @@ let dbPath = null
 if (process.env.IS_ELECTRON_MAIN) {
   const { app } = require('electron')
   const { join } = require('path')
+  // this code only runs in the electron main process, so hopefully using sync fs code here should be fine 😬
+  const { existsSync, statSync, realpathSync } = require('fs')
   const userDataPath = app.getPath('userData') // This is based on the user's OS
-  dbPath = (dbName) => join(userDataPath, `${dbName}.db`)
+  dbPath = (dbName) => {
+    let path = join(userDataPath, `${dbName}.db`)
+
+    if (existsSync(path) && statSync(path).isSymbolicLink) {
+      path = realpathSync(path)
+    }
+
+    return path
+  }
 } else {
   dbPath = (dbName) => `${dbName}.db`
 }


PikachuEXE:
How to create symlinks in Windows?
I can provide instructions for MacOS
```
# Setup
cd ~/Library/Application\ Support/Electron
mkdir real_db_folder
mv history.db real_db_folder
ln -s real_db_folder/history.db ./history.db
ls -l
# ... history.db -> real_db_folder/history.db

# Test app here

# Cleanup/Revert (after app closed)
rm ./history.db
mv real_db_folder/history.db ./
rm -r real_db_folder
ls -l
# ... history.db

```

PikachuEXE (approved):
Tested on MacOS 12.6.2

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3167, opened 2023-02-07
Extract chapters from the description if they are missing

absidue:
# Extract chapters from the description if they are missing

## Pull Request Type

- [x] Bugfix

## Description
For some videos YouTube doesn't show any chapter markers on the player progress bar, as those are the ones that we currently extract with the local API that also means we don't show them in FreeTube. This pull request adds a fallback to extracting the chapters from the description if YouTube doesn't provide any, with the same mechanism that we have been using for the Invidious API.

## Testing 
https://youtu.be/zaeUG2k-PbI

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/views/Watch/Watch.vue
+++ b/src/renderer/views/Watch/Watch.vue
@@ -122,7 +122,7 @@
       />
       <watch-video-chapters
         v-if="!hideChapters && !isLoading && videoChapters.length > 0"
-        :compact="backendPreference === 'invidious'"
+        :compact="!videoChapters[0].thumbnail"
         :chapters="videoChapters"
         :current-chapter-index="videoCurrentChapterIndex"
         class="watchVideo"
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer (approved):

src/renderer/views/Watch/Watch.vue
```diff
@@ -122,7 +122,7 @@
       />
       <watch-video-chapters
         v-if="!hideChapters && !isLoading && videoChapters.length > 0"
-        :compact="backendPreference === 'invidious'"
+        :compact="!videoChapters[0].thumbnail"
```
PikachuEXE, line 125: Will `!videoChapters[0].thumbnail` blow up if `videoChapters === []`?

ChunkyProgrammer: The v-if above will prevent that 🙂

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3172, opened 2023-02-07
Changed undefined to null

ArturWagnerBusiness:
# Fixing external player not opening on Subscriptions page

## Pull Request Type
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
none

## Description
There is no edge case for undefined in the function that opens external players in `/src/renderer/store/modules/utils.js:openInExternalPlayer()` so I found where the undefined came from and changed it into a null.
This is because if playlistId is set to undefined it will break as it expects null or string. Undefined makes it believe it is a playlist and ignores the videoId
This was not an issue in the recent version of FreeTube (0.18.0).

## Screenshots 
Screenshot_20230207_162049

## Testing 
After changes I tested:
Opening a video on subscriptions page. (worked)
Opening a video on watch page. (worked)
Openng a playlist on search page. (worked)

## Desktop

- **OS:** Linux
- **OS Version:** 5.15.91-1-MANJARO (64-bit)
- **FreeTube version:** c349d6398598efa9a2e2ca4ba1782492d830f65d (Latest commit)

## Additional context
none

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -313,11 +313,11 @@ export default defineComponent({
       }
 
       // Get playlist ID from history ONLY if option enabled
-      if (!this.showVideoWithLastViewedPlaylist) { return }
-      if (!this.saveVideoHistoryWithLastViewedPlaylist) { return }
+      if (!this.showVideoWithLastViewedPlaylist) { return null }
+      if (!this.saveVideoHistoryWithLastViewedPlaylist) { return null }
       const historyIndex = this.historyIndex
       if (historyIndex === -1) {
-        return undefined
+        return null
       }
 
       return this.historyCache[historyIndex].lastViewedPlaylistId


PikachuEXE:
@ArturWagnerBusiness
The PR title look hard to understand so I use what you put inside the PR description

PikachuEXE:
IMO I think instead of changing one value supplier (there might be more)
Just change the consumer to treat `undefined` same as `null`

I personally usually use `variable == null` (yes two equals) which is the same as `variable === null || variable === undefined`

ArturWagnerBusiness:
I found using `grep -r "openInExternalPlayer" .` (root directory just in case)  and grepping `"this.playlistIdFinal"` that only ft-list-playlist.js and fl-list-video.vue uses the this.playlistIdFinal so others will not be effected.

Also usage of playlistIDFinal seems to be only passed to the externalplayer function OR put in a playlist string after being checked in a `if (this.playlistIdFinal && ...)` which will work whether undefined or null as both are falsy by javascript standard.
Screenshot_20230208_073735

I think this change should not cause any damage but if the "!==" were to be change to "!=" should other properties of the payload in openInExternalPlayer be changed too just in case?
image
 There is few more `item !== null` above this code as well. Which is why I originally decided with the smaller alteration which seemed safer in my opinion.

Although the option of doing the bigger change in openInExternalPlayer could be good for future profing in case new code is added that will somehow pass an undefined.

P.S.: Only other components with openInExternalPlayer are `fl-list-playlist` and `watch-video-info` which both seem to only pass string or null.

ArturWagnerBusiness:
I can do the change in the openInExternalPlayer() and do some testing if that is something you would prefer. I don't know if I should create a different PR or just revert+extend this one.

PikachuEXE:
I would say use `variable == null` for `playlistId` and other conditionals in `/src/renderer/store/modules/utils.js:openInExternalPlayer()`

Unless other devs have some use of `null` & `undefined` separately

ArturWagnerBusiness:
Reverted my first commit and changed the needed expressions in openInExternalPlayer()

Tested opening external player for single video:
- Subscriber page
- Watch page (main video and side videos)
- Search page
- Playlist page
- History page

Tested opening external player for playlists
- Search page

PikachuEXE (approved):

ChunkyProgrammer (approved):

absidue (approved):

PR #3174, opened 2023-02-08
Better solution for hiding the PIP button for the audio formats

absidue:
# Better solution for hiding the PIP button for the audio formats

## Pull Request Type

- [x] Other - Improve existing feature

## Related issue

originally implemented in #2227

## Description
Since the original pull request I've learnt a lot more about FreeTube's code and have found out that the ft-player component is destroyed and recreated when the formats are changed, so I've now changed the implementation so video.js doesn't even add the PIP button to the player if the audio formats are selected.

This gets rid of a setInterval call which was based on a randomly selected timeout, yay :)

## Testing 
Watch a video and change the formats.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -364,6 +364,12 @@ export default defineComponent({
           await this.determineDefaultQualityLegacy()
         }
 
+        if (this.format === 'audio') {
+          const controlBarItems = this.dataSetup.controlBar.children
+          const index = controlBarItems.indexOf('pictureInPictureToggle')
+          controlBarItems.splice(index, 1)
+        }
+
         this.player = videojs(this.$refs.video, {
           html5: {
             preloadTextTracks: false,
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer:
The more timeouts and intervals we remove from our code, the better 🙂

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -364,6 +364,12 @@ export default defineComponent({
           await this.determineDefaultQualityLegacy()
         }
 
+        if (this.format === 'audio') {
+          const controlBarItems = this.dataSetup.controlBar.children
```
PikachuEXE, line 368: Please help this goldfish memory Pikachu
I don't even remember `Array#splice`!
```suggestion
          // hide PIP button for audio format
          const controlBarItems = this.dataSetup.controlBar.children
```

PikachuEXE (changes requested):
image

src/renderer/components/ft-video-player/ft-video-player.js
PikachuEXE: I have tested locally and it works though

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #3183, opened 2023-02-12
Use a player event to trigger the DASH quality selector creation

absidue:
# Use a player event to trigger the DASH quality selector creation

## Pull Request Type

- [x] Other - Refactor

## Description
Yay one less setTimeout.

## Testing 
Change between formats and check that the DASH quality selector is created when the DASH formats are enabled.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -445,15 +445,6 @@ export default defineComponent({
           })
         }
 
-        if (this.useDash) {
-          // this.dataSetup.plugins.httpSourceSelector = {
-          // default: 'auto'
-          // }
-
-          // this.player.httpSourceSelector()
-          this.createDashQualitySelector(this.player.qualityLevels())
-        }
-
         if (this.autoplayVideos) {
           // Calling play() won't happen right away, so a quick timeout will make it function properly.
           setTimeout(() => {
@@ -498,8 +489,10 @@ export default defineComponent({
           this.player.on('wheel', this.mouseScrollSkip)
         }
 
-        this.player.on('fullscreenchange', this.fullscreenOverlay)
-        this.player.on('fullscreenchange', this.toggleFullscreenClass)
+        this.player.on('fullscreenchange', () => {
+          this.fullscreenOverlay()
+          this.toggleFullscreenClass()
+        })
 
         this.player.on('ready', () => {
           this.$emit('ready')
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -511,6 +504,15 @@ export default defineComponent({
         })
 
         this.player.one('loadedmetadata', () => {
+          if (this.useDash) {
+            // this.dataSetup.plugins.httpSourceSelector = {
+            // default: 'auto'
+            // }
+
+            // this.player.httpSourceSelector()
```
PikachuEXE, line 512: ???
Add code comment to explain or remove?

I know it's just moved from old place but if they are not used since commented maybe it should be removed entirely?

PikachuEXE (commented):
Works fine but... (see line comments)

I assume the DASH quality selector is this:
image

src/renderer/components/ft-video-player/ft-video-player.js
absidue: Preston explained not to long ago on in the Matrix chat that the code is there, if we want to switch back to using videojs-http-quality-selector as the quality selector instead of having a custom DASH quality selector inside FreeTube (yes we currently still import the dependency but we don't use it).

I can get rid of it if you really want me to, although I have to strip out a lot more commented out code that is related to it.

PikachuEXE: I am ok with keeping commented code but some code comments should be added
Something like `reserved for switching back to videojs-http-quality-selector if needed`
I think it's healthier to not have to ask every time I see commented code

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -511,6 +504,15 @@ export default defineComponent({
         })
 
         this.player.one('loadedmetadata', () => {
+          if (this.useDash) {
+            // this.dataSetup.plugins.httpSourceSelector = {
```
absidue, line 508:
```suggestion
            // Reserved for switching back to videojs-http-quality-selector if needed
            // this.dataSetup.plugins.httpSourceSelector = {
```

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
lgtm

ChunkyProgrammer (approved):

PikachuEXE (approved):

PR #3199, opened 2023-02-16
Use a computed property instead of a prop to specify the chapters layout

absidue:
# Use a computed property instead of a prop to specify the chapters layout

## Pull Request Type

- [x] Other - Refactor

## Description
As this the layout of the chapters is decided based on data that is also passed into the component, we can move that decision inside the component instead of it being made in the parent component and passed down.

## Testing 
Open a video with chapters on the local and the Invidious API and check that it is compact on the Invidious API and has thumbnails on the local API.

Only compact with Invidious: https://youtu.be/uJvCVw1yONQ
Compact on both APIs: https://youtu.be/zaeUG2k-PbI (see #3167)

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/watch-video-chapters/watch-video-chapters.js
+++ b/src/renderer/components/watch-video-chapters/watch-video-chapters.js
@@ -7,10 +7,6 @@ export default defineComponent({
     'ft-card': FtCard
   },
   props: {
-    compact: {
-      type: Boolean,
-      default: false
-    },
     chapters: {
       type: Array,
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):
Both videos tested

PR #3203, opened 2023-02-18
Setting loudness to default volume in new videos

### LINKED ISSUE #3125: [Bug]: Loudness set to full when unmuting
### Guidelines

### Describe the bug

1. Click on a Video.
2. Video starts unmuted (because I might have muted the last video).
3. I click unmute.
4. The Video continues with full loudness (100%).

### Expected Behavior

Loudness should be set to my standard value, which is 20%.

In general I would recommend to not set the Loudness to 100%, when unmuting.

### Issue Labels

inconsistent behavior

### FreeTube Version

v0.18.0.beta.9d679cd9

### Operating System Version

Manjaro Linux

### Installation Method

AUR

### Primary API used

Local API

-

### Additional Information

_No response_

### Nightly Build

predystopic-dev:
# Title

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/3125
Partially fixes #3125

## Description
When user mutes a video and opens new video, it will open with volume set to default volume set by user in settings. However, if video was not muted, and new video is opened, then volume will remain at it's original value.

## Screenshots 
n/a

## Testing 
Code is self-explanatory and it didn't affect anything else.

## Desktop

- **OS: Windows**
- **OS Version: Windows 11 22H2**
- **FreeTube version: development build**

## Additional context

Instead of starting the new video muted, and settings it to default volume on unmuting, I made volume be set directly to default volume for new videos if previous video was muted by user.  If someone can point me in the right direction, I can attempt to fix this issue as suggested in #3125 but I was unable to find the function responsible for settings the value back to 100% on unmuting, hence tried to partially fix the issue.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -89,6 +89,7 @@ export default defineComponent({
       id: '',
       powerSaveBlocker: null,
       volume: 1,
+      muted: false,
       player: null,
       useDash: false,
       useHls: false,
@@ -315,9 +316,16 @@ export default defineComponent({
   },
   mounted: function () {
     const volume = sessionStorage.getItem('volume')
+    const muted = sessionStorage.getItem('muted')
 
     if (volume !== null) {
       this.volume = volume
+    }
+
+    if (muted !== null) {
+      // as sessionStorage stores string values which are truthy by default so we must check with 'true'
+      // otherwise 'false' will be returned as true as well
+      this.muted = (muted === 'true')
     }
 
     this.dataSetup.playbackRates = this.playbackRates
@@ -400,6 +408,7 @@ export default defineComponent({
         })
 
         this.player.volume(this.volume)
+        this.player.muted(this.muted)
         this.player.playbackRate(this.defaultPlayback)
         this.player.textTrackSettings.setValues(this.defaultCaptionSettings)
         // Remove big play button
@@ -712,10 +721,12 @@ export default defineComponent({
     },
 
     updateVolume: function (_event) {
-      // 0 means muted
       // https://docs.videojs.com/html5#volume
-      const volume = this.player.muted() ? 0 : this.player.volume()
+      const volume = this.player.volume()
+      const muted = this.player.muted()
+
       sessionStorage.setItem('volume', volume)
+      sessionStorage.setItem('muted', muted)
     },
 
     mouseScrollVolume: function (event) {
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -349,6 +349,7 @@ const stateWithSideEffects = {
     defaultValue: 1,
     sideEffectsHandler: (_, value) => {
       sessionStorage.setItem('volume', value)
+      value === 0 ? sessionStorage.setItem('muted', true) : sessionStorage.setItem('muted', false)
     }
   },
 


predystopic-dev:
The issue is fixed, however there was too much trial and error and I have attempted to clean up the unused variables and other bits from code but not sure if I missed something.

absidue:
Do we actually need the default volume? I feel like it might be confusing or unexpected behaviour from a users perspective if the player sometimes uses the volume from the last video and other times uses the default volume.
I was thinking something more along the lines of this, that way both the muted state and the volume get carried over from one video to the next:

```diff
diff --git a/src/renderer/components/ft-video-player/ft-video-player.js b/src/renderer/components/ft-video-player/ft-video-player.js
index 332ceb13..bdaa74cc 100644
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -89,6 +89,7 @@ export default defineComponent({
       id: '',
       powerSaveBlocker: null,
       volume: 1,
+      muted: false,
       player: null,
       useDash: false,
       useHls: false,
@@ -315,11 +316,19 @@ export default defineComponent({
   },
   mounted: function () {
     const volume = sessionStorage.getItem('volume')
+    const muted = sessionStorage.getItem('muted')
 
     if (volume !== null) {
       this.volume = volume
     }
 
+    if (muted !== null) {
+      // sessionStorage only returns strings
+      // any string that isn't empty is considered truthy by the javascript runtime
+      // so we need to explicitly check for a string value of true, as the string "false" would be considered true as well
+      this.muted = muted === 'true'
+    }
+
     this.dataSetup.playbackRates = this.playbackRates
 
     this.createFullWindowButton()
@@ -400,6 +409,7 @@ export default defineComponent({
         })
 
         this.player.volume(this.volume)
+        this.player.muted(this.muted)
         this.player.playbackRate(this.defaultPlayback)
         this.player.textTrackSettings.setValues(this.defaultCaptionSettings)
         // Remove big play button
@@ -712,10 +722,11 @@ export default defineComponent({
     },
 
     updateVolume: function (_event) {
-      // 0 means muted
       // https://docs.videojs.com/html5#volume
-      const volume = this.player.muted() ? 0 : this.player.volume()
+      const volume = this.player.volume()
+      const muted = this.player.muted()
       sessionStorage.setItem('volume', volume)
+      sessionStorage.setItem('muted', muted)
     },
 
     mouseScrollVolume: function (event) {

```

predystopic-dev:
> Do we actually need the default volume? I feel like it might be confusing or unexpected behaviour from a users perspective

Actually, I was going for the fix as suggested in bug report. But yes, now I think this approach would be more suitable. from user's perspective. What should I do now? (Sorry I am a bit new to contributing on github in general), should I make these changes and commit them? or have you done that already in which case I reckon I should close this PR?

ChunkyProgrammer:
> Actually, I was going for the fix as suggested in bug report. But yes, now I think this approach would be more suitable. from user's perspective. What should I do now? (Sorry I am a bit new to contributing on github in general), should I make these changes and commit them? or have you done that already in which case I reckon I should close this PR?

You should make these changes and commit them 🙂

predystopic-dev:
Sorry I made a mistake and ran some wrong commands.

PikachuEXE:
Since this is declared to partially fixes #3125  
I have updated the description to avoid closing #3125 automatically

PikachuEXE (changes requested):
Default volume set to 70%

A) Not muted
- View a video, set volume to ~10%
- View another video, volume remains at ~10% (OK)

B) Muted
- View a video, set volume to ~10%, set as muted
- View another video, volume remains muted (OK)
- Unmute, volume jumps to 100% instead of default volume 70% (Not OK)

predystopic-dev:
> * View a video, set volume to ~10%, set as muted
> 
> * View another video, volume remains muted (OK)
> 
> * Unmute, volume jumps to 100% instead of default volume 70% (Not OK)

How are you muting it? If we directly click on mute icon or press 'm' to mute/unmute then volume should be set at previously muted volume however if we drag the volume slider to 0 then volume in sessionStorage is set to 0 and if you unmute now, it will set volume to 100%

PikachuEXE:
I only mute it by pressing the mute icon (not even 'm')
I use 'm' this time and it's the same result

predystopic-dev:
> I only mute it by pressing the mute icon (not even 'm') I use 'm' this time and it's the same result
I rechecked and it seems to be working fine? 

Attaching a file to show how it's working.
https://user-images.githubusercontent.com/92176809/220027014-738a0f32-66fd-46b7-8c63-10565eba206e.mp4

PikachuEXE:
@predystopic-dev I updated your comment so the video can be played inline
There must be a blank line before video URL to enable inline replaying

PikachuEXE:
Turns out I haven't checkout your branch yet ._.
image

PikachuEXE:
Default volume set to 70%, all in the same window

A) Not muted
- View a video, set volume to ~10%
- View another video, volume remains at ~10% (OK)

B) Muted
- View a video, set volume to ~10%, set as muted
- View another video, volume remains muted (OK)
- Unmute, volume jumps to ~10% instead of default volume 70% (Is this expected????) 👈

predystopic-dev:
> @predystopic-dev I updated your comment so the video can be played inline There must be a blank line before video URL to enable inline replaying

Oh thanks, I didn't know that was possible with videos. I'll keep that in mind from now.

absidue:
@PikachuEXE It copies the volume and muted states from the previous video, so if you unmute on a new video it goes to the volume that the previous video was at.

predystopic-dev:
> Default volume set to 70%, all in the same window
> 
> A) Not muted
> 
>     * View a video, set volume to ~10%
> 
>     * View another video, volume remains at ~10% (OK)
> 
> 
> B) Muted
> 
>     * View a video, set volume to ~10%, set as muted
> 
>     * View another video, volume remains muted (OK)
> 
>     * Unmute, volume jumps to ~10% instead of default volume 70% (Is this expected????) 👈

Yes, my earlier fix was setting volume to defaultVolume after unmuting but @absidue suggested in a comment above that it's rather un-intuitive from user's perspective, and suggested the new fix which will keep track of previous volume and muted condition separately and set them accordingly on player. However this fix didn't consider the possibility where user can drag the slider to 0 to mute which will set the volume at 100% on unmuting so I have fixed that by keeping track of defaultVolume separately and using that in this condition. 

P.S: I think this fixes the issue completely and not partially which I have edited in original PR description.

PikachuEXE:
OK I guess you mean this

Default volume set to 70%, all in the same window

A) Not muted
- View a video, set volume to ~10%
- View another video, volume remains at ~10% (OK)

B) Muted via mute button
- View a video, set volume to ~10%, set as muted via mute button (or keyboard shortcut)
- View another video, volume remains muted (OK)
- Unmute, volume jumps to ~10% (OK)

C) Muted via dragging volume slider to 0
- View a video, set volume to ~10%, set as muted via dragging volume slider to 0
- View another video, volume remains muted (OK)
- Unmute, volume jumps to 70% (OK)

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -712,10 +726,18 @@ export default defineComponent({
     },
 
     updateVolume: function (_event) {
-      // 0 means muted
       // https://docs.videojs.com/html5#volume
-      const volume = this.player.muted() ? 0 : this.player.volume()
-      sessionStorage.setItem('volume', volume)
+      if (sessionStorage.getItem('muted') === 'false' && this.player.volume() === 0) {
+        const volume = parseFloat(sessionStorage.getItem('defaultVolume'))
```
PikachuEXE, line 731:
```suggestion
        // Probably was muted via dragging volume slider to 0
        const volume = parseFloat(sessionStorage.getItem('defaultVolume'))
```

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -712,10 +726,18 @@ export default defineComponent({
...
+        const muted = true
+        sessionStorage.setItem('volume', volume)
+        sessionStorage.setItem('muted', muted)
+      } else {
+        const volume = this.player.volume()
```
PikachuEXE, line 736:
```suggestion
        // Probably was muted via mute button/keyboard shortcut
        const volume = this.player.volume()
```

PikachuEXE (changes requested):
Need some code comments for my goldfish memory :P 
image

PikachuEXE (dismissed):

toby63:
> OK I guess you mean this
> 
> Default volume set to 70%, all in the same window
> 
> A) Not muted
> 
>     * View a video, set volume to ~10%
> 
>     * View another video, volume remains at ~10% (OK)
> 
> 
> B) Muted via mute button
> 
>     * View a video, set volume to ~10%, set as muted via mute button (or keyboard shortcut)
> 
>     * View another video, volume remains muted (OK)
> 
>     * Unmute, volume jumps to ~10% (OK)
> 
> 
> C) Muted via dragging volume slider to 0
> 
>     * View a video, set volume to ~10%, set as muted via dragging volume slider to 0
> 
>     * View another video, volume remains muted (OK)
> 
>     * Unmute, volume jumps to 70% (OK)

Bug reporter here, I would like to add some thoughts about this.

So I don't really know whether it's such a good idea to use the value from the last played video.
But this is not a problem unless value>defaultvalue.

So maybe you could add something like:
if value>defaultvalue, use defaultvalue instead.

The practical example here is:
1. Video is very quiet, so you set the volume value higher.
2. You play another video and mute that.
3. You unmute and now the really high value is used.

Or did I misunderstand something?

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -315,11 +316,23 @@ export default defineComponent({
...
 
+    if (muted !== null) {
+      // as sessionStorage stores string values which are truthy by default so we must check with 'true'
+      // otherwise 'false' will be returned as true as well
+      this.muted = (muted === 'true')
+    }
+
+    if (muted === 'false' && volume === '0') {
+      this.muted = true
+      sessionStorage.setItem('volume', 0.25)
```
absidue, line 333: Why is there a hardcoded value of 0.25 here?

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -712,10 +726,21 @@ export default defineComponent({
     },
 
     updateVolume: function (_event) {
-      // 0 means muted
       // https://docs.videojs.com/html5#volume
-      const volume = this.player.muted() ? 0 : this.player.volume()
-      sessionStorage.setItem('volume', volume)
+      if (sessionStorage.getItem('muted') === 'false' && this.player.volume() === 0) {
```
absidue, line 730: Why are you checking the session storage here, instead of the player?

absidue (approved):

absidue (commented):
Didn't mean to approve before

src/renderer/components/ft-video-player/ft-video-player.js
predystopic-dev: Oh frick, i was testing smth didn't remove it. Sorry!!

src/renderer/components/ft-video-player/ft-video-player.js
predystopic-dev: sessionStorage is not set to true when we drag the slider to 0 (it does get set to true if we press 'm' or click on mute though) so in those cases I am setting volume to defaultVolume set by user in settings. This should happen when updateVolume function is called by volumechange event I reckon? Because if s2these conditions are checked in player, then the feature doesn't work as intended

PikachuEXE:
@toby63 Problem is there can be a use case of listening to a list of videos known to need larger volume value
Having to adjust volume every time (due to volume set higher than default volume) is quite annoying (if only using previous volume when <= default volume)

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -712,10 +721,21 @@ export default defineComponent({
...
+        // If video was muted by dragging volume slider, we reset the volume to default volume.
+        // dragging volume to 0 doesn't change muted in sessionStorage to true hence checking it with 'false'
+        const volume = parseFloat(sessionStorage.getItem('defaultVolume'))
+        const muted = true
+        sessionStorage.setItem('volume', volume)
+        sessionStorage.setItem('muted', muted)
+      } else {
+        // If video was muted by pressing 'm'/'M' or clicking on mute button, then we reset the volume to previous volume.
```
absidue, line 733: This comment should probably be removed or changed to be accurate.
the `volumechange` event is fired whenever the volume or the muted state of the player fires, this comment makes it seem like this bit of code is only run when you mute the player, which is not true
it will run in situations like this one below as well
before:
muted: false
volume: 0.25
_volumechange event fires_
muted: false
volume: 0.3

absidue:
@toby63 navigating from an unmuted video to another unmuted video, should keep the volume of the previous video, that's how it behaved before this pull request and that behaviour should really be kept, changing that now is going to annoy more people than it helps.

src/renderer/components/ft-video-player/ft-video-player.js
absidue: Video.js doesn't change anything in session storage. the values in session storage are set once at startup by the code in the settings file and then updated by the code in this function every time the player is muted/unmuted or the volume of the player changes. Without this updateVolume function nothing would be changed in session storage after setting the initial values at startup in the settings file.

toby63:
> @toby63 Problem is there can be a use case of listening to a list of videos known to need larger volume value Having to adjust volume every time (due to volume set higher than default volume) is quite annoying (if only using previous volume when <= default volume)

Just in theory (see also below): Could this not be solved by identifying list videos?

> @toby63 navigating from an unmuted video to another unmuted video, should keep the volume of the previous video, that's how it behaved before this pull request and that behaviour should really be kept, changing that now is going to annoy more people than it helps.

Ok, I see that this is a situation where a design decision has to be made.
But what about an option in the settings?

ChunkyProgrammer:
Something like this https://github.com/iv-org/invidious/issues/3600 could potentially be added but imo it's a different feature request and will probably solve the issues

src/renderer/components/ft-video-player/ft-video-player.js
predystopic-dev: Are there any changes that I should do here? I tried checking sessionStorage in player and update player volume accordingly but it doesn't work as intended.

predystopic-dev:
Are there any issues with commit that need be fixed now?

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Im a bit confused by reading all the comments. Could u explain to me what the expected behavior is with a few example scenarios

predystopic-dev:
> Could u explain to me what the expected behavior is with a few example scenarios

 Assume default volume set to 80% in user settings.
 
 1.  Not muted
 ```
- View a video, set volume to 20%

- View another video -> volume remains at 20%
```
2. Muted via mute button
```
- View a video, set volume to 20%. 

- Mute by clicking on mute button or pressing appropriate key on keyboard.

- View another video, volume remains muted 

- On unmuting volume jumps to 20% 
```

3. Muting via dragging volume slider to 0.
```
- View a video at any volume, but drag the slider to 0.

- View another video -> volume remains muted

- On unmuting, volume jumps to 80% (user default)  
```

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc ?

ChunkyProgrammer (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Uumm is there a reason for not closing #3125

PR #3213, opened 2023-02-22
Remove beforeRouteLeave in ft-video-player as it is never called

absidue:
# Remove beforeRouteLeave in ft-video-player as it is never called

## Pull Request Type

- [x] Bugfix

## Description
Vue Router only calls the navigation guards in the components that are passed to the router configuation, it doesn't call them on components inside of those components (docs: https://v3.router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards). That means that the navigation guard inside of the ft-video-player component was never being called, it also means that if you navigated away from the watch page to a different page e.g. watch page to channel page, the keyboard event handler wouldn't get removed.

## Testing 
Try watching a few videos and see if the keyboard shortcuts still work. (testing this is kind of unnecessary but I know that if I don't add a test case you'll make up your own ones anyway).

## Desktop

- **OS:** Windows
- **OS Version:** 10


ChunkyProgrammer (approved):

PikachuEXE (dismissed):
KB shortcuts seems fine after watching videos, switch different view and back

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #3214, opened 2023-02-22
Better idle mouse hiding implementation

### LINKED ISSUE #699: Mouse not dissapearing
**Behavior of the program**
When u go out of full screen mode and go back in full screen mode while the video is playing the Mouse is not dissapearing.

**Expected behavior**
Mouse pointer should dissapear after a few seconds just like the mediabar.

**To Reproduce**
1. Watch a video in fullscreen
2. While the video still keeps playing click on 'Exit fullscreen icon'
3. Click on 'fullscreen button'
4. Mediabar dissapears but the mouse pointer not!

**Screenshots**
None

 - OS: [Windows 10]
 - OS Version: [1909]
 - FreeTube version: [v0.9.0 Beta]

**Additional context**
None

absidue:
# Better idle mouse hiding implementation

## Pull Request Type

- [x] Bugfix

## Related issue
closes #699

## Description
When the mouse is over the player we want to hide it after a certain amount of time so that it isn't distracting for the user when watching the video. The current implementation is based on mouse movements, so it won't hide the mouse in situations where the mouse didn't start out over the player but is over the player later, like when you full screen the video or move the window around with keyboard shortcuts.

This new approach relies on video.js's own idle detection, which not only fires off two events `useractive` and `userinactive` but also helpfully adds respective CSS classes to the player. We can therefore take advantage of the `vjs-user-inactive` class to hide the cursor with CSS. This also means we don't need to essentially recreate video.js's idle detection inside the FreeTube code.

Also did I mention that this gets rid of a timeout 😊.

## Testing 
(shamelessly copied the reproduction steps from the issue 🙈)

1. Watch a video in fullscreen
2. While the video still keeps playing click on 'Exit fullscreen icon'
3. Click on 'fullscreen button' and don't move the mouse after.
4. Mediabar dissapears but the mouse pointer not, but Mouse pointer does dissapear when it registers a movement.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.css
+++ b/src/renderer/components/ft-video-player/ft-video-player.css
@@ -6,6 +6,10 @@
   width:100%;
 }
 
+.ftVideoPlayer.vjs-user-inactive {
+  cursor: none;
+}
+
 :deep(.sponsorBlockMarker), :deep(.chapterMarker) {
   position: absolute;
   opacity: 0.6;
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -28,7 +28,6 @@ export default defineComponent({
     if (this.player !== null && !this.player.isInPictureInPicture()) {
       this.player.dispose()
       this.player = null
-      clearTimeout(this.mouseTimeout)
     } else if (this.player.isInPictureInPicture()) {
       this.player.play()
     }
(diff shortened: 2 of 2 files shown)


ChunkyProgrammer (approved):

PikachuEXE (approved):
I think the **expected** behaviour is
> Mediabar dissapears and also the mouse pointer, on mouse movement both appears

Since you copied from issue (4) is the unexpected behaviour, which is confusing to be put in "Testing" section in this PR

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Yay it works 🎉

PR #3224, opened 2023-02-26
Differentiate links in readme.md

efb4f5ff-1298-471a-8973-3d47447115dc:
# Differentiate links in readme.md

## Pull Request Type

- [x] Documentation

## Description

U could get confused when looking at it and think it is one link or maybe a mistake or something like that. So changed it a bit. Not sure if the word `or` is suitable here. Lemme know if something else is better for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you have issues with the extension working with FreeTube, please create an is
 
 * [FreeTube Website](https://freetubeapp.io/#download)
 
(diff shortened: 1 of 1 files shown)


absidue:
Maybe "and" would be better?

absidue (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #3234, opened 2023-02-28
Fix video throttling because of range header

absidue:
# Fix video throttling because of range header

## Pull Request Type

- [x] Bugfix

## Related issue
https://github.com/yt-dlp/yt-dlp/issues/6369

## Description
YouTube now throttles the DASH formats if you use the [HTTP Range header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range), the Range header is the standardised way of telling a server you only want part of a file. For a while YouTube has been using a `range` query parameter instead of the standard header, however now they seem to mandate it for the DASH formats, if you don't want your streams to be throttled.

## Testing 
It looks like they haven't rolled the throttling out for everyone on every video yet.
So you can try multiple videos and also make sure that no new errors show up because of this change.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -18,6 +18,16 @@ import { calculateColorLuminance, colors } from '../../helpers/colors'
 import { pathExists } from '../../helpers/filesystem'
 import { getPicturesPath, showSaveDialog, showToast } from '../../helpers/utils'
 
+// YouTube now throttles if you use the `Range` header for the DASH formats, instead of the range query parameter
+// videojs-http-streaming calls this hook everytime it makes a request,
+// so we can use it to convert the Range header into the range query parameter for the streaming URLs
+videojs.Vhs.xhr.beforeRequest = (options) => {
+  if (options.headers?.Range && new URL(options.uri).hostname.endsWith('.googlevideo.com')) {
+    options.uri += `&range=${options.headers.Range.replace('bytes=', '')}`
+    delete options.headers.Range
+  }
+}
+
 export default defineComponent({
   name: 'FtVideoPlayer',
   props: {


ChunkyProgrammer:
Is the string always "bytes="? Asking because https://github.com/unixfox/invidious-custom/blob/master/patches/007-range-header.patch is slightly different

absidue:
in my testing it was, but i'll switch to the split way, as that seems more robust

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -18,6 +18,16 @@ import { calculateColorLuminance, colors } from '../../helpers/colors'
 import { pathExists } from '../../helpers/filesystem'
 import { getPicturesPath, showSaveDialog, showToast } from '../../helpers/utils'
 
+// YouTube now throttles if you use the `Range` header for the DASH formats, instead of the range query parameter
+// videojs-http-streaming calls this hook everytime it makes a request,
+// so we can use it to convert the Range header into the range query parameter for the streaming URLs
+videojs.Vhs.xhr.beforeRequest = (options) => {
+  if (options.headers?.Range && new URL(options.uri).hostname.endsWith('.googlevideo.com')) {
```
ChunkyProgrammer, line 25: I'm guessing we check if it's .googlevideo.com because we dont need this if the video is proxied through invidious?

absidue: The Invidious proxy accepts both, but by the looks of it, it only sends YouTube the Range header, so they should probably fix that. https://github.com/iv-org/invidious/blob/master/src/invidious/routes/video_playback.cr

absidue: Actually looks like the Invidious proxy dies after a few requests and just sends HTTP 416 status codes if you use the range query parameter.

absidue: invidious-proxy-error

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
DASH format video loading seems faster now

PR #3271, opened 2023-03-06
Use smaller Vue esm runtime build

absidue:
# Use smaller Vue esm runtime build

## Pull Request Type

- [x] Other - Build optimisation

## Description
As we use vue-loader our {component}.vue templates get compiled at build time, which means we don't need to compile them at runtime. Currently we are using a the full Vue build, which contains both the runtime and compiler, switching to the runtime only build allows us to shave off a chunk of our bundle (0.04MB for the Electron build and 0.03MB for the web build).

https://v2.vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only

## Screenshots 
renderer:
renderer-before renderer-after

web:
web-before web-after

## Testing 
Test various functionality to check that it still works.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-community-post/ft-community-post.js
+++ b/src/renderer/components/ft-community-post/ft-community-post.js
@@ -1,18 +1,16 @@
-import Vue from 'vue'
+import { defineComponent } from 'vue'
 import FtListVideo from '../ft-list-video/ft-list-video.vue'
 import FtListPlaylist from '../ft-list-playlist/ft-list-playlist.vue'
 
 import autolinker from 'autolinker'
 import VueTinySlider from 'vue-tiny-slider'
 
-import {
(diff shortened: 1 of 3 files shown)


src/renderer/components/ft-community-post/ft-community-post.js
```diff
@@ -1,18 +1,16 @@
-import Vue from 'vue'
+import { defineComponent } from 'vue'
```
PikachuEXE, line 1: Sneaky fixes!

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #1533, opened 2021-07-20
DASH format does not load


tiany4:
**Behavior of the program**
DASH format doesn't work in any video I've tried.

**Expected behavior**
DASH format should load and quality higher than 720p should be available.

**To Reproduce**
Please add all steps to reproduce the behavior:
1. Go to any video
2. Choose DASH format
3. The video reloads and immediately reloads again back to legacy

**Screenshots**
![image](https://user-images.githubusercontent.com/14116340/126347985-653d656f-cc20-4c45-b1e0-817f30322a6c.png)

**Environment Information (please complete the following information):**
 - OS: Windows
 - OS Version: Windows 10, latest update
 - FreeTube version: 0.13.2 Beta
 - Installation Method (If applicable): Portable
 - Primary API Used: Local API but have tried both

efb4f5ff-1298-471a-8973-3d47447115dc:
hmm i tried to reproduce this but i cant.

GilgusMaximus:
The error in the screenshot is from the Invidious API.

There was an error with dev mode not being able to play dash due to some weird bug after an update. The fix is already in the nightlies, so you could try the newest nightly version of the portable and check whether that fixed the problem

tiany4:
> 
> 
> The error in the screenshot is from the Invidious API.
> 
> There was an error with dev mode not being able to play dash due to some weird bug after an update. The fix is already in the nightlies, so you could try the newest nightly version of the portable and check whether that fixed the problem

So I have tried the nightly build `v0.13.2-nightly-1005 Beta`. I am able to play dash. The quality tends to drop somewhat often, and it seems like videos do not go over 720p. Is that the correct behavior?

GilgusMaximus:
Please see #595. These are all related to the same issue with the VideoJS plugin.

The 720p issue is not related tho. Videos, if available, should always go to 1080p with dash, some even > 1080p

tiany4:
I messed around some more. I experience the same issue as the top screenshot if I choose https://yewtu.be as the instance, but https://invidious.snopyta.org does not have the same issue.

GilgusMaximus:
No please don't. We are very well aware of it but no solution attempt from our side has yet worked as the plugin handling this or VideoJS can still do their own stuff under the hood.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @tiany4 can u try a nightly build and tell me if u still have this problem?

Issue #595, opened 2020-10-05
[Bug] Default quality setting does not work


ThatIsAPseudo:
**Behavior of the program**
My default quality setting is set to 240p with Dash player, but videos keep loading in 720/1080p.

**Expected behavior**
The videos play in 240p

**To Reproduce**
1. Go to 'Settings'
2. Select '240p' in the 'Default Quality' dropdown list
3. Play any video
4. Notice that quality is 1080p

**Desktop :**
 - OS: macOS 
 - OS Version: Big Sur Beta 9 (20A5384c)
 - FreeTube version: 0.8.0

arrowgent:
v 0.8.0

linux x64

720p selected as default
same issue, video will select 1080p automatically

efb4f5ff-1298-471a-8973-3d47447115dc:
It still persists in the current version of FreeTube.
Win 10
1909

esotericDisciple:
It's the opposite for me...it feels like the player is locked on 'auto' mode

sysdbugfactory:
Same issue here, default is ignored and freetube plays video in the highest possible setting making the computer hot, slow and the video stuttering.

Using v0.11.2 Beta under manjaro linux

ThatIsAPseudo:
Note : With Legacy player, video plays by default at 360p, not the highest value that is 720p

LosEagle:
> It's the opposite for me...it feels like the player is locked on 'auto' mode

Same here. I'd prefer some buffering over player randomly switching to 240p all the time to present me with unwatchable blurry video no matter how many times I click HD resolutions.

tugcanolgun:
> It's the opposite for me...it feels like the player is locked on 'auto' mode

This is a problem for me too. Sometimes the video switches to 240p no matter what I do. It makes some videos unwatchable.

spectrumgamer75:
Same issue here too, it goes way up to 1080p even though have set 360p in settings. Any fix?

EvilGremlin:
Same, it's doing auto stuff despite set 1080, nor setting quality in player really works. I expect immediate change of format when i say so.

nils-van-zuijlen:
Same, videos playing as in auto (144p for me) in Dash formats, seems to work in Legacy formats (I get higher resolutions, and buffering)

Linux, AUR package
FreeTube 0.12

PrestonN:
Hey all. I know that this has been problematic for a while. I'd like to try to explain a little bit about what's going on.

[...]

PR #3278, opened 2023-03-09
Fix DASH default quality and quality selection

absidue:
# Fix DASH default quality and quality selection

## Pull Request Type

- [x] Bugfix

## Related issue
potentially addresses #595 (it fixes the original issue but someone left comments requesting automatically switching to legacy when certain qualities aren't available in DASH, which this pull request most certainly doesn't do)
potentially addresses #1533 (the quality dropping part)

## Description
Fixes the default DASH quality selection and the DASH quality selector so that it should now be possible to switch between qualities in the quality selector when DASH is enabled. Fixing the default quality selection also fixes the slow loading times.

This also fixes the quality selector showing 1080p, even if you selected a different default quality.

Why wasn't the default quality selection working for DASH?
Well as it turns out even if you disable qualities videojs-http-streaming, it ignores your preferences if it thinks that it doesn't have enough bandwidth to play them (it's wrong but we can override it's decision as we know better). On my machine with the default quality set to 1080p it first requested 720p, then 144p and finally it switched up to 1080p, so the slow loading was caused by videojs-http-streaming wasting time doing pointless stuff.
The solution is to disable them as soon as possible and then also override videojs-http-streaming's default bandwidth with the one of the format, that way videojs-http-streaming will always select the correct quality at the start, as we've told it that it has enough bandwidth to play it.

As for why the quality selector works now, no idea it just started working while I was making all of these changes, so I'll take the win.

Loading and quality switching still isn't as fast as it is on YouTube but that's something for another day, if it's even possible to do with video.js.

## Testing 
Try various default qualites and switch between the qualities in the DASH quality selector (reminder for PikachuEXE as you asked last time: that's the quality selector that is shown when the DASH formats are selected).

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -28,6 +28,10 @@ videojs.Vhs.xhr.beforeRequest = (options) => {
   }
 }
 
+// videojs-http-streaming spits out a warning every time you access videojs.Vhs.BANDWIDTH_VARIANCE
+// so we'll get the value once here, to stop it spamming the console
+const VHS_BANDWIDTH_VARIANCE = videojs.Vhs.BANDWIDTH_VARIANCE
+
 export default defineComponent({
   name: 'FtVideoPlayer',
   props: {
@@ -320,6 +324,10 @@ export default defineComponent({
 
     this.dataSetup.playbackRates = this.playbackRates
 
+    if (this.format === 'dash') {
+      this.determineDefaultQualityDash()
+    }
+
     this.createFullWindowButton()
     this.createLoopButton()
     this.createToggleTheatreModeButton()
@@ -369,6 +377,15 @@ export default defineComponent({
           controlBarItems.splice(index, 1)
         }
 
+        // regardless of what DASH qualities you enable or disable in the qualitLevels plugin
+        // the first segments videojs-http-streamming request are chosen based on the available bandwidth, which is set to 0.5MB/s by default
+        // overriding that to be the same as the quality we requested, makes videojs-http-streamming pick the correct quality
+        const playerBandwidthOption = {}
+
+        if (this.useDash && this.defaultQuality !== 'auto') {
+          playerBandwidthOption.bandwidth = this.selectedBitrate * VHS_BANDWIDTH_VARIANCE + 1
+        }
+
         this.player = videojs(this.$refs.video, {
           html5: {
             preloadTextTracks: false,
@@ -376,7 +393,8 @@ export default defineComponent({
               limitRenditionByPlayerDimensions: false,
               smoothQualityChange: false,
               allowSeeksWithinUnsafeLiveWindow: true,
-              handlePartialData: true
+              handlePartialData: true,
+              ...playerBandwidthOption
             }
           }
         })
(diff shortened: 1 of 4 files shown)


PikachuEXE:
This seems a bit slow...
Local API no VPN
image

I changed to `this.selectedBitrate * VHS_BANDWIDTH_VARIANCE * 10 + 1` then it's much faster
image

Thoughts?

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -369,14 +377,24 @@ export default defineComponent({
           controlBarItems.splice(index, 1)
         }
 
+        // regardless of what DASH qualities you enable or disable in the qualitLevels plugin
+        // the first segments videojs-http-streamming request are chosen based on the available bandwidth, which is set to 0.5MB/s by default
+        // overriding that to be the same as the quality we requested, makes videojs-http-streamming pick the correct quality
+        const playerBandwidthOption = {}
+
+        if (this.useDash && this.defaultQuality !== 'auto') {
+          playerBandwidthOption.bandwidth = this.selectedBitrate * VHS_BANDWIDTH_VARIANCE + 1
```
PikachuEXE, line 386: My habit to add code comment about references to doc
```suggestion
          // https://github.com/videojs/http-streaming#vhsbandwidth
          playerBandwidthOption.bandwidth = this.selectedBitrate * VHS_BANDWIDTH_VARIANCE + 1
```

PikachuEXE (changes requested):

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -369,14 +378,25 @@ export default defineComponent({
           controlBarItems.splice(index, 1)
         }
 
+        // regardless of what DASH qualities you enable or disable in the qualitLevels plugin
```
ChunkyProgrammer, line 381:
```suggestion
        // regardless of what DASH qualities you enable or disable in the qualityLevels plugin
```

ChunkyProgrammer (commented):
Typo

PikachuEXE:
Just fixed a bug when default quality is set to auto
My IDE keeps warning me `this.defaultQuality` can only be `number` (before fix)

PikachuEXE:
Also fixed quality selection when video height > width (e.g. W720 x H1280)
a.k.a Shorts/Vertical videos?

absidue:
@PikachuEXE Your code will only work for the local API, Invidious doesn't return a height or width, it returns  a resolution attribute with the larger dimension which we currently use as the height.

PikachuEXE:
I got the quality selection working for both local API and Invidious but the quality selector is not displaying the quality being used
image
image
image

Video used for testing:
https://youtu.be/v3wm83zoSSY

PikachuEXE:
Do we have some example videos for HDR formats?

I can only find https://gist.github.com/AgentOak/34d47c65b1d28829bb17c24c04a0096f for format IDs
For test video https://youtu.be/v3wm83zoSSY higher bitrate might not be a better quality (I can't tell)
image

I will revert the change since this might need some tweaking later (requires another PR)

PikachuEXE:
Another followup issue
https://youtu.be/tO01J-M3g0U cannot be played with Invidious API with `Allow DASH AV1 formats`

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Is selecting between multiple of the same quality in scope of this PR?

https://user-images.githubusercontent.com/73130443/226863751-d5ce104b-cec8-4583-9baa-320175564d40.mp4

PikachuEXE:
Right now it just use highest bitrate one
It's selecting 720p on 2nd video coz it has higher bitrate (even 30 fps)
Not scope of this PR (mainly fix DASH quality on start, see issues)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Issue #1369, opened 2021-05-29
[feature] screensaver disabled only if video is playing (linux)


arrowgent:
**Important note**
Please check other open and closed issues for a similar feature request before proposing a new one. 
Also please leave all headers except this one in the issue, so that it is clear what belongs where.
yes none found

**Is your feature request related to a usage problem (not a bug)? Please describe.**
Please describe clear and concise of what the problem is. Ex. I'm always frustrated when [...]
screensaver disable is active if no video is playing or active

**Describe the solution you'd like to see implemented**
Please give a clear and concise description of what you want to happen.
if video is playing then disable screensaver
if videos STOPS or no video page is active then screensaver/powermanagement is allowed

~~ v0.12 & v0.13
more deliberate testing required

arrowgent:
only contents i find of "power" ie: screensaver, powersaveblocker

https://github.com/FreeTubeApp/FreeTube/blob/754a4a8aa320a753139ae370a607a2ed6756196b/src/renderer/components/ft-video-player/ft-video-player.js#L32

https://github.com/FreeTubeApp/FreeTube/blob/1c57c2b0d72eb90c9f75e7979ea12a85ae131f7a/src/main/index.js#L3

https://github.com/FreeTubeApp/FreeTube/blob/1c57c2b0d72eb90c9f75e7979ea12a85ae131f7a/src/main/index.js#L443



ah, "power" "power management"
https://github.com/FreeTubeApp/FreeTube/issues/279

DnickC:
I'm on linux mint 19 and latest version of freetube.
I notice powersaving switching on and off works as expected until a video is started and while still playing I navigate back to subscriptions or any other page. After this power management stays disabled until the window is closed.

Issue #3140, opened 2023-01-31
[Bug]: Freetube blocks MacOs from auto sleep


John4003:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Run MacOs 
2. Open Freetube v.0.18.0. 
3. Display channel /subsciption grid (video not playing)
4. Check Mac sleep controls settings...
5. Wait for a Mac to go to sleep...  
6. Results? Mac never sleeps...even displays  
7. still waiting? :)/

### Expected Behavior

Previous version worked fine. No problems with system/display auto sleep

### Issue Labels

feature stopped working

### FreeTube Version

v.0.18.0

### Operating System Version

MacOS 10.14.6

### Installation Method

.dmg

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

previous

### Additional Information

Tested on 3 macs

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

pklampros:
Just to add: This seems to be happening on linux as well. Flatpak version on Debian Bookworm

oldlamps:
This is a constant problem for me as well. I've tried both the flatpak, and appimage versions. If the program is running, it has a sytemd-inhibit opened that keeps the display open. 

There's been countless times over the past months where I've left the app open in the background (not playing a video), left for a few hours, and come back to find my monitors still on.  Very frustrating. 

I haven't figured out a way to bypass the functionality, but it's really sad because the application is killer.

openforfuture:
This bug is still present today: FreeTube prevented the MacBook (M1 Pro) to sleep and it drained completely the battery during the night.

PR #3286, opened 2023-03-11
Rework the screen saver blocking while watching a video

absidue:
# Rework the screen saver blocking while watching a video

## Pull Request Type

- [x] Bugfix

## Related issue
might address #1369 (will change to closes if people can confirm it works)
might address #3248

## Description
This should hopefully help with the screen saver blocking being turned on but never off.

(functions have their own `this` scope that can be overriden, so what was happening is that the play event was actually assigning the blocker token to the player instance instead of the ft-player component, so when other functions went to try and disable the blocker with the token, they didn't find it and therefore couldn't disable the blocker)

## Testing 
This is untested, as the screen saver seems to be flaky on my comptuer regardless of whether i'm using FreeTube or not.

https://github.com/absidue/FreeTube/actions/runs/4393284869

The screen saver should be disabled when the video is playing but should be reenabled if the video is paused, ends or a player error occurs or the ft-player component is destroyed.

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (dismissed):
I have no idea how to test screensaver quickly
Code looks fine though

efb4f5ff-1298-471a-8973-3d47447115dc:
Might also addresses https://github.com/FreeTubeApp/FreeTube/issues/3140

oldlamps:
Hi, thanks for addressing this issue. I've recently shared my experience over in https://github.com/FreeTubeApp/FreeTube/issues/3140#issuecomment-1464918560 .. I downloaded the appImage from https://github.com/absidue/FreeTube/actions/runs/4393284869 , and tested a few scenarios that were keeping the screen saver lock on. 

I'm glad to report that this seems to fix the issue on my end. 

Test case 1:
Playing a video creates a "user session inhibited block", pausing removes the block

Test case 2:
While playing I navigate away to the subscriptions page. The block is removed!

Test case 3:
Playing a video, the lock is create. Skip to the end of the video, and let it finish. Upon finishing the block is removed!

These are the 3 scenarios that were keeping the block in place. I have verified with the release version, and the bugs are persistent. 

Very happy about this , thank you for looking into it again. I'll be using this AppImage, and hopefully the fix will find its way in.

absidue:
@oldlamps Thank you so much for confirming it works for you :).

efb4f5ff-1298-471a-8973-3d47447115dc:
@John4003 is it possible for u to test this?

PikachuEXE (approved):

PikachuEXE:
Let me test on MacOS if this is not approved yet tomorrow
https://github.com/FreeTubeApp/FreeTube/issues/3140

PikachuEXE:
Tested with MacOS 12.6.3, all 3 test cases passed

Steps (per case):
- Update display to sleep after 1 minute
- Perform test case
- wait 1 minute

```
Test case 1:
Start playing a video, then pause it

Test case 2:
Start playing a video, then leave the watch page

Test case 3:
Start playing a video, seek almost to the end, wait until the video ends
```

Also tested that the display won't go to sleep when playing a video (played a video with 1:00 length)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

DnickC:
Great to see this issue being fixed!

> I have no idea how to test screensaver quickly Code looks fine though

I use 'inhibit applet' on mint 19, this shows me power management being blocked or not in a panel icon, great tool for testing, no need to wait for the screensaver.

I tested on mint 19.3,  all 3 of the test cases succeed now, but this fails for me: 
start playing a video, reload the page with ctrl+r

oldlamps:
> Great to see this issue being fixed!
> 
> > I have no idea how to test screensaver quickly Code looks fine though
> 
> I use 'inhibit applet' on mint 19, this shows me power management being blocked or not in a panel icon, great tool for testing, no need to wait for the screensaver.
> 
> I tested on mint 19.3, all 3 of the test cases succeed now, but this fails for me: start playing a video, reload the page with ctrl+r

I can confirm that is still one case where the block still sticks. While the video is loading / loaded and playing hitting ctrl+r will cause the inhibit block to stay once the page reloads, and continue despite pausing / unpausing, letting the video finish, or navigating away.

EvilGremlin:
+1, reload totally must remove lock, issue is unresolved until then.

absidue:
@EvilGremlin The issue is essentially solved if you use the nightly builds. There should never be a need to use the reload function anyway, pressing the backwards and forwards keys does the same but a lot less destructive (reloading reloads that entire window, forcing it to reinitialise everything, it's almost the same as closing and reopening FreeTube).

EvilGremlin:
pressing ctrl+r is much faster and convenient thann going to history to open that video again

absidue:
You can just use the backward and forward buttons.

Also avoids having load your settings, subscriptions, watch history, playlists and everything else again.

PR #3315, opened 2023-03-16
Fetch the videos tab directly for local API subscriptions

absidue:
# Fetch the videos tab directly for local API subscriptions

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
New Innertube ratelimits mentioned in Matrix

## Description
Currently the local API fetches the channel default tab first and then the videos tab, that is less prone to breakage if YouTube changes stuff, as you are always using the information YouTube provided in the first request to get the videos tab. However that does mean that two requests are made for every channel, making it more likely to get ratelimited unless we lower the forced RSS threshold to 75.

This pull request fetches the videos tab directly, halving the number of requests that are made, which also avoids having to lower the forced RSS threshold.

## Testing 
Select the local API and disable RSS
Refresh your subscriptions, it should work the same as before, but without hitting the ratelimits if you have lots of subscriptions.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -1,4 +1,4 @@
-import { Innertube, ClientType, Misc, Utils } from 'youtubei.js'
+import { Innertube, ClientType, Misc, Utils, YT } from 'youtubei.js'
 import Autolinker from 'autolinker'
 import { join } from 'path'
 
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
9 subscriptions = 9 requests on refresh

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3289, opened 2023-03-12
Reset sort by when changing channels

absidue:
# Reset sort by when changing channels

## Pull Request Type

- [x] Bugfix

## Related issue
closes #3287 

## Description
When you change channels the sort by values currently don't get reset, this pull request fixes it (I've already pushed a fix for the live sort by to the live tab PR). That means that while the drop down shows it correctly sorted by newest, the actual API requests use the value from the previous channel page, which is fixed now.

## Testing 
Set the sort by values on the videos and playlists tabs to something other than newest, change to a different channel, make sure the videos and playlists are sorted by newest.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.18.0

PR #3341, opened 2023-03-22
Fix trending video list not updated without a workaround

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Similar "incorrect key messed up render update" issue occurred in #3289

## Description
Workaround is used to force render update when data (list of video entries) is changed
This PR makes it uses correct key to ensure render update is done without workaround
Since trending page only has videos (I assume) only video type is handled now, another PR will be created to handle all "result types"

## Screenshots 
Lazy!

## Testing 

Test with both Local/Invidious API plz
- Go Trending Page
- Switching to all tabs to warm up result cache
- Switching to all tabs randomly/sequentially to ensure rendered video list updated 

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/views/Trending/Trending.js
+++ b/src/renderer/views/Trending/Trending.js
@@ -122,14 +122,7 @@ export default defineComponent({
       // the ft-element-list component has a bug where it doesn't change despite the data changing
       // so we need to use this hack to make vue completely get rid of it and rerender it
       // we should find a better way to do it to avoid the trending page flashing
-      this.isLoading = true
-      setTimeout(() => {
-        this.shownResults = this.trendingCache[this.currentTab]
-        this.isLoading = false
-        setTimeout(() => {
-          this.$refs[this.currentTab].focus()
-        })
-      })
+      this.shownResults = this.trendingCache[this.currentTab]
     },
     getTrendingInfoInvidious: function () {
       this.isLoading = true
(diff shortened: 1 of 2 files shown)


src/renderer/views/Trending/Trending.js
```diff
@@ -122,14 +122,7 @@ export default defineComponent({
       // the ft-element-list component has a bug where it doesn't change despite the data changing
```
ChunkyProgrammer, line 122: I'm guessing this comment should probably be removed

PikachuEXE:
Done

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc That is different from the original issue, which caused the tabs, the title and the refresh button to briefly flash. I think that is just caused by the video elements and the lazy loading wrapper around them being expensive to process, the wrapper checks if the video element should be hidden because of various settings (e.g. hide live or hide based on channel id) and the video element itself checks the history and saved videos among other things, additionally each video element has it's own dropdown that needs to be created as well.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@absidue should i open an issue for that?

PR #3346, opened 2023-03-22
Prevent `channelInstance` from being `null` in community tab fallback

MarmadileManteater:
# Prevent `channelInstance` from being `null` in community tab fallback
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
The `getCommunityPostsLocal` uses `channelInstance`, but when it is called after the Invidious API fails inside of the `getCommunityPostsInvidious` method, the `channelInstance` is still `null`. This PR aims to address this by pulling the logic for getting a local API `channelInstance` out of the `getChannelLocal` method and into a new method called `getChannelInstanceLocal` (which is now called within `getChannelLocal`). Then, `getChannelInstanceLocal` can be called before the `getCommunityPostsLocal` inside of the `getCommunityPostsInvidious` fallback.

## Screenshots 

_before_
channel-instance-is-null-before
_after_
channel-instance-is-not-null-after

## Testing 
1. Make sure the backend preference is set to `Invidious API`
2. Make sure `Fallback to non-preferred backend` is enabled
3. Navigate to a channel where the community posts fail to load in the Invidious API (EX: https://www.youtube.com/channel/UC2bso7dShHmrlH9EzGyH-CQ)
4. Make sure the community posts load without any Local API errors

## Desktop
 - OS: Windows 10
 - FreeTube version: 0.18.0
## Additional Info
There may be better ways to accomplish this, but I felt like separating the logic for getting the local API channel instance from the `getChannelLocal` method made sense. The code still essentially functions the same for the Local API, but the structure should allow for the `channelInstance` to be set within fallbacks as well as the `getChannelLocal` method. It might make more sense to call `getChannelInstanceLocal` within `getCommunityPostsLocal` if `channelInstance` is `null`.

### DIFF
--- a/src/renderer/views/Channel/Channel.js
+++ b/src/renderer/views/Channel/Channel.js
@@ -396,160 +396,161 @@ export default defineComponent({
       this.$router.push({ path: `/channel/${id}` })
     },
 
+    getChannelInstanceLocal: async function (expectedId) {
+      const channel = await getLocalChannel(this.id)
+
+      let channelName
+      let channelThumbnailUrl
+
+      if (channel.alert) {
+        this.setErrorMessage(channel.alert)
+        return
+      } else if (channel.memo.has('ChannelAgeGate')) {
+        /** @type {import('youtubei.js/dist/src/parser/classes/ChannelAgeGate').default} */
+        const ageGate = channel.memo.get('ChannelAgeGate')[0]
+
+        channelName = ageGate.channel_title
+        channelThumbnailUrl = ageGate.avatar[0].url
+
+        this.channelName = channelName
+        this.thumbnailUrl = channelThumbnailUrl
+
+        document.title = `${channelName} - ${packageDetails.productName}`
+
+        this.updateSubscriptionDetails({ channelThumbnailUrl, channelName, channelId: this.id })
+
+        this.setErrorMessage(this.$t('Channel["This channel is age resticted and currently cannot be viewed in FreeTube."]'), true)
+        return
+      }
+
+      this.errorMessage = ''
+      if (expectedId !== this.id) {
+        return
+      }
+
+      let channelId
+      let subscriberText = null
+      let tags = []
+
+      switch (channel.header.type) {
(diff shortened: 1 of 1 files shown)


absidue:
Currently all other fallbacks call getChannelLocal, could you please modify those as well or switch the community fallback to be the same as the others?

MarmadileManteater:
> Moving most of the stuff in `getChannelInstanceLocal` into `getChannelLocal`, should make `getChannelInstance` a bit faster for the fallback and won't overwrite the information provided by Invidious, unless I'm misunderstanding what this change is supposed to do.

The reason I believe it makes sense to separate out the `getChannelInstanceLocal` logic is that calling `getChannelLocal` in the fallback for a single tab _(such as the community tab or playlist tab)_ can cause a screen flickering effect which I find visually unappealing, and the view ends up reloading data which was already loading in correctly from invidious with data which is loading in from the local API.

when-calling-get-channel-local-inside-of-community-posts-fallback

This is part of why I believe it makes sense to only call the corresponding local API call when fetching a single tab fails with invidious because that should prevent situations where the same data is fetched over again and the page _sort of_ refreshes. At the same time, I believe that calling `getChannelLocal` directly makes sense for the continuation fall-backs (`getChannelPlaylistsMore`, `channelInvidiousVideos`) because AFAIK there isn't any way to make the continuation data between invidious and youtube.js inter-compatible.

In the end, it may make the most sense to just call `getChannelLocal` instead of `getCommunityPostsLocal` and not add the extra method, but I just wanted to explain how I thought about the problem.

absidue:
The channel page continuation tokens have a GUID in them (that's why Invidious can't generate the tokens, to enable page number based pagination, like they do with other endpoints), so I suspect that they might not be interchangeable between machines. I haven't done any testing yet though.

absidue (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #3350, opened 2023-03-23
Fix adaptive format picked by video JS sometimes not showing up in quality selector

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Not if only happens after #3278, but discovered during testing #3278 anyway

## Description

Using Local API & Proxy video through Invidious would cause quality selector issue on some videos
Due to **different quality set** from Local API and Invidious API

## Screenshots 

Before fix
Screenshot 2023-03-23 at 09 21 13

After fix
Screenshot 2023-03-23 at 09 23 59

## Testing 

- Use Local API, enable Proxy Videos Through Invidious, set default quality to 720p
- Play https://youtu.be/v3wm83zoSSY
- Video quality should be 720p and quality selector should be something 720p (instead of undefined)

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -586,10 +586,15 @@ export default defineComponent({
             // we need to alter the result object so the toDash function uses the filtered formats too
             result.streaming_data.adaptive_formats = filterLocalFormats(result.streaming_data.adaptive_formats, this.allowDashAv1Formats)
 
-            this.adaptiveFormats = result.streaming_data.adaptive_formats.map(mapLocalFormat)
+            // When `this.proxyVideos` is true
+            // It's possible that the Invidious instance used only support only partial of the formats from Local API
+            // i.e. the value passed into `adaptiveFormats`
+            // e.g. Supports 720p60, but not 720p - https://[DOMAIN_NAME]/api/manifest/dash/id/v3wm83zoSSY?local=true
             if (this.proxyVideos) {
+              this.adaptiveFormats = await this.getAdaptiveFormatsInvidious()
               this.dashSrc = await this.createInvidiousDashManifest()
             } else {
+              this.adaptiveFormats = result.streaming_data.adaptive_formats.map(mapLocalFormat)
               this.dashSrc = await this.createLocalDashManifest(result)
             }
 
@@ -636,7 +641,7 @@ export default defineComponent({
       this.videoStoryboardSrc = `${this.currentInvidiousInstance}/api/v1/storyboards/${this.videoId}?height=90`
 
       invidiousGetVideoInformation(this.videoId)
-        .then(result => {
+        .then(async result => {
           if (result.error) {
             throw new Error(result.error)
           }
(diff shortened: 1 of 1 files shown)


src/renderer/views/Watch/Watch.js
```diff
@@ -1266,6 +1264,23 @@ export default defineComponent({
       ]
     },
 
+    getAdaptiveFormatsInvidious: async function(existingInfoResult = null) {
+      let getInfoPromise = Promise.resolve(existingInfoResult)
+      if (existingInfoResult == null) {
+        getInfoPromise = invidiousGetVideoInformation(this.videoId)
+      }
+      const result = await getInfoPromise
```
absidue, line 1272: Is there a specific reason you decided to go with wrapping the result in a promise? If not, something like this seems less complex.

```suggestion
      let result
      if (existingInfoResult) {
        result = existingInfoResult
      } else {
        result = await invidiousGetVideoInformation(this.videoId)
      }
```

src/renderer/views/Watch/Watch.js
```diff
@@ -1266,6 +1264,23 @@ export default defineComponent({
...
+
+      return filterInvidiousFormats(result.adaptiveFormats, this.allowDashAv1Formats)
+        .map((format) => {
+          format.bitrate = parseInt(format.bitrate)
+          if (typeof format.resolution !== 'undefined') {
+            format.height = parseInt(format.resolution.replace('p', ''))
+          }
```
absidue, line 1279: Instead of doing a negative check for undefined you could do a positive check for a string.
```suggestion
          if (typeof format.resolution === 'string') {
            format.height = parseInt(format.resolution.replace('p', ''))
          }
```

src/renderer/views/Watch/Watch.js
PikachuEXE: This is me (still) not getting used to `await`/`async`

src/renderer/views/Watch/Watch.js
PikachuEXE: Yup, not sure why I forgot

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

ChunkyProgrammer (approved):

PikachuEXE:
@absidue Applied workaround here as you asked
@efb4f5ff-1298-471a-8973-3d47447115dc @ChunkyProgrammer I have updated `Description ` & `Testing` for an additional workaround

src/renderer/views/Watch/Watch.js
```diff
@@ -586,10 +586,15 @@ export default defineComponent({
             // we need to alter the result object so the toDash function uses the filtered formats too
             result.streaming_data.adaptive_formats = filterLocalFormats(result.streaming_data.adaptive_formats, this.allowDashAv1Formats)
 
-            this.adaptiveFormats = result.streaming_data.adaptive_formats.map(mapLocalFormat)
+            // When `this.proxyVideos` is true
+            // It's possible that the Invidious instance used only support only partial of the formats from Local API
```
absidue, line 590:
```suggestion
            // It's possible that the Invidious instance used, only supports a subset of the formats from Local API
```

absidue (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3352, opened 2023-03-24
Add setting to auto load comment section

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

N/A

## Description

Add a setting to allow comment section to be auto loaded, default disabled
When enabled (with a percentage), comment section **bottom** when appears pass N% above bottom of viewport will have comment data auto loaded **once** (won't load it if data already being loaded, loaded)

## Screenshots 

New setting position (can be adjusted)
image

## Testing 

A. Disabled
- No setting change should be required, it's default
- View a video with comment
- Scroll~~~
- Ensure comment not auto loaded

B. 0% from window/view port bottom
- Update setting
- View a video with comment
- Scroll carefully and observe comment section
- Ensure comment auto loaded once comment section **bottom** is 0% above bottom of window/view port

C. 50% from window/view port bottom
- Update setting
- View a video with comment
- Scroll carefully and observe comment section
- Ensure comment auto loaded once comment section **bottom** is 50% above bottom of window/view port

You can test other settings as well, those percentage values (yes zero or negative values) are just passed into [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) [`rootMargin`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin)

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

- To prevent comment autoload on page load (when video player is minimized for a moment), a timeout is used (1000ms)
  I am not sure if that's good enough
- Updating setting won't adjust behaviour until video switched


ChunkyProgrammer:
I think a toggle should probably be used instead (percentage might be confusing for users)

PikachuEXE:
I don't understand
Toggle to enable and another setting to adjust percentage?

absidue:
Chunky means don't expose any percentages to the user, pick one and then have a single on and off switch.

Also instead of writing your own IntersectionObserver code, you can use vue-observe-visibility, which we already have to lazy load the videos etc in the lists. See [ft-list-lazy-wrapper](https://github.com/FreeTubeApp/FreeTube/tree/development/src/renderer/components/ft-list-lazy-wrapper) to see how it's used there.

PikachuEXE:
Problem is 
- small number like 0% = Auto load on large windows anyway
- large number = inconvenient

Let me take a look at vue-observe-visibility...

absidue:
If the comment section is always visible in large windows people will expect it to auto load straight away anyway.

PikachuEXE:
Updated

static/locales/en-US.yaml
```diff
@@ -280,6 +280,16 @@ Settings:
       Error:
         Forbidden Characters: Forbidden Characters
         Empty File Name: Empty File Name
+    Comment Auto Load Condition:
+      Comment Auto Load Condition: Comment Auto Load
+      Option Labels:
+        'disabled': Disabled
+        '0%': When 0% above window bottom
+        '-10%': When 10% above window bottom
```
ChunkyProgrammer, line 288: These can probably be removed since a toggle is being used now

src/renderer/components/player-settings/player-settings.js
```diff
@@ -266,6 +274,12 @@ export default defineComponent({
       })
     },
 
+    updateCommentAutoLoadConditionEnabled: function(enabled) {
+      this.updateCommentAutoLoadCondition(
+        enabled ? '0%' : 'disabled'
```
ChunkyProgrammer, line 279: Maybe just pass in the bool value?

src/renderer/components/player-settings/player-settings.js
```diff
@@ -195,7 +195,15 @@ export default defineComponent({
...
     screenshotFilenamePattern: function() {
       return this.$store.getters.getScreenshotFilenamePattern
-    }
+    },
+
+    commentAutoLoadConditionCurrentValue: function () {
+      return this.$store.getters.getCommentAutoLoadCondition
+    },
+
+    commentAutoLoadConditionEnabled: function () {
```
ChunkyProgrammer, line 204: This can be removed if you switch to bool

src/renderer/store/modules/settings.js
```diff
@@ -281,6 +281,7 @@ const state = {
   fetchSubscriptionsAutomatically: true,
   settingsPassword: '',
   allowDashAv1Formats: false,
+  commentAutoLoadCondition: 'disabled',
```
ChunkyProgrammer, line 284: Could be set to false instead

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
It would be nice to autoload comments when i almost reach the 'load more comments' button

PikachuEXE:
> It would be nice to autoload comments when i almost reach the 'load more comments' button
Do you mean mouse cursor?

src/renderer/components/player-settings/player-settings.js
PikachuEXE, line 204: Using string is intentional even though UI is a toggle
I am not excluding the possibility to introduce other values later
No DB change needed when that's the case

src/renderer/components/player-settings/player-settings.js
PikachuEXE, line 279: Intentional "frontend boolean to backend enum" conversion

PikachuEXE:
Unnecessary translation entries removed

PikachuEXE (commented):
It's my habit to ensure the values in DB is more flexible than the UI when possibility of future change is present

efb4f5ff-1298-471a-8973-3d47447115dc:
> > It would be nice to autoload comments when i almost reach the 'load more comments' button
> > Do you mean mouse cursor?

When i reach the bottom of the comment section i want it to auto load more comments


https://user-images.githubusercontent.com/73130443/227766930-57f91bb0-a622-4935-8a61-6ef750a0b92b.mp4

PikachuEXE:
Different feature, different PR please

absidue:
@PikachuEXE i doubt we'll be adding more values in the future, not having it be a boolean sounds like unnecessary complication to me.

Although considering that in most screenshots you've sent, you have had FreeTube at a massive window size, I'm guessing that the reason you don't want to switch to a boolean, is so that you can have a different percentage in your personal settings db, while everyone else uses 0%.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Different feature, different PR please Update 1: Or do you mean it should be done in this PR too??? Update 2: Will try to implement it if not merged within 2 days

As user i expect the following when i enable this feature: 

1. When i navigate to a video it loads the comments section (current behavior of PR)
2. When i scroll down the comment section it should load more comments

IMO (again from an user perspective) the PR in its current iteration is a half baked feature. So to come back to your question i dont see this as a different feature.

PikachuEXE:
Is there a path to update the setting from boolean to string if it ever happens?
Update reader code?

Also are 4 new methods instead of 2 really that complicated?

absidue:
I just don't think it's necessary to over engineer something, just in case we change our minds in the future, as that is extremely unlikely to happen.

PikachuEXE:
Updated

src/renderer/components/watch-video-comments/watch-video-comments.vue
```diff
@@ -251,6 +251,14 @@
     <ft-loader
       v-if="isLoading"
     />
+    <div
+      v-observe-visibility="observeVisibilityOptions"
+    >
+      <!--
+        Dummy element to be observed by Intersection Observer
+        Key is required to ensure NO rerender
```
absidue, line 259: Looks like this part of the comment is no longer needed.
```suggestion
```

PikachuEXE:
Deleted

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
If the hide comments distraction free setting gets enabled and auto load comments was enabled before. The toggle auto load toggle should be greyed out and disabled

PikachuEXE:
Setting updated

Also removed the delay on comment loading (I didn't know there is a ready event)

src/renderer/views/Watch/Watch.js
```diff
@@ -928,6 +929,11 @@ export default defineComponent({
       this.updateLastViewedPlaylist(payload)
     },
 
+    handleVideoReady: function () {
+      this.videoPlayerReady = true
```
absidue, line 933: please reset this to false when the video changes (in the watch handler for $route)

src/renderer/components/watch-video-comments/watch-video-comments.js
```diff
@@ -70,7 +78,41 @@ export default defineComponent({
...
+          // This is also fired when **hidden**
+          // No point doing anything if not visible
+          if (!isVisible) { return }
+          // It's possible the comments are being loaded/already loaded
+          if (this.canPerformInitialCommentLoading) {
+            this.getCommentData()
+          }
+          if (this.canPerformMoreCommentLoading) {
+            this.getMoreComments()
+          }
```
absidue, line 98: Could you please make the second if an else if, so that it is more obvious to someone reading the code, that only one of them will get triggered at a given time.

efb4f5ff-1298-471a-8973-3d47447115dc:
it shows being greyed out but enabled. That could potentially be confusing so i think u should also disable it.. AFAIK we also do that with other settings like that

PikachuEXE:
Done

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer ur powers are needed

ChunkyProgrammer (approved):

PR #3355, opened 2023-03-25
Improve performance of list elements and icon buttons

absidue:
# Improve performance of list elements and icon buttons

## Pull Request Type

- [x] Performance improvement

## Description
_Sorry in advance for the wall of text 🫂_

This pull request improves the initial creation performance of list elements (`ft-list-video`, `ft-list-channel`, `ft-list-playlist`) and `ft-icon-button`. Take a look at the screenshots below to see the before and after times for the list items and icon buttons, please keep in mind that the number will differ depending on your machine and what else it is currently doing, however on any machine the after numbers should be smaller than the before ones.

The biggest changes in this pull request are the following two things:

### 1. Move the data parsing in the list elements from the `mounted` hook to the `created` hook

The mounted hook is called after the component has been created, rendered and the initial DOM changes have been made.
The created hook on the other hand is called after the component instance has been created, the props are available and reactivity has been setup, before any rendering is done.

As we are currently parsing the data in the mounted hook, that means Vue first creates creates "empty" DOM elements and then later after it detects the changes to the data made by the parsing, it triggers an update, rerenders the element and changes the elements in the DOM to contain the data.

Moving the parsing to the created hook means that when Vue goes to render the component for the first time, all of the data needed for rendering is already available, so it can render it with the data and doesn't need the update (updates will still happen if the data changes, e.g. video is favourited or watched). This also means that the ft-icon-buttons don't need to get updated either.

### 2. Create the ft-icon-button dropdowns when they are needed

How often do you actually use the 3 dots dropdown on ft-list-videos? Probably not very often and when you do, you probably only do it on a few of them in the list and don't touch most of them.

Currently the ft-icon-button dropdowns are always generated and then hidden and shown with CSS (using `v-show` to toggle `display: none`). By switching to v-if, we only create them when the ft-icon-button is activated (by mouse or keyboard), which speeds up the intial load and also saves work creating dropdowns that might never be used. At the moment even for icon buttons without a dropdown and empty dropdown div is created.

## Screenshots 
<details><summary><b>ft-list-video</b></summary>

list-video-before
list-video-after

</details>

<details><summary><b>ft-list-video</b></summary>

icon-button-before
icon-button-after

</details>

## Testing 

To get the most repeatable results, these tests were performed on an fresh FreeTube setup by deleting the database files in the development data folder `%appdata%/Electron` before testing. That way we avoid any extra random delays that come into play when you have favourites and history (the delays will vary depending on how many history and saved items you have, as well as where the videos are in those lists, if they are in there at all).
Additionally I made sure to use a small enough window size `1186x761px`, just big enough to fit 6 videos inside the visible area of the screen (while the devtools are open, resizing a window will make the dimensions show up in the top right for a short moment), to avoid lazy loaded items making the results unpredictable (the first 16 items are always loaded, if you scroll down or have a large window size, other items get lazy loaded in later).

The first step is to make the vue devtools show up, you can do this by going to `node_modules\vue-devtools\lib\index.js` and changing line 5 (that dependency hasn't received an update in ages and Electron now requires an absolute path to the extension):

```diff
- const extPath = path.join(__dirname, '..', 'vender')
+ const extPath = path.resolve(path.join(__dirname, '..', 'vender'))
```
yarn will undo those changes the next time you use `yarn install`, `yarn add`, `yarn remove` and friends, commands like `yarn dev` don't make any changes to the dependencies, so they won't cause yarn to reset the vue-devtools package.

Now to the actual testing:

1. Start FreeTube with `yarn dev` (I recommend closing FreeTube between test runs, so that you get the "worst case" scenario, without giving V8 a chance to do JIT optimisations, which will result in faster numbers which will skew your results)
2. Click on the trending tab (we want to preload the trending cache)
3. Click back to the subscriptions tab
4. Start the Vue performance test/measuring
4.1. In the devtools go to the `Vue` tab
4.2. Inside that click on the `Performance` tab (has a bar graph icon, if you hover over the icon it says `Switch to Performance`)
4.3. Switch to `Component render`
4.4. Click `Start`
5. Click on the trending tab
6. Wait until it's loaded and the component list in the devtools has stopped changing
7. Click `Stop` on the performance test

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 3a997fa569bb7337a456a85459989f76b25223c1

### DIFF
--- a/src/renderer/components/ft-community-post/ft-community-post.js
+++ b/src/renderer/components/ft-community-post/ft-community-post.js
@@ -64,7 +64,7 @@ export default defineComponent({
       return this.$store.getters.getListType
     }
   },
-  mounted: function () {
+  created: function () {
     this.parseVideoData()
   },
   methods: {
--- a/src/renderer/components/ft-icon-button/ft-icon-button.js
+++ b/src/renderer/components/ft-icon-button/ft-icon-button.js
@@ -74,8 +74,10 @@ export default defineComponent({
       window.addEventListener('resize', this.handleResize)
     }
   },
-  destroyed: function () {
-    window.removeEventListener('resize', this.handleResize)
+  beforeDestroy: function () {
+    if (this.dropdownModalOnMobile) {
+      window.removeEventListener('resize', this.handleResize)
+    }
   },
   methods: {
     sanitizeForHtmlId,
--- a/src/renderer/components/ft-icon-button/ft-icon-button.vue
+++ b/src/renderer/components/ft-icon-button/ft-icon-button.vue
@@ -19,76 +19,78 @@
       @keydown.enter.prevent="handleIconClick"
       @keydown.space.prevent="handleIconClick"
     />
-    <ft-prompt
-      v-if="useModal"
-      v-show="dropdownShown"
-      :autosize="true"
-      :label="sanitizeForHtmlId(`iconButtonPrompt-${title}`)"
-      @click="dropdownShown = false"
+    <template
+      v-if="dropdownShown"
     >
-      <slot>
-        <ul
-          v-if="dropdownOptions.length > 0"
-          class="list"
-          role="listbox"
-          :aria-expanded="dropdownShown"
-        >
-          <li
-            v-for="(option, index) in dropdownOptions"
-            :id="sanitizeForHtmlId(title + '-' + index)"
-            :key="index"
-            role="option"
-            aria-selected="false"
-            tabindex="-1"
-            :class="option.type === 'divider' ? 'listItemDivider' : 'listItem'"
-            @click="handleDropdownClick({url: option.value, index: index}, $event)"
-            @keydown.enter="handleDropdownClick({url: option.value, index: index}, $event)"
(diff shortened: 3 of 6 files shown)


ChunkyProgrammer:
Should we fork vue devtools for our purposes? or maybe we can find an alternative

PikachuEXE:
If it's just for our dev use, just fork it (until it's active again)
No point making temp changes every time

PikachuEXE (approved):
Before
Screenshot 2023-03-26 at 09 18 05
Screenshot 2023-03-26 at 09 18 15

After
Screenshot 2023-03-26 at 09 19 01
Screenshot 2023-03-26 at 09 19 34

ChunkyProgrammer (dismissed):

PikachuEXE (approved):
Briefly tested after solving small merge conflict

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc The workflow doesn't add back label `PR: waiting for review`

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3383, opened 2023-03-29
Cleanup polyfills

absidue:
# Cleanup polyfills

## Pull Request Type

- [x] Cleanup

## Related issue
https://github.com/FreeTubeApp/FreeTube/projects/10#card-87323375

## Description
Now that we have fully migrated the local API to YouTube.js we don't need most of the polyfills anymore. We still need the `process` one as we have a couple of references to `process.platform` and co in the FreeTube code. The node `fs` and `path` modules are still used in some places too, additionally nedb imports them so it can do filesystem persistence.
That allows us to get rid of the browserify dependency, as long as we keep `process`, `path-browserify` and `buffer` around.

While I was at it, I took a look into why we actually need a polyfill for `Buffer` and it turns out that is because video.js' vhs-utils dependency supports both web browsers and node. As we only need web browser support I decided to patch out the check for the `atob()` function with the `DefinePlugin`, to say it's always available, that way we don't need the `Buffer` polyfill.

Another thing I noticed a while back and finally looked into was why the output `renderer.js` file still had stuff that looked like a DOM implementation, despite the web build of YouTube.js no longer using `linkedom`. Turns out that it's another case of a video.js module supporting node.js, this time it's the `mpd-parser` one, that is responsible for parsing the DASH manifest. As node doesn't have it's own DOM implementation they decided to use the `@xmldom/xmldom` package that works in both web browsers and node. `@xmldom/xmldom` provides an implementation of `DOMParser`, that `mpd-parser` uses, if you are think that you've heard of me mention `DOMParser` before, that's because web browsers have a native implementation, which we already use to parse the RSS feeds. So by telling webpack to resolve `@xmldom/xmldom` to our own small file that exports the native DOMParser, we can avoid including that dependency, shrinking our package sizes even further.

## Screenshots 
<details><summary>web</summary>

web-before web-after

</details>

<details><summary>renderer</summary>

renderer-before renderer-after

</details>

## Testing 
Try watching videos with the DASH formats enabled, if it still works, this pull request didn't break anything :)

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** c58f4e04faeba2596899783a2bd2f1e126db4d73


ChunkyProgrammer:
@MarmadileManteater (tagging you in case you want to review as as well)

ChunkyProgrammer (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #1568, opened 2021-08-06
Channel community page

GilgusMaximus:
---
Channel community page
---

**Pull Request Type**
Please select what type of pull request this is:
- [x] Feature Implementation

**Description**
This patch adds the community page tab to all channels. Items that can be displayed:
- Videos (Invidious and Local API)
- Images (Invidious and Local API)
- MultiImage (Invidious and Local API)
- Text (Invidious and Local API)
- Polls (Local API + Invidious when https://github.com/iv-org/invidious/pull/3635 is merged)
- Playlists (Local API)

Not supported:
- shared post

**Screenshots (if appropriate)**

Text
image

Image
image

Video
image

MultiImage
image

Poll
image

Playlist
image

**Testing (for code that is not small enough to be easily understandable)**
Has this pull request been tested?
Has been tested extensively for all kinds of channels. 

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: 10
 - FreeTube version: ~~v13.2~~ 0.18.0

**Additional context**
Still todo as discussed: Include playlist css into standard playlist css and check whether playlist is displayed in list or standalone mode

PR #3427, opened 2023-04-14
Communities category on Freetube

MuerteRoja:
Hello, good morning, today I am writing this post with the purpose of requesting that the category of "Communities" that YouTube has be introduced in the Freetube application, I think that function would be very useful, greetings. 
Captura de pantalla 2023-04-14 130356


efb4f5ff-1298-471a-8973-3d47447115dc:
already implemented in https://github.com/FreeTubeApp/FreeTube/pull/1568

PR #3451, opened 2023-04-20
Update History & UserPlaylists view to filter with delay

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

N/A

## Description
When using filtering on history view
I found it laggy due to list being filtered on every char input
This PR makes the filtering happens 1s after last input (except when query is clear) (feel free to suggest other values, UX changes...)

## Screenshots 

https://user-images.githubusercontent.com/1018543/233236060-72c89582-96b2-4336-ace2-da3e12d595a2.mov

## Testing 

- Browse History/UserPlaylist
- Type something to filter
- Wait a bit and ensure filtering is done

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


src/renderer/views/History/History.js
```diff
@@ -78,7 +79,26 @@ export default defineComponent({
...
+      }
+
+      if (this.query === '') {
+        // When query is empty it can be assumed that the user is clearing the query
+        // No need to wait
+        this.filterHistory()
+      } else {
+        // Updating list on every char input could be wasting resources on rendering
+        // So run it with delay (to be cancelled when more input received within time)
+        this.filterHistoryTimeout = setTimeout(this.filterHistory, 1000)
```
ChunkyProgrammer, line 96: Should we use lodash.debounce instead? (We use it elsewhere in the app)

PikachuEXE:
Updated to use `lodash.debounce`

absidue:
I think we should lower the delay, as 1 second looks slow in the video (I don't have enough items to test this properly). We should find a balance between reducing the lag and also making sure that it still feels snappy.
The search suggestions and setting the proxy both use 200ms delays, setting the Invidious instance uses a 500ms one. Maybe switching to one of those would be better, a different delay could also be picked.

PikachuEXE:
Updated to 500ms, video updated

absidue (approved):
feels good with 500ms, we can always choose a lower number in the future if we want to.

ChunkyProgrammer (dismissed):

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc ~

ChunkyProgrammer (approved):

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
Following feels a bit laggy now.

- Last removed character 
- crtl + A on text field and press backspace

https://user-images.githubusercontent.com/73130443/234727579-7467f784-510e-4655-8040-ee99304c8b3d.mp4

PikachuEXE:
This is due to my decision to update list immediately when the input is empty
If you think it's better to also have delay when it's empty I can make it so

efb4f5ff-1298-471a-8973-3d47447115dc:
> @efb4f5ff-1298-471a-8973-3d47447115dc This is due to my decision to update list immediately when the input is empty If you think it's better to also have delay when it's empty I can make it so

Yes i think its better to also have a delay when its empty. If the point of the PR is to not make it feel laggy anymore then i think the PR shouldnt make other parts feel more laggy to fix the issue.

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc updated

ChunkyProgrammer (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3452, opened 2023-04-20
Fix thumbnail URL for protocol-relative URLs

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

N/A

## Description

Sometimes the original thumbnail URLs are protocol-relative URLs, FT raises error since `URL` complains about this kind of URLs
This PR fixes it

## Screenshots 
Screenshot 2023-04-20 at 09 10 52

## Testing 

- Subscribe to https://www.youtube.com/channel/UC4R8DWoMoI7CAwX8_LjQHig
- Visit Channels View
- Ensure no error in console, list of channels displayed

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/views/SubscribedChannels/SubscribedChannels.js
+++ b/src/renderer/views/SubscribedChannels/SubscribedChannels.js
@@ -162,16 +162,20 @@ export default defineComponent({
 
     thumbnailURL: function(originalURL) {
       let newURL = originalURL
-      const hostname = new URL(originalURL).hostname
+      // Sometimes relative protocol URLs are passed in
+      if (originalURL.startsWith('//')) {
+        newURL = `https:${originalURL}`
+      }
+      const hostname = new URL(newURL).hostname
       if (hostname === 'yt3.ggpht.com' || hostname === 'yt3.googleusercontent.com') {
         if (this.backendPreference === 'invidious') { // YT to IV
-          newURL = youtubeImageUrlToInvidious(originalURL, this.currentInvidiousInstance)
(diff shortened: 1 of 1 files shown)


src/renderer/views/SubscribedChannels/SubscribedChannels.js
```diff
@@ -162,16 +162,20 @@ export default defineComponent({
 
     thumbnailURL: function(originalURL) {
       let newURL = originalURL
-      const hostname = new URL(originalURL).hostname
+      // Sometimes relative protocol URLs are passed in
+      if (originalURL.startsWith('//')) {
```
ChunkyProgrammer, line 166: Imo the whole thumbnailURL method can be removed as we already do invidious replacing through a function in api/invidious, we just need another function to fix up the local api urls (add missing https). Also if we save invidious links to the db I believe they will get overwritten by other parts of the app.

ChunkyProgrammer: Tldr: this method rewrites your entire subscription file if you change to invidious/change invidious instances and we should probably only use the youtube link

PikachuEXE: Make another PR I guess ~_~

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3456, opened 2023-04-20
Fix extracting chapters with handles from description

absidue:
# Fix extracting chapters with handles from description

## Pull Request Type

- [x] Bugfix

## Description
If YouTube doesn't provide chapters, we extract them from the description, for the local API we do it on the already parsed text runs, that means it can contain HTML. So if a creator uses channel handles in their chapters, we display the raw HTML for the links to the user (see first screenshot). This pull request instead extracts it from the text description, avoiding the problem.

Note: This doesn't affect the Invidious API as we already use the text description to extract the chapters.

## Screenshots 
before

after

## Testing 
https://www.youtube.com/watch?v=t1v050SqcD8

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 3597499df133348628a921a0ef7d05a2ef676b5f

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -368,7 +368,7 @@ export default defineComponent({
               })
             }
           } else {
-            chapters = this.extractChaptersFromDescription(this.videoDescription)
+            chapters = this.extractChaptersFromDescription(result.basic_info.short_description)
           }
 
           if (chapters.length > 0) {


ChunkyProgrammer (approved):

MarmadileManteater (approved):

PikachuEXE (approved):
image

Issue #2219, opened 2022-04-24
[Bug]: Cannot Import Subscriptions, Getting "Error: Request failed with status code 404"


GameGuyOnYT:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a bug report that matches the one I want to file, without success.

### Describe the bug

1. Go to Data Settings
2. Import [this](https://github.com/FreeTubeApp/FreeTube/files/8550156/abonelikler.csv) subscription .csv file
3. Get "Error: Request failed with status code 404"

### Expected Behavior

I expected that the subscriptions would **be in the app** after I imported them.

### Issue Labels

API issue, content not loading

### FreeTube Version

v0.16.0 Beta

### Operating System Version

Windows 11 Pro version 21H2

### Last Known Working FreeTube Version (If Any)

None

### Primary API used

Local API

### Additional Information

https://user-images.githubusercontent.com/81768350/164988863-f9729bf2-407a-4ccd-9d44-891c05092c9c.mp4

This is a video of me showcasing the bug.

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Lau-San:
I'm getting the exact same error here on Linux Mint 20.3

efb4f5ff-1298-471a-8973-3d47447115dc:
hi there have u tried to troubleshoot this issue like its described on https://docs.freetubeapp.io/usage/importing-subscriptions/#troubleshooting-issues

ChunkyProgrammer:
This issue is caused by one of the channels they're subscribed to having commas in their name.

> hi there have u tried to troubleshoot this issue like its described on https://docs.freetubeapp.io/usage/importing-subscriptions/#troubleshooting-issues

PR #3469, opened 2023-04-26
Fix vunerability in CSV import regex

absidue:
# Fix vunerability in CSV import regex

## Pull Request Type

- [x] Bugfix

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/security/code-scanning/62

## Description
GitHub's code scanning has been warning us about a potential vunerability in the CSV regex, so I finally decided to look into it when it flagged it again in #3455. Specifically an exponential/catastrophic backtracking one. If you are interested https://www.regular-expressions.info/catastrophic.html is a good read (the suggested fixes with atomic groups unfortunately don't work for us, as JavaScript doesn't support atomic groups).

https://regex101.com/ was very helpful in testing this as you can switch between different regex engines to try stuff out but also input test data and check if it correctly handles it.

## Testing 
I created this CSV based on the description of the vunerability given to us by GitHub:
> This part of the regular expression may cause exponential backtracking on strings starting with '"' and containing many repetitions of '!'.

Importing this CSV before this PR freezes FreeTube, so you'll have to force kill it from the task manager, with this PR it will now spit out errors from YouTube and Invidious about not being able to find the channel, without breaking anything.

[exponential-backtracking.csv](https://github.com/FreeTubeApp/FreeTube/files/11334948/exponential-backtracking.csv)

To test that it still works with a normal subscription CSV you can export, clear your subscriptions and import your own or use the one in this issue: #2219

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 45d679b84bfa8a797c6c586f0177efc4c3f498da

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -216,13 +216,13 @@ export default defineComponent({
       this.setProgressBarPercentage(0)
       let count = 0
 
-      const splitCSVRegex = /(?:,|\n|^)("(?:(?:"")*[^"]*)*"|[^\n",]*|(?:\n|$))/g
+      const splitCSVRegex = /(?:,|\n|^)("(?:(?:"")|[^"])*"|[^\n",]*|(?:\n|$))/g
 
       const ytsubs = youtubeSubscriptions.slice(1).map(yt => {
         return [...yt.matchAll(splitCSVRegex)].map(s => {
           let newVal = s[1]
           if (newVal.startsWith('"')) {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested [exponential-backtracking.csv](https://github.com/FreeTubeApp/FreeTube/files/11334948/exponential-backtracking.csv) and own exported csv

ChunkyProgrammer (approved):

PR #3483, opened 2023-05-03
Local API: Add support for hashtag pages

ChunkyProgrammer:
# Local API: Add support for hashtag pages

## Pull Request Type
- [x] Feature Implementation

## Description
This PR adds support for YouTube hashtag pages when the local api is enabled.

It is not added for Invidious because: 
- Invidious does not currently have api support for hashtag pages. See: https://github.com/iv-org/invidious/pull/3692

- The hashtag page is currently not working for invidious. See: https://github.com/iv-org/invidious/issues/3778

## Screenshots 
image

## Testing 
- type `https://www.youtube.com/hashtag/cats` in the search/nav bar
- scroll all the way down
- click fetch more

## Desktop
- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/views/Hashtag/Hashtag.js
+++ b/src/renderer/views/Hashtag/Hashtag.js
@@ -1,0 +1,99 @@
+import { defineComponent } from 'vue'
+import FtCard from '../../components/ft-card/ft-card.vue'
+import FtElementList from '../../components/ft-element-list/ft-element-list.vue'
+import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
+import FtLoader from '../../components/ft-loader/ft-loader.vue'
+import packageDetails from '../../../../package.json'
+import { getHashtagLocal, parseLocalListVideo } from '../../helpers/api/local'
+import { isNullOrEmpty } from '../../helpers/utils'
+export default defineComponent({
+  name: 'Hashtag',
+  components: {
+    'ft-card': FtCard,
+    'ft-element-list': FtElementList,
+    'ft-flex-box': FtFlexBox,
+    'ft-loader': FtLoader
+  },
+  data: function() {
+    return {
+      hashtag: '',
+      hashtagContinuationData: null,
+      videos: [],
+      apiUsed: 'local',
+      isLoading: true
+    }
+  },
+  computed: {
+    backendPreference: function () {
+      return this.$store.getters.getBackendPreference
+    },
+
+    backendFallback: function () {
+      return this.$store.getters.getBackendFallback
+    },
+
+    showFetchMoreButton() {
+      return !isNullOrEmpty(this.hashtagContinuationData)
+    }
+  },
+  watch: {
+    $route() {
+      this.isLoading = true
+      this.hashtag = ''
+      this.hashtagContinuationData = null
+      this.videos = []
+      this.apiUsed = 'local'
+      this.getHashtag()
+    }
+  },
+  mounted: async function() {
+    this.isLoading = true
+    await this.getHashtag()
+  },
+  methods: {
+    getHashtag: async function() {
+      const hashtag = this.$route.params.hashtag
+      if (this.backendFallback || this.backendPreference === 'local') {
+        await this.getLocalHashtag(hashtag)
+      } else {
+        this.getInvidiousHashtag(hashtag)
+      }
+      document.title = `${this.hashtag} - ${packageDetails.productName}`
+    },
+    getInvidiousHashtag: function(hashtag) {
+      this.hashtag = '#' + hashtag
+      this.apiUsed = 'Invidious'
+      this.isLoading = false
+    },
+
+    getLocalHashtag: async function(hashtag) {
+      const hashtagData = await getHashtagLocal(hashtag)
+      this.hashtag = hashtagData.header.hashtag
+      this.videos = hashtagData.contents.contents.filter(item =>
+        item.type !== 'ContinuationItem'
+      ).map(item =>
+        parseLocalListVideo(item.content)
+      )
+      this.apiUsed = 'local'
+      this.hashtagContinuationData = hashtagData.has_continuation ? hashtagData : null
+      this.isLoading = false
+    },
+
+    getLocalHashtagMore: async function() {
+      const continuation = await this.hashtagContinuationData.getContinuationData()
+      const newVideos = continuation.on_response_received_actions[0].contents.filter(item =>
+        item.type !== 'ContinuationItem'
+      ).map(item =>
+        parseLocalListVideo(item.content)
+      )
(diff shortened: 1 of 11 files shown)


PikachuEXE (commented):
Tested https://www.youtube.com/hashtag/pikachu
image

However I think we need more stuff tested. such as hashtag in video description
Example video with tags in description: https://youtu.be/v6jw6KRIOmw

Also here is screenshot for hashtag view when using IV
image

src/renderer/views/Hashtag/Hashtag.js
```diff
@@ -0,0 +1,99 @@
+import { defineComponent } from 'vue'
+import FtCard from '../../components/ft-card/ft-card.vue'
+import FtElementList from '../../components/ft-element-list/ft-element-list.vue'
+import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
+import FtLoader from '../../components/ft-loader/ft-loader.vue'
+import packageDetails from '../../../../package.json'
+import { getHashtagLocal, parseLocalListVideo } from '../../helpers/api/local'
+import { isNullOrEmpty } from '../../helpers/utils'
+export default defineComponent({
```
absidue, line 9:
```suggestion
import { isNullOrEmpty } from '../../helpers/utils'

export default defineComponent({
```

src/renderer/views/Hashtag/Hashtag.js
```diff
@@ -0,0 +1,99 @@
...
+      this.hashtagContinuationData = null
+      this.videos = []
+      this.apiUsed = 'local'
+      this.getHashtag()
+    }
+  },
+  mounted: async function() {
+    this.isLoading = true
+    await this.getHashtag()
+  },
```
absidue, line 52: I don't think Vue 2 supports async lifecycle hooks, so it won't `await` this hook, so it doesn't need to be async with await
```suggestion
  mounted: function() {
    this.isLoading = true
    this.getHashtag()
  },
```

static/locales/en-US.yaml
```diff
@@ -868,6 +868,11 @@ Downloading failed: 'There was an issue downloading "{videoTitle}"'
 Screenshot Success: Saved screenshot as "{filePath}"
 Screenshot Error: Screenshot failed. {error}
 
+Hashtag:
+  Hashtag: Hashtag
+  This hashtag does not currently have any videos: This hashtag does not currently
+      have any videos
+  You can only view hashtag pages through the local api: You can only view hashtag pages through the local api
```
ChunkyProgrammer, line 875:
```suggestion
  You can only view hashtag pages through the Local API: You can only view hashtag pages through the Local API
```

src/renderer/views/Hashtag/Hashtag.vue
```diff
@@ -0,0 +1,54 @@
...
+              {{ $t("Hashtag.This hashtag does not currently have any videos") }}
+            </p>
+          </ft-flex-box>
+          <ft-flex-box
+            v-else
+          >
+            <p
+              class="message"
+            >
+              {{ $t("Hashtag.You can only view hashtag pages through the local api") }}
```
ChunkyProgrammer, line 34:
```suggestion
              {{ $t("Hashtag.You can only view hashtag pages through the Local API") }}
```

PikachuEXE (dismissed):

src/renderer/views/Hashtag/Hashtag.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <div>
+    <ft-loader
+      v-if="isLoading"
+      :fullscreen="true"
+    />
+    <div v-else>
+      <ft-card>
+        <h1>{{ hashtag }}</h1>
```
absidue, line 9: Noticed that this title looks a lot larger than other places in FreeTube, turns out those use `h3` instead of `h1`.
```suggestion
        <h3>{{ hashtag }}</h3>
```

absidue (commented):
Would be cool if you could do something similar to what we do with the [channel handles](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/store/modules/utils.js#L293-L295), so that you can search for `#shorts` without needing to type out the full `https://youtube.com/hashtag/shorts` URL.

absidue:
The `#shorts` thing is just a nice to have, happy to approve without that too.

efb4f5ff-1298-471a-8973-3d47447115dc:
LGTM, just curious before approval. Is there a reason u didnt add video and channel count?

Capture

absidue:
Probably because YouTube returns it in this single line of text, which is annoying to parse (we need to parse it if we want it to be translatable) shorts-info

ChunkyProgrammer:
> Probably because YouTube returns it in this single line of text, which is annoying to parse (we need to parse it if we want it to be translatable) shorts-info

That and invidious currently does not parse that

ChunkyProgrammer:
> The `#shorts` thing is just a nice to have, happy to approve without that too.

I'll add that later tonight (if i have time)

efb4f5ff-1298-471a-8973-3d47447115dc:
Im not sure if i like searching for hashtags like that. With @[insert channel name] its clear u searching for the channel because it is mostly used for that. With hashtags i get other results if i search it normally vs this special hashtag tab. 

https://user-images.githubusercontent.com/73130443/236650158-139377bb-28ea-4143-94c9-b042bfce683e.mp4

Lets say im looking for that episode but only remembering finding it on yt using these keywords. On FT u can never find it again.

src/renderer/views/Hashtag/Hashtag.vue
```diff
@@ -0,0 +1,54 @@
...
+  <div>
+    <ft-loader
+      v-if="isLoading"
+      :fullscreen="true"
+    />
+    <div v-else>
+      <ft-card>
+        <h3>{{ hashtag }}</h3>
+      </ft-card>
+      <ft-card>
```
absidue, line 11: Should probably put the title and the content in the same card, so it matches the design of the rest of FreeTube
```suggestion
```

ChunkyProgrammer:
Okay, i think i'll remove the hashtag search option. 
This is the way youtube does it when you search a hashtag. 
Screenshots_2023-05-07-14-39-27

efb4f5ff-1298-471a-8973-3d47447115dc:
Is it possible for u to implement the #search come up in the search results in this PR?

ChunkyProgrammer:
> Is it possible for u to implement the #search come up in the search results in this PR?
I dont think it's supported by invidious nor youtube.js at the moment

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

PR #3523, opened 2023-05-09
Use POST for DASH requests

absidue:
# Use POST for DASH requests

## Pull Request Type

- [x] Feature Implementation

## Description
The official clients all use POST requests instead of GET, so we should too. Might improve loading speed a bit too.

## Testing 
Try various videos with DASH to check if they work

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** c93c1d065da230fdac040952ea73a49f0607d3da

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -22,9 +22,16 @@ import { getPicturesPath, showSaveDialog, showToast } from '../../helpers/utils'
 // videojs-http-streaming calls this hook everytime it makes a request,
 // so we can use it to convert the Range header into the range query parameter for the streaming URLs
 videojs.Vhs.xhr.beforeRequest = (options) => {
-  if (options.headers?.Range && new URL(options.uri).hostname.endsWith('.googlevideo.com')) {
-    options.uri += `&range=${options.headers.Range.split('=')[1]}`
-    delete options.headers.Range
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -22,9 +22,16 @@ import { getPicturesPath, showSaveDialog, showToast } from '../../helpers/utils'
...
+  // pass in the optional base so it doesn't error for `dashFiles/videoId.xml` (DASH manifest in dev mode)
+  if (new URL(options.uri, window.location.origin).hostname.endsWith('.googlevideo.com')) {
+    // The official clients use POST requests with this body for the DASH requests, so we should do that too
+    options.method = 'POST'
+    options.body = 'x\x00' // protobuf: { 15: 0 } (no idea what it means but this is what YouTube uses)
+
+    if (options.headers?.Range) {
+      options.uri += `&range=${options.headers.Range.split('=')[1]}`
+      delete options.headers.Range
+    }
```
PikachuEXE, line 34: I think we can consider removing this
Unless YT clients uses this form
image
image

absidue: YouTube doesn't use the Range header (they were even using it for throttling a while back), so we should stick to using the query parameter too.

ChunkyProgrammer (approved):

PR #3488, opened 2023-05-04
Allow using piped api for comments and playlists

ChunkyProgrammer:
# Allow using piped api for comments

## Pull Request Type
- [x] Feature Implementation

## Related issue
#1527

## Description
This PR will add support for using the piped api for comments.

### Comments
- set preferred api backend to piped
- go to a video
- load comments
- load replies
- load more comments

### Playlists
-set preferred api backend to piped
- search a playlist
- go to the playlist page
- load more
- click on a video
- notice that all playlist videos are shown (unlike Invidious)

## Desktop
- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** 0.18.0

PR #3545, opened 2023-05-14
Use script to create Invidious instance list file

ChunkyProgrammer:
# Use script to create Invidious instance list file

## Pull Request Type
- [x] Other - Development chore

## Related issue
https://github.com/FreeTubeApp/FreeTube/pull/3488#discussion_r1188884002
(I'll add the piped instance list script to the same file as the one for invidious)

## Description
Instead of manually updating the Invidious instances every once in a while, we now have a script to update the lists for us

## Testing 
- turn off wifi
- start freetube
- turn wifi on
- set an invidious instance

## Desktop
- **OS:** Windows
- **OS Version:** 11

## Additional context
I also simplified the instance list file


PikachuEXE (dismissed):

_scripts/getInstances.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs/promises')
+const invidiousApiUrl = 'https://api.invidious.io/instances.json'
+
+fetch(invidiousApiUrl).then(e => e.json()).then(res => {
+  const data = res.filter((instance) => {
+    return !(instance[0].includes('.onion') ||
+      instance[0].includes('.i2p') ||
+      !instance[1].api ||
+      (!process.env.IS_ELECTRON && !instance[1].cors))
```
absidue, line 9: `process.env.IS_ELECTRON` will always be `undefined` here, in the app we set it with webpack. (you just copy pasted it without checking it didn't you 🙈)

PikachuEXE: Running `yarn run get-instances` yields no error so I didn't notice this 😬

efb4f5ff-1298-471a-8973-3d47447115dc:
List seems to lgtm but i have one question, https://yt.floss.media has cors disabled and is included in the list. Does this matter to us in anyway?

PikachuEXE:
I guess it's filtered out for web build
image
image

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Should we fix it in this PR or followup?

absidue (approved):
As that issue wasn't introduced by this pull request, it can be fixed in a separate pull request and shouldn't block merging this one.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Just realised the backup file will never be used in the web build, it will error before that when it tries to access the filesystem. Needs to be done in a similar way to how the region data works, if we want it to work.
> 
> Just to be clear that's a thing that has existed before this PR.

@absidue could u create an issue for it. Lets not let this disappear into the GH void

PR #3527, opened 2023-05-11
Fix locale files so yaml parsing won't cause errors

ChunkyProgrammer:
# Try to fix Weblate issue

## Pull Request Type
- [x] Bugfix
- [x] Other- fixed json linting and added basic yaml linting

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/3522

## Description
Fix locale yaml files so weblate can parse them + add linting for yaml/yml and fix json linting

## Testing 
`yarn run dev` with this patch

```diff

diff --git a/_scripts/ProcessLocalesPlugin.js b/_scripts/ProcessLocalesPlugin.js
index 58bf5acc..5ac86b24 100644
--- a/_scripts/ProcessLocalesPlugin.js
+++ b/_scripts/ProcessLocalesPlugin.js
@@ -1,4 +1,4 @@
-const { existsSync, readFileSync } = require('fs')
+const { existsSync, readFileSync, readdirSync } = require('fs')
 const { brotliCompress, constants } = require('zlib')
 const { promisify } = require('util')
 const { load: loadYaml } = require('js-yaml')
@@ -72,7 +72,10 @@ class ProcessLocalesPlugin {
   loadLocales() {
     this.locales = []
 
-    const activeLocales = JSON.parse(readFileSync(`${this.inputDir}/activeLocales.json`))
+    const activeLocales = readdirSync(this.inputDir)
+      .filter(e => e.endsWith('.yaml'))
+      .map(e => e.replace('.yaml', ''))
+    // const activeLocales = JSON.parse(readFileSync(`${this.inputDir}/activeLocales.json`))
 
     for (const locale of activeLocales) {
       const contents = readFileSync(`${this.inputDir}/${locale}.yaml`, 'utf-8')

```

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** 0.18.0

Issue #3522, opened 2023-05-09
[Bug]: Weblate issue


rddim:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

When I try do translate the new strings on Weblate I see an error message at top of the page.

https://hosted.weblate.org/projects/free-tube/translations/#alerts

### Expected Behavior

I have expect to have no errors

### Issue Labels

text/string issue

### FreeTube Version

v0.18.0-nightly-2736

### Operating System Version

Win 7

### Installation Method

PortableApps

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

v0.18.0-nightly-2736

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
Related https://github.com/FreeTubeApp/FreeTube/issues/673#issuecomment-1527615575

efb4f5ff-1298-471a-8973-3d47447115dc:
@rddim it should be fixed now, could u verify this for us?

efb4f5ff-1298-471a-8973-3d47447115dc:
Issue isnt gone yet

```
static/locales/as.yaml	'could not find expected \':\' in "<unicode string>", line 54, column 1:\n Change Format:\n ^ (line: 54)'
static/locales/hi.yaml	'could not find expected \':\' in "<unicode string>", line 210, column 1:\n Change Format:\n ^ (line: 210)'
static/locales/ku.yaml	'could not find expected \':\' in "<unicode string>", line 200, column 1:\n Comments:\n ^ (line: 200)'
static/locales/or.yaml	'could not find expected \':\' in "<unicode string>", line 144, column 1:\n Toggle Theatre Mode: \'\'\n ^ (line: 144)'
[...]

PR #3546, opened 2023-05-14
Fix locale files so yaml parsing won't cause errors (Part 2)

ChunkyProgrammer:
# Fix locale files so yaml parsing won't cause errors (Part 2)

## Pull Request Type
- [x] Bugfix

## Related issue
#3522 

## Description
Fix locale yaml files so weblate can parse them

## Testing 
See #3527 

## Desktop
- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/static/locales/as.yaml
+++ b/static/locales/as.yaml
@@ -48,9 +48,6 @@ Video:
   Published:
     Jun: ''
     Months: ''
-Videos:
-  #& Sort By
-{}
 Change Format:
   Use Dash Formats: ''
 Comments:
--- a/static/locales/hi.yaml
+++ b/static/locales/hi.yaml
@@ -204,9 +204,6 @@ Video:
   Published:
     Sep: ''
     Year: ''
-Videos:
-  #& Sort By
-{}
 Change Format:
   Change Media Formats: ''
 Share:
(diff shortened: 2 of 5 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue:
I suspect this will just keep coming back until it's either fixed on the weblate side or we manually rip out all empty strings, so that weblate can't do it and break it.

PikachuEXE:
Looking at commit https://github.com/FreeTubeApp/FreeTube/commit/40d52594a3982b388e2ee95887a9600138ebb537
I think this might be caused by webplate trying to convert
```yaml
Videos:
  #& Sort By
  Sort By:
    Oldest: ''
    #& Most Popular
#& Playlists
```
into
```yaml
Videos: {}
```
but somehow bugged into 
```
Videos:
  #& Sort By
{}
```

So let's remove those and see if the same thing happens again

ChunkyProgrammer:
> I suspect this will just keep coming back until it's either fixed on the weblate side or we manually rip out all empty strings, so that weblate can't do it and break it.

Should we ask Preston to disable the weblate plugin to delete empty values?

PR #3557, opened 2023-05-16
Ensure screen saver blocker removed when window closed

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Similar to #1217 
Might work for reload case in #3286 too

## Description

`beforeDestroy` doesn't work for window close
https://github.com/nklayman/vue-cli-plugin-electron-builder/issues/389

## Screenshots 

Lazy! (Well or a video with at least 1 min is needed, definitely too big to be uploaded)

## Testing 

- Change screen saver to minimum (1 min for MacOS I think)
- Use FT to play random video
- Close window
- Wait 1 min to ensure screen saver/sleep triggered

## Desktop

- **OS:** MacOS
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -357,10 +357,7 @@ export default defineComponent({
       navigator.mediaSession.playbackState = 'none'
     }
 
-    if (process.env.IS_ELECTRON && this.powerSaveBlocker !== null) {
-      const { ipcRenderer } = require('electron')
-      ipcRenderer.send(IpcChannels.STOP_POWER_SAVE_BLOCKER, this.powerSaveBlocker)
-    }
+    this.stopPowerSaveBlocker()
   },
   methods: {
     initializePlayer: async function () {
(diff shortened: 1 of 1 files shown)


src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -42,6 +42,10 @@ const VHS_BANDWIDTH_VARIANCE = videojs.Vhs.BANDWIDTH_VARIANCE
 
 export default defineComponent({
   name: 'FtVideoPlayer',
+  beforeRouteLeave: function (_to, _from, next) {
+    window.removeEventListener('beforeunload', this.stopPowerSaveBlocker)
```
PikachuEXE, line 46: I am not sure if this can/should be moved into `beforeDestroy`
But this is how it's done currently in `Watch` view

absidue: This has to be moved to `beforeDestroy`, the route hooks only get triggered in components that are directly inside the router view. So for us those are the components in the `src/renderer/views`. Additionally the player gets destroyed and a new instance created when you change formats (maybe we can change that in the future, but for the moment that makes it easier for us), so you want this removed as soon as it's destroyed, so that you get rid of the event handler for all instances of the player.

PikachuEXE: Updated

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer need more approval chunks

ChunkyProgrammer (approved):

Issue #3532, opened 2023-05-13
[Bug]: `Uncaught TypeError: Cannot read properties of null (reading 'play')`


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. i watched a few videos back to back
2. used shortcut ALT + <- to go back
3. See error

`Uncaught TypeError: Cannot read properties of null (reading 'play')`

https://github.com/FreeTubeApp/FreeTube/assets/73130443/cd4c1491-328f-4e6c-9ff4-bccb8ed5245b

### Expected Behavior

No error.

### Issue Labels

inconsistent behavior

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/a4d45b5fa8a8a1726808c9eebe307643f8dde9c9

### Operating System Version

Win10 22H2

### Installation Method

.exe

### Primary API used

Local API

### Additional Information

@absidue comment on this matter, retrieved from IRC.

> that's presumably caused by the autoplay functionality currently relies on timeout/delays in the player and watch page, rather than using the autoplay functionality baked into video.js, So when the player is destroyed before the timeout fires, it tries calling play on something that doesn't exist anymore

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

PikachuEXE:
I got a different error but probably related?
![image](https://github.com/FreeTubeApp/FreeTube/assets/1018543/4920a24d-9b77-45b7-b99d-6a67afcb6dc4)

efb4f5ff-1298-471a-8973-3d47447115dc:
I think it the same but just outputting differently

PikachuEXE:
Not exactly
Your error is caused by `this.player.play()` while `this.player` is `null`
[...]

PR #3563, opened 2023-05-17
local API: Support multiple audio tracks

### LINKED ISSUE #2419: [Feature Request]: Choose audio track
### Guidelines

### Problem Description

Apparently, YouTube is currently testing a new feature to let users choose between several audio tracks provided by the video creator. For example, the “I Built Willy Wonka's Chocolate Factory!” [video](https://www.youtube.com/watch?v=Hwybp38GnZw) by Mr. Beast has got russian, Spanish, Hindi, French and Portuguese translations. However, it's not possible to choose one within FreeTube. In fact, YouTube apparently automatically selects a language: In my case, I got the russian version for some reason.

### Proposed Solution

Add an audio track switcher.

### Alternatives Considered

...

### Issue Labels

new feature

### Additional Information

_No response_

absidue:
# local API: Support multiple audio tracks

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #2419

https://github.com/iv-org/invidious/pull/3620 (WIP Invidious PR to support multiple audio tracks)

https://github.com/LuanRT/YouTube.js/pull/308, https://github.com/LuanRT/YouTube.js/pull/338, https://github.com/LuanRT/YouTube.js/pull/366, https://github.com/LuanRT/YouTube.js/pull/402 (relevant YouTube.js PRs)

## Description
Support for audio track switching is finally here 🥳!
_Disclaimer: local API only 🙈_

The DASH side of things is handled by videojs-http-streaming, through the information in the DASH manifest (related YouTube.js PRs linked above, if you are interested in what I did over there). As we handle the audio formats, we also need to handle the multiple audio tracks ourselves, so that's what the majority of the code in this pull request is for, as the legacy formats don't have multiple audio tracks, we don't have to do anything about them.

Unfortunately we can't support them for Invidious yet, as Invidious doesn't support them itself yet, however they do have a work in progress pull request to add support (linked above). This pull request is written in a way to keep the Invidious stuff unaffected by the changes.

## Screenshots

Audio formats:
audio

DASH formats:
dash

## Testing
For the local API test that it works and for Invidious check that the behaviour is still the same as it was before.
Test both DASH and audio formats to check that it works and the legacy ones to check that they didn't break.

same language, with original, dubbed and audio descriptions:
https://youtu.be/Kn56bMZ9OE8

mutliple languages:
https://youtu.be/WTOm65IZneg

normal video:
https://youtu.be/Xzhp1HTt43M

## Desktop
- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** d98a66c5c2186835c821e6ae1950dd7ef814ebe7

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -82,6 +82,10 @@ export default defineComponent({
     chapters: {
       type: Array,
       default: () => { return [] }
+    },
+    audioTracks: {
+      type: Array,
+      default: () => ([])
     }
   },
   data: function () {
@@ -89,6 +93,7 @@ export default defineComponent({
       powerSaveBlocker: null,
       volume: 1,
       muted: false,
+      /** @type {(import('video.js').VideoJsPlayer|null)} */
       player: null,
       useDash: false,
       useHls: false,
@@ -135,6 +140,7 @@ export default defineComponent({
             'pictureInPictureToggle',
             'toggleTheatreModeButton',
             'fullWindowButton',
+            'audioTrackButton',
             'qualitySelector',
             'fullscreenToggle'
           ]
@@ -309,6 +315,24 @@ export default defineComponent({
       this.toggleScreenshotButton()
     }
   },
+  created: function () {
+    this.dataSetup.playbackRates = this.playbackRates
+
+    if (this.format === 'audio') {
+      // hide the PIP button for the audio formats
+      const controlBarItems = this.dataSetup.controlBar.children
+      const index = controlBarItems.indexOf('pictureInPictureToggle')
+      controlBarItems.splice(index, 1)
+    }
+
+    if (this.format === 'legacy' || this.audioTracks.length === 0) {
+      // hide the audio track selector for the legacy formats
+      // and Invidious(it doesn't give us the information for multiple audio tracks yet)
+      const controlBarItems = this.dataSetup.controlBar.children
+      const index = controlBarItems.indexOf('audioTrackButton')
+      controlBarItems.splice(index, 1)
+    }
+  },
   mounted: function () {
     const volume = sessionStorage.getItem('volume')
     const muted = sessionStorage.getItem('muted')
@@ -322,8 +346,6 @@ export default defineComponent({
       // otherwise 'false' will be returned as true as well
       this.muted = (muted === 'true')
     }
-
-    this.dataSetup.playbackRates = this.playbackRates
 
     if (this.format === 'dash') {
       this.determineDefaultQualityDash()
@@ -368,13 +390,6 @@ export default defineComponent({
         if (!this.useDash) {
           qualitySelector(videojs, { showQualitySelectionLabelInControlBar: true })
           await this.determineDefaultQualityLegacy()
-        }
-
-        if (this.format === 'audio') {
-          // hide the PIP button for the audio formats
-          const controlBarItems = this.dataSetup.controlBar.children
-          const index = controlBarItems.indexOf('pictureInPictureToggle')
-          controlBarItems.splice(index, 1)
         }
 
         // regardless of what DASH qualities you enable or disable in the qualityLevels plugin
@@ -424,6 +439,36 @@ export default defineComponent({
           const qualityLevels = this.player.qualityLevels()
           qualityLevels.on('addqualitylevel', ({ qualityLevel }) => {
             qualityLevel.enabled = qualityLevel.bitrate === this.selectedBitrate
+          })
+        }
+
+        // for the DASH formats, videojs-http-streaming takes care of the audio track management for us,
+        // thanks to the values in the DASH manifest
+        // so we only need a custom implementation for the audio only formats
+        if (this.format === 'audio' && this.audioTracks.length > 0) {
+          /** @type {import('../../views/Watch/Watch.js').AudioTrack[]} */
+          const audioTracks = this.audioTracks
+
+          const audioTrackList = this.player.audioTracks()
+          audioTracks.forEach(({ id, kind, label, language, isDefault: enabled }) => {
+            audioTrackList.addTrack(new videojs.AudioTrack({
+              id, kind, label, language, enabled,
+            }))
+          })
+
+          audioTrackList.on('change', () => {
+            let trackId
+            // doesn't support foreach so we need to use an indexed for loop here
+            for (let i = 0; i < audioTrackList.length; i++) {
+              const track = audioTrackList[i]
+
+              if (track.enabled) {
+                trackId = track.id
+                break
+              }
+            }
+
+            this.changeAudioTrack(trackId)
           })
         }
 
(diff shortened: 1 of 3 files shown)


src/renderer/views/Watch/Watch.js
```diff
@@ -881,6 +940,42 @@ export default defineComponent({
       }
     },
 
+    /**
+     * @param {import('youtubei.js').Misc.Format[]} audioFormats
+     * @returns {AudioSource[]}
+     */
+    createLocalAudioSourceList: function (audioFormats) {
+      return audioFormats.sort((a, b) => {
+        return a.bitrate - b.bitrate
```
ChunkyProgrammer, line 949: Why not return `b.bitrate - a.bitrate` so you dont have to reverse it later?

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -835,6 +880,51 @@ export default defineComponent({
...
+        // need to call play to restore the player state, even if we want to pause afterwards
+
+        if (isPaused) {
+          this.player.play()
+            .then(() => {
+              this.player.pause()
+            })
+        } else {
+          this.player.play()
+        }
```
PikachuEXE, line 922:
```suggestion
        this.player.play().then(() => {
          if (isPaused) { this.player.pause() }
        })
```

PikachuEXE (commented):
Minor

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Please move the audio switch icon to the left of the caption. The icons in the player a grouped a bit. On the right side there is everything related to the window so fullscreen, fullwindow, theater and pip (dont look at the qualities one thats an exception -.- )

efb4f5ff-1298-471a-8973-3d47447115dc:
Some questions:

When u change format it will reset to original language why is that? I mean if i set my video to hindi and for some reason i want to switch to audio now i have to switch it to hindi again. 

https://github.com/FreeTubeApp/FreeTube/assets/73130443/75e5655c-b573-4c4d-90c9-3ac8fbf6699c

Why is the sorting on our side different then YT?

https://github.com/FreeTubeApp/FreeTube/assets/73130443/6acc19f2-45a3-47b2-a362-6d43e50ce89d

absidue:
The current format switching implementation completely destroys the player and recreates it with the new formats, this makes it a lot easier to have things like the default quality selection working and using the right quality selector (the DASH formats use a different one than the audio and legacy ones, as switching qualities with DASH works completely differently in video.js than it does with single file formats). However that means that we have to explicitly back up the settings we want to copy over from the previous player and restore them on the new player instance.

Currently the only player information we backup and reapply for format switches, is the playback position.

The volume and muted state get carried over because they are stored in the current window so that they are the same across videos. The subtitle settings e.g. font size, also get reapplied as they are stored with the other app settings in the settings database).

absidue:
As for the ordering, we don't do any sorting, we just use them in the order they are in the response from YouTube. Don't think it really matters, as YouTube's sorting doesn't seem to make sense either (it's not alphabetically or anything like that).

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

src/renderer/views/Watch/Watch.js
efb4f5ff-1298-471a-8973-3d47447115dc: @absidue

PikachuEXE:
Got this when 
- https://youtu.be/Kn56bMZ9OE8
- switch to legacy
- switch to audio format
- switch quality

Cannot reproduce reliably 
image

efb4f5ff-1298-471a-8973-3d47447115dc:
Isnt thag the same as https://github.com/FreeTubeApp/FreeTube/issues/3532

absidue:
Yes, that's an issue that existed before this PR.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

src/renderer/views/Watch/Watch.js
absidue: That was the existing logic, I didn't make any changes to it other than removing the arrow function around the switch statement.

I'll take a look if that's possible to do, just need to keep in mind that the list of audio formats varies per video, usually there are 4. For videos that only go up to 480,p there are only the low and medium one and some other videos have way more than 4, so only the 4 lowest bitrate ones get a text label and all the others get a bitrate label.

absidue: If I do the logic change, I'll add more test cases, so you'll have to test those too, to check that it didn't break existing behaviour.

absidue: We would have to get rid of the switch statement and do a bunch of stuff based on array length, as we would need to count from the end of the array instead of the start. It's doable, just makes the logic more complicated.

### Current implementation (sort lowest to highest and reverse later):
#### 2 formats:
| Index | Label |
| --- | --- |
| `0` | Low |
| `1` | Medium |

#### 4 formats:
| Index | Label |
| --- | --- |
| `0` | Low |
| `1` | Medium |
| `2` | High |
| `3` | Best |

#### 6 formats:
| Index | Label |
| --- | --- |
| `0` | Low |
| `1` | Medium |
| `2` | High |
| `3` | Best |
| `4` | {bitrate} |
| `5` | {bitrate} |

### Suggestion (sort highest to lowest):
#### 2 formats:
| Index | Label |
| --- | --- |
| `0` | Medium |
| `1` | Low |

#### 4 formats:
| Index | Label |
| --- | --- |
| `0` | Best |
| `1` | High |
| `2` | Medium |
| `3` | Low |

#### 6 formats:
| Index | Label |
| --- | --- |
| `0` | {bitrate} |
| `1` | {bitrate} |
| `2` | Best |
| `3` | High |
| `4` | Medium |
| `5` | Low |

ChunkyProgrammer (approved):

src/renderer/views/Watch/Watch.js
efb4f5ff-1298-471a-8973-3d47447115dc: @absidue @ChunkyProgrammer could one of u create an issue for this so it doesnt get lost in the GH void

absidue: No, this is not a change worth making.

Ein-Tim:
When will this feature be actually in the app?

absidue:
@Ein-Tim it already is, if you switch to the nightly builds, you can use it.

https://docs.freetubeapp.io/development/nightly-builds

hboetes:
Interesting case: With this video https://youtu.be/S4L_0CDsd1I which definitely also has an English audio track, I only get the French track whilst using a French invidious mirror.

absidue:
@hboetes As mentioned in the pull request title and description they are only supported on the local API, Invidious doesn't give us enough information yet.

hboetes:
> @hboetes As mentioned in the pull request title and description they are only supported on the local API, Invidious doesn't give us enough information yet.

@absidue, I am using the local API.

absidue:
You just said you are using a french Invidious mirror though?

hboetes:
I was, but that's because whenever I clear the invidious mirror, after the next time FT starts up another invidious mirror is used. Is it even possible to disable invidious?
https://0x0.st/HLJJ.png

absidue:
When the local API is set to preferred Invidious is only used if something goes wrong with the local API and "Revert to non-preferred backend on failure" is turned on.

You are presumably using 0.18.0 which was released in November 2022, this pull request was merged in May this year.

hboetes:
Nope, downloaded a nightly yesterday.
```
%  rpm -q freetube
freetube-0.18.0~nightly~3260-1.x86_64
```

PR #3565, opened 2023-05-19
Update README.md

ovari:


ovari:
https://hosted.weblate.org/widgets/free-tube/?lang=&component=glossary#multi

ovari:
1. To enable people to easily see what languages FreeTube is available in
2. To encourage new, or additional translations to, languages (as clicking on the image will open Weblate at the correct location to translate FreeTube)
3. An image is a thousand words

Thank you

absidue:
We want to know why this pull request is a positive change, currently from our point of view, it just bloats the README making it excessively longer.

1. As weblate allows people to add new languages without any supervision or ability to control it from our side, we have a whitelist inside FreeTube, so the list of languages on weblate is not reflective of the languages you can actually use inside FreeTube.

Point 2 is irrelevant as you didn't change the link and point 3 still doesn't give us a good reason to add that **gigantic** image to the README.

ovari:
Thank you for your comments and understanding.

PR #3591, opened 2023-05-27
Add subsections to the distraction free settings

absidue:
# Add subsections to the distraction free settings

## Pull Request Type

- [x] Feature Implementation

## Description
We have a lot of distraction free settings, which makes it hard to find the ones you need but also makes it confusing because some are similarly named but apply to different parts of the app. To solve that, I decided to split them up into subsections with the following headings:
- Side Bar
- Channel Pages
- Watch Pages
- General

Settings in the general subsection apply to multiple places in FreeTube, e.g. "Hide Video Views" applies to video lists and on the watch page.

Please let me know what you think of the headings and whether you want me to reword them.

## Screenshots 
before:
before

after:
after

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 8dbad618f9549ee075fe70572c07e7a7a978ced6

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -317,6 +317,11 @@ Settings:
     Fetch Automatically: Fetch Feed Automatically
   Distraction Free Settings:
     Distraction Free Settings: Distraction Free Settings
+    Sections:
+      Side Bar: Side Bar
+      Channel Pages: Channel Pages
+      Watch Pages: Watch Pages
+      General: General
     Hide Video Views: Hide Video Views
     Hide Video Likes And Dislikes: Hide Video Likes And Dislikes
     Hide Channel Subscribers: Hide Channel Subscribers
--- a/src/renderer/components/distraction-settings/distraction-settings.css
+++ b/src/renderer/components/distraction-settings/distraction-settings.css
@@ -1,3 +1,7 @@
+.groupTitle {
+  text-align: center;
+}
+
 @media only screen and (max-width: 800px) {
   br.hide-on-mobile {
     display: none;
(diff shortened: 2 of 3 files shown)


static/locales/en-US.yaml
```diff
@@ -317,6 +317,11 @@ Settings:
     Fetch Automatically: Fetch Feed Automatically
   Distraction Free Settings:
     Distraction Free Settings: Distraction Free Settings
+    Sections:
+      Side Bar: Side Bar
+      Channel Pages: Channel Pages
```
PikachuEXE, line 322: Channels? Channel Views?
I am not sure if we should call it "Pages"

static/locales/en-US.yaml
```diff
@@ -317,6 +317,11 @@ Settings:
...
+      Watch Pages: Watch Pages
```
PikachuEXE, line 323: `Watch Pages` sounds even more weird than `Channel Pages`

PikachuEXE (commented):
Pages?

static/locales/en-US.yaml
PikachuEXE: After some thought I can go with `Channel Page` or `Channel View`
No `s` (that's the point that makes me feel weird)

static/locales/en-US.yaml
PikachuEXE: Same as above `Watch Page` or `Watch View`

static/locales/en-US.yaml
absidue: Decided to go with page instead of view, as we use page in other strings too.

static/locales/en-US.yaml
absidue: see above

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Damn dis is prettyyy, lgtm!

ChunkyProgrammer (approved):

PR #3639, opened 2023-06-07
Fix `proxyVideos` setting for legacy formats & hls urls

MarmadileManteater:
# Fix `proxyVideos` setting for legacy formats & hls urls

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Currently, when the `proxy videos through invidious` setting is enabled, only DASH manifests are effected, but legacy formats and hls urls don't proxy.  This PR addresses this by replacing the `host` of stream urls with the current invidious instance and adding the appropriate `host` URL param. This is accomplished in a map for the `activeSourceList` which ensures all legacy formats go through the proxy when `proxyVideos` is true. Additionally, the links that come back in hls manifests from invidious don't always respect the `local` flag, so I added code to `videojs.Vhs.xhr.beforeRequest` to ensure that all requests that videojs makes go through the current invidious instance.

## Screenshots 
_(legacy format)_
|before|after|
|--|--|
|before-legacy|after-legacy|

_(hls url)_
|before|after|
|--|--|
|before-hls|after-hls|

## Testing 
_(this should probably be tested with both `invdious` and `local` API preference)_
1. Enable `Proxy Videos Through Invidious`
2. Navigate to a video (ex: https://youtu.be/SqB0lUcqFbA)
3. Select `Use legacy formats`
4. Open the network tab and ensure the request is being sent to the currently selected invidious instance instead of something like `rr--3.googlevideo.com`
5. Navigate to a live stream with a hls url (ex: https://youtu.be/jfKfPfyJRdk)
6. Open the network tab and ensure the request is being sent to the currently selected invidious instance

_(as an added note, if you test this in a web browser using `yarn dev:web`, you don't have to check the network tab because the video will just fail to play if it is not proxied)_

## Desktop

 - OS: Windows 10
- **FreeTube version:** 5295934

## Additional context
Currently, there is a much hackier version of this fix in FTCordova, but I just never got around to making an upstream PR about this issue until now even though it technically effects electron builds.


PikachuEXE:
Local API (with or without fallback) - https://youtu.be/jfKfPfyJRdk will result in 
image

No issues when using Invidious API
Preferred format set to Legacy during testing

MarmadileManteater:
> Local API (with or without fallback) - https://youtu.be/jfKfPfyJRdk will result in image
> 
> No issues when using Invidious API Preferred format set to Legacy during testing

When I was refactoring my code just before I made the PR, 🤦‍♀️I accidentally moved the logic that modifies the URI of the request to before the condition that checks if the host contains `googlevideo.com`, and I believe that is what caused the issues with the Local API.

PikachuEXE (approved):
No issues using all APIs

Tested with https://youtu.be/jfKfPfyJRdk

absidue (approved):

PR #3640, opened 2023-06-08
Playlists

sec-19:
# Title

## Pull Request Type

- [ ] Bugfix
- [ **] Feature Implementation
- [ ] Documentation
- [ ] Other

## Desktop

- **OS:*Debian Linux amd64*
- **OS Version:**
- **FreeTube version:*0.18.0*

## Additional context

Dear Team , Please Add the Functionality To add Whole Playlist Of A channel to Either a Separate Bookmarks Tab or Already Existing Playlist Tab, Because As for Now we can Only add A single Video To playlist Tab, Which is a tedious Task.
Or In more simple terms , We should be able to add say for example (A study Channel's  Particular Playlist  as a whole to Playlists Tab, So that We can Straight Go to Playlist Tab And find That Playlist as whole series of Videos).
I hope My request would be taken care Of! Thanks...
Screenshot from 2022-11-19 16-06-13


efb4f5ff-1298-471a-8973-3d47447115dc:
U copied an outdated branch and didnt commit anything to it.

PR #3700, opened 2023-06-25
Fix focus related race conditions

### LINKED ISSUE #3444: [Bug]: `Uncaught TypeError: Cannot read properties of undefined (reading 'focus')`
### Guidelines

### Describe the bug

1. Go to trending tab
2. press and hold `r`
3. see error in console

https://user-images.githubusercontent.com/73130443/232827473-d73eb784-6811-4d12-b12d-18474ac57468.mp4

### Expected Behavior

No error.

### Issue Labels

feature stopped working, inconsistent behavior

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/2984df111222e4c849f099e8b0e974f96a998f60

### Operating System Version

win10 22h2

### Installation Method

.exe

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

absidue:
# Fix focus related race conditions

## Pull Request Type

- [x] Bugfix

## Related issue
closes #3444

## Description
Fixes errors caused by focus being called on elements that Vue hasn't created yet. `setTimeout()` with no delay, just makes code run asynchronously, regardless of whether Vue has had time to create the elements or not. When you hold the <kbd>R</kbd> key on the trending tab, you trigger the refresh before Vue has had time to modify any HTML, so the element that we want to focus is never created. You might think that using Vue's `nextTick()` function would be more appropriate, however as that delays until after the next DOM update, it would mean that in this case the code would be triggered multiple times (however many times the refresh finished while you were holding the <kbd>R</kbd> key), when you finally let go of the <kbd>R</kbd>.

While I was at it, I decided to make the same change in other places that we call focus with a delay, which should hopefully fix similar errors caused by spamming/abusing FreeTube in those areas (looking at you @efb4f5ff-1298-471a-8973-3d47447115dc with your auto clicker 🙄).

## Testing 
On the trending page, hold the <kbd>R</kbd> key and check that no errors appear in the console.

### DIFF
--- a/src/renderer/components/ft-icon-button/ft-icon-button.js
+++ b/src/renderer/components/ft-icon-button/ft-icon-button.js
@@ -94,8 +94,8 @@ export default defineComponent({
           // wait until the dropdown is visible
           // then focus it so we can hide it automatically when it loses focus
           setTimeout(() => {
-            this.$refs.dropdown.focus()
-          }, 0)
+            this.$refs.dropdown?.focus()
+          })
         }
       } else {
         this.$emit('click')
--- a/src/renderer/components/ft-profile-selector/ft-profile-selector.js
+++ b/src/renderer/components/ft-profile-selector/ft-profile-selector.js
@@ -42,7 +42,7 @@ export default defineComponent({
         // wait until the profile list is visible
         // then focus it so we can hide it automatically when it loses focus
         setTimeout(() => {
-          this.$refs.profileList.$el.focus()
+          this.$refs.profileList?.$el?.focus()
         })
       }
     },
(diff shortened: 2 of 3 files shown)


PikachuEXE (approved):

PikachuEXE:
I found it strange to keep refreshing when I hold down `R` key
Isn't this another bug? (not saying it should be fixed in this PR)

PikachuEXE:
I only know now `keydown` event is also fired repeatedly when button is being hold

I guess FT is not using the `repeat` property?
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat

Reloading continuously on key holding is an unexpected behaviour for me, not sure what other people think

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3708, opened 2023-06-26
Support YouTube's new layout for shorts only playlists

absidue:
# Support YouTube's new layout for shorts only playlists

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
https://github.com/yt-dlp/yt-dlp/pull/7425

## Description
YouTube is now using a new layout for playlists that only contain shorts, it's the same one which is used for the shorts tab on channel pages and therefore has the same missing information: published date, channel (id and name). On channel pages we can use the name and id of the current channel, however as playlists can contain videos from multiple channels we can't do that.

Unfortunately it looks like YouTube hasn't implemented continuations for the new layout yet, so you will only see 100 videos, even if the playlist has more. Hopefully when they do add them, it'll just start working, otherwise we can do a follow up pull request to support those then.

As the channel information is missing, this pull request also hides the copy channel link options in the 3 dots menu.

This only makes changes to the local API, any changes for the Invidious API will need to be done on the Invidious side.

## Screenshots 
before:
before

after:
after

## Testing 
Linus Tech Tips autogenerated channel shorts playlist (95 videos): https://youtube.com/playlist?list=UUSHXuqSBlHAE6Xw-yeJA0Tunw
Gibi ASMR autogenerated channel shorts playlist (135 videos, only 100 will show up): https://youtube.com/playlist?list=UUSHE6acMV3m35znLcf0JGNn7Q

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -58,8 +58,8 @@ export default defineComponent({
     return {
       id: '',
       title: '',
-      channelName: '',
-      channelId: '',
+      channelName: null,
+      channelId: null,
       viewCount: 0,
       parsedViewCount: '',
       uploadedTime: '',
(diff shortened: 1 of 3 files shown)


src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -190,30 +190,34 @@ export default defineComponent({
...
-          {
-            label: this.$t('Video.Open Channel in YouTube'),
-            value: 'openYoutubeChannel'
-          },
-          {
-            label: this.$t('Video.Open Channel in Invidious'),
-            value: 'openInvidiousChannel'
           }
         )
+        if (this.channelId !== null) {
```
PikachuEXE, line 195:
```suggestion
        if (this.channelId != null) {
```

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -78,13 +78,14 @@
...
         <router-link
+          v-if="channelId !== null"
           class="channelName"
           :to="`/channel/${channelId}`"
         >
           <span>{{ channelName }}</span>
         </router-link>
         <template v-if="!isLive && !isUpcoming && !isPremium && !hideViews">
-          <span class="viewCount"> • {{ parsedViewCount }} </span>
+          <span class="viewCount"><template v-if="channelId !== null"> •</template> {{ parsedViewCount }} </span>
```
PikachuEXE, line 88:
```suggestion
          <span class="viewCount"><template v-if="channelId != null"> •</template> {{ parsedViewCount }} </span>
```

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -78,13 +78,14 @@
       </router-link>
       <div class="infoLine">
         <router-link
+          v-if="channelId !== null"
```
PikachuEXE, line 81:
```suggestion
          v-if="channelId != null"
```

PikachuEXE (changes requested):
I think it's less trouble treating `undefined` value same as `null` unless there `undefined` got some specific purpose in some context

PikachuEXE (approved):
If it's already normalized to be `null` when input is `undefined` then it's fine

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Local API LGTM

Not sure what it should look like on the IV side but i this is kind of interesting

https://github.com/FreeTubeApp/FreeTube/assets/73130443/fe5bb171-b53b-4cac-bf8c-6a230259b120

absidue:
Invidious doesn't seem to be affected by the new layout yet, but as it shows up for the local API, that's what this PR aims to address.

ChunkyProgrammer (approved):

Issue #3663, opened 2023-06-13
[Feature Request]: Use playlist thumbnail provided by YT


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

1. Go to https://www.youtube.com/channel/UCXuqSBlHAE6Xw-yeJA0Tunw/playlists
2. Compare thumbnails in FT vs YT

In FT the thumbnail of the playlist is based on the first video in that list

![DefaultThumbnail1](https://github.com/FreeTubeApp/FreeTube/assets/73130443/045d4e29-9956-4d35-910b-a131147499e0)
![DefaultThumbnail](https://github.com/FreeTubeApp/FreeTube/assets/73130443/b7363bcb-254e-4699-a5d7-3b61efeec9ba)

### Proposed Solution

If possible use the thumbnail provided by YT

### Alternatives Considered

None.

### Issue Labels

visual improvement

jadenet:
I've looked a bit on what could cause this and noticed that the local API gives the "correct" playlist thumbnails under thumbnail_renderer instead of thumbnails. Unfortunately though the Invidious API doesn't give the correct thumbnails, which may be a deal breaker. If it's not though, I'd be happy to make a pull request to fix only the local API.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @jadenet we're happy to accept a PR addressing only the Local API. I guess i have to create an issue over at Invidious if we want to have an fix for the Invidious API.

AtulPatidar1709:
Hello 👋 @jadenet, I want to work on this Issue, can you please assign it to me.

PR #3710, opened 2023-06-27
Use playlist thumbnail provided by YT

jadenet:
# Use playlist thumbnail provided by YT

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Addresses #3663 though only applies to local API 

## Description

The local API provides `thumbnail_renderer`, acting similarly to the previously used `thumbnails`, but also provides custom playlist thumbnails rather than the first video of the playlist by default. This only applies to the local API as Invidious has not yet implemented this feature.

### Original (YT):
Screenshot from 2023-06-27 05-06-56

### Before:
Screenshot from 2023-06-27 05-00-55

### After:
Screenshot from 2023-06-27 05-01-47

## Testing 

Linus Tech Tips playlist tab (sorted by last video added to better see effects): https://www.youtube.com/@LinusTechTips/playlists?view=1&sort=lad&flow=grid

Linus Tech Tips search with playlist filter: https://www.youtube.com/results?search_query=ltt&sp=EgIQAw%253D%253D
This wasn't a problem before but I noticed that playlists here don't have `thumbnail_renderer` included which required me to add a check to avoid errors.

## Desktop

- **OS:** Fedora Linux Workstation
- **OS Version:** 38

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -353,6 +353,8 @@ function parseShortDuration(accessibilityLabel, videoId) {
 export function parseLocalListPlaylist(playlist, author = undefined) {
   let channelName
   let channelId = null
+  /** @type {import('youtubei.js').YTNodes.PlaylistVideoThumbnail} */
+  const thumbnailRenderer = playlist.thumbnail_renderer
 
   if (playlist.author) {
     if (playlist.author instanceof Misc.Text) {
@@ -374,7 +376,7 @@ export function parseLocalListPlaylist(playlist, author = undefined) {
     type: 'playlist',
     dataSource: 'local',
     title: playlist.title.text,
-    thumbnail: playlist.thumbnails[0].url,
+    thumbnail: thumbnailRenderer ? thumbnailRenderer.thumbnail[0].url : playlist.thumbnails[0].url,
     channelName,
     channelId,
     playlistId: playlist.id,


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM, thank you @jadenet for addressing this issue!

absidue:
Looks good in lists. Could you please also make this change on the playlist page (the big thumbnail on the side)?

PikachuEXE:
Broken in playlist view for all playlists
image

e.g. https://youtube.com/playlist?list=PL8mG-RkN2uTxLsQhOyM5TBgMHF9V4Gfqa

jadenet:
Whoops I forgot a word, will fix it real quick.

efb4f5ff-1298-471a-8973-3d47447115dc:
Thumbnail isnt broken anymore but doesnt reflect YT provided thumbnail

https://github.com/FreeTubeApp/FreeTube/assets/73130443/fc252ca8-201e-4ff1-8f0b-9cb8b8f31691

efb4f5ff-1298-471a-8973-3d47447115dc:
Found more places where the thumbnail needs to be changed out

https://github.com/FreeTubeApp/FreeTube/assets/73130443/5a61f964-b19d-4db1-b101-30f21d9feab6

https://github.com/FreeTubeApp/FreeTube/assets/73130443/a4b886c1-ad34-4422-a557-0960684998fe

jadenet:
Screenshot from 2023-06-28 03-12-38

On initial impressions it looks like the WAN show is specifically built different than the others because it's podcast based(?), so it looks like I'll have to find a workaround for it.

jadenet:
Weirdly Youtube also uses the first video for playlist page and regular search, assuming because it has to do with it being a podcast. For searching the playlist through an account though (the 3rd video), I believe that it has to do with the library youtubei.js ignoring the thumbnail_renderer property when parsing a Playlist type despite it being available when fetching the data.

Screenshot from 2023-06-28 06-15-48

Screenshot from 2023-06-28 06-16-05

efb4f5ff-1298-471a-8973-3d47447115dc:
> Found more places where the thumbnail needs to be changed out

Have retested this so i could exclude the WAN Show from being a weird one. Tried this with the playlist Monitors & Display Tech and these concerns are still valid

Rikthepixel:
> For searching the playlist through an account though (the 3rd video), I believe that it has to do with the library youtubei.js ignoring the thumbnail_renderer property when parsing a Playlist type despite it being available when fetching the data.

I looked into this issue, and it is true. In the response of the search tab on Freetube it does get the renderer and it IS the correct thumbnail. The problem is that it uses the Playlist class, which doesn't define the thumbnail_renderer. If Youtubei.js fixes this the changes in `parseLocalPlaylistVideo` should work, since the thumbnail_renderer would be defined.

**Search response**
image

**Behavior**
In some cases the API response doesn't give the thumbnail_renderer, so it might need to be defined as an optional prop. It seems like that has to to with the playlist being a Mix instead of a user-created playlist.
image

**Contents of the Playlist YTNode**
image

Rikthepixel:
I think I am going to make a Github issue on the Youtubei.js repo and then send in a pull request. This should be a simple fix (famous last words).

Rikthepixel:
I created a PR in youtubei.js (https://github.com/LuanRT/YouTube.js/pull/424) that should add the thumbnail_renderer to the Playlist class.

This should fix the issue of it not displaying correctly in the channel search. But it won't fix it on the main search, since youtube itself doesn't show it properly itself.

jadenet:
Wow they fixed it really quickly, thank you so much! :D I'll add the update

absidue:
@Rikthepixel thanks for creating an issue and pull request on the YouTube.js side :) For the future though, adding functionality should be considered a feature request and feature implementation, not a bug and bug fix (calling a feature that hasn't been implemented yet a bug, tends to come across as a complaint to maintainers: "why isn't this feature already implemented", not a "please add this feature").
Thank you for creating the pull request on their side though.

absidue:
@jadenet thanks for opening this pull request and sticking with us through all the requested changes 🤗.

package.json
```diff
@@ -77,7 +77,7 @@
     "vue-router": "^3.6.5",
     "vue-tiny-slider": "^0.1.39",
     "vuex": "^3.6.2",
-    "youtubei.js": "^5.1.0"
+    "youtubei.js": "5.2.0"
```
PikachuEXE, line 80: Please keep the `^` prefix to avoid code conflict (and allow dependency auto update by bots) :)
```suggestion
    "youtubei.js": "^5.2.0"
```

PikachuEXE (changes requested):
Small one

Rikthepixel:
> @Rikthepixel thanks for creating an issue and pull request on the YouTube.js side :) For the future though, adding functionality ...

Thanks for the feedback! I am a little new to this type of thing. I'll keep this in mind in the future!

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM, Thanks @jadenet and @Rikthepixel !

absidue (approved):

PR #3712, opened 2023-06-27
Fix reading instances from the Invidious fallback file

absidue:
# Fix reading instances from the Invidious fallback file

## Pull Request Type

- [x] Bugfix

## Description
The current implementation was trying to JSON.parse an object, as the JSON had already been parsed, this was causing it to convert that object to a string `[object Object]` which of course failed to parse as JSON.

## Screenshots 
error

## Testing 
Open FreeTube, without the fix FreeTube won't finish starting even if you don't have Invidious selected as the default API (we really need to change that behaviour at some point). Alternatively if the invidious instances API is back up when you test this, turn your internet off and then launch FreeTube.

### DIFF
--- a/src/renderer/store/modules/invidious.js
+++ b/src/renderer/store/modules/invidious.js
@@ -45,7 +45,7 @@ const actions = {
       const filePath = `${fileLocation}${fileName}`
       if (!process.env.IS_ELECTRON || await pathExists(filePath)) {
         console.warn('reading static file for invidious instances')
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):

PikachuEXE (approved):

PikachuEXE:
> we really need to change that behaviour at some point

I might not agree with that
If it fails, might as well fail loud (It's a file bundled with FT, not some optional stuff)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #3457, opened 2023-04-22
[Bug]: Network download and buffering is significantly slower compared to browser


RevAngel7:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

Watching videos on FreeTube on 1080p often runs into buffer-underruns. Watching the same videos on browser there is no such problem. 

This can confirmed with any video that buffer-underruns on FreeTube - youtube on browser is not running into the same issue.

Tested on local API. Ubuntu .deb version v0.18.0-nightly-2844 Beta on ubuntu 22.04, kernel 6.2.12 mainline, mesa 23.1, all common codecs present. 100+mbit/s download speed. 12/16GB free RAM, fast SSD with enough free space.

### Expected Behavior

A "as fast as" download speed as viewing a video on my browser on youtube - with the local API of FreeTube. No buffer-underruns on 1080p (or higher).

### Issue Labels

content not loading, usability issue

### FreeTube Version

v0.18.0-nightly-2844 Beta

### Operating System Version

ubuntu 22.04 (LTS)

### Installation Method

.deb

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

None, issue was with previous versions as well.

### Additional Information

As I understood, the local API and FreeTube itself uses no special API key to access content from youtube. So why is it considerably slower than viewing content from youtube itself to a point where it buffer-underruns on 1080p on most videos in an environment where 100mbit/s and other hard- and software issues should not throttle in any way?

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

RevAngel7:
Definition to clarify: "1080p" means in my case 1080p60, 1080p30, 1080p23. Buffer underruns occur in all these 3 cases (even tho the contents bitrate varies).

Joerg123:
[...]

PR #3715, opened 2023-06-28
Use android formats when available, web for everything else

absidue:
# Use android formats when available, web for everything else

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
may help with #3457 until youtube breaks stuff again

## Description
This pull request makes FreeTube use the android formats when available (will work for the moment but YouTube loves to break the android client, not just with throttling but returning 403s and other unpleasant things), so if anything is wrong with the android formats, it uses the web ones. So that we still have all the same video metadata as before, only the formats are used from the android client, everything else is taken from the web client.

Known issue:
The android client has ultra low formats in addition to the normal ones, so when you select the audio formats, our current labelling is only designed for 4 formats (low, medium, high, best) instead of the android client's 7, so we need to add new labels for the ultra low ones, so that the top 4 actually get the correct labels instead of bitrates.

can be fixed in a separate pull request

## Testing 
Try various videos with the local API **without proxing through Invidious**, seeking should also be a lot quicker.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "vue-router": "^3.6.5",
     "vue-tiny-slider": "^0.1.39",
     "vuex": "^3.6.2",
-    "youtubei.js": "^5.1.0"
+    "youtubei.js": "^5.2.0"
   },
   "devDependencies": {
     "@babel/core": "^7.22.5",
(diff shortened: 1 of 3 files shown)


ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE ⚡

PikachuEXE (approved):
I remembering typing review comment but maybe not submitted? Strange

PR #3718, opened 2023-07-01
Various changes to allow us to blend in more

absidue:
# Various changes to allow us to blend in more

## Pull Request Type

- [x] Bugfix

## Description
This pull requests makes various changes to allow us to blend in more with YouTube's requests, like removing `Electron/<version>` and `FreeTube/<version>` from the user agent, setting the referer and origin headers for requests to the Innertube and the googlevideo domains.
I also decided to switch to letting YouTube generate the session for the watch page Innertube requests as those are the most sensitive ones.

This pull request doesn't fix the throttling issue, #3715 addresses that.

## Testing 
Check that everything still works/is as broken as it was before.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -287,10 +287,12 @@ function runApp() {
 
     // make InnerTube requests work with the fetch function
     // InnerTube rejects requests if the referer isn't YouTube or empty
-    const innertubeRequestFilter = { urls: ['https://www.youtube.com/youtubei/*'] }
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):
Observed a network request's headers
image

Edit: for `videoplayback`
image

Issue #2852, opened 2022-11-10
[Bug]: Live streams do not appear on subscriptions page nor on channel page


bpiec:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

There was recently a live stream on one of the channels. Youtube shows it like that:

![obraz](https://user-images.githubusercontent.com/1611112/201068234-eb49cce8-caa4-4777-89f5-deeae1ff720c.png)
(it says 13 hours ago).

FreeTube does not show it at. This is a channel page:
![obraz](https://user-images.githubusercontent.com/1611112/201068642-55e38c79-7f90-4b1f-acdc-5662524ce4c6.png)
This is sorted by newest. The first movie is a scheduled one, the second one is from 6 days ago. There is no the live stream from 13 hours ago. I am able to search for it however, when I put its name in the search box on this channel page.

This video is also not shown in the subscription page in FreeTube.

The possible cause of this is that YT now has a separate section for live streams:

![obraz](https://user-images.githubusercontent.com/1611112/201069199-d7a55f6a-6c2c-4ff0-8e73-f0f0915364eb.png)

I don't really care about the LIVE section added as long the live streams appear in the Subscriptions view…

### Expected Behavior

The video should be visible in the Subscriptions page and probably also in the Channel view.

### Issue Labels

API issue

### FreeTube Version

v0.18.0 Beta

### Operating System Version

Windows 11 22H2 22621.819

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
[...]

PR #3725, opened 2023-07-03
Add Tabs to subscriptions page for live streams and shorts

PrestonN:
# Add Tabs to subscriptions page for live streams and shorts

## Pull Request Type
- [ ] Bugfix
- [ X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
None IIRC

## Description
Adds tabs to the subscriptions page to show live streams and YouTube Shorts along the usual videos. Shorts exclusively use RSS however live streams can be fetched through normal means and via RSS

## Screenshots
videosTab
liveTab
shortsTab

## Testing

- Test Fetching via LocalAPI
- Test Fetching via RSS
- Test Fetching via Invidious
- Test Fetching via InvidiousRSS

## Desktop
- **OS:** Fedora
- **OS Version:** 37
- **FreeTube version:** v0.18.0 Nightly


efb4f5ff-1298-471a-8973-3d47447115dc:
closes https://github.com/FreeTubeApp/FreeTube/issues/2852

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Errors i've found 

https://github.com/FreeTubeApp/FreeTube/assets/73130443/61a3b936-c6ba-47bc-b20f-566d8b7737f0


https://github.com/FreeTubeApp/FreeTube/assets/73130443/5ee38237-1cf9-4790-ac91-a99a9b855f03


https://github.com/FreeTubeApp/FreeTube/assets/73130443/e8d4ec4c-1c6d-4324-b053-948592b24e51


https://github.com/FreeTubeApp/FreeTube/assets/73130443/81274553-28ef-4505-9e0f-6d10192da4b8


https://github.com/FreeTubeApp/FreeTube/assets/73130443/94f4a00c-11c2-439e-a5ba-78eaea47e891


https://github.com/FreeTubeApp/FreeTube/assets/73130443/6a21eb9f-076c-454b-8117-6fe48285e817

src/renderer/components/subscriptions-live/subscriptions-live.js
```diff
@@ -0,0 +1,489 @@
...
+        showToast(`${errorMessage}: ${error}`, 10000, () => {
+          copyToClipboard(error)
+        })
+        switch (failedAttempts) {
+          case 0:
+            return this.getChannelLiveInvidious(channel, failedAttempts + 1)
+          case 1:
+            if (process.env.IS_ELECTRON && this.backendFallback) {
+              showToast(this.$t('Falling back to the local API'))
+              return this.getChannelLiveLocal(channel, failedAttempts + 1)
```
PikachuEXE, line 409:
```suggestion
              return this.getChannelLiveLocalRSS(channel, failedAttempts + 1)
```

PikachuEXE (changes requested):
Should be RSS

src/renderer/components/subscriptions-live/subscriptions-live.js
```diff
@@ -0,0 +1,515 @@
...
+
+          if (liveTab.videos === null) {
+            this.errorChannels.push(channel)
+            return []
+          }
+
+          return parseLocalChannelVideos(liveTab.videos, channelInstance.header.author)
+        } else {
+          return []
+        }
```
PikachuEXE, line 299: Reduce no. of requests by half like video tab (2 => 1 request per channel)
```suggestion
        const entries = await getLocalChannelLiveStreams(channel.id)

        if (entries === null) {
          this.errorChannels.push(channel)
          return []
        }

        return entries
```

Plus adding following code to `src/renderer/helpers/api/local.js`
```js
export async function getLocalChannelLiveStreams(id) {
  const innertube = await createInnertube()

  try {
    const response = await innertube.actions.execute(Endpoints.BrowseEndpoint.PATH, Endpoints.BrowseEndpoint.build({
      browse_id: id,
      params: 'EgdzdHJlYW1z8gYECgJ6AA%3D%3D'
      // protobuf for the live tab (this is the one that YouTube uses,
      // it has some empty fields in the protobuf but it doesn't work if you remove them)
    }))

    const liveStreamsTab = new YT.Channel(null, response)

    // if the channel doesn't have a live tab, YouTube returns the home tab instead
    // so we need to check that we got the right tab
    if (liveStreamsTab.current_tab?.endpoint.metadata.url?.endsWith('/streams')) {
      return parseLocalChannelVideos(liveStreamsTab.videos, liveStreamsTab.header.author)
    } else {
      return []
    }
  } catch (error) {
    console.error(error)
    if (error instanceof Utils.ChannelError) {
      return null
    } else {
      throw error
    }
  }
}
```

PikachuEXE:
When using Local/Invidious RSS, video tab content overlaps with live tab

Known limitation of RSS Feed?

src/renderer/components/subscriptions-videos/subscriptions-videos.js
```diff
@@ -0,0 +1,492 @@
...
+          case 2:
+            return await this.getChannelVideosLocalRSS(channel, failedAttempts + 1)
+          default:
+            return []
+        }
+      }
+    },
+
+    getChannelVideosLocalRSS: async function (channel, failedAttempts = 0) {
+      const feedUrl = `https://www.youtube.com/feeds/videos.xml?channel_id=${channel.id}`
```
ChunkyProgrammer, line 306: We could use the UULF playlist here instead of channel rss (it'll exclude live + shorts).

I'm also wondering if we should include an all tab?(This can be done in a different PR if needed)

src/renderer/components/subscriptions-live/subscriptions-live.js
```diff
@@ -0,0 +1,515 @@
...
+      return new Promise((resolve, reject) => {
+        const subscriptionsPayload = {
+          resource: 'channels',
+          id: channel.id,
+          subResource: 'streams',
+          params: {}
+        }
+
+        invidiousAPICall(subscriptionsPayload).then(async (result) => {
+          resolve(await Promise.all(result.videos.map((video) => {
```
PikachuEXE, line 370: Invidious endpoint for live streams returns non video entries too
Channel tested: https://www.youtube.com/channel/UCnKbs5xffaEzzdKdPZvfhdQ
```suggestion
          // Sometimes `channel`, 'category' entries are mixed in the result
          resolve(await Promise.all(result.videos.filter(e => e.type === 'video').map((video) => {
```

PikachuEXE (changes requested):

src/renderer/components/subscriptions-videos/subscriptions-videos.js
PikachuEXE: All tab with refresh button = easier to get rate limited
No for me :S

ChunkyProgrammer: I guess the all tab may only work if it uses rss in that case and should probably be added in a future PR

PikachuEXE: I don't know if RSS feed includes all types of videos (videos + live + short)
But I would consider RSS feed as a different type of "stream" and put into another tab / view
That's what I did in my branch of live implementation (but less complete than this PR)

PikachuEXE:
- RSS Feed content overlap fixed
- Other change requests fixed

Hmmm what else could be broken :)

src/renderer/views/Subscriptions/Subscriptions.js
```diff
@@ -1,492 +1,45 @@
...
   data: function () {
     return {
-      isLoading: false,
-      dataLimit: 100,
-      videoList: [],
-      errorChannels: [],
-      attemptedFetch: false,
+      currentTab: 'videos'
     }
   },
```
PikachuEXE, line 23: Add remembering last tab
Quite common to browse a video (live/short) and then go back to feed
```suggestion
  },
  watch: {
    currentTab(value) {
      // Save last used tab, restore when view mounted again
      sessionStorage.setItem('Subscriptions/currentTab', value)
    },
  },
  mounted: async function () {
    // Restore currentTab
    const lastCurrentTabId = sessionStorage.getItem('Subscriptions/currentTab')
    if (lastCurrentTabId != null) { this.changeTab(lastCurrentTabId) }
  },
```

PikachuEXE (changes requested):
Better UX

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
You're gonna hate me for this request but can u switch the placement of shorts and live?
So shorts in the middle and live to the right.
This way it is consistent with the order its in on the channel pages

efb4f5ff-1298-471a-8973-3d47447115dc:
Not sure if we want to address it here but i mention it anyway. 

For the Local API the Upcoming videos Live videos are shown first and then the Upcoming videos. IMO it would make more sense to show the Upcoming first and the the Live videos

Local-API

Local API + RSS strips the upcoming labels which is expected but the current live videos are not shown on top of the page anymore

https://github.com/FreeTubeApp/FreeTube/assets/73130443/e03911c4-a74f-4e0b-af2d-5307dbab8941

For the IV API the Live and Upcoming labels are stripped from the thumbnails

IV-API

IV API + RSS shows the same stuff as Local API + RSS

https://github.com/FreeTubeApp/FreeTube/assets/73130443/c5a008e6-5b87-453c-860b-90961ae33432

src/renderer/views/Subscriptions/Subscriptions.js
```diff
@@ -1,492 +1,56 @@
...
   watch: {
-    activeProfile: async function (_) {
-      this.isLoading = true
-      this.loadVideosFromCacheSometimes()
+    currentTab(value) {
+      // Save last used tab, restore when view mounted again
+      sessionStorage.setItem('Subscriptions/currentTab', value)
     },
   },
   mounted: async function () {
```
PikachuEXE, line 30: By @absidue 
image


```suggestion
  created: async function () {
```

PikachuEXE (changes requested):

static/locales/en-US.yaml
```diff
@@ -93,6 +93,10 @@ Subscriptions:
   'Getting Subscriptions. Please wait.': Getting Subscriptions. Please wait.
   Refresh Subscriptions: Refresh Subscriptions
   Load More Videos: Load More Videos
+  Tabs:
```
absidue, line 96: It might make sense to reuse the existing strings from the channel page, as they have the same meaning. That way we won't have to wait for everyone to translate them before the RC, also stops us from having too many duplicate strings for the same thing.

absidue (commented):
There is a lot of code duplication but that could be fixed in a separate pull request later on, as that doesn't make any difference to the functionality provided by this pull request.

PikachuEXE:
Including this in my [own test build](https://github.com/PikachuEXE/FreeTube/actions/runs/5470256711) (with [playlist](https://github.com/PikachuEXE/FreeTube/tree/feature/playlist-2023-05) too)
Will use for a few days and see if there is any noticeable problem

static/locales/en-US.yaml
PrestonN: It's awkward either way we do this, as the strings are technically located in a place that's meant for the Channel view only. The right way to do it would be to move the strings outside of the Channel / Subscription objects and move it to a Neutral / Global location for reusable strings.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
IMO we should also add the distraction free settings that hides video/shorts/live tab

static/locales/en-US.yaml
absidue: I like the idea of moving them to a global location, as we have a bunch of other strings that are duplicated, so having a global location we could move stuff to would be good.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Would like this to be addressed
1. I saw that u opted for a more global approach on hiding shorts/live but i think people would prefer the granularity of our settings so they can dial it in how they like. So i would still like to see -> Create a new header in the distraction free settings called Subscription page with three toggles instead of the current implementation
2. For the IV API the Live and Upcoming labels are stripped from the thumbnails

IV-API

PikachuEXE:
> I saw that u opted for a more global approach on hiding shorts/live but i think people would prefer the granularity of our settings so they can dial it in how they like.

I would be one of them. I want to see shorts for some channels but not in subscriptions (which got way too many shorts)

absidue:
Looks like the live and upcoming labels not showing up on the live tab is broken on the Invidious side, as the API response says they aren't live.

iv-live-broken

I don't use the sites often, so I don't know if it's broken there too but it doesn't seem to be showing a label either:
iv-live-broken-site

absidue:
Looks like this Invidious issue might be relevant: https://github.com/iv-org/invidious/issues/3315

absidue:
Added dedicated distraction free settings, reduced some of the duplicate code and improved the accessibility of the tabs.

This pull request is now ready for review! :)

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Sorry that i have to do this but in the Destraction free settings in the general section there is a Hide Live Streams toggle. That one is a global one that hides channel live tab and live streams in the subscriptions. This doesnt hide the live streams in the subscriptions anymore but does hide live stream tab on the channel page. IMO we should remove this toggle and add the hide live stream tab to the channel page section.

Also accessibility wise a bit weird, here im pressing right arrow key `->` but i go in the opposite direction

https://github.com/FreeTubeApp/FreeTube/assets/73130443/5b70d332-bb5f-4d6a-bcb7-1d0e909636cc

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Everything LGTM, one question before approval 

App wide hide live streams doesn't hide previous streamed live streams like the Wan Show. It does hide currently live streams like the one from lofigirl.

As the subscription live streams tab also includes previous streamed live streams like the Wan Show would it make sense to also hide these on search?

absidue:
Theoretically yes, but not as part of this pull request.

For the local API the only way to detect previously live content in the search results, is to check if the published text starts with `Streamed`. I don't think we have any way of knowing what was previously live with the Invidious API.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc Sorry pushed another commit 🙈. Now I'm actually done.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
No worries, LGTM :)

absidue:
This pull requests will keep getting conflicts because people are translating the podcast and release tab strings. Either we speed run the approvals or wait a few weeks until everyone has translated stuff.

efb4f5ff-1298-471a-8973-3d47447115dc:
Speed run lets goooooo

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested new stuff

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer pls review ASAP as we dont want to get conflicts because of translating strings

ChunkyProgrammer (approved):

alex-milanov:
Hey guys, have you considered that some people would prefer to see all the video types from their subscriptions in one tab rather than to have to click every time to see if there are updates in the other tabs. Maybe the tabs can act as filters. And/or the behavior can be changed in the settings pane.
The reason is some channels produce more normal videos, other focus on shorts and third mostly live videos.  When I open my subscription page I want to see an overview of the updates. I see the video types as filters. If it's configurable some people might want to focus on live videos or shorts.

absidue:
@alex-milanov See the discussion here as for why that is not possible without loosing functionality (the tabs don't act as filters, they are completely separate tabs): https://github.com/FreeTubeApp/FreeTube/discussions/4225

alex-milanov:
> @alex-milanov See the discussion here as for why that is not possible without loosing functionality (the tabs don't act as filters, they are completely separate tabs): #4225

My comment was ignored in the other topic and it was closed and locked. I was also considering contributing but with this attitude my attention might be well spent elsewhere. Best of luck.

absidue:
Implementing support for every niche use case with a worse UX than the current implementation is not our goal no.

Your cache idea wouldn't work either because there is no way to know if there is a new upload without firing requests off to YouTube, bypassing the cache.

alex-milanov:
> Implementing support for every niche use case with a worse UX than the current implementation is not our goal no.
> 
> Your cache idea wouldn't work either because there is no way to know if there is a new upload without firing requests off to YouTube, bypassing the cache.

Worse UX is a very subjective statement. The default UX for the feed at least at 0.17.1 had all the video types combined in one feed. Which I guess some users got accustomed to. And I would argue that introducing tabs instead of "type filters" is actually a step in the wrong direction UX wise and this is actually part of what I am engaged in day to day basis. 

In any case the previous paragraph is an actual discourse with the goal of solving an issue or coming up with a more usable UI, subjective or not. My main issue was that the discussion was shut down in the other issue that you referred me to.  That type of interaction is less that constructive.

MarmadileManteater:
> > Implementing support for every niche use case with a worse UX than the current implementation is not our goal no.
> > Your cache idea wouldn't work either because there is no way to know if there is a new upload without firing requests off to YouTube, bypassing the cache.
> 
> Worse UX is a very subjective statement. The default UX for the feed at least at 0.17.1 had all the video types combined in one feed. Which I guess some users got accustomed to. And I would argue that introducing tabs instead of "type filters" is actually a step in the wrong direction UX wise and this is actually part of what I am engaged in day to day basis.
> 
> In any case the previous paragraph is an actual discourse with the goal of solving an issue or coming up with a more usable UI, subjective or not. My main issue was that the discussion was shut down in the other issue that you referred me to. That type of interaction is less that constructive.

Tabs are a **youtube** change, not a freetube change. Youtube changed their channel page fundamentally, and so, comparing UI of previous versions of freetube is irrelevant to this problem. Previous versions of freetube expected youtube to work differently than it does now. 

The problem is that youtube no longer returns everything as one big feed, and it is important to remember that if you make too many requests, you will get rate-limited, so making extra requests isn't really a viable option.

Even if you go back to 0.17.1 today, you will still only see the videos feed, no shorts, no live, and so, it is extremely unfair to compare the UX of 0.17.1 to 0.19.1 because the difference in UX you are describing is literally a change on YouTube's end, not FreeTube's. Freetube only conformed to the changes made my YouTube.

PR #3734, opened 2023-07-08
Fix+Chore: Remove crop directives from community post's images

Svallinn:
## Pull Request Type
- Other

## Description
A picture speaks a thousand words.

## Screenshots 
Before:
before

After:
after

## Desktop
- **FreeTube version:** `development` (ca085628dec24d9f01064c009642d763502ea5fc)


src/renderer/components/ft-community-post/ft-community-post.js
```diff
@@ -111,7 +111,8 @@ export default defineComponent({
         return Number.parseInt(b.width) - Number.parseInt(a.width)
       })
 
-      return imageArrayCopy.at(0)?.url ?? ''
+      // Remove cropping directives when applicable
+      return imageArrayCopy.at(0)?.url?.replace?.(/-c-fcrop64=[\w,]+/i, '') ?? ''
```
PikachuEXE, line 115: Remove excess `?.`
```suggestion
      return imageArrayCopy.at(0)?.url?.replace(/-c-fcrop64=[\w,]+/i, '') ?? ''
```

PikachuEXE (changes requested):
It works but the code looks strange

Also as additional notes I found an answer about this cropping directives removal
https://webapps.stackexchange.com/questions/91659/how-to-view-the-full-background-image-of-a-youtube-channel

In case anyone interested (which at least there would be the future version of myself)

Svallinn:
> It works but the code looks strange
> 
> Also as additional notes I found an answer about this cropping directives removal https://webapps.stackexchange.com/questions/91659/how-to-view-the-full-background-image-of-a-youtube-channel
> 
> In case anyone interested (which at least there would be the future version of myself)

Just to clarify, these images have a slightly different link format, they're unamed images.
`https://yt3.ggpht.com/s_PAoLivyXmcenxmI5XWWEexj-T9hhCy1OMJLih_Nme2_6hHQDTpvpMEWunuLDreeWCHzBnRVXDy=s715-c-fcrop64=1,64d80000f3d7ffff-nd-v1`
You can pretty much remove `-c-fcrop64` and everything after it
You can also mess around with the `s(\d{3})`  portion right behind it (which corresponds to the width you want) to try to get a smaller or bigger image, but I wasn't willing to mess with that, because:

1. There's no way of knowing what the maximum size you can pull is.
p.e. An image that is originally 450px in width
s412 -> image with 412px in width
s700 -> image with 450px in width (since that's the maximum width of the image)

2. I'd probably try to improve the UI a bit based on that, but setting div widths based on unknown image sizes are hella finicky and I'm a noob when it comes to manipulating UIs.

If someone who's more confortable with UIs would like to mess around with that, by all means.

PikachuEXE (approved):
Tested with https://www.youtube.com/channel/UC8ZItosWC34jdsiPjX11CXA again

absidue (approved):

ChunkyProgrammer (approved):

PR #3743, opened 2023-07-09
Improve the performance of the side nav

absidue:
# Improve the performance of the side nav

## Pull Request Type

- [x] Performance

## Description
As Array.map creates a new array, it's best to only use it when you are creating an array with new objects, if you use it to only modify properties, then you end up with two arrays containing the exact same objects. For modifying existing objects, it's better to just use a for loop as it saves the extra array creation, unless you'll be sorting it afterwards and don't want to modify the order of the original one. We have this pattern in various places throught FreeTube but I thought I would start with fixing it in the side nav.

I also changed the cloning to only clone the subscriptions list, as we don't need the rest of the profile.

## Testing 
Check that the active subscriptions list in the side nav updates when you change profiles.

### DIFF
--- a/src/renderer/components/side-nav/side-nav.js
+++ b/src/renderer/components/side-nav/side-nav.js
@@ -29,8 +29,9 @@ export default defineComponent({
       return this.$store.getters.getActiveProfile
     },
     activeSubscriptions: function () {
-      const profile = JSON.parse(JSON.stringify(this.activeProfile))
-      return profile.subscriptions.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
lgtm

PikachuEXE (approved):
Tested with IV instance https://invidious.tiekoetter.com/
Not all instances return channel image (even URLs are right) e.g. https://vid.puffyan.us/

absidue:
I think piffyan.us not returning any images is probably because of the 502 error page you see when you visit the site in a browser.

ChunkyProgrammer (approved):

Issue #3738, opened 2023-07-08
[Bug]: Age restricted videos get speed throttled and comments don’t load


hollownights:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

When trying to play age restricted videos like [this one](https://www.youtube.com/watch?v=rrSnnhe53ks), the speed is terribly throttled and the comments just won’t load. I’ve tried with 0.18.0-nightly-3089, so this problem isn’t a thing of the past. My settings are: “Preferred API backend: Local API” is set, “Revert to non-preferred backend on failure” is enabled and “Proxy Videos Through Invidious” is disabled.

I’ve also tried playing the same video on NewPipe, which also gets speed throttled, but it can load the comments. Also tried downloading with yt-dlp 2023.07.06 and it is able to download the video with no problems, no throttling at all.

### Expected Behavior

No speed throttling and comments loaded.

### Issue Labels

content not loading, inconsistent behavior

### FreeTube Version

3089

### Operating System Version

Windows 10

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
This is expected, as the throttling bypass we use, only helps with the android client, the web one and the one we use to bypass age restrictions are still throttled.

This will hopefully be fixed when we do the major refactor to use a different video player that behaves more similarly to the official website, As that is a lot of work (basic playback is easy, but we'll have to reimplement all of the custom features that accumulated over time, that are written for the current player) and we are planning to do the next release soon, it won't be included in or worked on before the next release.

efb4f5ff-1298-471a-8973-3d47447115dc:
[...]

PR #3759, opened 2023-07-10
Fix web throttling because of double deciphering

### LINKED ISSUE #3457: [Bug]: Network download and buffering is significantly slower compared to browser

absidue:
# Fix web throttling because of double deciphering

## Pull Request Type

- [x] Bugfix

## Related issue
closes #3457 
partially #3738 (will create a separate pull request for the comments)

## Description
Finally figured out why DASH was getting throttled but audio only and the legacy formats were fine. Took multiple weeks and stabs in the dark but we finally solved it.

The cause was us deciphering the formats and overwriting the URL in the `Format` objects, to avoid having to decipher them in lots of places around FreeTube (e.g. downloads, audio only, legacy). Unfortunately when we generated the DASH manifest with toDash, YouTube.js deciphered the already deciphered formats again, causing the n-param to be correct everywhere but the DASH manifest.

I don't why we were completely fine without any throttling for months before that became an issue, all I can think is that YouTube made a change that resulted in the double deciphering becoming a problem.

As the web client is fine again, I'm removing the extra fetching of the android formats, so that we don't make excess requests. If we end up needing them again in the future, they can always be added back quite easily.

Thank you very much to @AudricV and @LuanRT for their help and putting up with me while we tried to figure out the issue! 🤗

## Testing 
Play videos on the local API without proxying with the DASH formats enabled. Additionally test some age restricted videos like this one mentioned in the issue https://youtu.be/rrSnnhe53ks or the videos in this playlist: https://www.youtube.com/playlist?list=PLze6c8Dxpjv4eaemhVXYLURZt3Pcrddw5

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -152,27 +152,6 @@ export async function getLocalVideoInfo(id, attemptBypass = false) {
     player = innertube.actions.session.player
 
     info = await innertube.getInfo(id)
-
-    // // the android streaming formats don't seem to be throttled at the moment so we use those if they are availabe
-    try {
-      const androidInnertube = await createInnertube({ clientType: ClientType.ANDROID, generateSessionLocally: false })
-      const androidInfo = await androidInnertube.getBasicInfo(id, 'ANDROID')
-
-      // Sometimes when YouTube detects a third party client or has applied an IP-ratelimit,
-      // they replace the response with a different video id
-      // https://github.com/TeamNewPipe/NewPipe/issues/8713
-      // https://github.com/TeamPiped/Piped/issues/2487
-      if (androidInfo.basic_info.id !== id) {
-        console.error(`Failed to fetch android formats. Wrong video ID in response: ${androidInfo.basic_info.id}, expected: ${id}`)
-      } else if (androidInfo.playability_status.status !== 'OK') {
-        console.error('Failed to fetch android formats', JSON.stringify(androidInfo.playability_status))
-      } else {
-        info.streaming_data = androidInfo.streaming_data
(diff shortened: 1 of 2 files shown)


ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PikachuEXE (approved):
Finally fixed!

PR #3767, opened 2023-07-11
Extract published date and views for playlists with the local API

absidue:
# Extract published date and views for playlists with the local API

## Pull Request Type

- [x] Feature Implementation

## Description
Add support for extracting the published date and views for videos in playlists fetched with the local API, this doesn't affect shorts only playlists as they don't have the necessary information.

As the only place this information can be found is in the accessibility label and text runs that can have different content depending on whether the video is live, upcoming or normal, this wasn't previously implemented. While relying on localised strings isn't ideal, I think it's probably worth the benefits, especially as we always request US english responses and we do so in a few other locations too e.g. video duration on the shorts tab and shorts only playlists and subscription counts on the channel and watch pages.

## Screenshots 
Normal:
before-normal

after-normal

Live:
before-live

after-live

## Testing 
normal: https://youtube.com/playlist?list=PL8mG-RkN2uTx9Er8AqFDnD6mFvje5fybs
live: https://youtube.com/playlist?list=UULVSJ4gkVC6NrvII8umztf0Ow

Upcoming videos look the same but should still be tested to make sure nothing is broken.
To get a playlist with upcoming videos:
1. Visit https://www.youtube.com/live
2. Pick a video from the `Upcoming Live Streams` section
3. Copy the channel ID
4. Remove the `UC` from the start of the channel ID and replace it with `UULV`
5. Use the modified ID as a playlist ID

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -437,11 +437,47 @@ export function parseLocalPlaylistVideo(video) {
     /** @type {import('youtubei.js').YTNodes.PlaylistVideo} */
     const video_ = video
 
+    let viewCount = null
+
+    // the accessiblity label contains the full view count
+    // the video info only contains the short view count
+    if (video_.accessibility_label) {
+      const match = video_.accessibility_label.match(/([\d,.]+|no) views?$/i)
+
+      if (match) {
+        const count = match[1]
+
+        // as it's rare that a video has no views,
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #3771, opened 2023-07-13
Update external-player-map.json: Add basic support for external SMPlayer

### LINKED ISSUE #2231: [Feature Request]: SMPlayer external player support
### Guidelines

### Problem Description

I use SMPlayer as my main player for everything and it would make a great cohesive experience if SMPlayer was supported as an external player.

### Proposed Solution

There would be a drop down option for SMPlayer and the ability to add arguments.

### Alternatives Considered

An easy way to create a custom entry for external players so users can use whichever player they want.

### Issue Labels

### Additional Information

_No response_

trostboot:
# Add basic functionality to open videos in SMPlayer

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #2231 

## Description

Adds basic functionality to open videos in SMPlayer.

Adding more parameters (e.g., startOffset or speed) is not possible at this time as SMPlayer expects different formatting compared to what FreeTube currently passes through.

## Testing 

Testing has been done on Windows 11 x64 only.

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** v0.18.0-nightly-3109

### DIFF
--- a/static/external-player-map.json
+++ b/static/external-player-map.json
@@ -58,5 +58,23 @@
             "playlistShuffle": "--mpv-shuffle",
             "playlistLoop": "--mpv-loop-playlist"
         }
+    },
+      {
+        "name": "SMPlayer",
+        "nameTranslationKey": "Settings.External Player Settings.Players.SMPlayer.Name",
+        "value": "smplayer",
+        "cmdArguments": {
+            "defaultExecutable": "smplayer",
+            "defaultCustomArguments": null,
+            "supportsYtdlProtocol": true,
+            "videoUrl": "",
+            "playlistUrl": "",
+            "startOffset": null,
+            "playbackRate": null,
+            "playlistIndex": null,
+            "playlistReverse": null,
+            "playlistShuffle": null,
+            "playlistLoop": null
+        }
     }
 ]


ChunkyProgrammer:
How do you normally set the start offset from the cli with SMPlayer?

trostboot:
By passing the -start switch with the desired offset, e.g.
`smplayer -start 15 %url%`
Accepts either seconds or hh:mm:ss timecodes.
Alternatively it also parses standard Youtube timecodes in the URL.

Setting startOffset to "-start " does not work as a = is appended anyways. 
Other options differ more substantially as they would have to be passed via the -actions switch, e.g.
`smplayer -actions "set speed 2" %url%`
for 2x playback speed. However, in my testing, passing options via -actions was not very reliable. Would have to further investigate potential causes.

ChunkyProgrammer:
This patch seems to be getting the start offset to work for me

``` diff
diff --git a/src/renderer/store/modules/utils.js b/src/renderer/store/modules/utils.js
index 1f4edda9..454c5936 100644
--- a/src/renderer/store/modules/utils.js
+++ b/src/renderer/store/modules/utils.js
@@ -523,7 +523,12 @@ const actions = {

     if (payload.watchProgress > 0 && payload.watchProgress < payload.videoLength - 10) {
       if (typeof cmdArgs.startOffset === 'string') {
-        args.push(`${cmdArgs.startOffset}${payload.watchProgress}`)
+        if (cmdArgs.startOffset.endsWith('=')) {
+          args.push(`${cmdArgs.startOffset}${payload.watchProgress}`)
+        } else {
+          // special handling for SMPlayer and similar
+          args.push(cmdArgs.startOffset, payload.watchProgress)
+        }
       } else if (!ignoreWarnings) {
         showExternalPlayerUnsupportedActionToast(externalPlayer, 'starting video at offset')
       }

diff --git a/static/external-player-map.json b/static/external-player-map.json
index f77ae13c..0885eb40 100644
--- a/static/external-player-map.json
+++ b/static/external-player-map.json
@@ -69,7 +69,7 @@
             "supportsYtdlProtocol": true,
             "videoUrl": "",
             "playlistUrl": "",
-            "startOffset": null,
+            "startOffset": "-start",
             "playbackRate": null,
             "playlistIndex": null,
             "playlistReverse": null,
```

trostboot:
I can confirm that works. I've added it to this PR, hope you don't mind.

ChunkyProgrammer:
I don't mind at all 😊

ChunkyProgrammer (dismissed):

src/renderer/store/modules/utils.js
```diff
@@ -523,7 +523,12 @@ const actions = {
 
     if (payload.watchProgress > 0 && payload.watchProgress < payload.videoLength - 10) {
       if (typeof cmdArgs.startOffset === 'string') {
-        args.push(`${cmdArgs.startOffset}${payload.watchProgress}`)
+        if (cmdArgs.startOffset.endsWith('=')) {
+          args.push(`${cmdArgs.startOffset}${payload.watchProgress}`)
+        } else {
+          // special handling for SMPlayer and similar
+          args.push(cmdArgs.startOffset, Math.trunc(payload.watchProgress))
+        }
```
PikachuEXE, line 531: Just better comments
```suggestion
        if (cmdArgs.startOffset.endsWith('=')) {
          // For players using `=` in arguments
          // e.g. vlc --start-time=xxxxx
          args.push(`${cmdArgs.startOffset}${payload.watchProgress}`)
        } else {
          // For players using space in arguments
          // e.g. smplayer -start xxxxx
          args.push(cmdArgs.startOffset, Math.trunc(payload.watchProgress))
        }
```

PikachuEXE (changes requested):

PikachuEXE (approved):
`Math.trunc` might only be applicable for SMPlayer
But I have no better alternative yet

trostboot:
@PikachuEXE Well, the question that needs to be asked is if there is any value in passing along fractions of a second to begin with.
For one, links within Youtube itself only have a precision of whole seconds.
For another, Freetube itself is currently inconsistent in how it handles this (which is why I initially didn't catch the issue with SMPlayer).

If you open a link to an external player from within the Freetube video player with some amount of progress in a video, it will only pass on integer seconds, so I would assume there is some truncating already going on somewhere.
If you then go back to e.g. your subscription page and open that same video with some saved progress in an external player, Freetube will pass along non-integer seconds.

If indeed the behaviour of the player page is the desired behaviour, then my fix is just a band-aid and this should probably be handled further up the chain.
Either way, the behaviour should probably be consistent across different sections of the app.

PikachuEXE:
I have no idea what `payload.watchProgress` can contain
If it contains non integer numbers and it breaks stuff on external players
I wonder why we don't use it in all conditions... (coz existing external players ignore decimal places?)

But I am fine with a band-aid for one player, we can fix it more properly when more external players added

ChunkyProgrammer (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Thanks @trostboot for contributing!

If u are interested there are more issues open to support external players so u might want to look into that.

PR #3099, opened 2023-01-21
Handle missing data in search results

absidue:
# Handle missing data in search results

## Pull Request Type

- [x] Bugfix

## Description
YouTube's search results unfortunately don't always contain all of the information. This pull request adds logic to fix the errors that were occurring when FreeTube encountered the missing data.

Channel search results can contain any combination of handle, subscribers or video count.

Video search results don't always contain the published text.

Playlist search results don't always have a channel id, sometimes it's just some text like "Various artists".

## Screenshots 
Different types:
types-1-2
type-3 type-4

video

playlist

## Testing 
Set the search filter to channels and search for `zephyr whispers`
Clear the search filters and search for `rihanna topic`
Clear the search filters and search for `zephyr vol 1`

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.18.0

PR #3791, opened 2023-07-18
IV: Show channel handle in search results

ChunkyProgrammer:
# IV: Show channel handle in search results

## Pull Request Type
- [x] Feature Implementation

## Related issue
Video Count for channels is almost always 0 (it might always be 0)

## Description
Display the channel handle when it exists + fix missing translations for ft-list-channel

## Testing
- Use Invidious API
- Search 'MrBeast'
- See channel handle instead of video count for the Mr Beast channel

## Desktop
- **OS:** Windows
- **OS Version:** 10

## Additional context
Releis on https://github.com/iv-org/invidious/pull/3994 to be merged

### DIFF
--- a/src/renderer/components/ft-list-channel/ft-list-channel.js
+++ b/src/renderer/components/ft-list-channel/ft-list-channel.js
@@ -1,6 +1,7 @@
 import { defineComponent } from 'vue'
 import { youtubeImageUrlToInvidious } from '../../helpers/api/invidious'
 import { formatNumber } from '../../helpers/utils'
+import { parseLocalSubscriberCount } from '../../helpers/api/local'
 
 export default defineComponent({
   name: 'FtListChannel',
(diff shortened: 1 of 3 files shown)


absidue:
The test cases in this pull request should be tested too: https://github.com/FreeTubeApp/FreeTube/pull/3099

efb4f5ff-1298-471a-8973-3d47447115dc:
Upstream PR finally merged letssgoooo!

efb4f5ff-1298-471a-8973-3d47447115dc:
If possible i would like it to be formatted this way Capture3

We do formatting like this everywhere but not here so it would be nice if that also could be addressed in here

PikachuEXE (approved):
Tested via searching LTT
(One capped with local one with IV)
Screenshot 2023-10-10 at 13 14 49
Screenshot 2023-10-10 at 13 16 32

absidue (approved):
Works with all cases I tested. Also tested both local and Invidious API, just to make sure it still works for the local API (cleared the serach cache in between).

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PR #3797, opened 2023-07-21
Fix playlists for external players with no index support

trostboot:
# Fix playlists for external players with playlist but no index support

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

The current logic doesn't consider external players that have support for playlists but not playlist indexes enabled, leading to entire playlists being sent to the external player when a single video is opened from within a playlist.

This adds an additional check to only send the video URL for these players when a playlist with an index is specified, for all other players or if no index is specified the playlist URL will be sent.

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** v0.18.0-nightly-3148

### DIFF
--- a/src/renderer/store/modules/utils.js
+++ b/src/renderer/store/modules/utils.js
@@ -585,7 +585,12 @@ const actions = {
         }
       }
 
-      args.push(`${cmdArgs.playlistUrl}https://youtube.com/playlist?list=${payload.playlistId}`)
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
Code seems fine but I haven't tested it

ChunkyProgrammer (approved):

absidue (approved):

Issue #1975, opened 2022-01-03
Advanced options to allow use of generic external player


Hiers:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Please describe clear and concise of what the problem is. Ex. I'm always frustrated when [...]

Right now, all external players have to be hard coded into Freetube to be supported. This is fine, but there are users that want to use lesser known players with Freetube that might not be worth the effort of hard coding.

**Describe the solution you'd like to see implemented**
Please give a clear and concise description of what you want to happen.

To add an advanced option in the external player options with that would let a user complete the fields in external-player-map.json. A user could write the command arguments for playlists, offsets, playback, etc for the player in each proper field according to the mapping that Freetube uses.

**Describe alternatives you've considered**
Are there other ways to solve the problem? Please give a clear and concise description of any alternative solutions or features you might have considered.

What has been done so far, hard coding it. For some cases (like iina) it might be unavoidable.

**Screenshots**
Sometimes people have screenshots from mock-ups or other projects. Add them here.

None

**Additional context**
Add any other context about the feature request here.

None

efb4f5ff-1298-471a-8973-3d47447115dc:
We support many external players since this was initially opened. To my knowledge it has been years since a request came in to support a player so i think this isnt really needed anymore

PR #3798, opened 2023-07-21
Add basic support for PotPlayer, MPC-HC, MPC-BE

### LINKED ISSUE #3614: [Feature Request]: Open videos in MPC-BE, MPC-HC, or PotPlayer
### Guidelines

### Problem Description

I prefer using MPC-BE, MPC-HC, or PotPlayer with FreeTube.

### Proposed Solution

To enable these players, you may need to add them appropriately in the 'external-player-map.json' file. Here's an example of how to add MPC-BE: 
` {
        "name": "MPC-BE",
        "nameTranslationKey": "Settings.External Player Settings.Players.MPC-BE.Name",
        "value": "mpc-be",
        "cmdArguments": {
            "defaultExecutable": "mpc-be64",
            "defaultCustomArguments": null,
            "supportsYtdlProtocol": false,
            "videoUrl": "",
            "playlistUrl": "",
            "startOffset": null,
            "playbackRate": null,
            "playlistIndex": null,
            "playlistReverse": null,
            "playlistShuffle": null,
            "playlistLoop": null
        }
    },` 
    By adding the above code snippet to the file, you should be able to launch videos in MPC-BE or PotPlayer successfully. However, configuring arguments for MPC-HC or MPC-BE is a bit tricky with just this modification. For example, the `startOffset` command switch for MPC-BE/MPC-HC is `/startpos`, which expects the value in the format hh:mm:ss. This format is not what FreeTube passes, as far as I know. Additionally, setting `/audiorenderer 0` as an external player argument breaks playback for some reason. This argument is required because changing the playback speed alters the voice pitch without MPC-AR. Although setting MPC-AR as the default audio renderer might solve this issue, it's not an ideal solution.
    The `--rate=1` argument is the reason why simply changing the path for VLC media player to MPC-BE/PotPlayer doesn't work. Having a nice playback rate at startup can be useful, so you may want to consider this. 
    MPC-BE and PotPlayer support native YouTube parsing and playback, while MPC-HC requires yt-dl/yt-dlp. However, MPC-BE can also use yt-dl/yt-dlp.

### Alternatives Considered

Attempted to play videos in MPC-BE and PotPlayer by configuring a custom path after setting VLC as the external player, since both players are capable of playing YouTube videos. However, it didn't work as the `--rate=1` option is not supported by either player. MPC-BE throws an error, while PotPlayer adds the video to the playlist but fails to play it.

### Issue Labels

ease of use improvement, improvement to existing feature, support for external software

### Additional Information

This issue seems to be related to problems discussed in issues https://github.com/FreeTubeApp/FreeTube/issues/418 and https://github.com/FreeTubeApp/FreeTube/issues/2005. The latter issue is quite similar but hasn't made any progress and instead suggested opening a feature request.

I want to clarify that I'm a newbie and don't have a strong understanding of programming. I included all this information simply to complete this form. Please feel free to disregard any of it if you find it irrelevant. I tried to solve the problem myself by reading the aforementioned issues, but ended up making a mess of it.

Thank you.

trostboot:
# Add basic support for PotPlayer, MPC-HC, MPC-BE

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #3614 

## Description

This PR adds basic external player functionality for PotPlayer, MPC-HC and MPC-BE.
It also sorts the list alphabetically, because UX.

Note that MPC-HC requires yt-dlp or similar to be present.
It will also fail to properly play back videos with a startOffset if the resolution is > 1080p (audio will play, video will be black). Unfortunately the default limit (View > Options > Advanced > YDLMaxHeight) MPC-HC ships with is 1440p.
**Feedback desired on whether startOffset should stay enabled given the current state.**

Note that MPC-BE requires yt-dlp or similar to be present and enabled in order to parse playlists.

Note that both MPC-HC and MPC-BE require #3797 to fix playback from within playlists.

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** v0.18.0-nightly-3148

### DIFF
--- a/src/renderer/store/modules/utils.js
+++ b/src/renderer/store/modules/utils.js
@@ -529,7 +529,10 @@ const actions = {
 
     if (payload.watchProgress > 0 && payload.watchProgress < payload.videoLength - 10) {
       if (typeof cmdArgs.startOffset === 'string') {
-        if (cmdArgs.startOffset.endsWith('=')) {
+        if (cmdArgs.defaultExecutable.startsWith('mpc')) {
+          // For mpc-hc and mpc-be, which require startOffset to be in milliseconds
+          args.push(cmdArgs.startOffset, (Math.trunc(payload.watchProgress) * 1000))
+        } else if (cmdArgs.startOffset.endsWith('=')) {
           // For players using `=` in arguments
           // e.g. vlc --start-time=xxxxx
           args.push(`${cmdArgs.startOffset}${payload.watchProgress}`)
--- a/static/external-player-map.json
+++ b/static/external-player-map.json
@@ -6,42 +6,6 @@
         "cmdArguments": null
     },
     {
-        "name": "mpv",
-        "nameTranslationKey": "Settings.External Player Settings.Players.mpv.Name",
-        "value": "mpv",
-        "cmdArguments": {
-            "defaultExecutable": "mpv",
-            "defaultCustomArguments": null,
-            "supportsYtdlProtocol": true,
-            "videoUrl": "",
-            "playlistUrl": "",
-            "startOffset": "--start=",
-            "playbackRate": "--speed=",
-            "playlistIndex": "--playlist-start=",
-            "playlistReverse": null,
-            "playlistShuffle": "--shuffle",
-            "playlistLoop": "--loop-playlist"
-        }
-    },
-    {
-        "name": "VLC",
-        "nameTranslationKey": "Settings.External Player Settings.Players.VLC.Name",
-        "value": "vlc",
-        "cmdArguments": {
-            "defaultExecutable": "vlc",
-            "defaultCustomArguments": null,
-            "supportsYtdlProtocol": false,
-            "videoUrl": "",
-            "playlistUrl": null,
-            "startOffset": "--start-time=",
-            "playbackRate": "--rate=",
-            "playlistIndex": null,
-            "playlistReverse": null,
-            "playlistShuffle": "--random",
-            "playlistLoop": "--loop"
-        }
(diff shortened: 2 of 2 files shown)


PikachuEXE (changes requested):
https://github.com/FreeTubeApp/FreeTube/pull/3798/commits/e9fae080c1e03eab144b350fb1befc971414a1d4
Moves entries around

Also the indent alignment is inconsistent with other entries

trostboot:
Thanks, fixed the indents. 

As for the reordering, that was intentional so the list would be alphabetical. It just makes more sense to me. Do you have any objections to that?

PikachuEXE:
It's harder to review the diff

If we want the **display order** to be updated it should be done in code not in data (with limited no. of entries)

trostboot:
Well, I've had a look around the relevant code, but I'm afraid that's above my pay grade. Until someone else feels the need to tackle that, here's the unsorted list back :)

PikachuEXE (commented):
Code looks fine but I will wait for https://github.com/FreeTubeApp/FreeTube/pull/3797 to be merged first

ChunkyProgrammer (approved):

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Thanks for your contribution @trostboot. If u want to address more external player issues u could take a look at https://github.com/FreeTubeApp/FreeTube/issues/1975

Issue #3735, opened 2023-07-08
[Bug]: Cannot import any data from Freetube, Newpipe, Youtube, or Invidious.


Cyanide-1:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

How to achieve bug
1. go to data settings
2. click on "import subscriptions"
3. click on the desired file
4. no data is imported

### Expected Behavior

clicking on the file should import the data into freetube.

### Issue Labels

feature stopped working

### FreeTube Version

0.18.0 Beta

### Operating System Version

Void Linux x86-64 (Rolling Release)

### Installation Method

Flathub

### Primary API used

Invidious API

### Last Known Working FreeTube Version (If Any)

0.18.0 Beta

### Additional Information

basically i have tried importing data from every available source i could, none of it worked, not sure how to resolve this issue, i have use my freetube data from my last installation, as well as imported data using both invidious and newpipe, none of it seemed to register into freetube.

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @Cyanide-1 could u test this out using the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds). Or provide the file u would like to import so we can take a look.

kwg:
I was able to duplicate the result with my subscription list from YT. After trial and error, I found a single entry to a channel that had been banned. I removed the channel and imported 477 entries without an issue. 

With the entry left in, the import would freeze and never provide an error or finish.

Confirmed with the current release build and nightly build 3106.

Cyanide-1:
[...]

PR #3816, opened 2023-07-26
Fix: importing subscriptions with terminated channels

ChunkyProgrammer:
# Fix: importing subscriptions with terminated channels

## Pull Request Type
- [x] Bugfix

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/issues/3735

## Description
Import subscriptions when terminated channels found, etc.

## Testing 

See: https://github.com/FreeTubeApp/FreeTube/issues/3735#issuecomment-1642286050

## Desktop
- **OS:** Windows
- **OS Version:** 10

## Additional context
@efb4f5ff-1298-471a-8973-3d47447115dc I could not find the ` freetube-subscriptions.json` that you mentioned

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -1035,10 +1035,9 @@ export default defineComponent({
         invidiousAPICall(subscriptionsPayload).then((response) => {
           resolve(response)
         }).catch((err) => {
-          console.error(err)
           const errorMessage = this.$t('Invidious API Error (Click to copy)')
-          showToast(`${errorMessage}: ${err.responseJSON.error}`, 10000, () => {
-            copyToClipboard(err.responseJSON.error)
+          showToast(`${errorMessage}: ${err}`, 10000, () => {
+            copyToClipboard(err)
           })
 
           if (process.env.IS_ELECTRON && this.backendFallback && this.backendPreference === 'invidious') {
@@ -1056,7 +1055,7 @@ export default defineComponent({
         const channel = await getLocalChannel(channelId)
 
         if (channel.alert) {
-          return undefined
+          return []
         }
 
         return {


src/renderer/components/data-settings/data-settings.js
```diff
@@ -1035,10 +1035,9 @@ export default defineComponent({
         invidiousAPICall(subscriptionsPayload).then((response) => {
           resolve(response)
         }).catch((err) => {
-          console.error(err)
```
absidue, line 1038: Why did you remove this line?

ChunkyProgrammer: It's already getting logged in the invidiousAPICall function

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer ah sorry about that. See https://github.com/FreeTubeApp/FreeTube/issues/3735#issuecomment-1634624748 to get json file

PikachuEXE:
When I import that `freetube-subscriptions.json` from https://github.com/FreeTubeApp/FreeTube/issues/3735#issuecomment-1634624748
FT does nothing beside this message (no channel imported)
image

efb4f5ff-1298-471a-8973-3d47447115dc:
That matches with my foundlings when looking at latest dev branch in https://github.com/FreeTubeApp/FreeTube/issues/3735#issuecomment-1642286050 

But idk why it does nothing

PikachuEXE:
Tested `[SubscriptionsError.csv](https://github.com/FreeTubeApp/FreeTube/files/12018412/SubscriptionsError.csv)`
image
image

PikachuEXE (dismissed):

ChunkyProgrammer:
`freetube-subscriptions.json` should work now

PikachuEXE:
Can someone explain what's `channel.service_id === 0`?

ChunkyProgrammer:
> Can someone explain what's `channel.service_id === 0`?

Newpipe has different services (bandcamp, youtube, etc.). 0 is the id for youtube but it can probably be removed so we check the url instead

PikachuEXE:
`freetube-subscriptions.json` now works (but no error, expected?)

ChunkyProgrammer:
> `freetube-subscriptions.json` now works (but no error, expected?)

I believe so (that file was only erroring because there was no service id in it so all subscriptions were filtered out)

PikachuEXE (approved):

src/renderer/components/data-settings/data-settings.js
```diff
@@ -1056,7 +1055,7 @@ export default defineComponent({
         const channel = await getLocalChannel(channelId)
 
         if (channel.alert) {
-          return undefined
+          return []
```
absidue, line 1058: I know that returning an empty array is inline with the other parts of this file, however I think we should change that in all places, because we don't check for the empty array anywhere.

e.g. on this line we check if the `author` property of the return value is `undefined`, so when we return an empty array we are checking if `[].author` is `undefined`, as that property never exists, the check succeeds. Checking an array for a property that never exists seems wrong to me, either we should be checking for an empty array, or change the return values to be something other than an empty array, so that the checks make more sense.
https://github.com/FreeTubeApp/FreeTube/blob/4a86fa1b718b2b01c0c5c1a35c5bb1820b0f4f76/src/renderer/components/data-settings/data-settings.js#L1112

PikachuEXE: Separate PR? (Since I have no idea what do you mean

PikachuEXE: Also more difficult to review with more changes, this is just to fix importing

absidue: Doing it in a separate pull request sounds good to me. The current solution works, so technically the bug is fixed, it's just that the way it's done in the whole file, including this pull request, is bad and makes the code confusing. I suspect it's leftover from a previous refactoring.

Hopefully chunky understands what I mean, otherwise I can do it myself.

absidue (approved):
Approving as per discussion above.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
- [x]     profiles.txt imported successfully without any errors
- [x]     subscriptions.csv imported successfully but i did get an error in the console but i think that was expected 
Capture

- [x]     SubscriptionsError.csv imported successfully but i did get an error in the console but i think that was expected 
Capture

- [x]     SubscriptionsNoError.csv imported successfully without any errors
- [x]     Test.Subscriptions.txt imported successfully without any errors
- [x]     https://jumpshare.com/v/hrxb5KFMfmqOda4egFFN imported successfully without any errors

ChunkyProgrammer:
@efb4f5ff-1298-471a-8973-3d47447115dc those errors are expected

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3838, opened 2023-08-01
Fix author for album playlists on the playlist page

absidue:
# Fix author for album playlists on the playlist page

## Pull Request Type

- [x] Bugfix

## Description
With the addition of the releases tab, it is now a lot easier to stumble across the auto-generated album playlists, previously you would only get them if you copied a link from YouTube Music. Unlike normal playlists, these album playlists don't belong to a channel, so the playlist page doesn't contain any author information for them (no name, id, link, avatar), instead it has a subtitle in this format: `{artist/artists separated by commas} • Album`.

This pull request adds support for those playlists in the local API, so it will now show the subtitle instead of a blank link with the channel ID `undefined` (before the latest YouTube.js update that added support for them it would say `N/A` instead). This doesn't show any text for the Invidious API as that currently doesn't return it, but it does avoid it generating the `undefined` link.

## Screenshots 
before:
before

after (local API):
after

## Testing 

Example playlists:
One artist: https://www.youtube.com/playlist?list=OLAK5uy_lM3m05hKVY9vbzZH5lJpVKZKUQ-v_iTz4
Two artists: https://www.youtube.com/playlist?list=OLAK5uy_mYnfMbIG74obySLdWtKKGj_MJ-YZqVTu0

### DIFF
--- a/src/renderer/components/playlist-info/playlist-info.js
+++ b/src/renderer/components/playlist-info/playlist-info.js
@@ -21,7 +21,7 @@ export default defineComponent({
       title: '',
       channelThumbnail: '',
       channelName: '',
-      channelId: '',
+      channelId: null,
       videoCount: 0,
       viewCount: 0,
       lastUpdated: '',
--- a/src/renderer/components/playlist-info/playlist-info.vue
+++ b/src/renderer/components/playlist-info/playlist-info.vue
@@ -41,6 +41,7 @@
       class="channelShareWrapper"
     >
       <router-link
+        v-if="channelId"
         class="playlistChannel"
         :to="`/channel/${channelId}`"
       >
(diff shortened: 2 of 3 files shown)


ChunkyProgrammer:
I don't think `Album` would get translated

absidue:
Nothing gets translated at the moment, it returns the string unchanged from YouTube.

PikachuEXE (approved):

absidue:
@ChunkyProgrammer Would you like it to display just the artists?

Happy to make changes, just need to know what I should do.

ChunkyProgrammer:
Imo we should only display the artist(s) unless we are translating "Album"

absidue:
Updated.

ChunkyProgrammer (approved):
LGTM

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #3839, opened 2023-08-01
Use video durations from the watch history for RSS

### LINKED ISSUE #3672: [Feature Request]: cache durations to show in Subscriptions when using RSS
### Guidelines

### Problem Description

The Subscriptions page [does not show video durations when using RSS](https://docs.freetubeapp.io/usage/subscriptions/).
FreeTube shows, and thus knows, the duration when playing a video or when fetching information about it using the Local API (before it had switched to RSS, or in a profile that is not using RSS).

### Proposed Solution

Cache duration and any other information you get hold on, in order to show it when needed and not available.

### Alternatives Considered

In case you don't want to cache information about unwatched videos for privacy reasons,
1. Make a way to expire it by time, by program shutdown, and/or with a button.
2. Cache it only in the watch history.

### Issue Labels

display more information to user, improvement to existing feature

### Additional Information

v0.18.0-nightly-3030 Beta

absidue:
# Use video durations from the watch history for RSS

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #3672

## Description
Currently if you have more than 125 subscriptions in a profile or have manually enabled RSS for your subscriptions, you won't see any video durations, because they don't exist in the RSS feed. As we already store video durations in the watch history, we can use those for the RSS feeds when the video exists in the watch history.

This will not show the live status for videos, as they we don't store that in the history (videos that might have been live at the time they were marked as watched, might no longer be live now).

Unlike other things that interact with the history, which we disable when the `Remember History` setting is disabled, even if there is a history entry (e.g. restoring watch progress), I think it's fine to display durations even when the setting is disabled, as we are not changing the app behaviour just displaying useful information we have cached.

## Screenshots 
before:
before

after:
after

## Testing 

Preparation:
1. Ensure that the `Hide Videos on Watch` subscription setting is disabled.
2. If you don't have any items in the watch history (like me), enable the `Remember History` privacy setting and click on a few videos

Testing:
The shorts subscription feed always uses RSS (channel tab doesn't have published dates), so you can test with that or enable `Fetch Feeds from RSS` and check the videos and live feed.
Durations should show up for any video that has been watched, that isn't live.

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -436,7 +436,13 @@ export default defineComponent({
 
       this.channelName = this.data.author ?? null
       this.channelId = this.data.authorId ?? null
-      this.duration = formatDurationAsTimestamp(this.data.lengthSeconds)
+
+      if (this.data.isRSS && this.historyIndex !== -1) {
+        this.duration = formatDurationAsTimestamp(this.historyCache[this.historyIndex].lengthSeconds)
+      } else {
+        this.duration = formatDurationAsTimestamp(this.data.lengthSeconds)
+      }
+
       this.description = this.data.description
       this.isLive = this.data.liveNow || this.data.lengthSeconds === 'undefined'
       this.isUpcoming = this.data.isUpcoming || this.data.premiere


PikachuEXE (approved):
Note: Those manually marked as watched (but not actually watched) won't have duration stored

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Issue #3846, opened 2023-08-04
[Feature Request]: Handle redirected YouTube search results URL


confluence:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

I have filed this as a feature request rather than a bug, since it's unclear to me whether this feature is broken or just unsupported (since my setup may be atypical).

I am using the LibRedirect extension in Firefox to redirect YouTube URLs to FreeTube. I also have YouTube's search added as a search engine directly in the browser. When I search for `something` using this built-in search (i.e. bypassing browsing to YouTube in order to perform a search), I am sent to a URL in the format `https://www.youtube.com/results?search_query=something`. This is redirected to FreeTube, but it appears that FreeTube cannot process URLs in this format, and I get a blank page and a bunch of API errors.

### Proposed Solution

I would like to be redirected to a FreeTube search with the same query.

### Alternatives Considered

I know that I can search directly in FreeTube, but this makes the process much less seamless. As far as I know it's not possible to integrate an external app as a search engine in Firefox, so I don't think that there's any way for me to add FreeTube directly as a search engine and bypass YouTube entirely.

### Issue Labels

improvement to existing feature

ChunkyProgrammer:
Hi, this should be fixed in the next version of libredirect

PR #3849, opened 2023-08-05
Parse IV search urls

ChunkyProgrammer:
# Parse IV search urls

## Pull Request Type
- [x] Bugfix

## Related issue
I found this issue while testing #3846

## Description
FreeTube was not properly parsing IV urls. 

## Testing
enter `https://yt.artemislena.eu/search?q=hi` into the FreeTube search bar

## Desktop
- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/store/modules/utils.js
+++ b/src/renderer/store/modules/utils.js
@@ -324,7 +324,7 @@ const actions = {
 
     const typePatterns = new Map([
       ['playlist', /^(\/playlist\/?|\/embed(\/?videoseries)?)$/],
-      ['search', /^\/results\/?$/],
+      ['search', /^\/results|search\/?$/],
       ['hashtag', hashtagPattern],
       ['channel', channelPattern]
     ])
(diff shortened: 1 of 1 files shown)


src/renderer/store/modules/utils.js
```diff
@@ -359,12 +359,21 @@ const actions = {
...
+          // https://redirect.invidious.io/search?q={QUERY}
+          searchQuery = url.searchParams.get('q')
+        }
+        if (searchQuery == null) {
           throw new Error('Search: "search_query" field not found')
         }
 
-        const searchQuery = url.searchParams.get('search_query')
         url.searchParams.delete('search_query')
+        url.searchParams.delete('q')
```
PikachuEXE, line 376: I wonder if the `delete` should be moved into the `if`s

PikachuEXE (approved):
Also tested `https://www.youtube.com/results?search_query=bye` to ensure existing one works

src/renderer/store/modules/utils.js
absidue: I agree, moving it to the ifs is a good idea, we don't need to run it if we know that the parameter doesn't exist.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

PR #3867, opened 2023-08-10
Add initial TypeScript support

magne4000:
# Add initial TypeScript support

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue
https://github.com/FreeTubeApp/FreeTube/projects/10#card-87661920

## Description
Setup what's necessary to progressively migrate the codebase to TypeScript.

## Desktop

- **OS:** Arch Linux
- **OS Version:** 6.4.9-arch1-1

## Additional context
- CI files updated to run type checking
- Renamed `src/constants.js` to `src/constants.ts` as a really simple example
- Replaced babel with esbuild webpack plugin (faster, compatible with TypeScript and simplifies dependencies)


PikachuEXE:
I think we will need to wait for next release before merging this
I don't have much knowledge on TS best practices or stuff I will leave to the others to review it first anyway `~_~`
(I use TS but I don't it much)

absidue:
Curious why you are adding another build tool esbuild, when we already use webpack?

absidue:
Ah just read the description it's to replace babel, except that you seem to only be using esbuild for typescript and haven't told it the targets you want. So the way this pull request is currently esbuild will only polyfill the code in the typescript files (0 files) but not for the javascript or vue files, which is all the code in the project.
Please either revert that change (seems unnecessary) or set it up properly so that it actually does it's job.

_scripts/webpack.main.config.js
```diff
@@ -15,12 +15,18 @@ const config = {
...
     rules: [
       {
-        test: /\.js$/,
-        use: 'babel-loader',
+        test: /\.[tj]s$/,
+        loader: 'esbuild-loader',
         exclude: /node_modules/,
-      },
+        options: {
+          target: 'esnext'
```
absidue, line 22: The electron version we are using still uses node 16, wouldn't esnext imply whatever the latest version of node is?

Probably best to specify the actual version of node we are using (the whole point of this is to not end up with code that doesn't work).

_scripts/webpack.renderer.config.js
```diff
@@ -35,9 +35,12 @@ const config = {
   module: {
     rules: [
       {
-        test: /\.js$/,
-        use: 'babel-loader',
+        test: /\[tj]s$/,
+        loader: 'esbuild-loader',
         exclude: /node_modules/,
+        options: {
+          target: 'esnext'
```
absidue, line 42: Again same as above we are still using an electron version that uses chromium 106

absidue (commented):
Okay so quickly reviewed the esbuild changes, would prefer to use the real targets instead of specifying the ever evolving esnext target, that supports stuff that is incompatible with the node and chromium versions that our electron version uses, if esbuild doesn't support that then we should definitely switch back to babel.

absidue:
Also wonder if we should delay the typescript migration until after we've migrated to Vue 3, since that claims to have better typescript support than Vue 2.

_scripts/webpack.main.config.js
magne4000: I'll set this setting to `es2020`, [all of its functionalities](https://caniuse.com/sr_es11) are supported by majors browsers since many version.

magne4000:
> Also wonder if we should delay the typescript migration until after we've migrated to Vue 3, since that claims to have better typescript support than Vue 2.

It would greatly help to migrate to Vue 3 if Typescript was already present in the project IMO.

magne4000:
@absidue I understand. I'll keep this PR opened until then, please ping me once you need me to rebase it before merge.

efb4f5ff-1298-471a-8973-3d47447115dc:
Migration progress can be seen in https://github.com/FreeTubeApp/FreeTube/pull/4428

This should be revisited after that one is merged

PR #3870, opened 2023-08-10
Fix audio formats on long videos getting throttled

absidue:
# Fix audio formats on long videos getting throttled

## Pull Request Type

- [x] Bugfix

## Description
YouTube throttles the adaptive formats when you request a chunk larger than 10MiB ([relevant yt-dlp line](https://github.com/yt-dlp/yt-dlp/blob/db7b054a6111ca387220d0eb87bf342f9c130eb8/yt_dlp/extractor/youtube.py#L3730)), for DASH that is not a problem, as videojs-http-streaming prefers to request tiny chunks instead of massive ones, but for the audio formats this is an issue. The audio only formats are handled by the browser/electron directly, which requests the range `0-` or `1234-0` if you seek, so the entire file. For short videos, ones under 10 minutes, this is not an issue, as the file sizes for the best audio quality are usually less than 10MiB but for long videos requesting the entire file, means going above that 10MiB threshold, meaning we get throttled.

This pull request works around the issue by intercepting the request in the main process and limiting the requested range to 10MiB. This thankfully works because YouTube returns HTTP 206 responses with the `Content-Range` header, which tells the browser, where the requested chunk is in the file, so the browser player doesn't care that we altered it's range. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range

Once we switch to shaka, we can use audio only DASH, to avoid the issue, which is something video.js doesn't support.
https://github.com/videojs/http-streaming/blob/de183c8f88189883eb62c146c983ddc75dda3377/docs/supported-features.md?plain=1#L139

## Testing 
1. Set the default formats to audio
2. Set the preferred API to local
3. Open a long video (anything longer than 12mins should be affected by the throttling) e.g this WAN show live stream replay: https://youtu.be/rnIeknursww
4. With this change, it should load faster and also load quicker when you seek around.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -289,7 +289,7 @@ function runApp() {
     // InnerTube rejects requests if the referer isn't YouTube or empty
     const innertubeAndMediaRequestFilter = { urls: ['https://www.youtube.com/youtubei/*', 'https://*.googlevideo.com/videoplayback?*'] }
 
-    session.defaultSession.webRequest.onBeforeSendHeaders(innertubeAndMediaRequestFilter, ({ requestHeaders, url }, callback) => {
+    session.defaultSession.webRequest.onBeforeSendHeaders(innertubeAndMediaRequestFilter, ({ requestHeaders, url, resourceType }, callback) => {
       requestHeaders.Referer = 'https://www.youtube.com/'
       requestHeaders.Origin = 'https://www.youtube.com'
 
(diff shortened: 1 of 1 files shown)


MarmadileManteater (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tried seeking on https://youtu.be/rnIeknursww
image
image
image

PR #3886, opened 2023-08-15
Fix the layout of the hashtag page

absidue:
# Fix the layout of the hashtag page

## Pull Request Type

- [x] Bugfix

## Description
Currently the hashtag page takes up the entire width of the window, going against the layout used for the rest of the app, which is 90% on narrow screens and 80% for everything else. This pull request aligns the layout with other pages in FreeTube.

I recommend turning on the "Hide Whitespace" option in the pull request UI while reviewing this pull request, so that you can tell what the changes are, without the indentation changes getting highlighted.

## Screenshots 

before

after

## Testing 
#shorts hashtag https://youtube.com/hashtag/shorts

If you get the page header error on the local API, just navigate back and forward until it disappears, there is already a change in YouTube.js to support the new header, but we need to wait for it to be released in an update on their side, before we can add support on our side.

### DIFF
--- a/src/renderer/views/Hashtag/Hashtag.css
+++ b/src/renderer/views/Hashtag/Hashtag.css
@@ -11,9 +11,14 @@
 .getNextPage:hover, .getNextPage:focus {
   background-color: var(--side-nav-hover-color);
 }
+
 .card {
-  box-sizing: border-box;
   margin: 0 auto 20px;
-  position: relative;
   width: 85%;
+}
+
+@media only screen and (max-width: 680px) {
+  .card {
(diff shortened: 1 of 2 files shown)


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
lgtm!

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #693, opened 2020-10-19
FreeTube accessibility issues


RastislavKish:
Hello,
I was not sure whether to report this as a feature request or a bug, after some thinking, I've settled on the bug variant.
I'm a blind user of FreeTube, and while the app is definitely very interesting, there are some accessibility issues which would be very pleasant to have fixed.
They're mostly just playing with used HTML elements, so I hope it will be possible to implement them easily:

* Search results should be marked as headings. All that is needed to do here is to wrap the name of each result in perhaps h4 tag (as Search results heading uses h3), like Youtube does. Information about the author, what time ago was the video posted and the video's length can be included too, but in this case, it's important to arrange it in a way, that the name goes first and everything other follows it. Youtube has a good example of this. I don't know the visual specifics of FreeTube, so I'll leave upto authors how invasive would such rearrangement be.
* In settings, for some reason, the currently used combo-boxes are not accessible for screenreaders. Like, I see that there are combo-boxes, but I can't read the currently selected item and when i focus one and try to select something, my screenreader doesn't announce what I've selected. May be the items are not properly labelled? If you give me the used input type, I can write here an example of accessible combo-box HTML code.
* Screenreaders in general can't deal very well with clickable elements, like infact the all entries of the main menu (Trending, Playlists, Subscriptions, Settings), they're used also on video pages (to view comments, to view / hide replies to particular comment). It would be best to replace these with either links or buttons, which can be activated directly by pressing enter.
* On the page with a video, the video name should be marked as a heading too (as it leads the section with informations about video's author, likes, dislikes, description etc.).

That's what I've noticed so far, these four issues are from accessibility point of view the most important to fix, as they directly influence the app's usability for blind people.
The interface is otherwise very nice and hassle-free, that's why I hope it will be possible to do something about these things and make the app usable for everyone. :)

My configuration is Ubuntu mate 20.04 64-bit, screenreader orca v3.39.1 pre. It doesn't matter much I guess, as Electron's Chromium backend is the same for all platforms, adding it just for completeness.
FreeTube version 0.9.0.

Thanks for the great project!

peepopoggers:
@RastislavKish do you have problems interpreting icons only (compared to with text)? thanks.

RastislavKish:
Some accessibility updates:
* The settings page seems to work now! I don't know the exact version when this happened as I usually don't need to tweak them very often, but the comboboxes seem usable now.
[...]

PR #3887, opened 2023-08-15
Add semantic roles to make the site more accessible

ChunkyProgrammer:
# Add semantic roles to make the site more accessible

## Pull Request Type
- [x] Bugfix

## Related issue
#1525
#693 

## Description
This PR :
- adds h3s to search results
- adds semantic roles to elements (nav, main)
- properly sets the locale of the webpage

## Testing 

FreeTube should look and behave the same besides some headings being larger now
image

## Desktop
- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -60,7 +60,7 @@ export default defineComponent({
       return this.$store.getters.getShowProgressBar
     },
     isRightAligned: function () {
-      return this.$i18n.locale === 'ar'
+      return this.locale === 'ar' || this.locale === 'he'
     },
     checkForUpdates: function () {
       return this.$store.getters.getCheckForUpdates
@@ -102,6 +102,10 @@ export default defineComponent({
       return this.$store.getters.getSecColor
     },
 
+    locale: function() {
+      return this.$i18n.locale
+    },
+
     systemTheme: function () {
       return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'
     },
@@ -126,6 +130,8 @@ export default defineComponent({
 
     secColor: 'checkThemeSettings',
 
+    locale: 'setLocale',
+
     $route () {
       // react to route changes...
       // Hide top nav filter panel on page change
@@ -135,6 +141,7 @@ export default defineComponent({
   created () {
     this.checkThemeSettings()
     this.setWindowTitle()
+    this.setLocale()
   },
   mounted: function () {
     this.grabUserSettings().then(async () => {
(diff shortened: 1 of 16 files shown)


src/renderer/components/ft-input/ft-input.vue
```diff
@@ -49,6 +49,7 @@
         :placeholder="placeholder"
         :disabled="disabled"
         :spellcheck="spellcheck"
+        :aria-label="!showLabel ? placeholder : false"
```
absidue, line 52: Vue 3 will turn `false` into `"false"` (so that we don't need the `String()` hacks for `aria-selected` anymore), so it's best to use `null` here so it still disappears in the future.

```suggestion
        :aria-label="!showLabel ? placeholder : null"
```

src/renderer/App.vue
```diff
@@ -7,10 +7,17 @@
...
-    <top-nav ref="topNav" />
-    <side-nav ref="sideNav" />
+    <top-nav
+      ref="topNav"
+      role="navigation"
+    />
+    <side-nav
+      ref="sideNav"
+      role="navigation"
+    />
```
absidue, line 17: It probably makes more sense to place the roles, inside the actual components, instead of relying on Vue's fallthrough attributes behaviour.

src/renderer/App.js
```diff
@@ -500,6 +507,10 @@ export default defineComponent({
       }
     },
 
+    setLocale: function() {
+      document.querySelector('html').setAttribute('lang', this.locale)
```
absidue, line 511: Supported in every browser (e.g. Chrome v1, Firefox v1), and should be slightly faster than searching the DOM with `querySelector`: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement

```suggestion
      document.documentElement.setAttribute('lang', this.locale)
```

src/renderer/scss-partials/_ft-list-item.scss
```diff
@@ -304,3 +304,11 @@ $watched-transition-duration: 0.5s;
 .upcoming {
   text-transform: uppercase;
 }
+
+// we use h3 for semantic reasons but don't want to keep the h3 style
+h3 {
```
absidue, line 309: Please use a class instead of an element selector: https://vue-loader.vuejs.org/guide/scoped-css.html#also-keep-in-mind

src/renderer/App.js
```diff
@@ -102,6 +102,10 @@ export default defineComponent({
       return this.$store.getters.getSecColor
     },
 
+    locale: function() {
+      return this.$i18n.locale
```
absidue, line 106: Don't we need to replace the underscore here too?

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PR #3890, opened 2023-08-16
Add option to change thumbnail appearance

### LINKED ISSUE #2814: [Feature Request]: Add Hide Thumbnails to Distraction Free Settings
### Guidelines

- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for open and closed issues that are similar to the feature request I want to file, without success.

### Problem Description

I keep thumbnails disabled in NewPipe because sometimes they distract/annoy me. I would like to see something like NewPipe's "Load Thumbnails" or Invidious "Compact Mode" on FreeTube.

### Proposed Solution

A button to enable/disable thumbnails.

### Alternatives Considered

None

### Issue Labels

new feature, new optional setting, visual improvement

### Additional Information

Screenshot_20221103-191709_NewPipe_1

da-batt:
# Add option to change thumbnail appearance

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #2814
## Description

Added drop-down option to distraction free settings named _Thumbnail Display Mode_ which changes thumbnail behavior depending on selected option. Current options are _Hidden_ (loads a placeholder image instead of thumbnail) and _Blurred_ (blurs thumbnail)
## Screenshots 

default
Thumbnail Display Mode - _Default_

hidden
humbnail Display Mode - _Hidden_

blurred
Thumbnail Display Mode - _Blurred_

## Testing 

- Ensured that every thumbnail display mode works correctly by looking at video and playlist thumbnails.
- Inspected element size in devtools to ensure that sizing is consistent

## Desktop

- **OS:** Windows 11
- **OS Version:** 22H2
- **FreeTube version:**


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Everything LGTM in testing.

There is only one thing im not sure about and that is how the settings is shown under the distraction free settings.

The distraction free settings are mostly done with toggles to indicate if the feature is off/on. This dropdown could give of the feeling that this setting is always on even though Default is selected. 

Maybe we should opt for a toggle to enable this feature and if u enable it u get a dropdown with 2 options blurred or hidden.

Not sure if it can be done better so its best to wait for the members the chime in on this.

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -88,6 +88,14 @@ export default defineComponent({
       return this.$store.getters.getThumbnailPreference
     },
 
+    thumbnailDisplayMode: function () {
+      return this.$store.getters.getThumbnailDisplayMode
+    },
+
+    thumbnailBlurStyle: function () {
+      return this.thumbnailDisplayMode === 'blurred' ? 'blur(20px)' : 'none'
```
PikachuEXE, line 96:
```suggestion
      return this.thumbnailDisplayMode === 'blurred' ? 'blur(20px)' : null
```

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -42,6 +42,14 @@ export default defineComponent({
       return this.$store.getters.getThumbnailPreference
     },
 
+    thumbnailDisplayMode: function () {
+      return this.$store.getters.getThumbnailDisplayMode
+    },
+
+    thumbnailBlurStyle: function () {
+      return this.thumbnailDisplayMode === 'blurred' ? 'blur(20px)' : 'none'
```
PikachuEXE, line 50:
```suggestion
      return this.thumbnailDisplayMode === 'blurred' ? 'blur(20px)' : null
```

PikachuEXE:
For making the setting look better (hide behind a toggle)
I made an attempt at https://github.com/PikachuEXE/FreeTube/tree/thumbnail-display-mode-pika

Text & style can be further adjusted

image
image

da-batt:
Just ended up implementing the same solution, not sure if there is a better way.

absidue:
Blurred could be converted to a switch and hidden moved into the thumbail preference drop down (burred can't go in there as you need a thumbnail to blur).

efb4f5ff-1298-471a-8973-3d47447115dc:
@absidue I'm not opposed to that but im not sure i understand the reasoning for it

PikachuEXE:
I don't even understand that suggestion...

da-batt:
@absidue I don't agree with this approach as the intended purpose of this feature is to minimize distraction and moving it to a different place would make it harder to find for the user.

(I also think that the thumbnail preference setting is a bit ambiguous right now and maybe should be renamed to something like  'thumbnail position' or 'thumbnail alignment')

efb4f5ff-1298-471a-8973-3d47447115dc:
What about 2 toggles in the distraction free settings?

- Hidden
- Blur

If hidden enabled then disable and grey out blur
If blur enabled then disable and grey out hidden

absidue:
My reasoning for proposing to move the hidden option to the thumbail preference setting, is that it already has options to use the default thumbail, the first frame in the video, the last frame or one in the middle, so having the hidden/no thumbnail setting in the same dropdown makes a lot of sense to me.

Additionally not downloading thumbnails will not only be used as a distraction free setting but also by users that want to save bandwidth or have slow internet connections.

da-batt:
@absidue Wow, I have realized I completely misunderstood the thumbnail preference setting. Your proposal makes a lot more sense now. Let's wait for what the others have to say.

PikachuEXE:
Then if it's hidden we shouldn't allow blurring the thumbnail?

da-batt:
Yes we'd have to disable it. I'm just not sure how we'd communicate to the user that blur thumbnail is disabled because of thumbnail preference being set to hidden.

absidue:
You could do it similarly to how the hide livestreams and the hide subscriptions live work.

PikachuEXE:
I created a SVG file via https://mediamodifier.com/svg-editor
thumbnail_placeholder
Much smaller file size (4xxKB vs 1KB) and can be used in any size (coz SVG)
Color, shape sizes can be adjusted if needed
Preview:
image

da-batt:
@PikachuEXE Good idea! Managed to clean the svg up so now it's ~200 bytes.

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -225,6 +233,9 @@ export default defineComponent({
     },
 
     thumbnail: function () {
+      if (this.thumbnailPreference === 'hidden') {
+        return require('../../assets/img/thumbnail_placeholder.svg')
+      }
```
absidue, line 238: Nitpick but would prefer this being moved inside the switch case that is only a few lines below it.

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -51,6 +59,9 @@ export default defineComponent({
     },
 
     thumbnail: function () {
+      if (this.thumbnailPreference === 'hidden') {
+        return require('../../assets/img/thumbnail_placeholder.svg')
+      }
```
absidue, line 64: ditto

absidue (commented):
Hidden is broken for ft-list-playlist components, as you forgot to add the computed property for the thumbnail preference, it also breaks the layout on the playlist page.
Also two small code nitpicks.

ft-list-playlist:
image

Playlist page:

image

src/renderer/components/ft-list-video/ft-list-video.js
da-batt: If we made it a switch case we would have to run another check in the else block above which is less clean IMO.

PikachuEXE: I guess you mean `baseUrl` is not required when `hidden`?

da-batt: My concern is actually related to the implementation of this change in `playlist-info.js` not `ft-list-video.js` (my bad). Making this change here in `ft-list-video.js` is actually not a problem. I've left a comment in the relevant thread.

src/renderer/components/playlist-info/playlist-info.js
da-batt: This is how my code looks based on this suggestion. (See comment on line 6)

```js
thumbnail: function () {
      let baseUrl
      if (this.backendPreference === 'invidious') {
        baseUrl = this.currentInvidiousInstance
      } else {
         // We still have to check whether or not thumbnails are hidden here
        return this.thumbnailPreference === 'hidden' ? require('../../assets/img/thumbnail_placeholder.svg')  : this.data.playlistThumbnail
      }

      switch (this.thumbnailPreference) {
        case 'start':
          return `${baseUrl}/vi/${this.firstVideoId}/mq1.jpg`
        case 'middle':
          return `${baseUrl}/vi/${this.firstVideoId}/mq2.jpg`
        case 'end':
          return `${baseUrl}/vi/${this.firstVideoId}/mq3.jpg`
        case 'hidden':
          return require('../../assets/img/thumbnail_placeholder.svg')
        default:
          return `${baseUrl}/vi/${this.firstVideoId}/mqdefault.jpg`
      }
    }
  }
```
I can't think of a way right now which wouldn't require making the same check twice different to my last implementation. Maybe there's a better way but idk right now.

PikachuEXE: Your code is actually correct. The order is
- No thumbnail
- Local API = use `playlistThumbnail`
- IV API

efb4f5ff-1298-471a-8973-3d47447115dc:
@da-batt is this ready for review?

da-batt:
@efb4f5ff-1298-471a-8973-3d47447115dc yes, everything should be fine now.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PikachuEXE (approved):

PikachuEXE:
@absidue / @ChunkyProgrammer ?

ChunkyProgrammer (approved):

PR #3895, opened 2023-08-18
Upgrade YouTube.js to version 6.0

absidue:
# Upgrade YouTube.js to version 6.0

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation
- [x] Dependency upgrade

## Description
This pull request upgrade YouTube.js to version 6 and implements the changes from #3838 on the watch page, as I only did it on the playlist page at the time (yes I forgot to do it on the watch page :see_no_evil:).

The YouTube.js upgrade also makes it possible to fix the hashtag header layout change that I noticed in #3886 (will create a separate PR, will need to dig out some visitor data that has the layout change).

None of the breaking changes affect us, they are mostly internal changes, but as YouTube.js exposes all internal functionality, it's possible that or projects were relying on certain things (we don't), which is why they were classified as breaking. To us those changes might even have a positive impact as it now uses less polyfills.

YouTube.js 6 change log: https://github.com/LuanRT/YouTube.js/releases/tag/v6.0.0
The breaking changes in detail: https://github.com/LuanRT/YouTube.js/pull/468

## Testing 
Playlist with two pages (test on both the watch and playlist pages): https://youtube.com/playlist?list=PLa4zmfXkL4P0RkofW6vbsqhOAKgjwLws1

Watch page for an album playlist (check that the author name shows up and can't be clicked on): https://www.youtube.com/watch?v=WG2nf9IvWVg&list=OLAK5uy_mYnfMbIG74obySLdWtKKGj_MJ-YZqVTu0

PR #3897, opened 2023-08-19
Fix playlist view on the watch page not reacting to playlist changes

absidue:
# Fix playlist view on the watch page not reacting to playlist changes

## Pull Request Type

- [x] Bugfix

## Related issue
https://github.com/FreeTubeApp/FreeTube/pull/3895#issuecomment-1684875711

## Description
Currently the playlist view on the watch page doesn't react to the playlist changing, when you jump directly from one playlist to another one, without leaving the watch page in between.

This bug exists because unlike the other components on the watch page, the playlist view stays mounted when the video changes, this is to avoid refetching the playlist for every video in the same playlist.

## Screenshots 
Screenshot from the comment:
image

## Testing 
Paste the first playlist URL into the search bar, afterwards the second one, the playlist view on the watch page should update to the second playlist instead of continuing to show the first one.

playlist 1: https://www.youtube.com/watch?v=xEyR5dCQs60&list=PLa4zmfXkL4P2khNoEOGA8NEOgFwHIWGIX

playlist 2: https://www.youtube.com/watch?v=HUlR6zQjTmw&list=PLa4zmfXkL4P0UaXti2bPFpsyi3K_IE0rB

### DIFF
--- a/src/renderer/components/watch-video-playlist/watch-video-playlist.js
+++ b/src/renderer/components/watch-video-playlist/watch-video-playlist.js
@@ -129,6 +129,15 @@ export default defineComponent({
         nextTick(() => this.scrollToCurrentVideo())
       }
     },
+    playlistId: function (newVal, oldVal) {
+      if (oldVal !== newVal) {
+        if (!process.env.IS_ELECTRON || this.backendPreference === 'invidious') {
+          this.getPlaylistInformationInvidious()
+        } else {
+          this.getPlaylistInformationLocal()
+        }
+      }
+    }
   },
   mounted: function () {
     const cachedPlaylist = this.$store.getters.getCachedPlaylist


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

src/renderer/components/watch-video-playlist/watch-video-playlist.js
```diff
@@ -129,6 +129,15 @@ export default defineComponent({
         nextTick(() => this.scrollToCurrentVideo())
       }
     },
+    playlistId: function (newVal, oldVal) {
+      if (oldVal !== newVal) {
```
PikachuEXE, line 133: I don't understand this conditional...

absidue: We don't want to refetch the playlist if the id is the same

ChunkyProgrammer (approved):

src/renderer/components/watch-video-playlist/watch-video-playlist.js
PikachuEXE: It's called when value not updated too? (It's inside `watch`)

absidue: I'm not sure, we definitely set the property every time the video changes, but i'm not sure if watch is triggered on every set or just when the value changes. We have this pattern inside watches in a few places which made me think it might get triggered for every set  but i'm not sure.

absidue: I can always remove the check later, if we find out that it's redundant.

PikachuEXE (approved):

PR #3942, opened 2023-08-25
Support multiple audio tracks and AV1 for Invidious by using the local API DASH manifest generator

### LINKED ISSUE #3382: [Bug]: Allow DASH AV1 formats + Invidious API / (Local API + Proxy Videos Through Invidious) = Incorrect resolution
### Guidelines

### Describe the bug

- Set default quality to 720p
- Enable `Allow DASH AV1 formats`
- Use 
  - `Invidious API` or 
  - `Local API` with `Proxy Videos Through Invidious` enabled
- View https://youtu.be/8IFwpI7OT6I

### Expected Behavior

720p selected

### Issue Labels

usability issue

### FreeTube Version

55f2af103ffc21a605dae0d3baa4161fcc1823e4

### Operating System Version

MacOS

### Installation Method

Portable

### Primary API used

Invidious API

_No response_

### Additional Information

Actually using custom build with https://github.com/FreeTubeApp/FreeTube/pull/3350

### Nightly Build

absidue:
# Support multiple audio tracks and AV1 for Invidious by using the local API DASH manifest generator

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
closes #3382

## Description
Currently Invidious doesn't support multiple audio tracks in their API response or DASH manifest, they are also missing the important fields `width`, `height` for all video streams and `fps` and `qualityLabel` for the AV1 formats in their API responses.

This pull request adds multiple audio track support for Invidious, without requiring any changes on their side, a side effect of that ,is that we can also support AV1 formats properly for Invidious.

This pull request advantage of YouTube.js exposing its DASH manifest generator and its ability to parse language and role information from the streaming URLs. Unfortunately due to the missing fields in the API responses (mentioned in the first paragraph), this pull request still makes a request to Invidious' manifest endpoint, so that it can extract the height, width and fps from it, for the locally generated manifest. As video.js was already making that request before, this doesn't cause any extra load on Invidious instances.

YouTube.js usually deciphers the formats as part of the manifest generation process, as that requires sending a request to YouTube to check if there is a new player and a second one to fetch it if there is (would use the one from the cache if it's the same one), I decided to make the deciphering a no-op while generating the Invidious manifest (it will still decipher for the local API one).
While we could use the deciphering to unthrottle age-restricted videos on Invidious (Invidious doesn't do its own deciphering and the bypass doesn't use the android endpoint, with the magic params that bypass throttling), I decided against it, to avoid making the extra requests. If unthrottling those is still desirable, I could see if there is a way to detect age restricted videos on Invidious and implement deciphering just for those ones (only if backend fallback is enabled of course) in a follow-up pull request.

## Testing 
dubbed and descriptive audio tracks: https://youtu.be/Kn56bMZ9OE8 (dubbed at 1:35, descriptive at 1:40)
audio tracks in multiple languages: https://youtu.be/jObOjhUkf50
AV1 formats: https://youtu.be/njX2bu-_Vw4

Also test proxying if you can find an Invidious instance it works on.

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
I noticed that the test videos didnt start in the original track by default. The Local API always starts with the original track.

Is this intended behavior?

absidue:
It should start with the default one, was messing around a lot in FreeTube and YouTube.js while making it, so maybe I broke something along the way or maybe it's dependant on one of the YouTube.js changes I made while making this.

I'll take a look and get back to you.

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc Should work now, created a pull request to YouTube.js with the bug fix and upgraded to the version with it now.

PikachuEXE (dismissed):

absidue:
Added language names with the Intl.DisplayNames API, so now the labels match the local API ones. This means they should now look and feel the same.

src/renderer/views/Watch/Watch.js
```diff
@@ -1384,10 +1409,72 @@ export default defineComponent({
...
+              hasMultipleAudioTracks = true
+            }
+          }
+
+          formats.push(localFormat)
+        }
+
+        if (hasMultipleAudioTracks) {
+          // match YouTube's local API response with English
+          const languageNames = new Intl.DisplayNames('en-US', { type: 'language' })
```
ChunkyProgrammer, line 1463: Should we use the user's current locale instead of using "en-US"?

absidue: My goal with this pull request was to match the local API, so that you would have the same user experience with both the local and the Invidious API.

If you want me to switch to using the users locale, I can do that, it'll just be a different user experience than with the local API or the Invidious API when it gets multiple audio track support.

efb4f5ff-1298-471a-8973-3d47447115dc:
> @efb4f5ff-1298-471a-8973-3d47447115dc Should work now, created a pull request to YouTube.js with the bug fix and upgraded to the version with it now.

So tested both videos again and both didnt start with the original audio

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc Merged the development branch in again, hopefully that'll help, also make sure you have ran `yarn install` so that you actually have a youtube.js version installed that works.

src/renderer/helpers/api/invidious.js
```diff
@@ -312,12 +313,18 @@ export function filterInvidiousFormats(formats, allowAv1 = false) {
...
-  //   return [...audioFormats, ...h264Formats]
-  // }
-  return [...audioFormats, ...h264Formats]
+  // As we generate our own DASH manifest (using YouTube.js) for multiple audio track support in Electron,
+  // we can allow AV1 in that situation. If we aren't in electron,
+  // we still can't use them until Invidious fixes the issue on their side
+  if (process.env.IS_ELECTRON) {
+    if (allowAv1 && av1Formats.length > 0) {
+      return [...audioFormats, ...av1Formats]
+    } else {
```
ChunkyProgrammer, line 322: Suggestion: Remove this else block (including return)

src/renderer/helpers/api/invidious.js
```diff
@@ -312,12 +313,18 @@ export function filterInvidiousFormats(formats, allowAv1 = false) {
...
+      return [...audioFormats, ...h264Formats]
+    }
+  } else {
```
ChunkyProgrammer, line 325: Suggestion: remove the else (keep the return statement)

src/renderer/helpers/api/invidious.js
PikachuEXE: I always wonder why early return pattern is not used more often

absidue:
Fixed

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

ChunkyProgrammer (approved):
LGTM, looked at XHR/Fetch in network tab and no requests to YouTube are made when proxying through invidious is enabled. We should probably be translating the languages to the user's locale but as local api doesn't currently do that, this can come later

PR #3948, opened 2023-08-26
Rounded channel thumbnails

### LINKED ISSUE #1452: Add option for rounded grid elements within app
It is a low-priority feature request, but should be easy to impliment.

My request would be a settings option to change the border radius of page elements (pictures below).
Would be optional.

It could be implimented by using a css variable which is passed into `border-radius: var(x)` for each tiling element (plus video player with `overflow:hidden` and the same method.
The same could be done for thumbnails too with another option.

**Screenshots**
(Tiles) Before:
image

(Tiles) After:
image

(Thumbnails) Before:
image

(Thumbnails) After:
image

**Additional context**
If one is chosen for addition, tiles is of more importance.

jasonhenriquez:
# Rounded Channel Thumbnails

## Pull Request Type
- [x] Feature Implementation

## Related issue
related to #1452 

## Description
This change adds the sleek rounded thumbnail look. It also updates the outline to be rounded, as well as the watch progress bar to cover the now-shortened bottom side. I am open to alternative suggestions for how we should handle the watch progress bar (this has the left and right sides clipped so as to avoid floating above the rounded edges. One alternative would be only rounding the top left and right.

## Screenshots 

Screenshots of the rounded channel thumbnails:
Screenshot_20230826_010910
Screenshot_20230826_011824

If enough reviewers prefer this one, I can push a different version that only make the top left and right appear rounded:
Screenshot_20230826_011611
Screenshot_20230826_011757

## Testing
Tested in card and list view with hover, focus, active, and watched styling.

## Desktop
- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.scss
+++ b/src/renderer/components/ft-list-video/ft-list-video.scss
@@ -1,5 +1,1 @@
 @use '../../scss-partials/_ft-list-item';
-
-.thumbnailLink:hover {
-  outline: 3px solid var(--side-nav-hover-color);
-}
--- a/src/renderer/components/ft-list-video/ft-list-video.vue
+++ b/src/renderer/components/ft-list-video/ft-list-video.vue
@@ -62,11 +62,13 @@
       >
         {{ $t("Video.Watched") }}
       </div>
-      <div
-        v-if="watched"
-        class="watchedProgressBar"
-        :style="{width: progressPercentage + '%'}"
-      />
+      <div class="watchedProgressBarContainer">
+        <div
+          v-if="watched"
+          class="watchedProgressBar"
+          :style="{width: progressPercentage + '%'}"
+        />
+      </div>
     </div>
     <div class="info">
       <router-link
--- a/src/renderer/scss-partials/_ft-list-item.scss
+++ b/src/renderer/scss-partials/_ft-list-item.scss
@@ -92,6 +92,8 @@ $watched-transition-duration: 0.5s;
     }
 
     .thumbnailImage {
+      border-radius: 12px;
+
       @include is-sidebar-item {
         height: 75px;
       }
@@ -150,6 +152,11 @@ $watched-transition-duration: 0.5s;
       margin-right: 3px;
       margin-top: 3px;
       opacity: $thumbnail-overlay-opacity;
+    }
+
+    .watchedProgressBarContainer {
+      width: calc(100% - 24px);
+      margin: auto;
     }
 
     .watchedProgressBar {
@@ -296,6 +303,13 @@ $watched-transition-duration: 0.5s;
       font-size: 13px;
       margin-top: 8px;
     }
+  }
+}
+
+.thumbnailImage {
+  border-radius: 12px;
+  &:hover {
+    box-shadow: 0 0 0 3px var(--side-nav-hover-color);
   }
 }
 


PikachuEXE:
#1452 is about adding an option
This one does not add an option though...

And I don't like both of them (upper = strange progress bar, lower = strange looking in sidebar)

jasonhenriquez:
> #1452 is about adding an option This one does not add an option though...
> 
> And I don't like both of them (upper = strange progress bar, lower = strange looking in sidebar)

You're right, let me remove the "closes" and ask their current opinion on moving forward. I personally am reluctant to make minor aspects like these into configurable settings, as it ends up making open source projects' settings sections quite intimidating and hard to navigate, but I can understand the viewpoint. @peepo5 Is this something you would still like to leave as an option once implemented rather than just the new design?

I do think the progress bar could use improvement, but I do like the rounded look in general. I'll keep experimenting with it a bit; anyone that has a recommendation, please feel welcome to share it.

jasonhenriquez:
@PikachuEXE Here's an approach that only adds the rounded bottom left and right border radisu when the video is watched.
Screenshot_20230826_021939
Screenshot_20230826_022032
Screenshot_20230826_022330

jasonhenriquez:
Here's an approach D that uses a smaller `border-radius` of 6px unconditionally, without changing the watch progress bar. The watch progress bar looks like it fits in just fine. I think this one is the most broadly likeable one.
Screenshot_20230826_023302

PikachuEXE:
I prefer approach D if this must be added ~_~
But having no option to revert is... arrr

jasonhenriquez:
Hm, I didn't think it would be something people had a strong preference, but I suppose I'm wrong. Would it be fine for it to be an option that is enabled by default, or would that be unfavorable for any reason?

PikachuEXE:
The combination of watch progress bar and rounded thumbnail always look weird to me no matter how you make it (just how much)
But let's see how others think...

efb4f5ff-1298-471a-8973-3d47447115dc:
Yeah i think if we go with the rounded elements through the UI it needs a toggle. I dont care tbh but i do think allot of people do.

efb4f5ff-1298-471a-8973-3d47447115dc:
I think i would prefer a PR that addresses every element within FT to be round (with the use of a toggle ofcourse). Dont think its wise to ship a feature into release that would make some elements round.

jasonhenriquez:
Interesting, what other things do you have in mind?

efb4f5ff-1298-471a-8973-3d47447115dc:
1. Border of notification could be round
round1

2. Border of subscription/channels/trending/playlist/history/settings/about/search
round2

3. Watched card border
round3

4. menu border
round4

5. filter border
round5

6. profile select border
round6


7. profile manager border
round7

8. Upper section of main channel page
round8

9. community posts border
round9

10. drop down border
round10

11. player border
round11

12. chapters/description/comments border
round12

13. Active chapter border
round13

14. lower channel page border
round14

15. Edit profile border
round15

16. livechat upnext borders
round16

17. Borders of these menu's
https://github.com/FreeTubeApp/FreeTube/assets/73130443/2e3b6b75-7ea2-47fb-81b7-3acef8e307a3

efb4f5ff-1298-471a-8973-3d47447115dc:
Do you think these suggestions are realistic?

jasonhenriquez:
I think it's more than what was originally intended for the issue / would have to be applying some basic high-level rules for the whole app for the level of effort to match the utility of the feature, are my thoughts

efb4f5ff-1298-471a-8973-3d47447115dc:
Cant deny that its way more than original intention. What is the way forward here?

1. Leave PR as is
2. Implement my proposal to make the elements previously posted rounded (which is obviously not the way)
3. Implement some of the things in the list i created that makes sense?

PikachuEXE:
Please add setting to toggle this

jasonhenriquez:
> Cant deny that its way more than original intention. What is the way forward here?
> 
>     1. Leave PR as is
> 
>     2. Implement my proposal to make the elements previously posted rounded (which is obviously not the way)
> 
>     3. Implement some of the things in the list i created that makes sense?

Probably 3. Modifying `ft-card` stuff fine, but modifying things like the icon button dropdown just seems visually odd to me personally. When I get back to this PR after clearing the other higher priority ones, it will probably be only high-level stuff like `ft-card`s and video/playlist thumbnails. And with a better solution to the watch progress bar. And maybe making it more visually subtle.

> Please add setting to toggle this
Yes, for sure

jasonhenriquez:
Closing this one because I don't think it's worthwhile, and the optionality would be excessive.

efb4f5ff-1298-471a-8973-3d47447115dc:
i think its fair to say that we can close #1452, waiting on consensus :)

Issue #3593, opened 2023-05-27
[Feature Request]: Access all subscribed youtuber's posts from a community viewer button


Growebis:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The community tab has already been implemented  for YouTube channels, but its very tedious to go to each subscribed channel to go check for if they posted anything on their community tab, especially if you have a lot of subscribed channels.

### Proposed Solution

Add a button on the sidebar under Subscriptions that makes FreeTube fetch community posts from all your subscribed channels so you can see them all in one nice place.

### Alternatives Considered

Alternatively you could try implementing it inside Subscriptions as some sort of either independent tab or mix the fetched YouTube videos with community channel posts, like every couple videos fetched one community post is there, or something like that.

### Issue Labels

improvement to existing feature

Growebis:
Awesome!

Issue #3252, opened 2023-03-04
[Bug]: Shared posted on community tab isnt visible 


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

Here the creator shared an post on her timeline but it isn't showing on ours

https://user-images.githubusercontent.com/73130443/222265814-edc75d65-6ca1-48c8-bcca-79d8404167a7.mp4

### Expected Behavior

Implement fix when https://github.com/LuanRT/YouTube.js/pull/334 is merged

### Issue Labels

content not loading

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/4ef2f709ee8983029033a0952dfd22ba86bef6ac

### Operating System Version

Win10 22h2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
https://github.com/LuanRT/YouTube.js/releases/tag/v3.2.0 has been released!

efb4f5ff-1298-471a-8973-3d47447115dc:
Not quite happy how it is being displayed right now. It doesnt say it is shared and the shared text isnt being displayed

![Screenshot (303)](https://user-images.githubusercontent.com/73130443/236672883-39e67f97-c712-463d-9694-040f00235ec8.png)

efb4f5ff-1298-471a-8973-3d47447115dc:
This is a better look of it and u also see where the shared comment of the channel owners is.

https://github.com/FreeTubeApp/FreeTube/assets/73130443/aaff7f48-770c-4a9e-b7c4-22752797017f

Edit: also tested with the channel i posted in the issue and can confirm it is the same.

[...]

Issue #3253, opened 2023-03-04
[Feature Request]: View post on community tab


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

On YouTube u can click on a post in the community tab and it will bring u to the individual post with the comment section. This isn't a thing on FT right now.

### Proposed Solution

User should be able to click on a post and FT brings u to the post with the comments section.

### Alternatives Considered

None.

### Issue Labels

display more information to user, improvement to existing feature

ChunkyProgrammer:
We should probably also support opening the post link

ChunkyProgrammer:
Relies on https://github.com/LuanRT/YouTube.js/issues/352

PR #3973, opened 2023-08-29
Subscriptions: Add community tab

ChunkyProgrammer:
# Subscriptions: Add community tab

## Pull Request Type
- [x] Feature Implementation

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/issues/3593

## Description
If you have less than 125 subscriptions in your profile then a tab is added on the subscription page for community posts.

## Screenshots 

image

## Testing 
(Local + Invidious)
- view subscription page (less than 125 subscriptions)
- see community tab
- click on community tab
- see posts ordered by date

- disable community tab is Distraction Free Settings
- view subscription page
- dont see community tab

- have over 125 subscriptions
- view subscription page
- dont see community tab

## Desktop
- **OS:** windows
- **OS Version:** 10


src/renderer/views/Subscriptions/Subscriptions.js
```diff
@@ -50,6 +61,11 @@ export default defineComponent({
         tabs.push('live')
       }
 
+      // community does not support rss
+      if (!this.hideSubscriptionsCommunity && this.activeSubscriptionList.length < 125) {
```
absidue, line 65: RSS can also be manually enabled in the subscriptions settings

static/locales/en-US.yaml
```diff
@@ -45,6 +45,7 @@ Global:
   Videos: Videos
   Shorts: Shorts
   Live: Live
+  Community: Community
```
absidue, line 48: Please move the existing Community translations in all translation files, to avoid duplicate strings and to reduce the amount of work the translators have to do, the same thing was done in the original subscriptions tab pull request (#3725), which introduced the global object. You'll need to update the channel page to use the Global.Community string too. (Basically there is no point having global strings if we are only going to use them in one place)

src/renderer/components/ft-community-post/ft-community-post.vue
```diff
@@ -8,16 +8,39 @@
...
+      >
+        <img
+          :src="getBestQualityImage(authorThumbnails)"
+          class="communityThumbnail"
+          alt=""
+        >
+      </router-link>
       <img
-        v-if="authorThumbnails.length > 0"
+        v-else-if="authorThumbnails.length > 0"
```
absidue, line 24: Instead of checking the length of authorThumbnails in both the `v-if` and `v-else-if`, you could have a template with a `v-if` that wraps both and then a `v-if` authorId and `v-else` inside.

src/renderer/views/Subscriptions/Subscriptions.js
ChunkyProgrammer: True but when RSS fails it defaults to not using RSS for the other subscription lists

absidue: But this isn't about whether RSS fails or not, if we can't get community posts from RSS then shouldn't we hide the tab when the user has selected RSS as the preferred method of getting the subscriptions?

PikachuEXE:
IV API = useless error messages
image

PikachuEXE:
Changed `this.activeSubscriptionList.length < 125` to 5 so I don't have to subscribe that many channels
But tab still visible
image

PikachuEXE (approved):

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
I dont want to downplay the value this PR bring to FT but it feels incomplete because https://github.com/FreeTubeApp/FreeTube/issues/3253 isnt implemented

efb4f5ff-1298-471a-8973-3d47447115dc:
Message needs to be changed

Capture

ChunkyProgrammer:
> I dont want to downplay the value this PR bring to FT but it feels incomplete because #3253 isnt implemented. Thoughts?

#3253 requires Youtube.js to support it and for invidious to support it (I have a PR open for adding support to Invidious https://github.com/iv-org/invidious/pull/4010).

src/renderer/components/subscriptions-tab-ui/subscriptions-tab-ui.js
```diff
@@ -97,8 +107,10 @@ export default defineComponent({
       switch (event.key) {
         case 'r':
         case 'R':
+          console.error(this.dataLimit)
+          console.error(this.videoList.length)
           if (!this.isLoading) {
-            this.$emit('refresh')
+            // this.$emit('refresh')
           }
```
absidue, line 114: Looks like there is some debug code leftover here.

PikachuEXE: Also please use `console.debug` or `console.log` for debugging
Otherwise linter won't catch those

efb4f5ff-1298-471a-8973-3d47447115dc:
> > I dont want to downplay the value this PR bring to FT but it feels incomplete because #3253 isnt implemented. Thoughts?
> 
> #3253 requires Youtube.js to support it and for invidious to support it (I have a PR open for adding support to Invidious [iv-org/invidious#4010](https://github.com/iv-org/invidious/pull/4010)).

Sorry I forgot that this is basically out of our control. Is it maybe possible to fix https://github.com/FreeTubeApp/FreeTube/issues/3252#issuecomment-1546675781 first? Otherwise user might see allot of shared posts at the bottom of the community page.

absidue:
> > > I dont want to downplay the value this PR bring to FT but it feels incomplete because #3253 isnt implemented. Thoughts?
> > 
> > 
> > #3253 requires Youtube.js to support it and for invidious to support it (I have a PR open for adding support to Invidious [iv-org/invidious#4010](https://github.com/iv-org/invidious/pull/4010)).
> 
> Sorry I forgot that this is basically out of our control. Is it maybe possible to fix [#3252 (comment)](https://github.com/FreeTubeApp/FreeTube/issues/3252#issuecomment-1546675781) first? Otherwise user might see allot of shared posts at the bottom of the community page.

@ChunkyProgrammer Do you want to do this as part of this pull request or is that something that should be done later? (just wondering if I should review this pull request already or whether there is more to come)

ChunkyProgrammer:
Took a quick look into it. From what I can tell, we don't really have any logic for parsing/rendering shared posts in our code (i think there's some old code from when we were using yt-channel-info still there but I don't think it actually gets used?). So, I think it might be an issue with youtube.js ( i'm away from my computer so I can't really test it right now but we have no logic in the api/local.js nor in the .vue component for shared posts)

ChunkyProgrammer:
Shared posts are now hidden and IV shouldn't be giving the `No posts` error message

src/renderer/helpers/api/invidious.js
```diff
@@ -24,7 +24,12 @@ export function invidiousAPICall({ resource, id = '', params = {}, doLogError =
       .then((response) => response.json())
       .then((json) => {
         if (json.error !== undefined) {
-          throw new Error(json.error)
+          // community is empty, no need to display error.
+          if (json.error === 'This channel hasn\'t posted yet') {
```
absidue, line 28: Could you please add a link to the issue on the invidious repo or the one on ours, just so that it's mentioned and the special handling could be removed in the future if the issue is solved.

PikachuEXE (dismissed):
Assuming nothing much change since last reviewed

absidue (approved):

absidue (approved):

PikachuEXE (approved):

jasonhenriquez (approved):

PR #3980, opened 2023-08-31
Make ft-channel-bubble an actual link when it is being used as one

absidue:
# Make ft-channel-bubble an actual link when it is being used as one

## Pull Request Type

- [x] Feature Implementation

## Description
The channel bubble component is used in 4 places, twice for channel selection on the profile edit page and as links on the channel about tab for the featured channels and errored channels on the subscriptions page. Currently we just set the role of the element to link (`role="link"`) but it doesn't act like other links in FreeTube, right, middle and control/command clicking does nothing.

This pull request replaces the `role="link"` in favour of using router-links, that way they behave as expected, meaning you can control/command and middle click on them to open the channel in a new window and right clicking brings up the context menu that allows you to copy the link or open it in a new window.

## Testing 
Check that selecting and deselecting still work on the profile edit page, in both the "Subscriptions List" and "Other Channels" sections.

Check that the channel bubbles on the about tab of channels act as links, so clicking on them still takes you to the channel but also that the context menu with the copy link and open in new window entries show up.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/channel-about/channel-about.js
+++ b/src/renderer/components/channel-about/channel-about.js
@@ -61,10 +61,5 @@ export default defineComponent({
     formattedViews: function () {
       return formatNumber(this.views)
     },
-  },
-  methods: {
-    goToChannel: function (id) {
-      this.$router.push({ path: `/channel/${id}` })
-    },
   }
 })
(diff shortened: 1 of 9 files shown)


PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PR #3985, opened 2023-09-02
Fix/improve community post list styling

### LINKED ISSUE #3984: [Bug]: Community page broken with list view enabled
### Guidelines

### Describe the bug

1. Under General Settings, set Video View Type to "List"
2. [Navigate to this channel](https://www.youtube.com/channel/UCgwv23FVv3lqh567yagXfNg)
3. Navigate to the Community tab
4. Encounter bug (pictured below)

Screenshot_20230902_011647

### Expected Behavior

The Community page should appear normally and legibly.

### Issue Labels

visual bug

### FreeTube Version

v0.19.0

### Operating System Version

OpenSUSE Tumbleweed

### Installation Method

Portable

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

jasonhenriquez:
# Improve community post styling & fix list styling

## Pull Request Type

- [x] Bugfix

## Related issue
closes #3984

## Description
Fixes community post styling on the list view (as in, is legible and actually pretty). Wraps each community post in its own ft-card for list and grid views, and equalizes the list and grid views (given that there does not seem to be a good way to actually display community posts in a grid).

## Screenshots 
Screenshot_20230902_021401
localhost_9080_(iPad Air)
localhost_9080_(iPhone 12 Pro)

## Testing 
1, Under General Settings, set Video View Type to "List"
2. [Navigate to this channel](https://www.youtube.com/channel/UCgwv23FVv3lqh567yagXfNg)
3. Navigate to the Community tab
4. Ensure that view appears correctly
5. Check with smaller device size (Chrome devtool)
6. Repeat for Grid view
7. Check with smaller device size (Chrome devtool)

## Desktop

- **OS:** OpenSUSE
- **OS Version:** Tumbleweed

### DIFF
--- a/src/renderer/components/ft-auto-grid/ft-auto-grid.js
+++ b/src/renderer/components/ft-auto-grid/ft-auto-grid.js
@@ -7,7 +7,7 @@ export default defineComponent({
       type: Boolean,
       required: true
     },
-    usingFtCardWrapper: {
+    useFtCardWrapper: {
       type: Boolean,
       default: false
     }
--- a/src/renderer/components/ft-auto-grid/ft-auto-grid.js
+++ b/src/renderer/components/ft-auto-grid/ft-auto-grid.js
@@ -6,6 +6,10 @@ export default defineComponent({
     grid: {
       type: Boolean,
       required: true
+    },
+    usingFtCardWrapper: {
+      type: Boolean,
+      default: false
     }
   }
 })
--- a/src/renderer/components/ft-auto-grid/ft-auto-grid.scss
+++ b/src/renderer/components/ft-auto-grid/ft-auto-grid.scss
@@ -10,4 +10,19 @@
     display: grid;
     grid-gap: 4px;
   }
+
+  &.card {
+    position: relative;
+    width: 85%;
+    margin-block: 0 20px;
+    margin-inline: auto;
+    box-sizing: border-box;
+
+    .ft-card {
+      margin-inline: 0;
+      margin-block: 0 16px;
+      max-width: 850px;
+      padding-block-end: 3px;
+    }
+  }
 }
--- a/src/renderer/components/ft-auto-grid/ft-auto-grid.scss
+++ b/src/renderer/components/ft-auto-grid/ft-auto-grid.scss
@@ -22,7 +22,7 @@
       margin-inline: 0;
       margin-block: 0 16px;
       max-width: 850px;
-      padding-block-end: 3px;
+      padding-block-end: 6px;
     }
   }
 }
--- a/src/renderer/components/ft-auto-grid/ft-auto-grid.vue
+++ b/src/renderer/components/ft-auto-grid/ft-auto-grid.vue
@@ -4,7 +4,7 @@
     :class="{
       grid: grid,
       list: !grid,
-      card: usingFtCardWrapper
+      card: useFtCardWrapper
     }"
   >
     <slot />
--- a/src/renderer/components/ft-auto-grid/ft-auto-grid.vue
+++ b/src/renderer/components/ft-auto-grid/ft-auto-grid.vue
@@ -3,7 +3,8 @@
     class="ft-auto-grid"
     :class="{
       grid: grid,
-      list: !grid
+      list: !grid,
+      card: usingFtCardWrapper
     }"
   >
     <slot />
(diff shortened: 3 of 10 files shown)


absidue:
Community posts were only ever meant to be displayed in a list view, but I'm guessing it was only ever tested with global grid, forced listed, not global list + forced list.

absidue:
I guess the question is, instead of letting the the ft-community-post component even accept a grid and list value, maybe just doing it's own thing always in list mode would require less workarounds?

jasonhenriquez:
> I guess the question is, instead of letting the the ft-community-post component even accept a grid and list value, maybe just doing it's own thing always in list mode would require less workarounds? 

Yeah I was of two minds about this - it technically isn't doing any good right now, but I thought I'd keep it this way in case someone does have a good way of differentiating between the two styles (+ didn't know if overriding the user's list/grid preference would be a bad idea to start doing). The only difference is that a handful of CSS properties have been added to the grid view version to equalize it. But I don't really care much either way, so I can remove that if we have any reason to at all.

> (can this be fixed in the ft-community-post component without requiring "global" overrides?) Especially requiring the community-posts list on the channel page to be almost at the root, is going to make it impossible to create a shared tab component, because it can't be in a normal nested layout anymore.

Is this in reference to the putting it outside of the `ft-card` object covering a lot of the tabs in `Channel.vue`? Yeah, this one was giving me a lot of decision grief - I think it's nice to have a way to allow these ft-cards to have nested ft-cards, but taking it out of there is messier for sure. How about this: I put it back in there, and add styling to ft-card that overrides the parent styling if it has any `ft-card` children (with a certain class)?

absidue:
We already override the preference in a few places, e.g. the playlists page. The community posts were always meant to override the users global choice, it was just only made for when the global choice is grid, which is why it worked fine for that but not list.

jasonhenriquez:
> We already override the preference in a few places, e.g. the playlists page. The community posts were always meant to override the users global choice (that's why the `display="list"` prop exists), it was just only made for when the global choice is grid, which is why it worked fine for that but not list.
> 
> I partially fixed it a while back, by making the community posts think that the global choice was always grid, but never got round to fully fixing it.

Ah okay, I was thinking that to some extent when I saw it at the top level, but then I saw we were still grabbing their preference on the `ft-community-post` page and was confused. I can just remove it from the community posts directly then.

absidue:
Yeah it's a bit weird, the list has to be in list layout but the post itself, has to be in grid layout for it to work well.

absidue:
I was able to get the community posts to look identical in both the list and grid global layout, with these simple changes. Maybe I'm misunderstanding what this pull request is supposed to achieve, but it seems like it might be overcomplicating this rather simple change (modifying the way ft-card works and using component is, seems like the wrong approach to me).

```diff
diff --git a/src/renderer/components/ft-community-post/ft-community-post.js b/src/renderer/components/ft-community-post/ft-community-post.js
index a7086c29..c7d1c7f7 100644
--- a/src/renderer/components/ft-community-post/ft-community-post.js
+++ b/src/renderer/components/ft-community-post/ft-community-post.js
@@ -52,10 +52,6 @@ export default defineComponent({
         slideBy: 'page',
         navPosition: 'bottom'
       }
-    },
-
-    listType: function () {
-      return this.$store.getters.getListType
     }
   },
   created: function () {
diff --git a/src/renderer/components/ft-community-post/ft-community-post.vue b/src/renderer/components/ft-community-post/ft-community-post.vue
index 870a6680..d528af50 100644
--- a/src/renderer/components/ft-community-post/ft-community-post.vue
+++ b/src/renderer/components/ft-community-post/ft-community-post.vue
@@ -1,9 +1,8 @@
 <template>
   <div
     v-if="!isLoading"
-    class="ft-list-post ft-list-item outside"
+    class="ft-list-post ft-list-item outside grid"
     :appearance="appearance"
-    :class="{ list: listType === 'list', grid: listType === 'grid' }"
   >
     <div
       class="author-div"
```

jasonhenriquez:
So it's doing a bit more than that if you see the picture and description; but I will remove the complicating stuff.

efb4f5ff-1298-471a-8973-3d47447115dc:
Is there anything blocking this except for the conflicts?

jasonhenriquez:
Community posts are still broken, I think, so that needs to be handled

jasonhenriquez:
This one is a nice lil visual improvement for all users, and it does fix Community Post styling for people who use List view, but honestly, I haven't heard a single soul complain about this, meaning very few people probably actually use List view (and/or read community posts on FreeTube). Too low priority for me to care about fixing compared to other items on my list.

kommunarr:

PR #3986, opened 2023-09-02
Chore: Update vscode configurations

Svallinn:
## Pull Request Type

- Other

## Description
Nothing special, just some utilities to be able to debug both main and renderer processes inside VSCode and/or its derivatives.
npm related tasks are auto-detected, so I deleted the `tasks.json` file.
Additionally, fixed a very small bug, not worth placing it in its own commit/PR.


PikachuEXE (approved):
Looks good but I don't use VSCode :)

absidue:
Was able to get `dev-runner (Electron)` to work but `Attach to renderer process (Electron)` gave me this error (lran `yarn dev` and then ran that debug task):
error

Svallinn:
@absidue 
I have to ask
Did you wait for the window to spawn before trying to attach the debugger to it?

absidue:
Yes. Do I need to spawn FreeTube in a special way for the renderer debugger to work? I just launched it like I usually do by running `yarn dev` in my terminal.

Svallinn:
I assumed it would be obvious for anyone looking at the configs, but I guess not.
The whole point of these configs is being able to use them concurrently from the debug tab.
1. Select the `dev-runner` option from the dropdown and launch it (which btw includes the `--remote-debug` argument, so the yarn equivalent would actually be `yarn debug-runner` or  `yarn debug`)
2. Wait for the window to spawn
3. Select the `Attach to renderer process` option from the dropdown and run that as well

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #3970, opened 2023-08-29
Implement right-to-left interface

jasonhenriquez:
# Implement right-to-left interface

Co-authored by @ChunkyProgrammer

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #1550

## Description

At a high level, this PR updates the interface of FreeTube to accommodate the proper patterns for right-to-left languages throughout the entire app. This PR looks quite large, and it is, but >80-85% of the changes are migrating CSS properties to their 1:1 directionality-agnostic equivalents.

Implementation-wise:
- **Doesn't touch VideoJS.css**
- Replaces `margin-left` / `-right` with `margin-inline-start` / `-end` & `margin-top` / `-bottom` with `margin-block-start` / `-end`
- Replaces `padding-left` / `-right` with `padding-inline-start` / `end` & `padding-top` / `-bottom` with `padding-block-start` / `-end`
- Replaces `left`/`right` with `inset-inline-start`/`-end` & `top`/`bottom` with `inset-block-start`/`-end`
- Replaces `border-left*`/`border-right*` with `border-inset-start`/`-end` & `border-top*`/`border-bottom*` with `border-block-start`/`-end`
- Replaces `text-align: left`/`right` with `text-align: start`/`end`
- Custom: replaces `float: left`/`right` with two CSS variables (because [`float: inline-start`/`-end` is still experimental in Chrome](https://developer.mozilla.org/en-US/docs/Web/CSS/float))
- Custom: modifies `transform()` x values to conditionally be inverted by multiplying them with a CSS variable of 1 or -1 based on language directionality (090f392)
- Custom styling was needed/added in one place (`ft-input.css` to fix an emergent issue with the search bar `magnifying-glass` icon positioning on languages with right-to-left directionality
- Fixes settings switches not being aligned when the locale has right-to-left directionality
- Updates arrow keys when the current locale has right-to-left directionality such that left arrow is the "history forward" arrow, and right arrow is the "history backward" arrow
- Refactors the arrow key logic to use class binding to resolve emergent bug with history arrows showing as disabled when changing the locale to one with different directionality. It's also better practice than whatever the original implementer was doing (me, in 2021, to be clear)

## Screenshots 

Screenshot_20230829_022443
Screenshot_20230829_022536
Screenshot_20230829_022631
Screenshot_20230829_022711

### Test Plan
- Under General Settings, set locale to one with right-to-left directionality (e.g., Arabic / العربية )
- Check that sliders in, e.g., Player Settings, are right-to-left
[...]

PR #4011, opened 2023-09-08
Fix localization of counts

ChunkyProgrammer:
# Fix localization of counts

## Pull Request Type
- [x] Bugfix

## Related issue
See: https://github.com/FreeTubeApp/FreeTube/pull/3970#issuecomment-1697564131

## Description
Use TC to properly handle plurals of view count, subscriber count, channel count, etc.

## Testing 
Test the following on both local and invidious api (remember to clear search cache if searching the same thing)
- view a channel in search, look at subscriber and video count (ex: search `the replacements - topic`)
image
- view a video on search, look at view count
image
- click on a video, look at view count
image
- look at the subscribers count on a channel page
image

## Desktop
- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.vue
+++ b/src/renderer/components/ft-list-video/ft-list-video.vue
@@ -90,9 +90,7 @@
           <span>{{ channelName }}</span>
         </router-link>
         <template v-if="!isLive && !isUpcoming && !isPremium && !hideViews">
-          <span class="viewCount"><template v-if="channelId !== null"> •</template> {{ parsedViewCount }} </span>
-          <span v-if="viewCount === 1">{{ $t("Video.View").toLowerCase() }}</span>
-          <span v-else>{{ $t("Video.Views").toLowerCase() }}</span>
+          <span class="viewCount"><template v-if="channelId !== null"> •</template>{{ $tc('Global.Counts.View Count', viewCount, {count: parsedViewCount}) }}</span>
         </template>
         <span
           v-if="uploadedTime !== '' && !isLive && !inHistory"
@@ -105,7 +103,7 @@
         <span
           v-if="isLive && !hideViews"
           class="viewCount"
-        > • {{ parsedViewCount }} {{ $t("Video.Watching").toLowerCase() }}</span>
+        > • {{ $tc('Global.Counts.Watching Count', viewCount, {count: parsedViewCount}) }}</span>
       </div>
       <ft-icon-button
         class="optionsButton"
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -45,6 +45,12 @@ Global:
   Videos: Videos
   Shorts: Shorts
   Live: Live
+  Counts:
+    Video Count: 1 video | {count} videos
+    Channel Count: 1 channel | {count} channels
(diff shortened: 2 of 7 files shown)


PikachuEXE (approved):

absidue:
Looking at the second screenshot it looks like you've removed the whitespace between the divider and the view count.

absidue:
Also you should probably get rid of the old strings as they are presumably unused now.

PikachuEXE:
Space between divider & view count still missing
image

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -90,9 +90,7 @@
           <span>{{ channelName }}</span>
         </router-link>
         <template v-if="!isLive && !isUpcoming && !isPremium && !hideViews">
-          <span class="viewCount"><template v-if="channelId !== null"> •</template> {{ parsedViewCount }} </span>
-          <span v-if="viewCount === 1">{{ $t("Video.View").toLowerCase() }}</span>
-          <span v-else>{{ $t("Video.Views").toLowerCase() }}</span>
+          <span class="viewCount"><template v-if="channelId !== null"> •</template>{{ $tc('Global.Counts.View Count', viewCount, {count: parsedViewCount}) }}</span>
```
PikachuEXE, line 93: Probably the previous method produces leading spaces and new one does not
Following change tested locally
Also make the long line not so long
```suggestion
          <span class="viewCount">
            <template v-if="channelId !== null"> • </template>
            {{ $tc('Global.Counts.View Count', viewCount, {count: parsedViewCount}) }}
          </span>
```
image

static/locales/en-US.yaml
```diff
@@ -48,6 +48,9 @@ Global:
   Counts:
     Video Count: 1 video | {count} videos
     Channel Count: 1 channel | {count} channels
+    Subscriber Count: 1 subscriber | {count} subscribers
+    View Count: 1 view | {count} views
+    Watching Count: 1 watching | {count} watching
```
PikachuEXE, line 53: Though not mentioned by
https://kazupon.github.io/vue-i18n/guide/pluralization.html#pluralization

When zero value not provided and count is zero, the plural version is used (I tested locally) (for en at least)

PikachuEXE (approved):

static/locales/en_GB.yaml
```diff
@@ -649,15 +647,12 @@ Video:
...
   Open in Invidious: 'Open in Invidious'
   Copy Invidious Link: 'Copy Invidious Link'
-  View: 'View'
   Views: 'Views'
   Loop Playlist: 'Loop Playlist'
   Shuffle Playlist: 'Shuffle Playlist'
   Reverse Playlist: 'Reverse Playlist'
   Play Next Video: 'Play Next Video'
   Play Previous Video: 'Play Previous Video'
-  # Context is "X People Watching"
```
jasonhenriquez, line 659: I do see that we're not including this comment anymore, which I think is fine as the pluralization and placement makes the context of the translation label more intuitive. As a sidenote, comments like these do make me wonder if having implicit subjects/objects in our labels is a burden for [a subset of] our translators; I have no idea.

jasonhenriquez (approved):

absidue (approved):

PR #4013, opened 2023-09-09
Set new consent cookie and supress tracking cookies on the watch page

absidue:
# Set new consent cookie and supress tracking cookies on the watch page

## Pull Request Type

- [x] Bugfix

## Description
This pull request set the new consent cookie, based on what yt-dlp uses, although the old cookie still seems to be used too.

It also gets rid of tracking cookies that YouTube is returning, maybe we should consider changing the values of the consent cookies at some point, so that YouTube doesn't return those cookies.

## Screenshots 
before:
before

after:
after

## Testing 
1. Open a video
2. In the devtools look at the network log
3. Click on `https://www.youtube.com/sw.js_data` and `https://www.youtube.com/iframe_api`
4. Switch to their cookies tab
5. Check that only the `CONSENT` and `SOCS` cookies are listed


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

jasonhenriquez:
Could you explain the implications of this change and/or share a resource or another example of this? It's not that I don't trust you, but I'm not currently able to review this with knowledgeable confidence.

absidue:
yt-dlp uses the SOCS cookie too now (fixed some extraction errors for them), as for nuking tracking cookies before Electron can save them, so that it doesn't send them along with future requests, that's something that should be done regardless of whether you want the new consent cookie or not.

PikachuEXE:
PR for `SOCS` cookie on `yt-dlp`: https://github.com/yt-dlp/yt-dlp/pull/7774

PikachuEXE:
I got 1 extra: `VISITOR_INFO1_LIVE` via https://youtu.be/nXkgbmr3dRA
Expected?
image

absidue:
This pull request should block it, try restarting FT, after switching to this branch.

PikachuEXE:
To be sure it isn't my local env screwed up I tried `development` too (same video https://youtu.be/nXkgbmr3dRA)
image

Maybe I need to manually remove that key?

absidue:
development has the cookies, this pull request aims to get rid of them

PikachuEXE:
I think you need to add code to remove existing cookies 
Since they (e.g. `VISITOR_INFO1_LIVE`) have long expiry time (probably 1 year)

image

absidue:
@PikachuEXE The problem is that you are quitting FreeTube instead of closing all windows. When you close the last window FreeTube clears up browsing data including cookies, we can't run it when quitting because it's an asynchronous operation but the quit event handler only waits for synchronous code.

This sounds like an issue that will be exclusive to macOS as quitting without closing the windows isn't possible on Windows or Linux (unless you force kill it in the task manager but then you'll be fully aware that the app hasn't had time to clean up).

PikachuEXE:
https://www.electronjs.org/docs/latest/api/app#event-window-all-closed

>  If the user pressed Cmd + Q, or the developer called app.quit(), Electron will first try to close all the windows and then emit the will-quit event, and in this case the window-all-closed event would not be emitted.

See if you want this issue to be fixed in another PR or not
Or I can try to fix myself coz I got access to MacOS

PikachuEXE:
I have updated code locally and tested

<details>
<summary>Code</summary>

```js
  let resourcesCleanUpDone = false

  app.on('window-all-closed', () => {
    // Clean up resources (datastores' compaction + Electron cache and storage data clearing)
    cleanUpResources().finally(() => {
      if (process.platform !== 'darwin') {
        app.quit()
      }
    })
  })

  if (process.platform === 'darwin') {
    // `window-all-closed` will no fire on Cmd+Q
    // https://www.electronjs.org/docs/latest/api/app#event-window-all-closed
    // It's also fired when `app.quit` called
    // Not using `before-quit` since that one is fired before windows are closed
    app.on('will-quit', (e) => {
      // Let app quit of resource cleanup done
      if (resourcesCleanUpDone) { return }

      e.preventDefault()
      cleanUpResources().finally(() => {
        // Auto quit again AFTER resource cleanup done
        // Which will call this listener again
        app.quit()
      })
    })
  }

  function cleanUpResources() {
    if (resourcesCleanUpDone) {
      return Promise.resolve()
    }

    return Promise.allSettled([
      baseHandlers.compactAllDatastores(),
      session.defaultSession.clearCache(),
      session.defaultSession.clearStorageData({
        storages: [
          'appcache',
          'cookies',
          'filesystem',
          'indexdb',
          'shadercache',
          'websql',
          'serviceworkers',
          'cachestorage'
        ]
      })
    ])
      .then(() => {
        resourcesCleanUpDone = true
      })
  }
```

</details>

My test steps:
- Run in dev, ensure cookie present (`VISITOR_INFO1_LIVE`)
- Run in new branch, ensure cookie present in request (cleanup not done yet)
- Cmd+Q, start again, ensure cookie absent in request

Not tested in other OSes

ChunkyProgrammer:
@absidue are you planning on adding the macos code to this pr?

PikachuEXE:
I can put it into a new PR if you prefer
But I guess I am still the only one testing MacOS :P

absidue:
I can add it to this one, I just can't test it.

PikachuEXE:
Please add and I will test it again

absidue:
Updated this pull request.

PikachuEXE (approved):

PikachuEXE:
Re-tested on MacOS with steps in
https://github.com/FreeTubeApp/FreeTube/pull/4013#issuecomment-1716812876

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer @jasonhenriquez please review when u have time

ChunkyProgrammer (approved):

PR #3006, opened 2022-12-27
Add feature to remember video history with last viewed playlist when enabled

PikachuEXE:
# Add feature to remember video history with last viewed playlist when enabled

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
N/A

## Description
This is a feature I want XD
I constantly listen to specific playlists and it's quite annoying to sometimes having to close the app
And after app restart I go back to history but unable to continue my playback progress with the last viewed playlist
So I write this feature for myself :D 
There is a new option too which can be disabled (but enabled by default)

The playlist would **only** be applied when visited via history
If video viewed via URL pasted (with/without playlist) or non-history list it should **not** restore last viewed playlist

## Screenshots 
Settings
image

### Disabled
Should be old behaviour

### Enabled
- Find a video (without playlist) to view, go to history, view that same video, no playlist restored
- Find a video (with playlist PL-A) to view, go to history, view that same video, PL-A restored
  - Then view the same video with different playlist (PL-B), go to history, view that same video, PL-B restored
  - Then view the same video with no playlist, go to history, view that same video, no playlist restored
- Find a video (with playlist PL-A) to view, go to channel, view that same video, no playlist restored

If you need actual videos & playlists provided as test cases let me know

## Desktop

- **OS:** MacOS
- **OS Version:** 12.6.2
- **FreeTube version:** 75046777e937eb7a94ce8457d37449fa1be9324a

## Additional context
WIP coz `If video viewed via non-history list it should **not** restore last viewed playlist` is not done yet

PR #4017, opened 2023-09-10
Improve performance of the watch history handling

### LINKED ISSUE #3796: [Bug]: Slow rendering with large history
### Guidelines

### Describe the bug

After importing my youtube history of ~32k videos rendering subscription pages slowed down a lot.

Profiling one load 1.5s is spent initializing the page to display, 80% of that in `historyIndex`, rendering each video calls it twice taking ~40ms each time.

### Expected Behavior

Increasing history size should barely affect page rendering speed.

### Issue Labels

usability issue

### FreeTube Version

v0.18.0-nightly-3112 Beta

### Operating System Version

Windows 10

### Installation Method

.exe

### Primary API used

Local API

_No response_

### Additional Information

It looks like it's doing a linear scan through the history, building an index by video-id would speed this up greatly.

### Nightly Build

absidue:
# Improve performance of the watch history handling

## Pull Request Type

- [x] Bugfix
- [x] Performance Improvement

## Related issue
closes #3796

## Description
Currently all watch history interactions go through a chronologically sorted list, which is great for the history page, so that it displays items in the right order, but less than ideal if you need to check if a video is watched. For most users this will likely never be a problem, but for users like the ones in the issue, that have gigantic watch histories (32k+), this means that video lists and especially the subscriptions page, slow down.

This pull request introduces an additional history object, this time it is an object (vuex doesn't support [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)s) with the video ids as the keys. This should allow for almost instant lookups, which should speed up video lists and any other place that accesses the watch history to check an entry exist and read properties from them.

## Testing 
Check that history handling still works as expected:
- Watch history and watch progress get saved and restored on the next playback (well the watch progress might not get restored, as it is now fast enough that it might try to do it before the video is loaded, yes it's a race condition that I am fully aware of, using a link with a timestamp doesn't work either if you have no history)
- Adding and removing and item from the history should live update it's state in a second window
- Watch history page still works
- Exporting history still works
- The hide watched items on the subscription page setting still works

@Nemo157 @apatheticslacker @StrangestNoob
As none of the devs have a large enough history to encounter any performance problems, it would be really helpful if you could test this pull request and check if it solves the problem for you. (Ideally all 3 of you, but mainly @Nemo157 as you opened the bug report).

Here is a test build you 3 can download, so that you don't have to set up a development environment to test it: https://github.com/absidue/FreeTube/actions/runs/6136849377

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -76,8 +76,8 @@ export default defineComponent({
     }
   },
   computed: {
-    historyCache: function () {
-      return this.$store.getters.getHistoryCache
+    historyEntry: function () {
+      return this.$store.getters.getHistoryCacheById[this.id]
     },
 
     listType: function () {
@@ -313,12 +313,6 @@ export default defineComponent({
       }
     },
 
-    historyIndex: function() {
-      return this.historyCache.findIndex((video) => {
-        return video.videoId === this.id
-      })
-    },
-
     playlistIdFinal: function () {
       if (this.playlistId) {
         return this.playlistId
@@ -327,12 +321,8 @@ export default defineComponent({
       // Get playlist ID from history ONLY if option enabled
       if (!this.showVideoWithLastViewedPlaylist) { return }
       if (!this.saveVideoHistoryWithLastViewedPlaylist) { return }
-      const historyIndex = this.historyIndex
-      if (historyIndex === -1) {
-        return undefined
-      }
 
-      return this.historyCache[historyIndex].lastViewedPlaylistId
+      return this.historyEntry?.lastViewedPlaylistId
     },
 
     currentLocale: function () {
@@ -348,7 +338,7 @@ export default defineComponent({
     }
   },
   watch: {
-    historyIndex() {
+    historyEntry() {
       this.checkIfWatched()
     },
   },
@@ -451,8 +441,8 @@ export default defineComponent({
       this.channelName = this.data.author ?? null
       this.channelId = this.data.authorId ?? null
 
-      if (this.data.isRSS && this.historyIndex !== -1) {
-        this.duration = formatDurationAsTimestamp(this.historyCache[this.historyIndex].lengthSeconds)
+      if (this.data.isRSS && typeof this.historyEntry !== 'undefined') {
+        this.duration = formatDurationAsTimestamp(this.historyEntry.lengthSeconds)
       } else {
         this.duration = formatDurationAsTimestamp(this.data.lengthSeconds)
       }
@@ -538,22 +528,25 @@ export default defineComponent({
     },
 
     checkIfWatched: function () {
-      const historyIndex = this.historyIndex
+      const historyEntry = this.historyEntry
 
-      if (historyIndex !== -1) {
+      if (typeof historyEntry !== 'undefined') {
         this.watched = true
         if (this.saveWatchedProgress) {
           // For UX consistency, no progress reading if writing disabled
-          this.watchProgress = this.historyCache[historyIndex].watchProgress
+          this.watchProgress = historyEntry.watchProgress
         }
 
-        if (this.historyCache[historyIndex].published !== '') {
-          const videoPublished = this.historyCache[historyIndex].published
+        if (historyEntry.published !== '') {
+          const videoPublished = historyEntry.published
           const videoPublishedDate = new Date(videoPublished)
           this.publishedText = videoPublishedDate.toLocaleDateString()
         } else {
           this.publishedText = ''
         }
+      } else {
+        this.watched = false
+        this.watchProgress = 0
       }
     },
 
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -123,8 +123,8 @@ export default defineComponent({
     }
   },
   computed: {
-    historyCache: function () {
-      return this.$store.getters.getHistoryCache
+    historyEntry: function () {
+      return this.$store.getters.getHistoryCacheById[this.videoId]
     },
     rememberHistory: function () {
       return this.$store.getters.getRememberHistory
@@ -1040,9 +1040,7 @@ export default defineComponent({
     },
 
     checkIfWatched: function () {
-      const historyIndex = this.historyCache.findIndex((video) => {
-        return video.videoId === this.videoId
-      })
+      const historyEntry = this.historyEntry
 
       if (!this.isLive) {
         if (this.timestamp) {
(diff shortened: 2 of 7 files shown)


Nemo157:
I can confirm this version gets switching between profiles on the subscriptions page back to being instant, from the ~2s it is currently.

StrangestNoob:
It works.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM

jasonhenriquez:
All seems great, except I see a new error when importing history that was exported from this PR (Picture 2) compared to history that was exported from the `development` branch (Picture 1).

Picture 1
Screenshot_20230910_183053

Picture 2
Screenshot_20230910_183753

PikachuEXE:
@jasonhenriquez
Could it be the issue of https://github.com/FreeTubeApp/FreeTube/pull/3006 not updating the import logic? (which would be a separate issue)
`lastViewedPlaylistId` not in required keys (well it's not required, there should be optional keys)

absidue:
I think the issue is probably unrelated as this doesn't touch any of the import code.

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -76,8 +76,8 @@ export default defineComponent({
     }
   },
   computed: {
-    historyCache: function () {
-      return this.$store.getters.getHistoryCache
+    historyEntry: function () {
+      return this.$store.getters.getHistoryCacheById[this.id]
     },
```
PikachuEXE, line 81:
```suggestion
    },
    historyEntryExists: function() {
      return typeof this.historyEntry !== 'undefined'
    },
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -451,8 +441,8 @@ export default defineComponent({
       this.channelName = this.data.author ?? null
       this.channelId = this.data.authorId ?? null
 
-      if (this.data.isRSS && this.historyIndex !== -1) {
-        this.duration = formatDurationAsTimestamp(this.historyCache[this.historyIndex].lengthSeconds)
+      if (this.data.isRSS && typeof this.historyEntry !== 'undefined') {
```
PikachuEXE, line 444:
```suggestion
      if (this.data.isRSS && this.historyEntryExists) {
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -538,22 +528,25 @@ export default defineComponent({
     },
 
     checkIfWatched: function () {
-      const historyIndex = this.historyIndex
+      const historyEntry = this.historyEntry
 
-      if (historyIndex !== -1) {
+      if (typeof historyEntry !== 'undefined') {
```
PikachuEXE, line 533:
```suggestion
      if (this.historyEntryExists) {
```

src/renderer/views/Watch/Watch.js
```diff
@@ -123,8 +123,8 @@ export default defineComponent({
     }
   },
   computed: {
-    historyCache: function () {
-      return this.$store.getters.getHistoryCache
+    historyEntry: function () {
+      return this.$store.getters.getHistoryCacheById[this.videoId]
     },
```
PikachuEXE, line 128:
```suggestion
    },
    historyEntryExists: function() {
      return typeof this.historyEntry !== 'undefined'
    },
```

src/renderer/views/Watch/Watch.js
```diff
@@ -1053,9 +1051,9 @@ export default defineComponent({
           } else {
             this.$refs.videoPlayer.player.currentTime(this.timestamp)
           }
-        } else if (this.saveWatchedProgress && historyIndex !== -1) {
+        } else if (this.saveWatchedProgress && typeof historyEntry !== 'undefined') {
```
PikachuEXE, line 1054:
```suggestion
        } else if (this.saveWatchedProgress && this.historyEntryExists) {
```

src/renderer/views/Watch/Watch.js
```diff
@@ -1066,8 +1064,8 @@ export default defineComponent({
       if (this.rememberHistory) {
         if (this.timestamp) {
           this.addToHistory(this.timestamp)
-        } else if (historyIndex !== -1) {
-          this.addToHistory(this.historyCache[historyIndex].watchProgress)
+        } else if (typeof historyEntry !== 'undefined') {
```
PikachuEXE, line 1067:
```suggestion
        } else if (this.historyEntryExists) {
```

src/renderer/views/Watch/Watch.js
```diff
@@ -1040,9 +1040,7 @@ export default defineComponent({
     },
 
     checkIfWatched: function () {
-      const historyIndex = this.historyCache.findIndex((video) => {
-        return video.videoId === this.videoId
-      })
+      const historyEntry = this.historyEntry
```
PikachuEXE, line 1043: Is this `historyEntry` supposed to be used below?

PikachuEXE (changes requested):
Readability

PikachuEXE (commented):
Tested others except

> Watch history and watch progress get saved and restored on the next playback (well the watch progress might not get restored, as it is now fast enough that it might try to do it before the video is loaded, yes it's a race condition that I am fully aware of, using a link with a timestamp doesn't work either if you have no history)

PikachuEXE (dismissed):

jasonhenriquez (dismissed):

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -451,8 +445,8 @@ export default defineComponent({
       this.channelName = this.data.author ?? null
       this.channelId = this.data.authorId ?? null
 
-      if (this.data.isRSS && this.historyIndex !== -1) {
-        this.duration = formatDurationAsTimestamp(this.historyCache[this.historyIndex].lengthSeconds)
+      if (this.data.isRSS && typeof this.historyEntryExists) {
```
ChunkyProgrammer, line 448: Is there a reason typeof is being used here?

absidue: forgot to remove it in my last commit 🙈

absidue, line 448:
```suggestion
      if (this.data.isRSS && this.historyEntryExists) {
```

jasonhenriquez (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #4023, opened 2023-09-10
Implement Stylelint linting of CSS logical properties

jasonhenriquez:
# Implement Stylelint linting of CSS logical properties

## Pull Request Type

- [x] Feature Implementation

## Related issue
See [comment](https://github.com/FreeTubeApp/FreeTube/pull/3970#issuecomment-1697564131)https://github.com/FreeTubeApp/FreeTube/pull/3970#issuecomment-1697564131

## Description
This change implements the Stylelint plugin [stylelint-use-logical-spec](https://github.com/Jordan-Hall/stylelint-use-logical-spec) with all rules enabled except for requiring `float: inline-start` and `float: inline-end`, which as of today is still an experimental feature in Chrome.

To prevent current issues blocking contributors going forward, this PR also enacts this rule throughout FreeTube. This mostly just changes all CSS properties with `height` and `width` to become `block-size` and `inline-size` respectively, which is only functionally relevant for different [`writing-mode`s](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode) (note that the default `horizontal-tb` is the only `writing-mode` currently supported by FreeTube, but that is subject to change in the future).

This PR enables Stylelint for `npm run lint` and `npm run lint-fix`, including minor modification of the other scripts to make space for this. This PR also runs Stylelint on pre-commit. This was never enabled beforehand due to concerns about which other rules should be needed. As a result, this PR removes the other plugins and rules, and they can be added back as necessary in their own tested PRs.

## Screenshots 
Screenshot_20230910_161625

## Testing 
1. Add a non-logical property (e.g., `padding-right: 55px`) to a CSS file (e.g., `App.vue`).
2. Save the file and try to commit the change.
3. Encounter pre-commit hook notifying you of the issue with the non-logical property.
4. Run `npm run lint` to see the same assessment as in 3.
5. Run `npm run lint-fix` to see the issue automatically resolved by the linter.
6. Commit the fixed change; you should encounter no issues.
7. (Optional) run other slightly modified commands to ensure they work correctly (`eslint-lint`, `eslint-lint-fix`, `lint-style-fix`)

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

## Additional context
Given that `transform` and `float` are still vulnerable to introducing directionality concerns, code contributors and reviewers must take special care to [use these CSS horizontal directionality variables going forward](https://github.com/FreeTubeApp/FreeTube/blob/c96f8c58c915feb23355a227e15f311ed5f96531/src/renderer/themes.css#L1008-L1010). We currently only care about the x-dimension for right-to-left language support. Code contributors and reviewers must also take care to use [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) inside `style` blocks and properties in `.js` and `.vue` files. Currently, the [stylelint-use-logical-spec](https://github.com/Jordan-Hall/stylelint-use-logical-spec) plugin is not capable of detecting these instances. We should probably add this note to the FreeTube Docs as well.


.stylelintignore
```diff
@@ -5,3 +5,4 @@ src/renderer/videoJS.css
 dist/
 static/
 node_modules/
+src/renderer/components/ft-video-player
```
ChunkyProgrammer, line 8: Is there a reason that this is ignored?

jasonhenriquez: The intent was to avoid mucking up the video player with flipping the directionality of elements, which seems to be how YouTube handles it as well (except for things like the caption menu items, which we have as centered in FT). But you bring a good point, in that this linting should just be disabled for this specific rule: implemented now.

On an unrelated note, I did keep VideoJS.css in the ignore file for that reason as well & because it was there previously, but we could just disable the logical property linting rule there. Is there a reason why we wanted this file to be ignored by Stylelint? It should not be super relevant with a replacement on the horizon, but I couldn't find a rationale anywhere on the original PR.

ChunkyProgrammer: Iirc I ignored VideoJS.css because a lot of that file is from videojs (with some customizations added). So, I was thinking it might be desirable to move our custom css to a different file and use the css directly from videojs

jasonhenriquez: I didn't know that, thanks for the info. That's reasonable. Do you have any other questions or concerns with the PR?

ChunkyProgrammer:
This mostly has to do with changes that I added before this PR but I think these changes should be done here as we are enforcing stylelint now: (other than the following changes, the PR looks good to me)
``` diff --git a/package.json b/package.json
index 3e6ccdf5..cae7d2d8 100644
--- a/package.json
+++ b/package.json
@@ -33,18 +33,14 @@
     "dev-runner": "node _scripts/dev-runner.js",
     "get-instances": "node _scripts/getInstances.js",
     "get-regions": "node _scripts/getRegions.mjs",
-    "lint-all": "run-p lint lint-json lint-style",
+    "lint-all": "run-p lint lint-json",
     "lint": "run-p eslint-lint lint-style",
     "lint-fix": "run-p eslint-lint-fix lint-style-fix",
     "eslint-lint": "eslint --ext .js,.vue ./",
     "eslint-lint-fix": "eslint --fix --ext .js,.vue ./",
     "lint-json": "eslint --ext .json ./",
-    "lint-style": "run-p lint-style:scss lint-style:css",
-    "lint-style:scss": "stylelint \"**/*.scss\"",
-    "lint-style:css": "stylelint \"**/*.css\"",
-    "lint-style-fix": "run-p lint-style-fix:scss lint-style-fix:css",
-    "lint-style-fix:scss": "stylelint --fix \"**/*.scss\"",
-    "lint-style-fix:css": "stylelint --fix \"**/*.css\"",
+    "lint-style": "stylelint \"**/*.{css,scss}\"",
+    "lint-style-fix": "stylelint --fix \"**/*.{css,scss}\"",
     "lint-yml": "eslint --ext .yml,.yaml ./",
     "pack": "run-p pack:main pack:renderer",
     "pack:main": "webpack --mode=production --node-env=production --config _scripts/webpack.main.config.js",

**Review by ChunkyProgrammer — APPROVED** [review]

**Review by absidue — APPROVED** [review]

**Review by PikachuEXE — APPROVED** [review]

File: package.json [hunk 2]
```diff
@@ -66,6 +65,7 @@
     "marked": "^8.0.0",
     "path-browserify": "^1.0.1",
     "process": "^0.11.10",
+    "stylelint-use-logical-spec": "^5.0.0",
```
ChunkyProgrammer, line 68: Just realized that this was added as a Dependency and not a DevDependency

jasonhenriquez: Darn, you're right

PR #4036, opened 2023-09-11
Sort list of external players

trostboot:
# Sort list of external players

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

As more players get added, the list should be sorted for better UX. This PR sorts the elements before display and keeps the default entry for no external player at the top.

## Screenshots 

{4A13AAC2-618E-41ED-B76D-6BE209D6D40C}

## Testing 

Has been checked by opening the settings and it works as designed.

## Desktop

- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** v0.19.0

## Additional context

Due to the way this setting appears to be saved, changing the order will mix up previously existing configurations and users would have to go back into the settings and set their preferred player again.
I can't really think of an elegant and simple way to avoid this. If you do, I'm all ears and happy to make changes.

If you feel that breaking this setting is too much, I'd understand. But I do believe it's necessary in the long run to keep the list from becoming too unwieldy.

### DIFF
--- a/src/renderer/components/external-player-settings/external-player-settings.js
+++ b/src/renderer/components/external-player-settings/external-player-settings.js
@@ -21,9 +21,20 @@ export default defineComponent({
   computed: {
     externalPlayerNames: function () {
       const fallbackNames = this.$store.getters.getExternalPlayerNames
-      const nameTranslationKeys = this.$store.getters.getExternalPlayerNameTranslationKeys
+      const nameTranslationKeys = this.$store.getters.getExternalPlayerNameTranslationKeys.map((translationKey, idx) => this.$te(translationKey) ? this.$t(translationKey) : fallbackNames[idx])
 
-      return nameTranslationKeys.map((translationKey, idx) => this.$te(translationKey) ? this.$t(translationKey) : fallbackNames[idx])
+      // Sort list of external players alphabetically & case-insensitive and keep the default 'None' in the first index
+      const nameTranslationKeysSorted = [...nameTranslationKeys]
+
+      nameTranslationKeysSorted.sort((a, b) => a.localeCompare(b, undefined, { sensitivity: 'base' }))
+
+      const playerNone = nameTranslationKeys[0]
+      const playerNoneIndex = nameTranslationKeysSorted.indexOf(playerNone)
+
+      nameTranslationKeysSorted.splice(playerNoneIndex, 1)
+      nameTranslationKeysSorted.unshift(playerNone)
+
+      return nameTranslationKeysSorted
     },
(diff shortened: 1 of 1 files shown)


trostboot:
Actually I just realized, this would also potentially break the setting for a user anytime a new player is added, so the way the setting is saved needs to be reworked. I'll turn this into a draft for now.

absidue (commented):
Instead of leaving None in the list while sorting and then correcting the position afterwards, removing it from the list, sorting and then adding it back might save you a bit of code.

src/renderer/components/external-player-settings/external-player-settings.js
```diff
@@ -21,9 +21,15 @@ export default defineComponent({
   computed: {
     externalPlayerNames: function () {
       const fallbackNames = this.$store.getters.getExternalPlayerNames
-      const nameTranslationKeys = this.$store.getters.getExternalPlayerNameTranslationKeys
+      const nameTranslationKeys = this.$store.getters.getExternalPlayerNameTranslationKeys.map((translationKey, idx) => this.$te(translationKey) ? this.$t(translationKey) : fallbackNames[idx])
 
-      return nameTranslationKeys.map((translationKey, idx) => this.$te(translationKey) ? this.$t(translationKey) : fallbackNames[idx])
+      // Sort list of external players alphabetically & case-insensitive and keep the default 'None' in the first index
+      const playerNone = nameTranslationKeys[0]
+      nameTranslationKeys.splice(0, 1)
```
absidue, line 28:
```suggestion
      const playerNone = nameTranslationKeys.shift()
```

absidue (commented):
Also to answer your comment about it breaking the setting, it's because you are only changing the order of the names but not reordering the values to match

so instead of it being vlc -> vlc, it might now be smplayer -> vlc

absidue:
If you do the sorting in `src/renderer/store/modules/utils.js` in `getExternalPlayerCmdArgumentsData` instead, before it gets split up into invididual names and values arrays, you won't have the problem with the setting breaking.

trostboot:
All excellent suggestions, thank you. I had already realized the value issue, but doing it in utils.js instead was clearly the better choice to begin with.
It now works as it is, but the code is a bit ugly, I'm sure there's a nicer way to do it. I'll sleep on it and go over it again tomorrow.

jasonhenriquez (approved):
LGTM!

PikachuEXE (approved):
Confirmed option values matching labels
image

absidue (approved):

PR #4045, opened 2023-09-14
Change icon color of FT logo

### LINKED ISSUE #4020: [Feature Request]: Change icon color of FT logo on About page based on set theme
### Guidelines

### Problem Description

Depending on what theme u pick the FT icon in the top bar will change color. This isnt the case for the FT icon on the about page. Some themes are not compatible with the default FT icon. So it would make sense to use the same icon color everywhere based on the set theme.

### Proposed Solution

Make icon consistent with the icon color of the set theme

### Alternatives Considered

Dont implement this

### Issue Labels

visual improvement

### Additional Information

_No response_

Shubh8899:
Change icon color of FT logo

Pull Request Type

- [ ] Feature Implementation

Related issue

closes #4020 

Description
Changed icon color of FT logo on based on set theme

### DIFF
--- a/src/renderer/views/About/About.scss
+++ b/src/renderer/views/About/About.scss
@@ -13,8 +13,29 @@
 }
 
 .logo {
-  max-inline-size: 100%;
-  inline-size: 500px;
+  block-size: 200px;
+  inline-size: 640px;
+  display: inline-flex;
+  background-image: url('../../../../_icons/logoWhite.png');
+
+  @at-root {
+    .dracula &,
+    .CatppuccinMocha &,
+    .hotPink & {
+      background-image: url('../../../../_icons/logoWhite.png');
+    }
+
+    .system &,
+    .light &,
+    .dark &,
+    .black & {
+      background-image: url('../../../../_icons/logoColor.png');
+    }
+
+    .pastelPink & {
+      background-image: url('../../../../_icons/logoBlack.png');
+    }
+  }
 }
 
 .version {
--- a/src/renderer/themes.css
+++ b/src/renderer/themes.css
@@ -141,8 +141,8 @@
   --side-nav-hover-color: #57596B;
   --side-nav-active-color: #3D4051;
   --search-bar-color: #3E3F4A;
-  --logo-icon: url("../../_icons/iconColorSmall.png");
-  --logo-text: url("../../_icons/textColorSmall.png");
+  --logo-icon: url("../../_icons/iconDraculaLightSmall.png");
+  --logo-text: url("../../_icons/textDraculaLightSmall.png");
 }
 
 .catppuccinMocha {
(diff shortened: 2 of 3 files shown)


jasonhenriquez:
Hi @Shubh8899, I believe there might be a misunderstanding. #4020 was asking for the color of the FreeTube icon on the About page (`src/renderer/views/About`) to match the color in the top navigation bar for each theme, not to make all of the top navigation colors the same. Could you please revert this last commit and modify the About page?

Screenshot_20230914_083116

Shubh8899:
@jasonhenriquez  okay got it, will make the change

src/renderer/views/About/About.scss
```diff
@@ -13,8 +13,29 @@
...
+  block-size: 200px;
+  inline-size: 640px;
+  display: inline-flex;
+  background-image: url('../../../../_icons/logoWhite.png');
+
+  @at-root {
+    .dracula &,
+    .CatppuccinMocha &,
+    .hotPink & {
+      background-image: url('../../../../_icons/logoWhite.png');
```
jasonhenriquez, line 25: I see that you're using mapping to the these three logo icons because we don't have full logo images for all of our themes. This is probably the best easily available solution, but it may need adjustment to meet FreeTube's muster. 

I'll talk with the FreeTube team about what we can do next. In the meantime, thank you for your initiative in working on this, and I'll see what we can do to work this out.

jasonhenriquez (changes requested):

src/renderer/views/About/About.scss
jasonhenriquez: As far as I can see, you have two choices:
1. I used an online tool to vectorize our existing logo. You can add it to `_icons`, add it inline to the template, and use the `fill` CSS property to dynamically change the color to the `--primary-text-color` for that theme.
logoBlack
2. You can use an editor like GIMP to create logo images for themes missing one (with the fill color of that theme's `--primary-text-color`).

Shubh8899: I am little confused, what exactly do i need to do?

jasonhenriquez: Essentially, we need to have the appropriate theme color logo for each theme instead of mapping to the white, black, and default logos. These are the three ways that that can be achieved, and any of them should work. I apologize that this is a bit harder of an issue than was originally thought.

Shubh8899: okay got it, clear now. I would like to continue working on this. I will update this MR in couple of days with the listed solution.

efb4f5ff-1298-471a-8973-3d47447115dc: Hi @Shubh8899 just checking in, are u running in any issues with this?

PR #4058, opened 2023-09-18
Fix buttons intersecting on Subscribed Channels tab

jasonhenriquez:
# Fix buttons intersecting on Subscribed Channels tab

## Pull Request Type

- [x] Bugfix

## Related issue
None

## Description
Fixes buttons on the Subscribed Channels tab intersecting in many languages. In all languages where this issue occurs, there are multiple words in the Unsubscribe button label text, which causes the buttons to intersect. This change has the words wrap in such cases and updates the default behavior for all `ft-subscribe-button`s for future-proofing. As a result, this wrapping behavior goes into effect on other instances of `ft-subscribe-button` when the button would otherwise overflow to outside of the container (e.g., the subscribe button on the Channel page on very small viewport widths), but this is desired behavior in those cases.

## Screenshots 

Before:
Screenshot_20230918_182239

After:
Screenshot_20230918_182221

## Testing 
1. Change locale to a language with multiple words in the "Unsubsribe" button label text (e.g., `Svenska` / Swedish)
2. Navigate to Channels tab
3. See normal behavior

## Desktop
- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

### DIFF
--- a/src/renderer/components/ft-subscribe-button/ft-subscribe-button.css
+++ b/src/renderer/components/ft-subscribe-button/ft-subscribe-button.css
@@ -1,6 +1,7 @@
-.subscribeButton {
+.subscribeButton.btn {
   align-self: center;
   block-size: 50%;
   margin-block-end: 10px;
   min-inline-size: 150px;
+  white-space: initial;
 }


src/renderer/components/ft-subscribe-button/ft-subscribe-button.css
```diff
@@ -1,6 +1,7 @@
-.subscribeButton {
+.subscribeButton.btn {
```
PikachuEXE, line 1:
```suggestion
/* Ensures style here overrides style of .btn */
.subscribeButton.btn {
```

PikachuEXE (changes requested):
Tested with changing the text manually
image

Suggested a comment for the selector

src/renderer/components/ft-subscribe-button/ft-subscribe-button.css
jasonhenriquez: Even though this is a pretty obvious/borderline case, I can see exactly why you're suggesting this, and I need to be more conscientious of this. Adding it in.

absidue (approved):

src/renderer/components/ft-subscribe-button/ft-subscribe-button.css
PikachuEXE: Project specific stuff (which cannot be searched via doc) should be explained when possible
(like the `.btn` style got higher priority than `.subscribeButton` here)

We spend more time reading code than writing code :)

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #3893, opened 2023-08-17
Fix locales getting unnecessarily reprocessed for incremental builds

absidue:
# Fix locales getting unnecessarily reprocessed for incremental builds

## Pull Request Type

- [x] Bugfix

## Description
Currently when you are in dev mode and make a change to the file, it is just meant to recompile changed files. As FreeTube caches languages files in memory once they have been loaded, it requires a restart of the app to pick up changes, so we don't watch for changes to those files in dev mode.

Unfortunately I misunderstood how webpack's plugin system works and didn't realise it calls our plugin for all builds, even the incremental ones. Currently the plugin reads the language files from the file system once, but does the processing (removing the language name property, removing empty strings and objects and converting it to a JSON string) for every single build.

This pull request changes the behaviour in dev mode (release builds still work the same) to cache the results of the language file processing the first time it's done, when an incremental build is triggered it now gives webpack the cached results instead of processing them again. This should result in the incremental builds being slightly faster, as they now have to do less work.

## Screenshots 

<details><summary>Incremental build output before:</summary>

```
Compiled renderer script!

Watching file changes for renderer script...
assets by status 478 KiB [cached] 28 assets
assets by path static/locales/*.json 1.75 MiB
  asset static/locales/el.json 65.9 KiB [emitted]
  asset static/locales/ru.json 59.9 KiB [emitted]
  asset static/locales/bg.json 59.6 KiB [emitted]
  asset static/locales/uk.json 55.4 KiB [emitted]
  asset static/locales/ar.json 53.8 KiB [emitted]
  asset static/locales/he.json 49.7 KiB [emitted]
  asset static/locales/ja.json 47.5 KiB [emitted]
  + 37 assets
assets by path *.js 20.8 MiB
  asset renderer.js 20.8 MiB [emitted] (name: renderer)
  asset renderer.1c5334bb4e2c056e53c4.hot-update.js 47 KiB [emitted] [immutable] [hmr] (name: renderer)
asset index.html 523 bytes [emitted]
asset renderer.1c5334bb4e2c056e53c4.hot-update.json 32 bytes [emitted] [immutable] [hmr]
Entrypoint renderer 21.1 MiB (213 KiB) = renderer.css 266 KiB renderer.js 20.8 MiB renderer.1c5334bb4e2c056e53c4.hot-update.js 47 KiB 27 auxiliary assets
cached modules 6.83 MiB (javascript) 209 KiB (css/mini-extract) 13.2 KiB (asset) [cached] 1395 modules
[...]

PR #4066, opened 2023-09-24
Ensure updated locales are auto-reloaded in electron renderer dev mode

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

Ensure updated locales are auto-reloaded in electron renderer dev mode

Webpack does not watch locale files to rebuild
So added [`webpack-watch-external-files-plugin`](https://www.npmjs.com/package/webpack-watch-external-files-plugin)
Also updated internal locale processing plugin to only read changed files (mtime comparison)

## Screenshots 

Limited emit for locale JSON after en-US.yaml changed
image

No emit for locale JSON after renderer file changed
image

### Basic
- `yarn && yarn dev`
- Update translation entry value (e.g. `You have disabled automatic subscription fetching`)
- Confirm value update displayed on screen

### Fix locales getting unnecessarily reprocessed for incremental builds
See #3893 

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Only affects renderer not web one (different config file)


absidue:
Didn't mean to click that.

I haven't tested this yet, but I don't think this change will do what you think it will, as FreeTube itself caches them too, it only loads each locale once. So you'll need changes inside FreeTube along with the webpack ones.

PikachuEXE:
Watch external file plugin now only added in dev mode

But I have tested the behaviour with multiple locales and it's working for me
Please test first :)

efb4f5ff-1298-471a-8973-3d47447115dc:
Note to self: test after @absidue review is in

absidue:
Okay, so it works and as I suspected, it does so by reloading the entire application, basically the same that would happen when you open a new window, so very expensive but at least it works.

Just remember not to change any strings while you are on the subscriptions page and have automatic subscription fetching enabled, otherwise you'll have fun with your screen filled with falling back to RSS toasts after a few edits.

Maybe in the future we can look into a more efficient solution but for now this works.

_scripts/webpack.renderer.config.js
```diff
@@ -146,4 +147,14 @@ const config = {
   target: 'electron-renderer',
 }
 
+if (isDevMode) {
+  config.plugins.push(
+    new WatchExternalFilesPlugin({
+      files: [
+        './static/locales/*.yaml',
```
absidue, line 154: As only 45 out of the 65 locale files are actually enabled, it might make more sense to read the list from the activeLocales.json file, instead of having it watch all of them. (you'll need to read the activeLocales.json file too, but here is how you could tell it to only watch the necessary files.
```suggestion
        `./static/locales/{${activeLocales.join(',')}}.yaml`,
```

_scripts/ProcessLocalesPlugin.js
```diff
@@ -37,66 +38,88 @@ class ProcessLocalesPlugin {
...
+              const cacheEntry = this.cache[locale]
+              if (cacheEntry != null) {
+                const { filename, source, mtimeMs: cachedMtimeMs } = cacheEntry
+                if (cachedMtimeMs === mtimeMs) {
+                  compilation.emitAsset(filename, source, { minimized: true })
+                  resolve()
+                  return
+                }
               }
+            }
```
absidue, line 60: Please add some whitespace to make this easier to read.

```suggestion
        Object.values(this.locales).forEach((localeEntry) => {
          const { locale, data, mtimeMs } = localeEntry

          promises.push(new Promise(async (resolve) => {
            if (IS_DEV_SERVER) {
              const cacheEntry = this.cache[locale]

              if (cacheEntry != null) {
                const { filename, source, mtimeMs: cachedMtimeMs } = cacheEntry

                if (cachedMtimeMs === mtimeMs) {
                  compilation.emitAsset(filename, source, { minimized: true })
                  resolve()
                  return
                }
              }
            }
```

_scripts/ProcessLocalesPlugin.js
```diff
@@ -37,66 +38,88 @@ class ProcessLocalesPlugin {
...
-          }
-        } else {
-          const promises = []
-
-          for (const { locale, data } of this.locales) {
-            promises.push(new Promise(async (resolve) => {
-              if (Object.prototype.hasOwnProperty.call(data, 'Locale Name')) {
-                delete data['Locale Name']
+        const promises = []
+        this.loadLocales(true)
```
absidue, line 45: Currently what this does is calls loadLocales once when the plugin is created in the constructor (can't be removed as we need to be able to pass the locale names to the DefinePlugin) and then straight away again when this hook is called, when webpack goes to compile the files, so please add an `if` around this, so it is only called when the hook is called the second, third, fourth, etc times, not on the first time. (that will also prevent loadLocales getting called twice for release builds).

Can be a simple if else e.g.

```js
// in constructor
this.isIncrementalBuild = false

// here
if (this.isIncrementalBuild) {
  this.loadLocales(true)
} else {
  this.isIncrementalBuild = true)
}

File: _scripts/ProcessLocalesPlugin.js [hunk 4]
```diff
@@ -37,66 +38,88 @@ class ProcessLocalesPlugin {
...
+        }
+      }
+      const contents = readFileSync(filePath, 'utf-8')
       const data = loadYaml(contents)
+      this.locales[locale] = { locale, data, mtimeMs: mtimeMsFromStats }
 
-      this.localeNames.push(data['Locale Name'] ?? locale)
-      this.locales.push({ locale, data })
+      const localeName = data['Locale Name'] ?? locale
+      if (!this.localeNames.includes(localeName)) { this.localeNames.push(localeName) }
```
absidue, line 122: `this.localeNames` is read once when webpack loads the `renderer.config.js` file and passed to the `DefinePlugin`, so you can change this to:
```suggestion
      if (!loadModifiedFilesOnly) {
        this.localeNames.push(localeName)
      }
```

absidue (changes requested):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4129, opened 2023-10-10
Fix playlists not showing up on artist topic channels with the local API

absidue:
# Fix playlists not showing up on artist topic channels with the local API

## Pull Request Type

- [x] Bugfix

## Description
Currently when you visit an artist topic channel e.g. `Rihanna - Topic` on the local API, the playlists tab will be empty. This is caused by those channels having the `Created Playlists` category despite not actually having any playlists that they own, instead we need to use the `Albums & Singles` category, which isn't empty and contains Album playlists, which aren't tied directly to any channel (playlist ID starts with `OL`).

## Screenshots 
before:
before

after:
after

## Testing 
https://www.youtube.com/channel/UCvWtix2TtWGe9kffqnwdaMw

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/views/Channel/Channel.js
+++ b/src/renderer/views/Channel/Channel.js
@@ -1191,13 +1191,20 @@ export default defineComponent({
         // for the moment we just want the "Created Playlists" category that has all playlists in it
 
         if (playlistsTab.content_type_filters.length > 1) {
+          let viewId = '1'
+
+          // Artist topic channels don't have any created playlists, so we went to select the "Albums & Singles" category instead
+          if (this.channelName.endsWith('- Topic') && channel.metadata.music_artist_name) {
+            viewId = '50'
+          }
+
           /**
            * @type {import('youtubei.js').YTNodes.ChannelSubMenu}
            */
           const menu = playlistsTab.current_tab.content.sub_menu
           const createdPlaylistsFilter = menu.content_type_sub_menu_items.find(contentType => {
             const url = `https://youtube.com/${contentType.endpoint.metadata.url}`
-            return new URL(url).searchParams.get('view') === '1'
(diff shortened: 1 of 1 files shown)


absidue:
Do we use YouTube's labels for the categories (they'll always be English US) or do we hardcode specific ones and only display those?

ChunkyProgrammer:
> Do we use YouTube's labels for the categories (they'll always be English US) or do we hardcode specific ones and only display those?

I think a mix could be good (use hard coded specific ones when possible so we can have them translated, use youtube ones if they cant be mapped)

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Lets just go for the route of adding the things @ChunkyProgrammer suggested. After implementing it update testing description if needed

absidue:
🙈
Guess I'll have to go down that rabbit hole of YouTube edge cases then and figure out when I have to hide the sorting options and get the UI working properly with two drop downs. So that it doesn't turn into too much of a shit show, I'll just hard code allowed options, that way if it encounters unknown ones it can ignore them, feels safer than trusting YouTube, especially after the recent published date fiasco.

(was really hoping you peeps would decided that this bug fix can go ahead as is, instead of turning it into a feature implementation that will introduce new bugs 🙈, cause that's what always happens when I implement a feature.)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
On second thought lets leave that for a future PR. Sounds way to complicated to add that here

ChunkyProgrammer (approved):

PR #4200, opened 2023-10-21
Make landing page configurable

### LINKED ISSUE #183: Allow user to choose which page view to start on
When you start FreeTube <=0.40b it will automatically open the `subscription` page.   It would be good to have a settings option allowing the choice of which page to open on - Subscription, Trending, Most Popular, Favorites, History,....

jasonhenriquez:
# Make landing page configurable

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #183 

## Description
Implements setting to customize your landing page on startup. Only includes routes without params. Also slightly reorders routes to closer match the more logical order of how they appear in the application and adds `name` properties to a handful (not used here, but nice to have handy).

## Screenshots 
Screenshot_20231021_011521

## Testing 
- Go to General Settings and change the default landing page. Confirm that it is populated correctly with a default value.
- Change to one of the values. Restart application -- not just refresh (i.e., NOT CTRL+R) -- and ensure page is used as starting page. Repeat for all values.

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -111,6 +111,10 @@ export default defineComponent({
       return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light'
     },
 
+    landingPage: function() {
+      return '/' + this.$store.getters.getLandingPage
+    },
+
     externalLinkOpeningPromptNames: function () {
       return [
         this.$t('Yes'),
@@ -136,7 +140,7 @@ export default defineComponent({
     $route () {
       // react to route changes...
       // Hide top nav filter panel on page change
-      this.$refs.topNav.hideFilters()
+      this.$refs.topNav?.hideFilters()
     }
   },
   created () {
@@ -178,7 +182,13 @@ export default defineComponent({
       })
 
       this.$router.afterEach((to, from) => {
-        this.$refs.topNav.navigateHistory()
+        this.$refs.topNav?.navigateHistory()
+      })
+
+      this.$router.onReady(() => {
+        if (this.$router.currentRoute.path !== this.landingPage && this.landingPage !== '/subscriptions') {
+          this.$router.push({ path: this.landingPage })
+        }
       })
     })
   },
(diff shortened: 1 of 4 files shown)


absidue:
I think it's probably better to have a hard coded list, that way it won't include stuff that people are unlikely to ever select like About, Create New Profile and the Profile Settings page.

jasonhenriquez:
> I think it's probably better to have a hard coded list, that way it won't include stuff that people are unlikely to ever select like About, Create New Profile and the Profile Settings page.

I can get that reasoning. My reasoning was "eh, if someone wants it, who am I to stand in the way." That and the elegance of reusing the `meta.title`s. Certainly have no problem changing this part if others agree.

PikachuEXE:
Agree on whitelist

efb4f5ff-1298-471a-8973-3d47447115dc:
If u click the FT logo in the top bar it takes u to the subscriptions page (current main landing page). IMO that should be based on the user configuration.

Just a unrelated note: Maybe accessibility issue? when i hover over the FT logo it doesn't show a hover effect like with the other icons in the top bar

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
- Most Popular should only be in there if IV API is set as primary or fallback is enabled
- If distraction free setting Hide Trending/Popular/Playlist is enabled -> hide them from dropdown

jasonhenriquez:
> * Most Popular should only be in there if IV API is set as primary or fallback is enabled
> 
>     * If distraction free setting Hide Trending/Popular/Playlist is enabled -> hide them from dropdown

Ah, you're right, although it will be a bit more than that, as those can be changed in any order. Looking at it now.

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #3086, opened 2023-01-16
[Feature Request]: Option for filtering out certain videos from creators on the main subscription page


Sandyran:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Occasionally, creators, that I do not want to unsubscribe to, create videos that are of no interest to me. Because of this, they tend to clutter up the main "subscribtion" page in FreeTube.

### Proposed Solution

This function would hide videos, with certain keywords in the title, from the main subscription page in FreeTube. Some example of keywords could be: "#shorts", "XXX Podcast", "#live", etc.

Imagine, if there was four videos on the main page. Two of these have "#shorts" in this title, while the other two have very generic video titles.

![FreeTube](https://user-images.githubusercontent.com/31732894/212643764-37c8d3c5-59db-4401-9ecf-1f3a94864083.PNG)

This feature would hide the YouTube shorts videos from the main FreeTube page.

![FreeTube without shorts](https://user-images.githubusercontent.com/31732894/212643782-975b7fab-dd79-45b1-8a01-67f57f598a94.PNG)

### Alternatives Considered

An alternative solution would be to expand the "distraction free" setting, with settings such as "hide shorts" or other kinds of videos, etc.

### Issue Labels

new feature

Flashwalker:
Option for filtering shorts is a good idea. There are too many of them

RageGamerBoi:
:+1:  Yes please!
I believe there are better ways to detect shorts however, for example, videos with an aspect ratio of square or more vertical and less than 60 seconds in length are YouTube's current criteria.

Issue #1018, opened 2021-02-05
Export all data


codewizkid:
Hello again again,

Sorry to keep spaming with ideas, but this app could be game changing for me, so I want to help as much as possible. Anyway, this idea is adding a export all data option that will let you export you subscriptions, playlists, history etc. to one file. Then you can just take that file and import it in any other FreeTube app and you'll never lose all you data!

emmeemmeesse:
I was basically looking for this option I believe.
The possibility to import and export my account(s) settings in case I, for example, change laptop or have to reset it or whatever.

I believe there isn't such an option so far since this thread is stillopen?

brainstorm:
All of the above plus having it importable/exportable via URL, so you can point to it directly on the app as a URL (not necessarily local filesystem-based only).

jorp:
This would be useful for moving 'Distraction Free Settings' between devices. The list of hidden channels/keywords could get quite long, and as of now needs to be done manually.

kuligs2:
Been waiting for years, still waiting for years to come for this option.

Optekah:
It's awesome being able to import/export playlists and subscriptions, but being able to do it with settings would be the chef's kiss.

PR #4202, opened 2023-10-21
Video title filter / blacklist

### LINKED ISSUE #1070: Video Title Filter / Blacklist

jasonhenriquez:
# Video title filter / blacklist

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #1070

## Description
Enter a fragment, word, or phrase (case insensitive) to hide all videos with titles containing it in subscriptions, trending, most popular, recommended, and searches. Special exclusion text appears for community posts (only from the Subscriptions -> Community tab; nothing is hidden on the Channel route). Incidentally: to set a minimum character length of 3 for blacklisted inputs, this PR implements a `minInputLength` prop in `ft-input` and `ft-input-tags`. 

## Screenshots 

Screenshot_20231021_044819
Screenshot_20231021_045130
Screenshot_20231021_045248

## Testing 
- Ensure videos whose titles contain blacklisted tags are visible in the Channels tabs, the History tab, the Playlists tab, Watch Video Playlist Boxes, and Playlists
- Ensure videos whose titles contain blacklisted tags are not visible in trending, subscriptions, most popular, recommended, and searches
- Ensure tags are case-insensitive and can be substrings of words
- Try inputting tags of length < 3

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

## Additional context
- I enabled this filtering on the specific pages mentioned by the author of the issue. Please speak up on if you feel this is too little or too much in terms of its expansiveness across different routes. 
- This also specifically targets videos as per the original issue author's request.

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -52,6 +52,8 @@ Global:
     Subscriber Count: 1 subscriber | {count} subscribers
     View Count: 1 view | {count} views
     Watching Count: 1 watching | {count} watching
+  Input Tags:
+    Length Requirement: Tag must be at least {number} characters long
 
 # Search Bar
 Search / Go to URL: Search / Go to URL
@@ -371,6 +373,8 @@ Settings:
     Hide Channel Shorts: Hide Channel Shorts
     Hide Channel Podcasts: Hide Channel Podcasts
     Hide Channel Releases: Hide Channel Releases
+    Hide Videos Containing Text: Hide Videos Containing Text
+    Hide Videos Containing Text Placeholder: Fragment, Word, or Phrase
     Hide Subscriptions Videos: Hide Subscriptions Videos
     Hide Subscriptions Shorts: Hide Subscriptions Shorts
     Hide Subscriptions Live: Hide Subscriptions Live
@@ -605,6 +609,7 @@ Channel:
     votes: '{votes} votes'
     Reveal Answers: Reveal Answers
     Hide Answers: Hide Answers
+    Video hidden by FreeTube: Video hidden by FreeTube
 Video:
   Mark As Watched: Mark As Watched
   Remove From History: Remove From History
@@ -868,6 +873,7 @@ Tooltips:
     Hide Channels: Enter a channel name or channel ID to hide all videos, playlists and the channel itself from appearing in search, trending, most popular and recommended.
        The channel name entered must be a complete match and is case sensitive.
     Hide Subscriptions Live: 'This setting is overridden by the app-wide "{appWideSetting}" setting, in the "{subsection}" section of the "{settingsSection}"'
+    Hide Videos Containing Text: Enter a fragment, word, or phrase (case insensitive) to hide all videos with titles containing it in subscriptions, trending, most popular, recommended, and searches.
   Subscription Settings:
     Fetch Feeds from RSS: When enabled, FreeTube will use RSS instead of its default
       method for grabbing your subscription feed. RSS is faster and prevents IP blocking,
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -374,7 +374,7 @@ Settings:
     Hide Channel Podcasts: Hide Channel Podcasts
     Hide Channel Releases: Hide Channel Releases
     Hide Videos Containing Text: Hide Videos Containing Text
-    Hide Videos Containing Text Placeholder: Fragment, Word, or Phrase
+    Hide Videos Containing Text Placeholder: Word, Word Fragment, or Phrase
     Hide Subscriptions Videos: Hide Subscriptions Videos
     Hide Subscriptions Shorts: Hide Subscriptions Shorts
     Hide Subscriptions Live: Hide Subscriptions Live
@@ -873,7 +873,7 @@ Tooltips:
     Hide Channels: Enter a channel name or channel ID to hide all videos, playlists and the channel itself from appearing in search, trending, most popular and recommended.
        The channel name entered must be a complete match and is case sensitive.
     Hide Subscriptions Live: 'This setting is overridden by the app-wide "{appWideSetting}" setting, in the "{subsection}" section of the "{settingsSection}"'
-    Hide Videos Containing Text: Enter a fragment, word, or phrase (case insensitive) to hide all videos with titles containing it in subscriptions, trending, most popular, recommended, and searches.
+    Hide Videos Containing Text: Enter a word, word fragment, or phrase (case insensitive) to hide all videos with titles containing it in subscriptions, trending, most popular, recommended, and searches.
   Subscription Settings:
     Fetch Feeds from RSS: When enabled, FreeTube will use RSS instead of its default
       method for grabbing your subscription feed. RSS is faster and prevents IP blocking,
--- a/src/renderer/components/distraction-settings/distraction-settings.js
+++ b/src/renderer/components/distraction-settings/distraction-settings.js
@@ -107,13 +107,16 @@ export default defineComponent({
     channelsHidden: function () {
       return JSON.parse(this.$store.getters.getChannelsHidden)
     },
+    forbiddenVideoTitleText: function() {
+      return JSON.parse(this.$store.getters.getForbiddenVideoTitleText)
+    },
     hideSubscriptionsLiveTooltip: function () {
       return this.$t('Tooltips.Distraction Free Settings.Hide Subscriptions Live', {
         appWideSetting: this.$t('Settings.Distraction Free Settings.Hide Live Streams'),
         subsection: this.$t('Settings.Distraction Free Settings.Sections.General'),
         settingsSection: this.$t('Settings.Distraction Free Settings.Distraction Free Settings')
       })
-    }
+    },
   },
   methods: {
     handleHideRecommendedVideos: function (value) {
@@ -125,6 +128,9 @@ export default defineComponent({
     },
     handleChannelsHidden: function (value) {
       this.updateChannelsHidden(JSON.stringify(value))
+    },
+    handleForbiddenVideoTitleText: function (value) {
+      this.updateForbiddenVideoTitleText(JSON.stringify(value))
     },
 
     ...mapActions([
@@ -148,6 +154,7 @@ export default defineComponent({
       'updateHideSharingActions',
       'updateHideChapters',
       'updateChannelsHidden',
+      'updateForbiddenVideoTitleText',
       'updateShowDistractionFreeTitles',
       'updateHideFeaturedChannels',
       'updateHideChannelShorts',
--- a/src/renderer/components/distraction-settings/distraction-settings.vue
+++ b/src/renderer/components/distraction-settings/distraction-settings.vue
@@ -245,6 +245,17 @@
         @change="handleChannelsHidden"
       />
     </ft-flex-box>
+    <ft-flex-box>
+      <ft-input-tags
+        :label="$t('Settings.Distraction Free Settings.Hide Videos Containing Text')"
+        :placeholder="$t('Settings.Distraction Free Settings.Hide Videos Containing Text Placeholder')"
+        :show-action-button="true"
+        :tag-list="forbiddenVideoTitleText"
+        :min-tag-length="3"
+        :tooltip="$t('Tooltips.Distraction Free Settings.Hide Videos Containing Text')"
+        @change="handleForbiddenVideoTitleText"
+      />
+    </ft-flex-box>
   </ft-settings-section>
 </template>
 
--- a/src/renderer/components/distraction-settings/distraction-settings.vue
+++ b/src/renderer/components/distraction-settings/distraction-settings.vue
@@ -251,7 +251,7 @@
         :placeholder="$t('Settings.Distraction Free Settings.Hide Videos Containing Text Placeholder')"
         :show-action-button="true"
         :tag-list="forbiddenVideoTitleText"
-        :min-tag-length="3"
+        :min-input-length="3"
         :tooltip="$t('Tooltips.Distraction Free Settings.Hide Videos Containing Text')"
         @change="handleForbiddenVideoTitleText"
       />
--- a/src/renderer/components/ft-community-post/ft-community-post.js
+++ b/src/renderer/components/ft-community-post/ft-community-post.js
@@ -27,6 +27,10 @@ export default defineComponent({
     appearance: {
       type: String,
       required: true
+    },
+    hideVideosWithForbiddenTextInTitle: {
+      type: Boolean,
+      default: true
     }
   },
   data: function () {
@@ -57,6 +61,15 @@ export default defineComponent({
 
     listType: function () {
       return this.$store.getters.getListType
+    },
+
+    forbiddenVideoTitleText() {
+      if (!this.hideVideosWithForbiddenTextInTitle) { return [] }
+      return JSON.parse(this.$store.getters.getForbiddenVideoTitleText)
+    },
+
+    hideVideo() {
+      return this.forbiddenVideoTitleText.some((text) => this.data.postContent.content.title?.toLowerCase().includes(text.toLowerCase()))
     }
   },
   created: function () {
--- a/src/renderer/components/ft-community-post/ft-community-post.scss
+++ b/src/renderer/components/ft-community-post/ft-community-post.scss
@@ -13,6 +13,12 @@
   box-sizing: border-box;
 }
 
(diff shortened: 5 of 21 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
A bunch of issues regarding shorts/live videos were closed as a duplicate of #1070 

So we should also take that into account somehow

- A way to block a shorts on all pages
- A way to block a live streams on all pages
- A way to block upcoming live streams on all pages

U could also just remove Closes behind issue number

jasonhenriquez:
Whoops, I'll remove the closing. Although to be fair, those other aspects aren't really present in the title or original issue.

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc With the current information it's impossible to hide shorts on all pages, the only way we know something is a short, is if it's on a shorts specific page.

efb4f5ff-1298-471a-8973-3d47447115dc:
Would it be possible to hide live, premiere and upcoming streams

absidue:
We can probably hide currently live stuff and upcoming stuff, we won't be able to differentiate between live stream or premiere and we also can't do any of those for RSS.

jasonhenriquez:
> Would it be possible to hide live, premiere and upcoming streams
From what I can discern, given that they have a `title` property on their data, I think they are being hidden appropriately if they contain a banned title, which I believe is what is being asked. Do you know of any videos I can use to test / how I can find videos matching these conditions easily?

absidue:
I think having this pull request just do the title filtering is fine.

jasonhenriquez:
Yeah, that sounds good to me. If #1070 is serving a secondary purpose of improving the expansiveness of the existing `Hide Upcoming Premieres` & `Hide Live Streams` features, it certainly doesn't seem clear to me why that is the case, but it's not in scope for this PR IMO.

efb4f5ff-1298-471a-8973-3d47447115dc:
I agree it doesnt have to be added in this PR 

but

Should we even implement that if we cant differentiate them when RSS is enabled?

Feels like a half baked feature

jasonhenriquez:
Seems like a separate issue should be created for it. If we have sub-issues that are not referenced in main issue titles or descriptions, it's hard for people to search for those & inevitably leads to dupes.

efb4f5ff-1298-471a-8973-3d47447115dc:
Doesnt this also close https://github.com/FreeTubeApp/FreeTube/issues/3086 ?

efb4f5ff-1298-471a-8973-3d47447115dc:
Why we have so many dupes

Closes https://github.com/FreeTubeApp/FreeTube/issues/95

gelatinbomb:
> Why we have so many dupes
> 
> Closes #95

It seems a few users might want this feature. I mean I was just about to open an issue asking for this exact thing until I found this PR so it's nice to see someone implemented the thing :D @jasonhenriquez Nice job!

There is only one thing I would like to ask: It is possible to add a way to export the "blocklist" as a text file? You know as a backup file.  That would be nice in the case you had a lot of blocked words.

jasonhenriquez:
@gelatinbomb That will be possible once #1018 is implemented - a user-friendly means of sharing `settings.db` files. Until then, you can make a backup of [the `settings.db` file on your device](https://docs.freetubeapp.io/usage/data-location/#notable-files) and grab it from there.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
- Copy original video title from subscriptions -> Put it in the blacklist -> check subscription page -> video gone -> Enable DeArrow -> video still gone even though DeArrow title is different
- Enable DeArrow -> Go to subscription page -> copy a word/title of a vid that has a title change -> insert into blacklist -> go to subscription page -> See video not gone
- Should we even do `word fragments`? What if i really hate `ores` for some reason so i insert `ore` but i really love `cores`, now i dont get to see all the beautiful vids about `cores` :(
- Why not hide playlist titles and video titles within a playlist?

jasonhenriquez:
> * Copy original video title from subscriptions -> Put it in the blacklist -> check subscription page -> video gone -> Enable DeArrow -> video still gone even though DeArrow title is different
> 
>     * Enable DeArrow -> Go to subscription page -> copy a word/title of a vid that has a title change -> insert into blacklist -> go to subscription page -> See video not gone

Implementation-wise, the irksome part of that is that the DeArrow logic is in the `ft-list-video`, but our video hiding logic bits have been in the video list container classes a level or two above that, which I believe is done that way for performance & logical reasons, correct me if I'm wrong (if you can't show the child, you can't show the parent wrapper, so that introduces problems that can't be solved easily by performance).

Behavior-wise, the main question is what would we want the expected behavior to be in both of these cases. Would you really want to see a video if either the DeArrow title or the original contained a banned word? I would think not, so that would imply that the most preferred behavior would be to filter if either matched. But then again, that does put a lot of power in DeArrow contributors, so maybe we shouldn't care about it to that extent. So I'm currently leaning on just ignoring the DeArrow title consistently, which is what the code is currently doing. Let me know what you think.

>     * Should we even do `word fragments`? What if i really hate `ores` for some reason so i insert `ore` but i really love `cores`, now i dont get to see all the beautiful vids about `cores` :(

I do think this is a worthwhile hypothetical. It's a trail that often ends in regex, which I think is probably a bit of overkill for the initial feature & isn't always good for usability (How do you indicate that this is a regex tag versus just a regular tag? If all tags are regex tags, will non - power users like it if they can't use special characters like they want?  How do you handle errors? How does it affect performance?). So I think we should let it simmer and see how it goes for people.

On if you want to be able to differentiate it without regex, it becomes a question of "is the banning of partial words banning too much stuff for me" versus "is the banning of only full words banning too little". It's a good question to ask. I think the second one is closer to where people are, but I can't really know that for sure. Is it okay if my answer to this is "let's see how many people complain" and leave it there for now?

>     * Why not hide playlist titles and video titles within a playlist?

Good question! I was going off of what we were doing in our components, as discussed in the Matrix chat. I figure that the same push to add it to playlists will also be adding the channel logic as well. Unsure of the considerations there, and this PR is a little big, so I left it be. But I can add this and/or channel filtering logic to there as well if desired.

Alexander-Wilms:
Regex support would be great. BlockTube supports it like [this](https://github.com/amitbl/blocktube/wiki/Filters-Options#filters-syntax):

```
// all-caps
/[A-Z ]15+/
// no caps
/^[a-z ]{30}/
// indian characters
/[\u0900-\u097F]+/
// french characters
/[àâçéèêëîïôûùüÿñæœ]+/
```

It's very useful if YouTube decides to only recommend foreign-language videos for some unknown reason and to block videos with block all-caps clickbait titles.

efb4f5ff-1298-471a-8973-3d47447115dc:
> > * Copy original video title from subscriptions -> Put it in the blacklist -> check subscription page -> video gone -> Enable DeArrow -> video still gone even though DeArrow title is different
> >   
> >   * Enable DeArrow -> Go to subscription page -> copy a word/title of a vid that has a title change -> insert into blacklist -> go to subscription page -> See video not gone
> 
> Implementation-wise, the irksome part of that is that the DeArrow logic is in the `ft-list-video`, but our video hiding logic bits have been in the video list container classes a level or two above that, which I believe is done that way for performance & logical reasons, correct me if I'm wrong (if you can't show the child, you can't show the parent wrapper, so that introduces problems that can't be solved easily by performance).
> 
> Behavior-wise, the main question is what would we want the expected behavior to be in both of these cases. Would you really want to see a video if either the DeArrow title or the original contained a banned word? I would think not, so that would imply that the most preferred behavior would be to filter if either matched. But then again, that does put a lot of power in DeArrow contributors, so maybe we shouldn't care about it to that extent. So I'm currently leaning on just ignoring the DeArrow title consistently, which is what the code is currently doing. Let me know what you think.

Hmm i think when someone explicitly toggles DeArrow titles to show up they dont want to see click bait titles and want a title that tells them what the contents of the video will look like. If that title contains a word that is in the blacklist i think it should hide it. I also think it shouldnt hide a video with a replaced DeArrow title if the original title contains the blacklisted word

> 
> > ```
> > * Should we even do `word fragments`? What if i really hate `ores` for some reason so i insert `ore` but i really love `cores`, now i dont get to see all the beautiful vids about `cores` :(
> > ```
> 
> I do think this is a worthwhile hypothetical. It's a trail that often ends in regex, which I think is probably a bit of overkill for the initial feature & isn't always good for usability (How do you indicate that this is a regex tag versus just a regular tag? If all tags are regex tags, will non - power users like it if they can't use special characters like they want? How do you handle errors? How does it affect performance?). So I think we should let it simmer and see how it goes for people.
> 
> On if you want to be able to differentiate it without regex, it becomes a question of "is the banning of partial words banning too much stuff for me" versus "is the banning of only full words banning too little". It's a good question to ask. I think the second one is closer to where people are, but I can't really know that for sure. Is it okay if my answer to this is "let's see how many people complain" and leave it there for now?

I agree lets let this simmer and come back to it when people open issues or start asking questions.

> > ```
> > * Why not hide playlist titles and video titles within a playlist?
> > ```
> 
> Good question! I was going off of what we were doing in our components, as discussed in the Matrix chat. I figure that the same push to add it to playlists will also be adding the channel logic as well. Unsure of the considerations there, and this PR is a little big, so I left it be. But I can add this and/or channel filtering logic to there as well if desired.

Will test soon!

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
`2023` is in my blocklist -> i click on the big playlist thumbnail -> video plays that contains blacklisted word

`2023` is in blocklist -> cant see any in the playlist that contains `2023` -> click on random vid -> see that some videos that contain `2023` are still visible in the playlist queue view

https://github.com/FreeTubeApp/FreeTube/assets/73130443/cfcd9ab4-4937-4769-9fb5-2826f4ba86a1




https://github.com/FreeTubeApp/FreeTube/assets/73130443/2503faae-4ea2-459e-9aeb-df0d624494cb

jasonhenriquez:
I'll look into this. Out of curiosity, how is this problem handled with blocked channels?

jasonhenriquez:
Pre-existing issue carried over: when autoplay is enabled and a blocked channel's video (using existing channel block feature) is next up in the pre-filtered recommended videos, that video is loaded up next. Same issue occurs now with text-hidden channels and hidden title text. Almost certain that the same occurs for playlist autoplay as well.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Sometimes i get this error when i navigate to the WAN Show playlist

Capture

jasonhenriquez:
Thank for pointing that out, @efb4f5ff-1298-471a-8973-3d47447115dc - this feature is actually no longer supposed to work on History and videos inside of playlists, and I just had to fix the logic for not using this feature on certain routes. We can minimize that second part from inside playlists to just user playlists, but I'm thinking we may want to do that as a separate PR / effort where we enable both hidden channel & text preferences for playlists.

static/locales/en-US.yaml
```diff
@@ -880,7 +880,7 @@ Tooltips:
     Hide Channels: Enter a channel ID to hide all videos, playlists and the channel itself from appearing in search, trending, most popular and recommended.
        The channel ID entered must be a complete match and is case sensitive.
     Hide Subscriptions Live: 'This setting is overridden by the app-wide "{appWideSetting}" setting, in the "{subsection}" section of the "{settingsSection}"'
-    Hide Videos and Playlists Containing Text: Enter a word, word fragment, or phrase (case insensitive) to hide all videos & playlists with titles containing it throughout all of FreeTube.
+    Hide Videos and Playlists Containing Text: Enter a word, word fragment, or phrase (case insensitive) to hide all videos & playlists with titles containing it throughout all of FreeTube, excluding only History and videos inside of playlists.
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 883:
```suggestion
Hide Videos and Playlists Containing Text: Enter a word, word fragment, or phrase (case insensitive) to hide all videos & playlists with titles containing it throughout all of FreeTube, excluding only your Playlists, History and videos inside of playlists.
```

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Maybe this can be worded better because the wording tricked me into believing it was also blocking vids in my playlists even though i knew it did not do that

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Video is hidden in community post but creator was nice enough to include the url in the description of the post. Should we maybe hide the whole post to prevent someone from accessing the content they blacklisted?

Capture

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc: @jasonhenriquez

jasonhenriquez: Went with this but just moved the order: `Enter a word, word fragment, or phrase (case insensitive) to hide all videos & playlists with titles containing it throughout all of FreeTube, excluding only History, Your Playlists, and videos inside of playlists.`

jasonhenriquez:
> Hmm thought of an usecase that we maybe can consider in the decision of hiding it completely or not:
> 
> Having a child so the parents puts something in the blacklist -> password protect the settings because dont want them to change the blacklist -> child goes to community post but sees that vid is hidden but in the description there is a url to the video
> 
> The child could also have ez access to a browser to retrieve that url from the community post and drop it into FT so idk if this is even worth considering. I know this might be a stretch but i just wanted to throw it in here.

The part that I'm stuck at is that this isn't a YouTube feature to have those links present; this "duplicate link" is something some authors put in by choice on some of their videos. So we're kinda just adding logic to check all of the different text / links of a community post with a matching ID of a blocked video (if they shared the non vanity link) in the same post just because some authors put a duplicate video link there. It would need a more compelling use case to check for such an informal case, IMO.

Screenshot_20231123_065749

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
All seems fine but one UX issue:
When I test  `Try inputting tags of length < 3`
There is no response from app when text < 2 chars input and enter pressed which makes it look broken
Better to have a toast

jasonhenriquez:
Weirdly am seeing this error when trying to add `showToast` into `ft-input`; have never seen anything like this before. 
```
renderer.js:15792 Uncaught ReferenceError: Cannot access '__WEBPACK_DEFAULT_EXPORT__' before initialization
    at Module.default (ft-input.vue:3:42)
    at eval (SubscribedChannels.js:15:1)
    at ./node_modules/babel-loader/lib/index.js!./src/renderer/views/SubscribedChannels/SubscribedChannels.js?vue&type=script&lang=js& (renderer.js:2863:1)
    at __webpack_require__ (renderer.js:15789:33)
    at fn (renderer.js:15998:21)
    at eval (SubscribedChannels.js?vue&type=script&lang=js&:5:144)
    at ./src/renderer/views/SubscribedChannels/SubscribedChannels.js?vue&type=script&lang=js& (renderer.js:5980:1)
    at __webpack_require__ (renderer.js:15789:33)
    at fn (renderer.js:15998:21)
    at eval (SubscribedChannels.vue:6:105)
```

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
We also need a toast when something already exists in the blocklist

https://github.com/FreeTubeApp/FreeTube/assets/73130443/ce14d05c-3145-4b6c-986e-efd099ad65b2

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
I am not sure if clearing input when input already exists is a good UX
Maybe user wants to correct the string (make it more specific)
It's easier to delete the whole input text then to retype the whole input text

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested
- Video list (subscription)
- Play Next in watch page (with some video hidden) to ensure only visible video played next
- Tested community post (with video) hidden

ChunkyProgrammer (approved):

BarbzYHOOL:
seems interesting

Lezney:
Can we add an option to hide podcasts?

PR #4204, opened 2023-10-21
Open/close all settings tabs persistent toggle

### LINKED ISSUE #1743: Add "Open All" button to settings page to open all tabs
It would make it much easier to look through the settings for a miscellaneous button if there was an easy way to open all of the tabs. It could float to the side of the settings tab, or could have a similar location.

To add a button to open all settings tabs at once

Opening them manually (takes 10s)

**Screenshot (mockup)**
image

jasonhenriquez:
# Open/close all settings tabs persisent toggle

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #1743 

## Description
Adds toggle in Theme Settings determining whether all settings tabs or opened or closed by default. The result is immediate, but the Theme Settings tab is kept open in both cases for convenience.

## Testing 
- Toggle it on & see state persist across sessions, & vice versa

## Desktop
- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx


absidue:
I wonder if it's possible to do that check inside the `ft-settings-section` component, instead of inside every settings section separately, that would reduce the duplicate code.

My suggested solution would be to add a prop to the ft-settings-section component so that you can still pass in a value from the theme settings one (have it optional with the default value being false), then in the `created` lifecyle hook set a data property based on the computed property and the prop from the theme settings. That data property can them be used in the vue template to set the value of the open attribute.

jasonhenriquez:
Just updated now to something like that (but still uses `computed` to have the change reflect immediately, like we do with our other theme changes). -44 LOC; that's the benefit of getting good sleep!

src/renderer/store/modules/settings.js
```diff
@@ -162,6 +162,7 @@ const defaultSideEffectsTriggerId = settingId =>
 /*****/
 
 const state = {
+  allSettingsTabsOpenedOrClosedDefault: false,
```
PikachuEXE, line 165:
```suggestion
  allSettingsSectionsExpandedByDefault: false,
```

src/renderer/components/theme-settings/theme-settings.vue
```diff
@@ -37,6 +38,12 @@
           :default-value="hideHeaderLogo"
           @change="updateHideHeaderLogo"
         />
+        <ft-toggle-switch
+          :label="$t('Settings.Theme Settings.Open All Settings Tabs By Default')"
+          :default-value="allSettingsTabsOpenedOrClosedDefault"
```
PikachuEXE, line 43:
```suggestion
          :default-value="allSettingsSectionsExpandedByDefault"
```

src/renderer/components/theme-settings/theme-settings.vue
```diff
@@ -37,6 +38,12 @@
...
+          :compact="true"
+          @change="handleAllSettingsTabsOpenedOrClosedDefault"
```
PikachuEXE, line 45:
```suggestion
          @change="handleAllSettingsSectionsExpandedByDefault"
```

src/renderer/components/theme-settings/theme-settings.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <ft-settings-section
+    :is-open-override="allSettingsTabsOpenedOrClosedDefaultModified"
```
PikachuEXE, line 3:
```suggestion
    :is-open-override="allSettingsSectionsExpandedByDefaultModified"
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -143,6 +147,11 @@ export default defineComponent({
       this.showRestartPrompt = true
     },
 
+    handleAllSettingsTabsOpenedOrClosedDefault: function (value) {
+      this.allSettingsTabsOpenedOrClosedDefaultModified = true
```
PikachuEXE, line 151:
```suggestion
      this.allSettingsSectionsExpandedByDefault = true
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -42,6 +43,9 @@ export default defineComponent({
     }
   },
   computed: {
+    allSettingsTabsOpenedOrClosedDefault: function () {
```
PikachuEXE, line 46:
```suggestion
    allSettingsSectionsExpandedByDefault: function () {
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -25,6 +25,7 @@ export default defineComponent({
       uiScaleStep: 5,
       disableSmoothScrollingToggleValue: false,
       showRestartPrompt: false,
+      allSettingsTabsOpenedOrClosedDefaultModified: false,
```
PikachuEXE, line 28:
```suggestion
      allSettingsSectionsExpandedByDefaultModified: false,
```

src/renderer/components/ft-settings-section/ft-settings-section.js
```diff
@@ -6,6 +6,15 @@ export default defineComponent({
...
       type: String,
       required: true
+    },
+    isOpenOverride: {
+      type: Boolean,
+      default: false
+    }
+  },
+  computed: {
+    allSettingsTabsOpenedOrClosedDefault: function () {
```
PikachuEXE, line 16:
```suggestion
    allSettingsSectionsExpandedByDefault: function () {
```

src/renderer/components/ft-settings-section/ft-settings-section.js
```diff
@@ -6,6 +6,15 @@ export default defineComponent({
...
+      return this.$store.getters.getAllSettingsTabsOpenedOrClosedDefault
```
PikachuEXE, line 17:
```suggestion
      return this.$store.getters.getAllSettingsSectionsExpandedByDefault
```

src/renderer/components/ft-settings-section/ft-settings-section.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <details class="settingsSection">
+  <details
+    :open="allSettingsTabsOpenedOrClosedDefault || isOpenOverride ? 'open' : null"
```
PikachuEXE, line 3:
```suggestion
    :open="allSettingsSectionsExpandedByDefault || isOpenOverride ? 'open' : null"
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -42,6 +43,9 @@ export default defineComponent({
     }
   },
   computed: {
+    allSettingsTabsOpenedOrClosedDefault: function () {
+      return this.$store.getters.getAllSettingsTabsOpenedOrClosedDefault
```
PikachuEXE, line 47:
```suggestion
      return this.$store.getters.getAllSettingsSectionsExpandedByDefault
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -143,6 +147,11 @@ export default defineComponent({
       this.showRestartPrompt = true
     },
 
+    handleAllSettingsTabsOpenedOrClosedDefault: function (value) {
+      this.allSettingsTabsOpenedOrClosedDefaultModified = true
+      this.updateAllSettingsTabsOpenedOrClosedDefault(value)
```
PikachuEXE, line 152:
```suggestion
      this.updateAllSettingsSectionsExpandedByDefault(value)
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -162,6 +171,7 @@ export default defineComponent({
     ...mapActions([
       'updateBarColor',
       'updateBaseTheme',
+      'updateAllSettingsTabsOpenedOrClosedDefault',
```
PikachuEXE, line 174:
```suggestion
      'updateAllSettingsTabsOpenedOrClosedDefault',
```

src/renderer/components/theme-settings/theme-settings.js
```diff
@@ -143,6 +147,11 @@ export default defineComponent({
       this.showRestartPrompt = true
     },
 
+    handleAllSettingsTabsOpenedOrClosedDefault: function (value) {
```
PikachuEXE, line 150:
```suggestion
    handleAllSettingsSectionsExpandedByDefault: function (value) {
```

PikachuEXE (changes requested):

jasonhenriquez:
I see examples of both expanded/collapsed & closed/open, for example [in this MDN article:](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details)
> This CSS creates a look similar to a tabbed interface, where activating the tab expands and opens it to reveal its contents.
> This CSS creates a look similar to a tabbed interface, where clicking the tab opens it to reveal its contents.
Note the use of the language of tabs as well. Still, I changed it to your suggestions because I have no strong preference.

PikachuEXE:
The linked example does look like a tab
image

I would argue our settings page have sections not tabs
Tabs: Can see content of one tab at a time only (e.g browser tabs)
Sections: Can see content of none-all sections a a time

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
I havent tested this yet but why does the user have to go into one of the setting to find this setting feels a bit weird to me. 

I would expect a toggle like this to be outside of all the settings that makes more sense in my mind

jasonhenriquez:
I was thinking that at first, but then again, it's a persistent setting that affects appearance of stuff, which places it in here. How would that external toggle's placement look like?

jasonhenriquez:
Here's what that would look like. It's a little too cute for me, but will go for it if others prefer it. If we do it this way, I'll remove the "By Default" text.

Screenshot_20231027_180714

PikachuEXE:
So it's no longer a setting?
Just a UI component to expand all?

jasonhenriquez:
Oh, to be clear, I was still thinking of it as a persistent setting in that case. I just think there's less need to call attention to that fact if we change the placement, if that makes sense.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Here's what that would look like. It's a little too cute for me

Looks a bit weird to me, would aligning it to the left make it better?

absidue:
If you put the A-Z switch from https://github.com/FreeTubeApp/FreeTube/pull/4231 next to it, it might look better, as then there are two switches using the space instead of just one.

jasonhenriquez:
> > Here's what that would look like. It's a little too cute for me
> 
> Looks a bit weird to me, would aligning it to the left make it better?

Here's what it looks like now:
Screenshot_20231028_114918

efb4f5ff-1298-471a-8973-3d47447115dc:
Looks better than placement in the middle but i would like the others to chime in also on this

1. Yes i agree it looks weird to have a single toggle outside all of the settings but in my mind at least it makes sense to have it outside

2. On the other hand i fully understand than if u have to put the toggle under one of the settings it would be theme settings. 

In all my life i just cant remember seeing such a toggle within the menu instead of outside of all the menus but if that is totally a thing maybe it would make more sense to go with nr 2

jasonhenriquez:
One point in favor of @absidue's suggestion is that both of those controls change the verticality of where the Theme Settings are, so you have to scroll back to find them. Keeping them on the outside makes them very easy to toggle on/off.

jasonhenriquez:
I think we should merge this one before #4231. Once this is merged, I'll move the settings sort in that PR to the top area as a toggle. As stated, much better user experience for those when the settings toggles are in one fixed space as opposed to having the manipulated section bouncing to a different height.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PikachuEXE (approved):

absidue (approved):

PR #4206, opened 2023-10-21
Display currently watching viewer count on live streams

### LINKED ISSUE #4081: [Feature Request]: Show viewer count on livestreams
### Guidelines

### Problem Description

On the youtube website, when you open a livestream, it shows the amount of people currently viewing the stream. Freetube does not do this.

### Proposed Solution

Add an active viewer count for livestreams in freetube, which increases and decreases as the amount of viewers changes.

### Alternatives Considered

n/a

### Issue Labels

### Additional Information

n/a

absidue:
# Display currently watching viewer count on live streams

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4081

## Description
Draft until https://github.com/LuanRT/YouTube.js/pull/527 is released.

Adds the currently the number of people currently watching the live stream to the live chat. It'll update about every 5 seconds similarly to YouTube. Respeccts the "Hide Video Views" setting.

While I know that visually the live chat view probably isn't the best place for this number, as it only works when the live chat is visible, I decided that placing in there would be less confusing to users (also keeps the code simpler as i don't have to move the value between 3 different components).

## Screenshots 
live-chat-with-watching-count

## Testing 
Open a live stream with a live chat and make sure the "Hide Video Views" setting is disabled.

e.g. https://youtu.be/jfKfPfyJRdk

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE:
Cannot review until PR released?

absidue:
I thought GitHub would just queue up the review requests until it was marked ready for review, but I guess not, sorry.

ChunkyProgrammer:
Should this PR be set to ready to review?

src/renderer/components/watch-video-live-chat/watch-video-live-chat.css
```diff
@@ -22,6 +22,10 @@
   row-gap: 16px;
 }
 
+.watchingCount {
+  font-weight: normal;
```
jasonhenriquez, line 26: **suggestion (non-blocking):** Add space between title text and count. Decrease font size of count and use tertiary color to better direct focus.

```suggestion
  font-weight: normal;
  margin-inline-start: 5px;
  font-size: 15px;
  color: var(--tertiary-text-color);
```

Screenshot_20231116_204817

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
- When live chat is disabled by the creator u cant see the viewer count anymore but on YT u can
- Users can also hide the live chat in the distraction free settings and that also hides the viewer count

U could move the element next to the started streaming element this is how YT does it idk if it is a better UX though

absidue:
We can't have it separate from the live chat because that view count comes from polling the live chat endpoints, I can move it in the UI, but I'll still have to hide it if the video doesn't have a live chat.

efb4f5ff-1298-471a-8973-3d47447115dc:
Ah ok, the only thing i maybe would like to see is the same as @jasonhenriquez suggestion about

`Add space between title text and count`

Maybe to make it look nice we can reuse the same separators we use in the img below?

Capture

absidue:
Updated the styling using @jasonhenriquez suggestion.

jasonhenriquez (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #3051, opened 2023-01-11
[Feature Request]: "Don't Recommend Channel" option in Search Results via "More Options" 3-dot context menu


AnnaOps:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

It would be nice to block channels in the search results by clicking on the 3-dot context menu "More Options" -> "Don't Recommend Channel".
The current options are "Mark As Watched", "Copy YouTube Link" etc.

https://www.makeuseof.com/how-to-block-youtube-channels/

Thank you for your good work.

### Proposed Solution

Add "Don't Recommend Channel" option to the 3-dot context menu "More Options" in Search Results.
The current options are "Mark As Watched", "Copy YouTube Link" etc.
Add "Search Settings" in Settings with "Manage Blocked Channels" button or list.

![FreeTube](https://user-images.githubusercontent.com/122434149/211753296-c91d6908-0108-4dc3-a5b4-ce0e1364042f.jpg)

### Alternatives Considered

None

### Issue Labels

new optional setting

BarbzYHOOL:
Very wanted (but there is no recommandation feature either)

absidue:
For anyone thinking about implementing this, it can be built on top of the "Hide videos from channels" feature introduced in https://github.com/FreeTubeApp/FreeTube/pull/2849.

MitchelPaulin:
I'd like to take a look at implementing this

PR #4228, opened 2023-10-24
Add hide channel option directly to More Options menu

MitchelPaulin:
# Add hide channel option directly to More Options menu

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/3051

## Description

Piggybacks off the feature that was implemented in https://github.com/FreeTubeApp/FreeTube/pull/2849 to add an option directly to the "More Options" menu to hide a channel. Options will show up in the "Distraction Settings" afterwards and can be removed from there. 

## Screenshots 

image

## Testing 

Selected the option in the context menu, the video is immediately filtered and a toast is thrown up to confirm it was hidden. Then I checked the "Distraction Settings" and confirmed the channel is there. 

## Desktop

- **OS:** Pop_OS
- **OS Version:** 22.04 LTS

## Additional Info

I decided not to go with "Hide Recommendation" like mentioned in the original issue as there really isn't recommendations, just trending.

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -211,6 +211,10 @@ export default defineComponent({
               type: 'divider'
             },
             {
+              label: this.$t('Video.Hide Channel'),
+              value: 'hideChannel'
+            },
+            {
               label: this.$t('Video.Copy YouTube Channel Link'),
               value: 'copyYoutubeChannel'
             },
@@ -434,6 +438,9 @@ export default defineComponent({
         case 'openInvidiousChannel':
           openExternalLink(this.invidiousChannelUrl)
           break
+        case 'hideChannel':
+          this.hideChannel(this.channelName)
+          break
       }
     },
 
@@ -621,12 +628,21 @@ export default defineComponent({
       showToast(this.$t('Video.Video has been removed from your saved list'))
     },
 
+    hideChannel: function(channelName) {
+      const hiddenChannels = JSON.parse(this.$store.getters.getChannelsHidden)
+      hiddenChannels.push(channelName)
+      this.updateChannelsHidden(JSON.stringify(hiddenChannels))
+
+      showToast(this.$t('Channel Hidden', { channel: channelName }))
+    },
+
     ...mapActions([
       'openInExternalPlayer',
       'updateHistory',
       'removeFromHistory',
       'addVideo',
-      'removeVideo'
+      'removeVideo',
+      'updateChannelsHidden'
     ])
   }
 })
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -623,6 +623,7 @@ Video:
   Copy YouTube Channel Link: Copy YouTube Channel Link
   Open Channel in Invidious: Open Channel in Invidious
   Copy Invidious Channel Link: Copy Invidious Channel Link
+  Hide Channel: Hide Channel
   Views: Views
   Loop Playlist: Loop Playlist
   Shuffle Playlist: Shuffle Playlist
@@ -920,6 +921,7 @@ Starting download: 'Starting download of "{videoTitle}"'
 Downloading failed: 'There was an issue downloading "{videoTitle}"'
 Screenshot Success: Saved screenshot as "{filePath}"
 Screenshot Error: Screenshot failed. {error}
+Channel Hidden: Channel {channel} Hidden
 
 Hashtag:
   Hashtag: Hashtag


PikachuEXE:
I think the new option should be added to the end with a separator
Also it might be better to add a confirmation modal/something to avoid misclick

jasonhenriquez:
Good work @MitchelPaulin!

On top of @PikachuEXE's suggestions, could you also add an "Unhide Channel" option for when the channel is hidden? That may sound silly, but hidden channel videos can still show up in other places like Subscriptions, the Channel page, and Playlists.

MitchelPaulin:
Moved it into its own divider at the bottom and added the ability to show hidden channels, good suggestions. 

Not sure how I feel about the confirmation popup, I feel like it might just get in the way more than anything. But if its a blocker I can add it.

jasonhenriquez:
I don't think it's strictly needed because you can always see and remove those channels from your Distraction Free Settings. Especially given that there's a toast letting you know what you did if you accidentally clicked it. @PikachuEXE Second thoughts?

PikachuEXE (approved):
Hmm? Forgot to approve after review

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -621,12 +644,28 @@ export default defineComponent({
       showToast(this.$t('Video.Video has been removed from your saved list'))
     },
 
+    hideChannel: function(channelName) {
+      const hiddenChannels = JSON.parse(this.$store.getters.getChannelsHidden)
+      hiddenChannels.push(channelName)
+      this.updateChannelsHidden(JSON.stringify(hiddenChannels))
+
+      showToast(this.$t('Channel Hidden', { channel: channelName }))
+    },
```
absidue, line 653: Please change this pull request to use the channel ID, not the name. Blocking by name is unreliable as the channel owner can change the channel name, so you would have to block it again when that happens, but the ID never changes, so it will continue to work even when the name changes.

absidue (changes requested):

src/renderer/components/ft-list-video/ft-list-video.js
MitchelPaulin: I went with the name since the Ids were very unreadable, if #4230 merges though that's no longer a concern, I can switch it Ids

ChunkyProgrammer (dismissed):

jasonhenriquez (approved):

PikachuEXE:
I will wait for https://github.com/FreeTubeApp/FreeTube/pull/4230 before merging this

MitchelPaulin:
@PikachuEXE #4230 was merged, can this be merged now?

PikachuEXE (approved):
Oops sorry I forgot
No wonder I feel like something is missing when testing https://github.com/FreeTubeApp/FreeTube/pull/4230 ...

PikachuEXE:
@absidue Please re-review 😅

jasonhenriquez (approved):

static/locales/en-US.yaml
```diff
@@ -920,6 +922,8 @@ Starting download: 'Starting download of "{videoTitle}"'
 Downloading failed: 'There was an issue downloading "{videoTitle}"'
 Screenshot Success: Saved screenshot as "{filePath}"
 Screenshot Error: Screenshot failed. {error}
+Channel Hidden: Channel {channel} added to filter
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 925: I think this messaging can be better now that #4202 is here

I think `{channel} added to Channel filter` is more explicit

static/locales/en-US.yaml
```diff
@@ -920,6 +922,8 @@ Starting download: 'Starting download of "{videoTitle}"'
...
+Channel Unhidden: Channel {channel} removed from filter
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 926: same here

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Maybe this can be made more explicit

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -228,8 +228,25 @@ export default defineComponent({
...
               label: this.$t('Video.Open Channel in Invidious'),
               value: 'openInvidiousChannel'
+            },
+            {
+              type: 'divider'
             }
           )
+
+          const hiddenChannels = JSON.parse(this.$store.getters.getChannelsHidden)
+          const channelShouldBeHidden = hiddenChannels.some(c => c === this.channelName || c === this.channelId)
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 238: AFAIK channelName is not a thing anymore

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
In the latest nightly this isnt a thing

src/renderer/components/ft-list-video/ft-list-video.js
MitchelPaulin: Would we not need the channel name checked here to support anyone that has legacy blocks via channel name?

efb4f5ff-1298-471a-8973-3d47447115dc: i could be missing something but

If im looking at the conversations in https://github.com/FreeTubeApp/FreeTube/pull/4230 i think everything is based on channel ID now

MitchelPaulin: I was just thinking of what would happen to a user who has blocked a channel by name then updates freetube

MitchelPaulin: Ah ok, scrolling through the other PR it does look like we are fine just not respecting legacy blocks https://github.com/FreeTubeApp/FreeTube/pull/4230#issuecomment-1783769677

In that case I can remove the channel name check

efb4f5ff-1298-471a-8973-3d47447115dc:
I see `channelName` being used in more places do they also need to be removed?

MitchelPaulin:
> I see `channelName` being used in more places do they also need to be removed?

Yes, missed a spot, the rest of the uses are for the toast message though, so those need to stay

absidue:
Looks like an error caused by invalid YAML. The newly added strings with the curly bracket at the start probably need to be quoted.

absidue (approved):

absidue:
Thank you for your contribution! sorry for the long back-and-forwards on this.

PikachuEXE (approved):
I tested after merging `development` into this branch otherwise it might not work properly
All seems fine

ChunkyProgrammer (approved):

BarbzYHOOL:
cool

Issue #4181, opened 2023-10-18
[Feature Request]: Resolve Channel ID for hidden channels


LangLangBart:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

- with the release of [v0.19.0 Beta](https://github.com/FreeTubeApp/FreeTube/releases/tag/v0.19.0-beta), the ability to hide channels was added
  - ref to #2849

---

- the description suggests a complete name match, so I began using only the channel IDs instead of their names to avoid typos or potential name changes

<img src="https://github.com/FreeTubeApp/FreeTube/assets/92653266/90ef1db7-6c87-479a-a0ce-213a5ef5d590" width="300">

- however, after adding about ~35 entries, it became difficult to identify the channels

<img src="https://github.com/FreeTubeApp/FreeTube/assets/92653266/f87db2c3-890b-4911-9c2f-c32692c9bd73" width="400">

---

- currently, the workaround is to run a command to reveal the channel name and then remove the channel I no longer wish to block, which is cumbersome

```bash
# macOS
for i in $(jq -r 'select(._id=="channelsHidden").value | fromjson[]' ~/Library/Application\ Support/FreeTube/settings.db); do
	echo "$i"
	curl -s "https://www.youtube.com/feeds/videos.xml?channel_id=$i" |
		sed -n 's|.*<name>\(.*\)</name>.*|\1|p' |
		sed q
done

# UCeeFfhMcJa1kjtfZAGskOCA
# TechLinked
# UCXuqSBlHAE6Xw-yeJA0Tunw
# Linus Tech Tips
# UCVT1tPkR-fUVlO652EcO3ow
# CBS Sunday Morning
# UC4qhj5cnX0MkbwXQ5vlI1NQ
# ...
```

### Proposed Solution

- when adding a new entry, verify the channel ID
  - if valid, append the channel's name in parentheses after the ID

  - example: 
```md
UCeeFfhMcJa1kjtfZAGskOCA (TechLinked)
```

[...]

PR #4230, opened 2023-10-25
Add identifiers to channels hidden by ID

Benjababe:
# Add identifiers to channels hidden by ID

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Closes https://github.com/FreeTubeApp/FreeTube/issues/4181

## Description
If a channel ID is provided (checks for `UC` at start and is 24 characters long), searches invidious or local API for the channel name and thumbnail and stores them. Final commit uses a tooltip instead of parenthesis to show the channel name, screenshots below to compare.

## Screenshots 
image

## Testing 

Tried with
- UC9_p50tH3WmMslWRWKnM7dQ (Adam Ragusea)
- UCfsOfLvadg89Bx8Sv_6WERg (NoobFromUA)
- UCXR5PCyUI2pG_a3ydnCIS-g (NesHacker)
- 3DSage
- engineerguy

## Desktop

- **OS:** Windows 10
- **OS Version:** 22H2
- **FreeTube version:**


src/renderer/components/ft-list-video-lazy/ft-list-video-lazy.js
```diff
@@ -58,12 +58,20 @@ export default defineComponent({
...
-      return JSON.parse(this.$store.getters.getChannelsHidden)
+      let hidden = JSON.parse(this.$store.getters.getChannelsHidden)
+      hidden = hidden.map((ch) => {
+        // Legacy support
+        if (typeof ch === 'string') {
+          return { name: ch, secondaryName: '', icon: '' }
+        }
+        return ch
+      })
+      return hidden
```
PikachuEXE, line 69:
```suggestion
      return JSON.parse(this.$store.getters.getChannelsHidden).map((ch) => {
        // Legacy support
        if (typeof ch === 'string') {
          return { name: ch, secondaryName: '', icon: '' }
        }
        return ch
      })
```

src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.js
```diff
@@ -53,7 +53,15 @@ export default defineComponent({
...
-      return JSON.parse(this.$store.getters.getChannelsHidden)
+      let hidden = JSON.parse(this.$store.getters.getChannelsHidden)
+      hidden = hidden.map((ch) => {
+        // Legacy support
+        if (typeof ch === 'string') {
+          return { name: ch, secondaryName: '', icon: '' }
+        }
+        return ch
+      })
+      return hidden
```
PikachuEXE, line 64:
```suggestion
      return JSON.parse(this.$store.getters.getChannelsHidden).map((ch) => {
        // Legacy support
        if (typeof ch === 'string') {
          return { name: ch, secondaryName: '', icon: '' }
        }
        return ch
      })
```

src/renderer/components/distraction-settings/distraction-settings.js
```diff
@@ -105,7 +114,15 @@ export default defineComponent({
...
-      return JSON.parse(this.$store.getters.getChannelsHidden)
+      let hidden = JSON.parse(this.$store.getters.getChannelsHidden)
+      hidden = hidden.map((ch) => {
+        // Legacy support
+        if (typeof ch === 'string') {
+          return { name: ch, secondaryName: '', icon: '' }
+        }
+        return ch
+      })
+      return hidden
```
PikachuEXE, line 125:
```suggestion
      return JSON.parse(this.$store.getters.getChannelsHidden).map((ch) => {
        // Legacy support
        if (typeof ch === 'string') {
          return { name: ch, secondaryName: '', icon: '' }
        }
        return ch
      })
```

PikachuEXE (commented):
Shorten the code a bit first?

LangLangBart:
Thank you. Works well.

Here are few points for consideration:

1.  The UI could be simplified by removing the ID and displaying only the icon and channel name, thereby eliminating the need for a tooltip.

2. In the process of migrating `channelsHidden` settings from `~/Library/Application\ Support/FreeTube/settings.db` to `~/Library/Application\ Support/Electron/settings.db`, I've noticed that the channels don't update automatically. Currently, the user is required to manually delete each item and re-add them to the list. It becomes cumbersome as the ID cannot be simply copied but requires constant referencing back to the `~/Library/Application\ Support/FreeTube/settings.db` file. This issue could benefit from a more efficient update process.

3. When the window is very small, the `?` shifts to the right, as shown in the video below:

https://github.com/FreeTubeApp/FreeTube/assets/92653266/e363446b-a9b9-4785-8ea5-331ce1d59094

---

everything above is UI related only and you can argue either way, the underlying problem is fixed with this PR

absidue:
@LangLangBart When you are editing the files on disk, make sure FreeTube is completely closed.

Maybe it's time we add a lock on those files while FreeTube is running, so people stop editing them while FreeTube is running and wondering why stuff breaks.

Benjababe:
Updated it to just show the channel name. 

Need suggestions on handling already blocked channels using ID. Like to do it automatically once on startup or have some button for the user to update them.

absidue:
@Benjababe The blocking should be done using the channel ID, that way it will still work when the channel name changes.

efb4f5ff-1298-471a-8973-3d47447115dc:
what about automatically when user opens the distraction free settings

absidue:
Ideally blocking by name should be prevented and only blocking by ID allowed, any legacy blocks that are done by name should be ignored and placed in a separate section in the settings labelled with "Incompatible legacy blocks" or something.

Benjababe:
Updated the hidden channels to only accept IDs.

For existing channel IDs, it will try to update them when distraction settings is mounted on the condition. Only IDs that return some form of `channel doesn't exist` will be skipped. Idk whether they should automatically be removed or not.

PikachuEXE (approved):
This PR should not handle name to ID conversion (too much to review in 1 PR and not sure if that's desired)

Otherwise looks good (`MrBeast` one is the old block by name entry)
image

absidue:
I agree this pull request, shouldn't try converting them. Especially as it can't be automated, as searching for a name, will likely bring up multiple channels and you would need the user to say which one.

What this pull request could do is add some indication to the legacy name ones, that makes it clear to the user that they aren't supported and get ignored during filtering.

e.g. maybe they could be displayed in a separate section or something, but again that doesn't need to be part of this pull request

ChunkyProgrammer (dismissed):

src/renderer/components/ft-input-tags/ft-input-tags.vue
```diff
@@ -13,15 +20,21 @@
...
       <ul>
         <li
           v-for="tag in tagList"
           :key="tag.id"
         >
-          <span>{{ tag }}</span>
+          <img
+            v-if="tag.icon"
+            :src="tag.icon"
+            alt="icon"
```
jasonhenriquez, line 32: [This is not meaningful alt text.](https://www.howtogeek.com/821673/what-is-alt-text-why-you-should-use-it-and-how-to-write-it/) You should generally use an empty alt tag when the image's purpose is decorative.
```suggestion
            alt=""
```

jasonhenriquez (commented):
**suggestion (non-blocking)**: It would be nice if clicking on the channel thumbnail brought you to the channel page.

**suggestion (non-blocking)**: This was pre-existing, but it would be nice if the `ft-input-tags` text was selectable.

jasonhenriquez:
Thanks for the speedy work! But getting some errors now.
Screenshot_20231108_072820

Benjababe:
My bad, forgot to check for tags without icons

PikachuEXE (approved):

jasonhenriquez (approved):

ChunkyProgrammer (approved):

PR #4233, opened 2023-10-25
Add a 15 second timeout for fetching the Invidious instances at startup

### LINKED ISSUE #4223: [Bug]: freetube wont load
### Guidelines

### Describe the bug

1. just open freetube
2. blank screen forever

### Expected Behavior

it should show my feed?

### Issue Labels

content not loading

### FreeTube Version

v0.19.1 Beta

### Operating System Version

Ubuntu 20.04.6 LTS

### Installation Method

.deb

### Primary API used

Local API

_No response_

### Additional Information

I've tried downgrading to v0.18.x but it won't work, as well...
image
image

### Nightly Build

absidue:
# Add a 15 second timeout for fetching the Invidious instances at startup

## Pull Request Type

- [x] Bugfix

## Related issue
closes #4223

## Description
When the Invidious instances API goes down, we want to fallback to our local copy of the instance list, unfortunately `fetch` waits for a really long time before it times out, unfortunately that means FreeTube sits on a blank screen at startup forever. I wasn't able to find an exact value for long chromium browsers wait, but the general consensus seems to be that it is a really long time.

To solve that issue, this pull request introduces a 15 second timeout for fetching the Invidious instances list before falling back to the local copy. I know 15 seconds seems quite a lot, but I wanted to make sure it wouldn't time out too early, even for people with the slowest proxies configured.

While we definitely shouldn't have the loading of the entire app dependant on a network request, refactoring all relevant parts is a lot more complicated. We would have to hide and deactivate certain functionality until the instance list is fetched, when you have the Invidious API selected and don't have a default instance configured. Top bar could be loaded but the search bar would have to be disabled, the side bar could be loaded but without the subscriptions list and with the popular icon hidden (also have to disable the popular keyboard shortcut), fetching subscriptions would have to be delayed until an instance is selected, etc etc.

We should probably setup a GitHub workflow that updates our list once a month, so it stays relatively up-to-date.

<details><summary>If you have a really slow proxy configured for your normal FreeTube use, firstly why??? secondly</summary>

meme

</details> 

## Testing 
Edit the timeout in the code to something short like 5ms, reload the FreeTube window with <kbd>CTRL</kbd>+<kbd>R</kbd> or <kbd>Command</kbd>+<kbd>R</kbd> and notice that the `Fetching the Invidious instance list timed out after 15 seconds. Falling back to local copy.` error shows up in the console. (you can try messing around with the throttling settings in the network dev tools if you want, but that was painfully slow for me, as it slows down the loading of FreeTube's own files too).

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -672,3 +672,35 @@ export function escapeHTML(untrusted) {
 export function deepCopy(obj) {
   return JSON.parse(JSON.stringify(obj))
 }
+
+/**
+ * Check if the `name` of the error is `TimeoutError` to know if the error was caused by a timeout or something else.
+ * @param {number} milliseconds
+ * @param {RequestInfo|URL} input
+ * @param {RequestInit=} init
+ */
+export async function fetchWithTimeout(milliseconds, input, init) {
+  const timeoutSignal = AbortSignal.timeout(milliseconds)
+
+  if (typeof init !== 'undefined') {
+    init.signal = timeoutSignal
+  } else {
+    init = {
+      signal: timeoutSignal
+    }
+  }
+
+  try {
+    return await fetch(input, init)
+  } catch (err) {
+    if (err.name === 'AbortError' && timeoutSignal.aborted) {
+      // According to the spec, fetch should use the original abort reason.
+      // Unfortunately chromium browsers always throw an AbortError, even when it was caused by a TimeoutError,
+      // so we need manually throw the original abort reason
+      // https://bugs.chromium.org/p/chromium/issues/detail?id=1431720
+      throw timeoutSignal.reason
+    } else {
+      throw err
+    }
+  }
+}
--- a/src/renderer/store/modules/invidious.js
+++ b/src/renderer/store/modules/invidious.js
@@ -1,6 +1,6 @@
 import fs from 'fs/promises'
 import { pathExists } from '../../helpers/filesystem'
-import { createWebURL } from '../../helpers/utils'
+import { createWebURL, fetchWithTimeout } from '../../helpers/utils'
 
 const state = {
   currentInvidiousInstance: '',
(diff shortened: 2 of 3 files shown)


ChunkyProgrammer (dismissed):

src/renderer/helpers/utils.js
```diff
@@ -672,3 +672,35 @@ export function escapeHTML(untrusted) {
...
 }
+
+/**
+ * Check if the `name` of the error is `TimeoutError` to know if the error was caused by a timeout or something else.
+ * @param {number} milliseconds
+ * @param {RequestInfo|URL} input
+ * @param {RequestInit=} init
+ */
+export async function fetchWithTimeout(milliseconds, input, init) {
+  const timeoutSignal = AbortSignal.timeout(milliseconds)
```
PikachuEXE, line 683: Not blocking issue but good to name parameters as their function (what to store)
For small functions this is fine (so it's not blocking) but it's a good practice for bigger functions
```suggestion
export async function fetchWithTimeout(timeoutMs, input, init) {
  const timeoutSignal = AbortSignal.timeout(timeoutMs)
```

PikachuEXE (approved):

src/renderer/helpers/utils.js
efb4f5ff-1298-471a-8973-3d47447115dc: @absidue want to implement this?

src/renderer/helpers/utils.js
```diff
@@ -672,3 +672,35 @@ export function escapeHTML(untrusted) {
 export function deepCopy(obj) {
   return JSON.parse(JSON.stringify(obj))
 }
+
+/**
+ * Check if the `name` of the error is `TimeoutError` to know if the error was caused by a timeout or something else.
+ * @param {number} milliseconds
```
absidue, line 678:
```suggestion
 * @param {number} timeoutMs
```

ChunkyProgrammer (approved):

PikachuEXE (approved):

jasonhenriquez (approved):

Issue #2728, opened 2022-10-15
[Bug]: Previous video doesnt work on the first video in playlist when loop is enabled


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for open and closed issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.

### Describe the bug

1. Go to any playlist
2. play first video of playlist
3. enable loop
4. click previous button
5. doesnt go to previous video

### Expected Behavior

Player should go from video 1/12 -> video 12/12

### Issue Labels

usability issue

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/62c70fbdd60435ab96113bdee9bb2c529df0eac7

### Operating System Version

win10 21H1

### Installation Method

.exe

### Primary API used

Local API

### Additional Information

https://github.com/FreeTubeApp/FreeTube/pull/2698#issuecomment-1279806479

This should also work with the shortcut that has been implemented for it in https://github.com/FreeTubeApp/FreeTube/pull/2698

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
Cant reproduce after a gazillion restarts

efb4f5ff-1298-471a-8973-3d47447115dc:
Closed wrong issue -.-

PR #4102, opened 2023-10-05
Fix playlist item fetching for local API

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes https://github.com/FreeTubeApp/FreeTube/issues/4068 https://github.com/FreeTubeApp/FreeTube/issues/4086

## Description

YT now sometimes returns useless continuation data which makes loading "next page" sometimes failed with error from `youtube.js`
We need to handle it somehow

## Screenshots 

Lazy

## Testing 

Example playlists/video with playlists to test
- https://www.youtube.com/watch?v=a52LaNBDV7Q&list=PL61FH1Fo4C7FPgWSYMaQj-OjWUB-buzUq
- https://youtube.com/playlist?list=PL61FH1Fo4C7FPgWSYMaQj-OjWUB-buzUq
- https://www.youtube.com/watch?v=MH2JHRBQ6ng&list=PLrBjj4brdIRzn3ja4BfqYUForg0p-p5wi (https://github.com/FreeTubeApp/FreeTube/issues/4068)
- https://www.youtube.com/watch?v=Ytkf3qZTj74&list=PLT1F2nOxLHOcmi_qi1BbY6axf5xLFEcit (https://github.com/FreeTubeApp/FreeTube/issues/4086)

### Watch video page, playlist component loading (normal)
- Watch a video with playlist **without** going to playlist view first (e.g. https://www.youtube.com/watch?v=a52LaNBDV7Q&list=PL61FH1Fo4C7FPgWSYMaQj-OjWUB-buzUq)
- Ensure no error, all items loaded

### Watch video page, playlist component loading (playlist cached)
- Watch a video with playlist **without** going to playlist view first (e.g. https://youtube.com/playlist?list=PL61FH1Fo4C7FPgWSYMaQj-OjWUB-buzUq)
- Ensure no error, all items loaded, **no duplicate**

### Playlist view
- Visit a playlist (e.g. https://youtube.com/playlist?list=PL61FH1Fo4C7FPgWSYMaQj-OjWUB-buzUq)
- Press load more, ensure no error shown
- No idea how to make it disappear on load without extra request

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

Issue #1815, opened 2021-10-12
Mark destructive buttons as red


ChunkyProgrammer:
**Behavior of the program**
We currently use the user's primary and secondary color preferences to inform how to distinguish destructive versus non-destructive actions. The cost of this design anti-pattern is that that is not a clear visual indicator of destructive actions, which means that it takes longer for users to process and comprehend the implications of these actions.

**Expected behavior**
[Destructive actions should be colored red](https://uxmovement.com/buttons/how-to-design-destructive-actions-that-prevent-data-loss/). If possible, we should attempt to make this red sufficiently distinct from any oher theme color (3:1 color contrast), but it is fine if it comes up a bit short against only the Red theme color. This includes certain buttons in the settings, some buttons in `ft-prompt`s, and probably a couple of other places I'm not thinking of.

**Screenshots**
![image](https://user-images.githubusercontent.com/78101139/137035364-3accd339-b709-4b02-82a3-169d159bd3f2.png)

**Environment Information (please complete the following information):**
 - OS: Windows
 - OS Version: 10
 - FreeTube version: Nightly build
 - Installation Method (If applicable): .exe
 - Primary API Used: Local API

PrestonN:
This inconsistency makes a little more sense with the default colors (red and blue) as I wanted it to be clear that these were different.

The buttons in question are more hard actions where irreversible changes can happen. In the default theme, these buttons are red, which makes sense in that context.

With the customizability of the app, I'm not really sure what other solution we should do at a glance. I don't really believe making all of the buttons the same color is the solution due to the nature of what some of the buttons do, but I also don't want to force the buttons to clash with the theme settings either.

I'm open to suggestions for this. I know it's a little strange.

jasonhenriquez:
Permission to change this issue name to "Mark destructive buttons as red"? @ChunkyProgrammer

efb4f5ff-1298-471a-8973-3d47447115dc:
That would mean that all buttons in the settings will basically be based on the Secondary Color Theme and the destructive ones will be red. Which im perfectly fine with just wanted to clarify this for myself

ChunkyProgrammer:
> Permission to change this issue name to "Mark destructive buttons as red"? @ChunkyProgrammer

Feel free to update this issue or close it and make a new one

PR #4234, opened 2023-10-26
(Multiple) Local Playlist (instead of saving video in one hidden list)

PikachuEXE:
## Pull Request Type

- [x] Feature Implementation

## Related issue

- No idea please update this list directly lol

## Description

Instead of saving videos (in one hidden playlist) and displaying only that list
This PR makes it possible to manage multiple local playlists and adding videos to it
You can also import export

## Tasks Left
- [x] Translation (already done but could have some entries missed)
- [ ] Code Cleanup

## Screenshots 

Lazy maybe post later on request

## Testing 

- Playlist migration (from old version of FT)
- Playlist import export (from old/new version of FT)
- Playlist management
- Video management in playlist (including adding video(s) to playlist)
- I18n (not sure how to test this)
- What else?

## Additional context

Code based on branch https://github.com/FreeTubeApp/FreeTube/tree/playlists
See https://github.com/PikachuEXE/FreeTube/issues for issues reported before this PR created


efb4f5ff-1298-471a-8973-3d47447115dc:
Note: See playlist related issues https://github.com/PikachuEXE/FreeTube/issues/44

Ill check later in that list if this PR closes more issues

efb4f5ff-1298-471a-8973-3d47447115dc:
It should close #2728 but i would like the others to confirm this

src/renderer/components/watch-video-info/watch-video-info.js
```diff
@@ -268,15 +256,32 @@ export default defineComponent({
...
-
-    toggleSave: function () {
-      if (this.inFavoritesPlaylist) {
-        this.removeFromPlaylist()
-      } else {
-        this.addToPlaylist()
+        playlistLoop: this.getPlaylistLoop(),
+      }
+      // Only play video in non playlist mode when user playlist detected
+      if (!this.inUserPlaylist) {
```
ChunkyProgrammer, line 262: Shouldn't this be
```suggestion
      if (this.inUserPlaylist) {
```

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,85 @@
...
+
+    this.playlistName = this.newPlaylistVideoObject.title
+    // Faster to input required playlist name
+    this.$refs.playlistNameInput.focus()
+  },
+  destroyed() {
+    if (this.lastActiveElement != null) {
+      this.lastActiveElement.focus()
+    }
+  },
```
jasonhenriquez, line 46:
```suggestion
    this.lastActiveElement?.focus()
  },
```

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,85 @@
...
+    this.lastActiveElement = document.activeElement
+
+    this.playlistName = this.newPlaylistVideoObject.title
+    // Faster to input required playlist name
+    this.$refs.playlistNameInput.focus()
+  },
+  destroyed() {
+    if (this.lastActiveElement != null) {
+      this.lastActiveElement.focus()
+    }
```
jasonhenriquez, line 45:
```suggestion
   this.lastActiveElement?.focus()
```

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
efb4f5ff-1298-471a-8973-3d47447115dc: I think you marked this as resolved on accident

src/renderer/components/ft-prompt/ft-prompt.js
```diff
@@ -64,6 +67,11 @@ export default defineComponent({
     })
     this.focusItem(0)
   },
+  destroyed() {
+    if (this.lastActiveElement != null) {
+      this.lastActiveElement.focus()
+    }
```
jasonhenriquez, line 73:
```suggestion
   this.lastActiveElement?.focus()
```

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -0,0 +1,267 @@
...
+  mounted: function () {
+    this.lastActiveElement = document.activeElement
+
+    this.updateQueryDebounce = debounce(this.updateQuery, 500)
+    // User might want to search first if they have many playlists
+    this.$refs.searchBar.focus()
+  },
+  destroyed() {
+    if (this.lastActiveElement != null) {
+      this.lastActiveElement.focus()
```
jasonhenriquez, line 200:
```suggestion
   this.lastActiveElement?.focus()
```

src/renderer/components/ft-prompt/ft-prompt.js
jasonhenriquez: Lots of examples of this, so just do a find and replace

src/renderer/components/watch-video-info/watch-video-info.js
PikachuEXE: Fixed

src/renderer/components/ft-prompt/ft-prompt.js
PikachuEXE: I only see 3 (search by `Element.focus()`)

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,83 @@
...
+import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
+import FtPrompt from '../ft-prompt/ft-prompt.vue'
+import FtButton from '../ft-button/ft-button.vue'
+import FtInput from '../ft-input/ft-input.vue'
+import FtPlaylistSelector from '../ft-playlist-selector/ft-playlist-selector.vue'
+import {
+  showToast,
+} from '../../helpers/utils'
+
+export default Vue.extend({
```
absidue, line 12: `defineComponent` gives us auto-complete and is compatible with Vue 3.
```suggestion
import { defineComponent } from 'vue'
import { mapActions } from 'vuex'
import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
import FtPrompt from '../ft-prompt/ft-prompt.vue'
import FtButton from '../ft-button/ft-button.vue'
import FtInput from '../ft-input/ft-input.vue'
import FtPlaylistSelector from '../ft-playlist-selector/ft-playlist-selector.vue'
import {
  showToast,
} from '../../helpers/utils'

export default defineComponent({
```

src/renderer/components/ft-playlist-selector/ft-playlist-selector.js
```diff
@@ -0,0 +1,75 @@
+import Vue from 'vue'
+import FtIconButton from '../ft-icon-button/ft-icon-button.vue'
+import { mapActions } from 'vuex'
+
+export default Vue.extend({
```
absidue, line 5: `defineComponent` gives us auto-complete and is compatible with Vue 3.
```suggestion
import { defineComponent } from 'vue'
import FtIconButton from '../ft-icon-button/ft-icon-button.vue'
import { mapActions } from 'vuex'

export default defineComponent({
```

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -0,0 +1,265 @@
...
+  NameDescending: 'name_descending',
+
+  LatestCreatedFirst: 'latest_created_first',
+  EarliestCreatedFirst: 'earliest_created_first',
+
+  LatestUpdatedFirst: 'latest_updated_first',
+  EarliestUpdatedFirst: 'earliest_updated_first',
+}
+
+export default Vue.extend({
```
absidue, line 25: `defineComponent` gives us auto-complete and is compatible with Vue 3.
```suggestion
import { defineComponent } from 'vue'
import { mapActions } from 'vuex'
import debounce from 'lodash.debounce'
import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
import FtPrompt from '../ft-prompt/ft-prompt.vue'
import FtButton from '../ft-button/ft-button.vue'
import FtPlaylistSelector from '../ft-playlist-selector/ft-playlist-selector.vue'
import FtInput from '../../components/ft-input/ft-input.vue'
import FtSelect from '../../components/ft-select/ft-select.vue'
import {
  showToast,
} from '../../helpers/utils'

const SORT_BY_VALUES = {
  NameAscending: 'name_ascending',
  NameDescending: 'name_descending',

  LatestCreatedFirst: 'latest_created_first',
  EarliestCreatedFirst: 'earliest_created_first',

  LatestUpdatedFirst: 'latest_updated_first',
  EarliestUpdatedFirst: 'earliest_updated_first',
}

export default defineComponent({
```

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.scss
absidue, line 1: This file doesn't contain any scss specific syntax, so it can be renamed to a `.css` file, please remember to change it in the template (`.vue`) file too.

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
```diff
@@ -0,0 +1,70 @@
...
+      :select-names="sortBySelectNames"
+      :select-values="sortBySelectValues"
+      :placeholder="$t('User Playlists.Sort By.Sort By')"
+      @change="sortBy = $event"
+    />
+    <div class="playlists-container">
+      <ft-flex-box>
+        <ft-playlist-selector
+          v-for="(playlist, index) in activePlaylists"
+          :key="`${playlist._id}-${index}`"
```
absidue, line 38: If playlists have unique `_id`s, you can remove the index from the key.

src/renderer/components/ft-playlist-selector/ft-playlist-selector.js
```diff
@@ -0,0 +1,75 @@
...
+    titleForDisplay: function () {
+      if (typeof this.title !== 'string') { return '' }
+      if (this.title.length <= 255) { return this.title }
+
+      return `${this.title.substring(0, 255)}...`
+    },
+  },
+  mounted: function () {
+    this.parseUserData()
+  },
```
absidue, line 52:
```suggestion
  created: function () {
    this.parseUserData()
  },
```

src/renderer/components/ft-prompt/ft-prompt.js
```diff
@@ -64,6 +67,9 @@ export default defineComponent({
     })
     this.focusItem(0)
   },
+  destroyed() {
+    this.lastActiveElement?.focus()
+  },
```
absidue, line 72: `destroyed` is called after the component instance has been destroyed, so `this` is no longer available at that point. `beforeDestroy` is called just before the component instance is destroyed, so `this` is still available.
```suggestion
  beforeDestroy() {
    this.lastActiveElement?.focus()
  },
```

src/renderer/components/ft-share-button/ft-share-button.js
```diff
@@ -62,6 +62,19 @@ export default defineComponent({
       return this.$store.getters.getCurrentInvidiousInstance
     },
 
+    selectedUserPlaylist: function () {
+      if (this.playlistId == null) { return null }
+      if (this.playlistId === '') { return null }
```
absidue, line 67: These two `if`s can be combined into a single `if`.
```suggestion
      if (this.playlistId == null || this.playlistId === '') { return null }
```

src/renderer/components/watch-video-info/watch-video-info.js
```diff
@@ -268,15 +256,32 @@ export default defineComponent({
...
+      switch (format) {
+        case 'dash':
+          this.$parent.enableDashFormat()
+          break
+        case 'legacy':
+          this.$parent.enableLegacyFormat()
+          break
+        case 'audio':
+          this.$parent.enableAudioFormat()
+          break
```
absidue, line 284: This function was removed a while ago (the template uses `$emit`, instead of calling those functions with `$parent`), I'm guessing this was because development was merged into here, instead of this branch being rebased on top of development.

src/renderer/views/Playlist/Playlist.js
```diff
@@ -138,8 +219,6 @@ export default defineComponent({
     },
 
     getPlaylistInvidious: function () {
-      this.isLoading = true
-
       invidiousGetPlaylistInfo(this.playlistId).then((result) => {
         this.infoData = {
```
absidue, line 223: `infoData` doesn't seem to be used anymore, so any code relating to it should probably be removed.

src/renderer/views/Playlist/Playlist.js
```diff
@@ -59,20 +72,78 @@ export default defineComponent({
...
+    },
+    playlistId: function() {
+      return this.$route.params.id
+    },
+    userPlaylistsReady: function () {
+      return this.$store.getters.getPlaylistsReady
+    },
+    selectedUserPlaylist: function () {
+      if (this.playlistId == null) { return null }
+      if (this.playlistId === '') { return null }
```
absidue, line 84: Can be combined to a single `if`.
```suggestion
      if (this.playlistId == null || this.playlistId === '') { return null }
```

absidue (commented):
Haven't reviewed everything yet but here is an initial code review.

Some of the things I noticed that I didn't leave individual comments on, is that especially in the CSS and SCSS files, the newly added lines don't have any whitespace between them, which makes them hard to read.

Also please go through your changes and make sure you aren't reintroducing code that was removed ages ago, I noticed one case in `watch-video-info` but there might be more.

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -79,49 +164,144 @@ export default defineComponent({
...
+      // Only online playlists can be shared
+      if (this.infoSource === 'user') { return false }
+
+      // Cannot delete protected playlist
+      return !this.hideSharingActions
+    },
+  },
+  mounted: function () {
+    this.newTitle = this.title
+    this.newDescription = this.description
```
absidue, line 188:
```suggestion
  created: function () {
    this.newTitle = this.title
    this.newDescription = this.description
```

PikachuEXE:
Changes are made
The new files are all from the existing old playlist branch so they contain old code
Please review them harder

src/renderer/store/modules/playlists.js
```diff
@@ -1,32 +1,83 @@
...
+      _id: 'favorites',
     },
     {
-      playlistName: 'WatchLater',
-      protected: true,
-      removeOnWatched: true,
-      videos: []
-    }
-  ]
+      playlistName: 'Watch Later',
```
jasonhenriquez, line 31: Discussion on this: It seems like the current Favorites functionality in FreeTube today is the equivalent of a Watch Later. Why is this being separated in this manner? More broadly, I want to pick your brain on what your plans for this are. My top concerns for this PR are:
1. Maintaining reverse compatibility for users who download the new version but want to revert for whatever reason
2. Adding back all of the removed instances of the favorite icon in the JS, template, and CSS files with a functionally equivalent or better Add to Watch Later button. I think having a release without this present would be a strong negative change, as our dedicated users are the ones who have adapted to saving videos in this one-click way we have set up. Because of the sheer importance of this, I want to make sure that your plans for this are made clear for both the sake of discussion and for allowing other devs to follow through with this in a way that's aligned with your vision if you were unable to complete development on this feature.

jasonhenriquez:
Not sure if this is pertinent, but changing from this branch to `development` is the first time I had to delete `dist` and `node_modules` in order to prevent `npm run dev` from failing with an error. It was persisting as an issue even after switching back to this branch again and clearing my `Electron` folder.

absidue:
When you go to add a video to a playlist, in the prompt hovering over the playlist shows the normal cursor instead of the pointer/hand, which you would get when you hover over a button or a link, it would be nice if the cursor changed, so it's clear that you can click on the playlists.

src/renderer/store/modules/playlists.js
```diff
@@ -64,13 +170,118 @@ const actions = {
...
+            DBPlaylistHandlers.upsert(playlist)
+          }
+        })
+
+        const findFavorites = payload.filter((playlist) => {
+          return playlist.playlistName === 'Favorites' || playlist._id === 'favorites'
+        })
+        const findWatchLater = payload.filter((playlist) => {
+          return playlist.playlistName === 'Watch Later' || playlist._id === 'watchLater'
+        })
```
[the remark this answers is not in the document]
PikachuEXE: Fixed
I blame old code :)

src/renderer/store/modules/playlists.js
```diff
@@ -1,32 +1,83 @@
...
+  return `ft-playlist--${generateRandomUniqueId()}`
+}
+
+function generateRandomPlaylistName() {
+  return `Playlist ${new Date().toISOString()}-${Math.floor(Math.random() * 10000)}`
+}
+
+function generateRandomUniqueId() {
+  return crypto.randomUUID()
+}
```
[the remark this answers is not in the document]
PikachuEXE: Fixed

src/renderer/scss-partials/_ft-list-item.scss
```diff
@@ -148,11 +148,29 @@ $watched-transition-duration: 0.5s;
...
+      block-size: fit-content;
+    }
+    .upArrowIcon {
+      font-size: 17px;
+      block-size: fit-content;
+    }
+    .downArrowIcon {
+      font-size: 17px;
       block-size: fit-content;
     }
```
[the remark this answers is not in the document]
PikachuEXE: If there are class specific style later I would re-expand this
For now it's ok to be in one

<details>

<summary>I saw style declared somewhere else like this:</summary>

```scss
.a, .b, .c {
  // common style
}

.a {
  // specific style
}
.b {
  // specific style
}
.c {
  // specific style
}
```

What I want:
```scss

.a {
  // common + specific style
}
.b {
  // common + specific style
}
.c {
  // common + specific style
}
```

</details>

src/renderer/components/watch-video-playlist/watch-video-playlist.vue
```diff
@@ -114,13 +118,13 @@
       >
         <div
           v-for="(item, index) in playlistItems"
-          :key="index"
-          :ref="currentVideoIndex === (index + 1) ? 'currentVideoItem' : null"
+          :key="`index-${item.playlistItemId || item.videoId}`"
```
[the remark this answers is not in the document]
PikachuEXE: Fixed
Totally forgot why there is a prefix

src/renderer/components/watch-video-playlist/watch-video-playlist.vue
```diff
@@ -17,27 +17,31 @@
...
-        v-if="channelId"
-        class="channelName"
-        :to="`/channel/${channelId}`"
-      >
-        {{ channelName }}
-      </router-link>
       <span
-        v-else
-        class="channelName"
+        v-if="channelName !== ''"
```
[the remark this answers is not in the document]
PikachuEXE: Fixed and TIL (maybe just forgot)

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,83 @@
...
+      if (this.playlistName === '') {
+        showToast('Playlist name cannot be empty. Please input a name.')
+        return
+      }
+
+      const nameExists = this.allPlaylists.findIndex((playlist) => {
+        return playlist.playlistName === this.playlistName
+      })
+      if (nameExists !== -1) {
+        showToast('There is already a playlist with this name. Please pick a different name.')
```
ChunkyProgrammer, line 56: Toast message needs localization

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,83 @@
...
+    // Faster to input required playlist name
+    this.$refs.playlistNameInput.focus()
+  },
+  beforeDestroy() {
+    this.lastActiveElement?.focus()
+  },
+  methods: {
+    createNewPlaylist: function () {
+      if (this.playlistName === '') {
+        showToast('Playlist name cannot be empty. Please input a name.')
```
ChunkyProgrammer, line 48: Toast message needs localization

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,83 @@
...
+      const playlistObject = {
+        playlistName: this.playlistName,
+        protected: false,
+        description: '',
+        videos: [],
+      }
+
+      try {
+        this.addPlaylist(playlistObject)
+        showToast(`Playlist ${this.playlistName} has been successfully created.`)
```
ChunkyProgrammer, line 69: Toast message needs localization

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.js
```diff
@@ -0,0 +1,83 @@
...
+      } catch (e) {
+        showToast('There was an issue with creating the playlist.')
+        console.error(e)
```
ChunkyProgrammer, line 72: Toast message needs localization

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -0,0 +1,265 @@
...
+          _id: playlist._id,
+          // Use [].concat to avoid `do not mutate vuex store state outside mutation handlers`
+          videos: [].concat(this.toBeAddedToPlaylistVideoList),
+        })
+        addedPlaylistIds.add(playlist._id)
+        // Update playlist's `lastUpdatedAt`
+        this.updatePlaylist({ _id: playlist._id })
+      })
+
+      showToast(`${this.toBeAddedToPlaylistVideoCount} video(s) added to ${addedPlaylistIds.size} playlist(s).`)
```
ChunkyProgrammer, line 237: Toast message needs localization

src/renderer/components/privacy-settings/privacy-settings.js
```diff
@@ -114,6 +115,14 @@ export default defineComponent({
       this.clearSubscriptionsCache()
     },
 
+    handleRemovePlaylists: function (option) {
+      this.showRemovePlaylistsPrompt = false
+      if (option !== 'yes') { return }
+
+      this.removeAllPlaylists()
+      showToast('All playlists have been removed')
```
ChunkyProgrammer, line 123: Toast message needs localization

src/renderer/views/Playlist/Playlist.js
```diff
@@ -210,8 +295,90 @@ export default defineComponent({
...
     },
 
+    moveVideoUp: function (videoId, playlistItemId) {
+      const playlistItems = [].concat(this.playlistItems)
+      const videoIndex = playlistItems.findIndex((video) => {
+        return video.videoId === videoId && video.playlistItemId === playlistItemId
+      })
+
+      if (videoIndex === 0) {
+        showToast('This video cannot be moved up.')
```
ChunkyProgrammer, line 305: Toast message needs localization

src/renderer/views/Playlist/Playlist.js
```diff
@@ -210,8 +295,90 @@ export default defineComponent({
...
+        protected: this.selectedUserPlaylist.protected,
+        description: this.playlistDescription,
+        videos: playlistItems,
+        _id: this.playlistId
+      }
+      try {
+        this.updatePlaylist(playlist)
+        this.playlistItems = playlistItems
+      } catch (e) {
+        showToast('There was an issue with updating this playlist.')
```
ChunkyProgrammer, line 325: Toast message needs localization

src/renderer/views/Playlist/Playlist.js
```diff
@@ -210,8 +295,90 @@ export default defineComponent({
...
+    },
+
+    moveVideoDown: function (videoId, playlistItemId) {
+      const playlistItems = [].concat(this.playlistItems)
+      const videoIndex = playlistItems.findIndex((video) => {
+        return video.videoId === videoId && video.playlistItemId === playlistItemId
+      })
+
+      if (videoIndex + 1 === playlistItems.length || videoIndex + 1 > playlistItems.length) {
+        showToast('This video cannot be moved down.')
```
ChunkyProgrammer, line 337: Toast message needs localization

src/renderer/views/Playlist/Playlist.js
```diff
@@ -210,8 +295,90 @@ export default defineComponent({
...
+        protected: this.selectedUserPlaylist.protected,
+        description: this.playlistDescription,
+        videos: playlistItems,
+        _id: this.playlistId
+      }
+      try {
+        this.updatePlaylist(playlist)
+        this.playlistItems = playlistItems
+      } catch (e) {
+        showToast('There was an issue with updating this playlist.')
```
ChunkyProgrammer, line 357: Toast message needs localization

src/renderer/views/Playlist/Playlist.js
```diff
@@ -210,8 +295,90 @@ export default defineComponent({
...
+        this.removeVideo({
+          _id: this.playlistId,
+          videoId: videoId,
+          playlistItemId: playlistItemId,
+        })
+        // Update playlist's `lastUpdatedAt`
+        this.updatePlaylist({ _id: this.playlistId })
+        showToast('Video has been removed')
+      } catch (e) {
+        showToast('There was a problem with removing this video')
```
ChunkyProgrammer, line 373: Toast message needs localization

src/renderer/views/Playlist/Playlist.js
```diff
@@ -210,8 +295,90 @@ export default defineComponent({
...
+    removeVideoFromPlaylist: function (videoId, playlistItemId) {
+      try {
+        this.removeVideo({
+          _id: this.playlistId,
+          videoId: videoId,
+          playlistItemId: playlistItemId,
+        })
+        // Update playlist's `lastUpdatedAt`
+        this.updatePlaylist({ _id: this.playlistId })
+        showToast('Video has been removed')
```
ChunkyProgrammer, line 371: Toast message needs localization

PikachuEXE:
@ChunkyProgrammer 
All toast messages (playlist related) should be translated now

Also fixed old code for "remove watched videos" which was broken by https://github.com/FreeTubeApp/FreeTube/pull/4017

absidue:
Yep, when you maintain a separate branch for that long, you need to stay on top of all changes that are made to the original branch and replicate them in your branch.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
This is kinda unfortunate


https://github.com/FreeTubeApp/FreeTube/assets/73130443/dafea57e-7f8f-4834-b3ac-04de2dde4a08

absidue:
Blame YouTube, we don't know that the continuation is going to be empty until we request it and get an empty response.

PikachuEXE:
That's related to YT returning fake continuation token

https://github.com/FreeTubeApp/FreeTube/pull/4102

But to fix the remote playlist issue you mentioned here requires eager loading all items in remote playlist which is probably not what we want (and it's also not caused by this PR, checked `development`)

src/renderer/views/UserPlaylists/UserPlaylists.vue
```diff
@@ -1,47 +1,61 @@
...
-        <ft-tooltip
-          class="selectTooltip"
-          position="bottom"
-          :tooltip="$t('User Playlists.Playlist Message')"
+      <div class="heading">
+        <h2 class="headingText">
+          {{ $t("User Playlists.Your Playlists") }}
+        </h2>
+        <ft-icon-button
+          title="Create New Playlist"
```
ChunkyProgrammer, line 16: This should be translated

src/renderer/components/playlist-info/playlist-info.vue
```diff
@@ -68,11 +98,71 @@
...
+      <ft-prompt
+        v-if="showDeletePlaylistPrompt"
+        :label="$t('User Playlists.Are you sure you want to delete this playlist? This cannot be undone.')"
+        :option-names="deletePlaylistPromptNames"
+        :option-values="deletePlaylistPromptValues"
+        @click="handleDeletePlaylistPromptAnswer"
+      />
+      <ft-prompt
+        v-if="showRemoveVideosOnWatchPrompt"
+        :label="$t('User Playlists.Are you sure you want to remove all watched videos from this playlist? This cannot be undone.')"
```
ChunkyProgrammer, line 162: the . at the end is causing issues when called like this. You can either remove the . from the key or do the following change
```suggestion
        :label="$t('User Playlists[\'Are you sure you want to remove all watched videos from this playlist? This cannot be undone.\']')"
```

src/renderer/components/playlist-info/playlist-info.vue
```diff
@@ -68,11 +98,71 @@
...
+          v-if="sharePlaylistButtonVisible"
+          :id="id"
+          :dropdown-position-y="description ? 'top' : 'bottom'"
+          share-target-type="Playlist"
+        />
+      </div>
+
+      <ft-prompt
+        v-if="showDeletePlaylistPrompt"
+        :label="$t('User Playlists.Are you sure you want to delete this playlist? This cannot be undone.')"
```
ChunkyProgrammer, line 155: same issue with `.` at the end

src/renderer/store/modules/playlists.js
ChunkyProgrammer: @PikachuEXE thoughts on these two points ?(in case you missed the notification)

src/renderer/components/playlist-info/playlist-info.vue
PikachuEXE: Fixed, hard to spot thx

src/renderer/store/modules/playlists.js
PikachuEXE: > It seems like the current Favorites functionality in FreeTube today is the equivalent of a Watch Later.

It might be used that way (or not, since I don't use it as watch later ever), since there is **only one playlist** before this PR.

1. Reverting is gonna be moving (copying is fine) all videos into **the first playlist** (well current version of FT just display the list of videos of first playlist, you can check exported playlist files). It's not automatic but if you want a version which has automatic revert function we will need to release a version just for that (before releasing a version with this PR)
2. I would NOT add more functionality to this PR (which is already a complicated one). The current UX is based on having **one playlist** only and I would not consider that is good for UX with privacy (over one-click features). Feel free to develop and propose another PR if anyone is willing to.

src/renderer/components/data-settings/data-settings.js
```diff
@@ -882,87 +882,116 @@ export default defineComponent({
...
         'title',
         'author',
         'authorId',
-        'published',
         'lengthSeconds',
         'timeAdded',
-        'isLive',
-        'type',
+
+        // `playlistItemId` should be optional for backward compatibility
```
ChunkyProgrammer, line 917: Shouldn't this be added to optionalKeys then? Or is there a reason it wasn't added?

PikachuEXE: If you look at the video import code, the whole video object is pushed as long as all the required keys are present (not whitelisting properties)

src/renderer/store/modules/playlists.js
```diff
@@ -1,32 +1,83 @@
...
+      _id: 'favorites',
     },
     {
-      playlistName: 'WatchLater',
-      protected: true,
-      removeOnWatched: true,
-      videos: []
-    }
-  ]
+      playlistName: 'Watch Later',
```
ChunkyProgrammer, line 31: Renaming 'Watch Later' to 'WatchLater' imported an empty playlist when importing an old version of FreeTube's playlists
image

I also wonder if we should keep these default playlists protected. IMO it could be a good idea to keep Watch Later and Favorites as protected playlists, then we can also add translations to them

PikachuEXE: You can reintroduce "default special playlists" later
Right now if you have protected, that means more functions & special playlist treatment need to be added and tested (which might still look incomplete due to playlist translation?)
Having an empty `WatchLater` (no space) playlist is an one time issue and I don't want to exclude that playlist (again that would be special playlist treatment)

PikachuEXE: Also the idea of supporting multiple playlists is to have multiple "favorites" & "watch later" playlists (however they want to use their created playlist, as long term or temporary bookmark is their thing, we just need to enable them to do both). Until you can come up with features linked to "special playlists", those playlists shouldn't even be introduced, not to say protected.

jasonhenriquez: The main existent feature linked to a special playlist for Favorites is maintaining reverse compatibility regardless of user actions, no? That, and signalling to the user that those playlists do possess some special importance (reverse compatibility now, Watch Later / Favorites shortcut down the road). + Allows us to control the (translated) labels for these playlists at any point down the road for all users.

PikachuEXE: Keeping non special playlists protected instead of allowing user to opt-in and choose the playlist(s) for later introduced features like "quick save video for purpose of long term bookmark and/or watch later"?

jasonhenriquez: I personally think customizing which playlist(s) the Watch Later button is for is a more debatable and lower priority feature than having one fixed & designated one. In any case, having one playlist that occupies the special position of representing legacy Favorites and one that occupies the special position of Watch Later (note: I'm fine with these two being the same) that are both protected is preferable. It provides a sense of constancy to the user and emphasizes their specialness for both reverse compatibility and Watch Later. The implementation details of the power user benefit of being able to delete and rename them is IMO not that important and can be revisited for when we do implement customizing which playlist(s) the Watch Later button is for.

PikachuEXE: If anyone have developed/is developing the quick save video feature, I am fine it's not selectable at the beginning but the special playlist should be able to be hidden (when the feature is disabled).

Reverse compatibility can be done IMO via either by instructions in document (move all videos to one playlist) or by code (not sure how but that means another release required before this is merged).

Either way there is still no place for special playlists in this PR.

jasonhenriquez:
Sorry about that; merged into the wrong branch a short time back & had some issues with fixing the history. Should be fine now, even if ugly. Not been having the best week. 🤦

PikachuEXE:
Force pushed with new commits from my local to fix it

jasonhenriquez (commented):
**issue (unsure if blocking):** when reverting to the development branch, the playlist feature is broken (can't remove any of my favorites). This is going to be a notable issue for the dozens or hundreds of people who choose to load up the new version and revert to an earlier one at some point. They are likely going to be very confused about this. Do we have the upgrade / downgrade issues catalogued in any created issue? Because I worry this could be a big problem.

Our options are to try to fix this, if possible, by making the conversion process more reverse compatible, or add documentation in the release notes and/or in the code itself notifying users that this will occur, and what they need to do to fix it (create a backup of their settings before/after (?), then delete your `settings.db` (?)). But this might just be a blocking issue.

jasonhenriquez:
**issue (non-blocking, minor):** sort order broken once search is in use on the `ft-playlist-add-video-prompt`

Screenshot_20231118_215226

src/renderer/scss-partials/_ft-list-item.scss
```diff
@@ -148,11 +148,20 @@ $watched-transition-duration: 0.5s;
...
+      display: grid;
+      grid-auto-flow: column;
+      justify-content: flex-end;
+    }
+    .addToPlaylistIcon,
+    .trashIcon,
+    .upArrowIcon,
+    .downArrowIcon {
+      font-size: 17px;
       block-size: fit-content;
```
jasonhenriquez, line 165:
```suggestion
      justify-content: flex-end;
      block-size: fit-content;
    }
    .addToPlaylistIcon,
    .trashIcon,
    .upArrowIcon,
    .downArrowIcon {
      font-size: 17px;
```
This bug was re-introduced. Here's the fix.

Screenshot_20231118_220012

jasonhenriquez: (Test in your own branch first - I tend to make minor linting mistakes on the Web UI - and you can get rid of the `block-size: fit-content;` on line 165 as well)

src/renderer/components/playlist-info/playlist-info.vue
```diff
@@ -19,19 +22,46 @@
...
       </h2>
       <p>
-        {{ videoCount }} {{ $t("Playlist.Videos") }} - <span v-if="!hideViews">{{ viewCount }} {{ $t("Playlist.Views") }} -</span>
+        {{ videoCount }} {{ $t("Playlist.Videos") }}
+        <span v-if="!hideViews && infoSource !== 'user'">
+          - {{ viewCount }} {{ $t("Playlist.Views") }}
+        </span>
+        <span>- </span>
         <span v-if="infoSource !== 'local'">
           {{ $t("Playlist.Last Updated On") }}
```
jasonhenriquez, line 48: **nitpick:** this text should be lowercase: `0 Videos - Last Updated On Nov 18, 2023` -> `0 videos - last updated on Nov 18, 2023`. Looks odd in all-caps & out of place with our other similar labels.

src/renderer/components/playlist-info/playlist-info.vue
```diff
@@ -68,11 +98,71 @@
...
+          v-if="videoCount > 0 && showPlaylists && !editMode"
+          :title="$t('User Playlists.Copy Playlist')"
+          :icon="['fas', 'copy']"
+          theme="secondary"
+          @click="toggleCopyVideosPrompt"
+        />
+        <ft-icon-button
+          v-if="!editMode && infoSource === 'user' && videoCount > 0"
+          :title="$t('User Playlists.Remove Watched Videos')"
+          :icon="['fas', 'eye-slash']"
```
jasonhenriquez, line 134: **issue (non-blocking):** The purpose of this button is incredibly unclear by looking at the icon. This should be less of an issue once we have icon labels, but please see if you can find a better one.

**issue (non-blocking):** On top of the previous issue, this is a destructive action, which we should have clearly marked. In the design world, destructive controls are colored red to clearly communicate to the user that they have a lasting destructive impact. The prompt will hopefully stop most people from making a mistake, but it's still a notable issue. My suggestion is to Implement a color override property on the `ft-icon-button` component and use enforce a red color here (e.g., the red primary color variable and the primary-text-color for that).

src/renderer/views/Playlist/Playlist.scss
```diff
@@ -30,6 +33,30 @@
...
+  }
+  // Prevent link click
+  :deep(.ft-list-item .videoThumbnail .thumbnailLink) {
+    // https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events
+    pointer-events: none;
+  }
+}
+.playlistItem-enter, .playlistItem-leave-to {
+  opacity: 0;
+  transform: translateX(10%);
```
jasonhenriquez, line 57:
```suggestion
  translateX(calc(10% * var(--horizontal-directionality-coefficient))
```

(disclaimer: validate in local before trying; I sometimes make mistakes typing in the GH UI)

static/locales/en-US.yaml
```diff
@@ -137,8 +137,78 @@ User Playlists:
     videos currently here will be migrated to a 'Favorites' playlist.
   Your saved videos are empty. Click on the save button on the corner of a video to have it listed here: Your saved videos are empty. Click on the save button on the corner of a video to have
     it listed here
+  You have no playlist. Click on the create new playlist button to create a new one.: You have no playlist. Click on the create new playlist button to create a new one.
```
jasonhenriquez, line 140:
```suggestion
  You have no playlists. Click on the create new playlist button to create a new one.: You have no playlists. Click on the create new playlist button to create a new one.
```

static/locales/en-US.yaml
```diff
@@ -331,6 +401,9 @@ Settings:
     Are you sure you want to remove all subscriptions and profiles?  This cannot be undone.: Are
       you sure you want to remove all subscriptions and profiles?  This cannot be
       undone.
+    Remove All Playlists: Remove All Playlists
+    All playlists have been removed: All playlists have been removed
+    Are you sure you want to remove all your playlists?: Are you sure you want to remove all your playlists?
```
jasonhenriquez, line 406:
```suggestion
    Are you sure you want to remove all of your playlists?: Are you sure you want to remove all of your playlists?
```

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -0,0 +1,268 @@
...
+      })
+
+      // Focus back to search input
+      // Allow search and easier deselecting new created playlist
+      this.$refs.searchBar.focus()
+    },
+
+    createPlaylistPromptShown(value) {
+      // Lock interactions when create playlist prompt opened
+      this.interactionsLocked = value
```
jasonhenriquez, line 182: **question:** For what purpose are we locking the search bar? Could you document it in the code?

jasonhenriquez: Also, it seems to be just making it non-keyboard-focusable, not disabling it. If you want to disable it, you can use the `disabled` prop on `ft-input`.

jasonhenriquez: **suggestion (blocking):** If you do end up needing to keep this in as is, please replace the `inputTabindex` `number` prop with a `boolean` `isInputKeyboardNavigable`. That way it's less declarative and has no need of handling odd inputs.

static/locales/en-US.yaml
PikachuEXE: Any reason to use `all of`?
If it feels inconsistent with `Are you sure you want to remove all subscriptions and profiles?` I can remove `your` (not sure why I add `your`)

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
PikachuEXE: Code removed due to the issue I experienced at early stage of dev seems no longer reproducible

src/renderer/views/Playlist/Playlist.scss
PikachuEXE: Seems missing `transform: ` prefix, and semicolon at the end...
Anyway I updated and validated it locally but you should check again

src/renderer/components/playlist-info/playlist-info.vue
PikachuEXE: https://fontawesome.com/search?q=slash&o=r&m=free 
Not that many icons have slash, help me find a better icon I guess ._.

PikachuEXE: Color is updated, I kind of waiting for any of you to comment on it

src/renderer/components/playlist-info/playlist-info.vue
PikachuEXE: Please perform case related text update for **existing** translation entries in a separate PR

src/renderer/scss-partials/_ft-list-item.scss
PikachuEXE: fixed

static/locales/en-US.yaml
PikachuEXE: fixed

PikachuEXE:
> sort order broken once search is in use on the ft-playlist-add-video-prompt

Fixed, got old code for sorting not removed (which was only run for filtered result only)

PikachuEXE:
For "can't remove any of my favorites", please see https://github.com/FreeTubeApp/FreeTube/pull/4353

src/renderer/components/playlist-info/playlist-info.vue
jasonhenriquez: Yeah, I don't see anything else better in the `fontawesome` list by itself ([without using layers, at least](https://origin.fontawesome.com/docs/web/style/layer)), so that's fine.

jasonhenriquez: The color is updated to use the primary color, not red, which is different than what I was intending. Main point is that forcing red in particular endows it with special meaning (for everyone else except someone with a `Secondary Color Theme` of `Red`, I guess, which you can mitigate by using a darker red.

src/renderer/views/Playlist/Playlist.scss
jasonhenriquez: Thank you lol. Tested by adding/removing in English and Arabic. Looks good now!

static/locales/en-US.yaml
jasonhenriquez: I personally think it sounds more formal, but after checking online, [English is weird and both are completely fine](https://english.stackexchange.com/questions/80412/is-of-necessary-in-all-of). Fine to close this one if you want.

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
jasonhenriquez: Thanks! Can you also remove the `input-tabindex` prop from `ft-input`s?

PikachuEXE: Just did

src/renderer/components/playlist-info/playlist-info.vue
PikachuEXE: I am following https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/privacy-settings/privacy-settings.vue#L43-L62
I think it's better to have later PRs to do this kind of minor changes which probably also affects other places

PikachuEXE:
For code to allow user to export playlist for older versions of FT, see
https://github.com/PikachuEXE/FreeTube/tree/feature/playlist-2023-05-w-old-ft-consumable-playlist-export

### Testing
- Checkout this PR, add some playlists and videos, probably also removed the default playlists
- Export playlists as usual (also as backup file for different tests
- Export playlists with new toggle on
- Remove all playlists
- Switch back to `development`, import the export with new toggle on
- Confirm all videos from all playlists are present

image

I prefer not to put it into this PR (already a mega thread...)

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Found errors:

https://github.com/FreeTubeApp/FreeTube/assets/73130443/1bbe0137-3dc8-45e8-b753-7f32c52bc569

[Should we mark these destructive buttons red?](https://github.com/FreeTubeApp/FreeTube/issues/1815)


https://github.com/FreeTubeApp/FreeTube/assets/73130443/ccabc2ee-38c5-4571-9d88-2f5f4af8ab19

PikachuEXE:
Vue warning fixed

For button color please see my previous reply in https://github.com/FreeTubeApp/FreeTube/pull/4234#discussion_r1398640085

src/renderer/components/ft-button/ft-button.js
```diff
@@ -18,6 +18,11 @@ export default defineComponent({
     id: {
       type: String,
       default: ''
-    }
+    },
```
jasonhenriquez, line 21: **question:** Why was this added?

src/renderer/components/ft-button/ft-button.js
```diff
@@ -18,6 +18,11 @@ export default defineComponent({
...
+    tabindex: {
+      type: Number,
+      // Relies on template to not output tabindex on values < -1
+      default: -2,
+    },
```
jasonhenriquez, line 26: **question:** Why was this added?

src/renderer/components/playlist-info/playlist-info.vue
jasonhenriquez: Sounds good, I've placed this into its own issue

src/renderer/components/ft-button/ft-button.js
PikachuEXE: Probably part of the workaround I introduced before
Let me see if they can be removed

PikachuEXE: Removed
Probably wanted to set as `-1` at some point as workaround to some issue which can no longer be reproduced
Forgot what issue

PikachuEXE:
Just merged https://github.com/PikachuEXE/FreeTube/pull/62 into this PR

Will copy stuff from it but to see the diff please visit that PR

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Everything LGTM but i will leave it to 3 others to give the approval because i helped tested this feature since the beginning so i could overlook bunch of stuff

jasonhenriquez (approved):

jasonhenriquez:
@absidue @ChunkyProgrammer

ChunkyProgrammer (dismissed):

PikachuEXE:
Just resolve conflict with https://github.com/FreeTubeApp/FreeTube/pull/4376 on thumbnail style (block for `.thumbnailImage` got changed, and to resolve the conflict I brought the old block back but without the `inline-size`)

Thumbnail (especially playlist thumbnail with placeholder image when empty) should still look fine

image
image

PikachuEXE (commented):
From @absidue 

> Not a fan of unconditionally delaying the loading of components like the watch page until local playlists are loaded, even when you aren't watching a video with any kind of playlist. 
Also the current implementation checks if the id is a local playlist by checking if it exists in the local playlists store. 
That could be done with a simple url parameter (playlistType=user, default if the url parameter wasn't passed is that it is a remote playlist), that way you would only have to delay the loading if the url parameter says it's a local playlist, instead of also delaying when it's remote playlist.

> Basically you can only get to a watch page with a local playlist, if you came from a page that already knows it's a local playlist, so you can just pass that information on in a url parameter, instead of having to detect it from scratch again. 
That also solves the problem of local playlist's having an id that potentially matches a remote one.

PikachuEXE:
Updated Watch view to only delay loading when user playlist requested
Also fixed single playlist view cover image link not passing playlist type for user playlist

Tested with
- video with no playlist
- video with remote playlist
- video with user playlist

efb4f5ff-1298-471a-8973-3d47447115dc:
Stock thumbnail should not be clickable because it cant take you to the first vid in the list.

https://github.com/FreeTubeApp/FreeTube/assets/73130443/05365669-19dc-40c9-9923-350bbb5817cd

efb4f5ff-1298-471a-8973-3d47447115dc:
If i hit enter instead of the create button im able to create mutiple playlists

https://github.com/FreeTubeApp/FreeTube/assets/73130443/c2921885-31b7-4b6a-9e4d-d7db86df7791

PikachuEXE:

I copied https://youtube.com/playlist?list=PLm4EwY0oINs9jnQg_xvMz1nfJJ_8TAfcS to a local playlist **twice** and see no such warning in that playlist
Please provide steps to reproduce

Will address other stuff later

PikachuEXE:
> If i hit enter instead of the create button im able to create multiple playlists

The `Create New Playlist` button receive a click event (`PointerEvent` to be precise) when enter is pressed on input box in create new playlist prompt input box
No idea how to fix this
image

PikachuEXE:
> Clear input field icon isnt centered anymore

I think one of those input tag related change might have done something
Anyway fixed

image

PikachuEXE:
> If i randomly click somewhere in the add to playlist window and press enter it closes the window

I think this is intentional
The code is from old branch so if you guys think this behaviour is unwanted I can remove it easily

image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
> Got this vue warning when opening up my playlist. I think this happened because i copied the remote playlist twice into the local one



https://github.com/FreeTubeApp/FreeTube/assets/73130443/41e3f767-e2d9-4666-814a-736cf75c245f



> Stock thumbnail should not be clickable because it cant take you to the first vid in the list.

Fixed

> If i hit enter instead of the create button im able to create multiple playlists

It still allows the user to do this a few times when the user doesnt have playlists 



https://github.com/FreeTubeApp/FreeTube/assets/73130443/df4dad6c-902a-44e8-bffc-d09c2d32f78e



> This button button color fades in and out kind of weirdly

Its not a bug that the button changes color from blue to red and red to blue but the bug is that u see the transition of the button color
 
> Clear input field icon isnt centered anymore

Also here not centered 

Capture

New bug: Add a video from channel page video tab to your playlist -> You will see the description of the video in your playlist. Add the same video to your playlist but this time the video originates from a playlist -> You will see that there is no description of the video. Expected behavior is that the description is always hidden.

Capture

PikachuEXE:
> Got this vue warning when opening up my playlist. I think this happened because i copied the remote playlist twice into the local one

Got a fix but might not be error proof fix (but solves your case)

PikachuEXE:
> It still allows the user to do this a few times when the user doesnt have playlists

No idea how to fix this help

PikachuEXE:
> It still allows the user to do this a few times when the user doesnt have playlists

To workaround this issue, I have changed the search input box on add video to playlist prompt to be always visible
Not a good UX but still better than having that issue

> This button button color fades in and out kind of weirdly

Please test on existing `ft-icon-button` with same theme to see if it happens too, I don't think a new button theme/style is introduced here

PikachuEXE:
> New bug: Add a video from channel page video tab to your playlist -> You will see the description of the video in your playlist. Add the same video to your playlist but this time the video originates from a playlist -> You will see that there is no description of the video. Expected behavior is that the description is always hidden.

Fixed, along with view count

> you can navigate to the deleted playlist

Replaced with toast message + loading screen
image

> Also here not centered

Fixed

efb4f5ff-1298-471a-8973-3d47447115dc:
> Got this vue warning when opening up my playlist. I think this happened because i copied the remote playlist twice into the local one 

Still getting this one with the same steps as seen in the video

>  If i hit enter instead of the create button im able to create multiple playlists
>>It still allows the user to do this a few times when the user doesnt have playlists

Fixed



> Clear input field icon isnt centered anymore
>>Also here not centered

Fixed

> New bug: Add a video from channel page video tab to your playlist -> You will see the description of the video in your playlist. Add the same video to your playlist but this time the video originates from a playlist -> You will see that there is no description of the video. Expected behavior is that the description is always hidden.

Fixed

>    you can navigate to the deleted playlist
>> Replaced with toast message + loading screen

Im still seeing the behavior of the video

**New Bug:** search suggestions blocked by thumbnail

Capture

PikachuEXE:
> Still getting this one with the same steps as seen in the video

Real fix applied (I know why it happens now see commit message for details

efb4f5ff-1298-471a-8973-3d47447115dc:
> Got this vue warning when opening up my playlist. I think this happened because i copied the remote playlist twice into the local one 
>> Still getting this one with the same steps as seen in the video

> you can navigate to the deleted playlist
>> Replaced with toast message + loading screen
>>>Im still seeing the behavior of the video

> New Bug: search suggestions blocked by thumbnail

All fixed. 

Not sure if the loading screen is the way to go in terms of providing a good UX. Maybe a blank page would be better but i would like the others to chime into that.

New Bug: Reverse playlist throws error


https://github.com/FreeTubeApp/FreeTube/assets/73130443/e2ca4d3b-e8ce-456a-a3c0-60c42590a3ad

PikachuEXE:
404 view for user playlist can be updated in later PRs (it's an edge case)

> Reverse playlist throws error

Fixed by creating a new array due to `this.playlistItems` can be arrays from data store state
Which affects **both** cached playlist (local API I guess) and user playlist
[`Array#reverse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse) does mutation in place (there is also [`toReversed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) but `reverse` browser compatibility is much better)

efb4f5ff-1298-471a-8973-3d47447115dc:
All seems fine now!

PikachuEXE:
Anything else to be fixed/changed?

Just merged `development` and [started a build](https://github.com/PikachuEXE/FreeTube/actions/runs/7148274043) in case those can't compile want to test 
(or [look for latest build for this branch as a new one might be created later](https://github.com/PikachuEXE/FreeTube/actions/workflows/build.yml?query=branch%3Afeature%2Fplaylist-2023-05))
**Always** use non official releases with backup, or even the official releases

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.css
```diff
@@ -0,0 +1,28 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    You should have received a copy of the GNU General Public License
+    along with FreeTube.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+/*
+* Credit goes to pavelvaravko for making this css.
+* https://codepen.io/pavelvaravko/pen/qjojOr
+*/
```
absidue, line 18: Is this credit message still relevant?

senntore:
@PikachuEXE I tested coping playlists with huge number of videos(something like 50+ & 100+) & when I click this
> Some videos in the playlist are not loaded yet. Click here to copy anyway.

almost all of the time, I was able to copy up to 100 videos in a playlist without trying to load more videos in the playlist by scrolling down. So it seems we don't need this message for playlist with less than 100 videos.
Though copying playlist with 100+ videos can be a pain as I had to scroll all the way to their last video, so that all videos can be copied.

absidue:
@senntore see this comment further up in the thread and the ones surrounding it: https://github.com/FreeTubeApp/FreeTube/pull/4234#issuecomment-1793199615

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.css
PikachuEXE: It's from old code
I guess it's some copy pasta
Probably can be removed

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.css
```diff
@@ -0,0 +1,50 @@
...
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    FreeTube is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    You should have received a copy of the GNU General Public License
+    along with FreeTube.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
```
PikachuEXE, line 14:
```suggestion
```

src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.css
```diff
@@ -0,0 +1,28 @@
...
+
+/* select starting stylings ------------------------------*/
```
PikachuEXE, line 20:
```suggestion
```

PikachuEXE (commented):
(Comment for remove comments that are absent in other current CSS files)

src/renderer/components/ft-playlist-selector/ft-playlist-selector.scss
```diff
@@ -0,0 +1,280 @@
...
+      border-radius: 50%;
+    }
+  }
+
+  .info {
+    flex: 1;
+    position: relative;
+
+    .optionsButton {
+      float: right; // ohhhh man, float was finally the right choice for something;
```
PikachuEXE, line 197:
```suggestion
      float: right;
```

src/renderer/components/ft-playlist-selector/ft-playlist-selector.scss
```diff
@@ -0,0 +1,280 @@
...
+  .info {
+    flex: 1;
+    position: relative;
+
+    .optionsButton {
+      float: right;
+    }
+
+    .externalPlayerButton {
+      float: right;
```
jasonhenriquez, line 201:
```suggestion
      float: var(--float-right-ltr-rtl-value);
```

src/renderer/components/ft-playlist-selector/ft-playlist-selector.scss
```diff
@@ -0,0 +1,280 @@
...
+      border-radius: 50%;
+    }
+  }
+
+  .info {
+    flex: 1;
+    position: relative;
+
+    .optionsButton {
+      float: right;
```
jasonhenriquez, line 197:
```suggestion
      float: var(--float-right-ltr-rtl-value);
```

PikachuEXE:
Fixed float (`yarn lint-fix` won't detect it

PikachuEXE:
Merged https://github.com/FreeTubeApp/FreeTube/pull/4437 and revert fixed in this branch
Clear text button style should still be fine

src/renderer/components/ft-playlist-selector/ft-playlist-selector.scss
```diff
@@ -0,0 +1,280 @@
+$thumbnail-overlay-opacity: 0.85;
+$watched-transition-duration: 0.5s;
```
jasonhenriquez, line 2: **issue (non-blocking):** Many of the styling rules in this file seem to not actually be matching anything. I would say it's copied from somewhere, but the styling seems to be unique in some parts for reasons I can't discern (e.g., the `videoDuration` class rules).

jasonhenriquez: If at least some of this is important but is just rote copying, my suggestion would be to create a `mixin` for that and just import it here.

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
```diff
@@ -0,0 +1,70 @@
...
+    <p class="selected-count">
+      {{ $tc('User Playlists.AddVideoPrompt.N playlists selected', selectedPlaylistCount, {
+        playlistCount: selectedPlaylistCount,
+      }) }}
+    </p>
+    <ft-input
+      ref="searchBar"
+      :placeholder="$t('User Playlists.AddVideoPrompt.Search in Playlists')"
+      :show-clear-text-button="true"
+      :clear-input-text-button-vertically-centered="true"
```
absidue, line 19:
```suggestion
```

src/renderer/components/ft-playlist-selector/ft-playlist-selector.scss
PikachuEXE: It's too early to call for style variable extraction
Since after removing unused style they become unused

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
PikachuEXE: Done

efb4f5ff-1298-471a-8973-3d47447115dc:
I do cant find any weird bug anymore. I do have a question though about the sorting of playlists.

When the user is on the playlist page, the default sorting state is Recently played. 
User changes the sorting to for example A-Z -> moves away from the playlist page -> navigate back to it -> still sorted by A-Z (which is good imo) -> user closes FT completely -> opens FT -> navigate to playlist page -> sorting is back to default state, Recently played

Q: Would it be better to always remember the last sorted state instead of default, so in this case that would mean the user gets to see sorted by A-Z when they open FT

Add to playlist prompt behaves differently as it will always default to Recently played even when you changed it

Same Q basically applies here, should we remember last set sorting state?

PikachuEXE:
I think people change sorting order to find specific playlist without filtering
But they don't find the same playlist every time
Which is why I default sorting to `latest played first` (playing last played playlist > find playlist in terms of frequency)

For add to playlist prompt let me take a look

PikachuEXE:
> Add to playlist prompt behaves differently as it will always default to Recently played even when you changed it

I doubt it when I read it (coz I code it)
It is `recently updated first` (always)

It's more common IMO to add to the same playlist, so setting `recently updated first` would be beneficial for that.

src/renderer/components/data-settings/data-settings.js
```diff
@@ -882,87 +886,116 @@ export default defineComponent({
...
-              if (hasAllKeys) {
+              if (videoObjectHasAllRequiredKeys) {
                 videoArray.push(video)
               }
             })
 
             playlistObject[key] = videoArray
+          } else if (ignoredKeys.includes(key)) {
+            // Do nothing for keys to be ignored
           } else {
```
jasonhenriquez, line 950:
```suggestion
          } else if (!ignoredKeys.includes(key)) {
            // Do nothing for keys to be ignored
```

<sub>**Obligatory disclaimer:** I am prone to lexical errors when writing in the GitHub UI</sub>

PikachuEXE: https://github.com/FreeTubeApp/FreeTube/pull/4234/commits/36e7ed883afc535d0004df0445a46da2e780ab94

jasonhenriquez:
**thoughts:** Thinking about UX considerations with copying an existing playlist. These are non-blocking and out of scope for this PR, but it is worth discussing if it can help inform our future direction.

i. This PR has a button to copy the videos in a YT playlist to a user playlist. This behavior will be possible through #4198 (Enable Video Selection Mode => Ctrl+A => Add to Playlist). Once we have that, it's worth considering if we want its own button for that, or if we want to repurpose it somehow.

ii. Continuing from i), I feel like "Saving a playlist" meaning "copy all of its videos to a user playlist" is not what a lot of people want, most of the time. To compare to YT, saving an existing YTer's playlist pretty much just adds a link to that playlist page on your sidebar. This is a pretty reasonable choice. I don't regularly want to do the equivalent of "forking" my YTer's playlist like we are allowing; rather, I want to "subscribe" to it to see what videos they've updated it with, and be able to access it with ease.

I think we at FreeTube could benefit from doing something similar here; basically just have a "Save Playlist" button that saves the link, and puts the playlist in my Playlists section. I would argue that making it immutable (as in, no removing or adding of videos, and they don't show up in the "Add to Playlist" modal), and non-renameable, are also worth doing to make this "link"/"subscription" relationship for these types of playlists even clearer to the user. It may also help to reduce the extra requests & UX annoyance of having to keep scrolling & loading more videos to "fully" copy a playlist, which is currently pretty clunky.

PikachuEXE:
You are describing subscribing/bookmarking a **remote** playlist which is totally different from managing **local** playlists.
I also don't understand "Saving a playlist" (ii) point when the button on remote playlist is labelled "Copy Playlist"
2023-12-16 15_53_44-Playlist - FreeTube

This PR is about **local** playlist only.

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -0,0 +1,258 @@
...
+          _id: playlist._id,
+          // Use [].concat to avoid `do not mutate vuex store state outside mutation handlers`
+          videos: [].concat(this.toBeAddedToPlaylistVideoList),
+        })
+        addedPlaylistIds.add(playlist._id)
+        // Update playlist's `lastUpdatedAt`
+        this.updatePlaylist({ _id: playlist._id })
+      })
+
+      showToast(this.$t('User Playlists.AddVideoPrompt.Toast["{videoCount} video(s) added to {playlistCount} playlist(s)."]', {
```
jasonhenriquez, line 234: **suggestion (non-blocking):** use `tc` and update label with pluralization in lieu of "(s)".

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -79,49 +168,160 @@ export default defineComponent({
...
+        const playlist = {
+          playlistName: this.title,
+          protected: this.selectedUserPlaylist.protected,
+          description: this.description,
+          videos: videosToWatch,
+          _id: this.id
+        }
+        try {
+          this.updatePlaylist(playlist)
+          showToast(this.$t('User Playlists.SinglePlaylistView.Toast["{videoCount} video(s) have been removed."]', {
```
jasonhenriquez, line 291: **suggestion (non-blocking):** use `tc` and update label with pluralization in lieu of "(s)".

PikachuEXE: https://github.com/FreeTubeApp/FreeTube/pull/4234/commits/e7729eecc4c88a28c0385b267ac233603eeba006

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
PikachuEXE: https://github.com/FreeTubeApp/FreeTube/pull/4234/commits/e7729eecc4c88a28c0385b267ac233603eeba006

There are 2 counts so 2 translation entries are used

jasonhenriquez (dismissed):

PikachuEXE:
Download latest build if you want to
https://github.com/PikachuEXE/FreeTube/actions/workflows/build.yml?query=branch%3Afeature%2Fplaylist-2023-05

PikachuEXE:
Just fixed a bug found when I use the build myself: playlist link in watch page not having `playlistType`

I did a search in code and there should be no other places having link to a local user playlist

[`6647099` (#4234)](https://github.com/FreeTubeApp/FreeTube/pull/4234/commits/6647099ba099093c7824dfdc9e8dbeb4b96c8de8)

jasonhenriquez (approved):

JohnBetaro:
What about the ability to go to any channel and save a whole playlist from there directly? I think this should be a necessary feature' any idea if it will be implemented? Thank you for your time.

absidue:
The "Remove Watched Videos" button is displayed inside of playlists, even when "Remember History" is disabled, is that intentional?

efb4f5ff-1298-471a-8973-3d47447115dc:
I think it is, disabling that feature applies only on future watched videos not previously watched videos

ChunkyProgrammer (approved):

absidue (approved):

Issue #568, opened 2020-10-03
[Feature Request] Add/remove channel to profile from channel/video page


ThatIsAPseudo:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
When I subscribe to a channel, I often want to add it to a profile immediately, so it's quite annoying you must go into the profile setting, look for the channel in the list, to finally add it to **one** profile.

**Describe the solution you'd like to see implemented**
There should be a dropdown list next to the subscribe button, to select in which profile(s) this channel is included.

GilgusMaximus:
Yep this is a good idea. Definitely worth to implement. However some new overlaying UI might be required for that, depending on the implementation. 
So this is probably not gonna make it from our side immediately into the next release. 

However if someone else want's to implement that, we are open for pull requests and can give you advise if needed.

PrestonN:
This isn't quite the solution you're looking for, however you can switch your active profile while you're on the channel / video page and subscribing again will add that channel to the new profile.

I do agree that this is something that should be added though.

PR #4238, opened 2023-10-27
A new way to subscribe

### LINKED ISSUE #1615: A new way of subscribe using profiles
I find it a bit counter-intuitive to subscribe using multiple profiles.
My path to subscribe to a channel consists in searching for a channel, clicking on the channel, and hitting "subscribe".
The way FreeTube is right now, leads me to end up not using the multiples profiles, because I have to go to the top-right and change profile, and then clicking subscribe, this breaks the flow of adding a subscription.

I think a good solution is to have a kind of a dropdown combined with the subscribe button so when the user clicks the 'subscribe' button, he can select multiples profiles which to add the channel.

**Screenshots**
A quick rough sketch of what it may looks like
A rough sketch

**Additional context**

jasonhenriquez:
# A New Way to Subscribe, Part II - An Even Newer Way to Subscribe: The Last of the Subscribe Buttons; A New Hope

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #1615 

## Description
Modifies the humble `ft-subscribe-button` with — dare I say it — a new way to subscribe. Click to subscribe/unsubscribe and see every profile's subscribed/unsubscribed state. Click any of those profiles to subscribe/unsubscribe them from that channel.

Does not change the default behavior of All Channels (if subscribing to a channel with another profile, All Channels automatically subscribes to it; if unsubscribing to a Channel with the All Channels profile, unsubscribe to it from all profiles). Due to its special functionality, All Channels and the currently active profile are always top of the list. Then goes unsubscribed profiles, then subscribed profiles.

## Screenshots 
Screenshot:
Screenshot_20231027_000839

Video:
https://github.com/FreeTubeApp/FreeTube/assets/84899178/039cc002-da4d-414d-8ef5-bac1e9bbf4cd

## Testing 
- Test that All Channels works as described in description (i.e., how it currently works)
- Test that profiles are successfully subscribed/unsubscribed
- Test that sort order of profiles is as described in description
- Test keyboard navigation of profile list and activation through space bar
- Test on emulated mobile view

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

## Additional context
There's an odd bug that occurs when unsubscribing to a channel for the active profile on the SubscribedChannels component that has at least one channel to the right of it: the subscribe button dropdown for the channel that takes its place is made open. In debugging this, I discovered that in no place is that second component's `isProfileDropdownOpen` ever set to true. It seems to have nothing to do with an error in the application code or even something we are able to create a workaround for. As a result, I think it's fine to pretend it's a feature for the time being.


efb4f5ff-1298-471a-8973-3d47447115dc:
Doesn't this also close https://github.com/FreeTubeApp/FreeTube/issues/568

src/renderer/components/ft-subscribe-button/ft-subscribe-button.scss
```diff
@@ -0,0 +1,146 @@
...
+    -moz-transition: background 0.2s ease-out;
+    -o-transition: background 0.2s ease-out;
+    transition: background 0.2s ease-out;
+
+    &:hover {
+      background-color: var(--side-nav-hover-color);
+      color: var(--side-nav-hover-text-color);
+      -moz-transition: background 0.2s ease-in;
+      -o-transition: background 0.2s ease-in;
+      transition: background 0.2s ease-in;
```
absidue, line 107: Are the vendor prefixes needed here? If the answer is yes, the webkit one is only used in the first group, not the second one, so that should probably be added.

src/renderer/views/SubscribedChannels/SubscribedChannels.vue
```diff
@@ -53,7 +53,7 @@
               class="unsubscribeContainer"
             >
               <ft-subscribe-button
-                class="btn"
+                :hide-profile-dropdown-toggle="false"
```
absidue, line 56: As the default value for that property is false, you can probably remove this line.

absidue (commented):
The solution for the bug you mentioned in the "Additional context" section, is to do this change in the SubscribedChannels.vue component:
```diff
<div
  v-for="channel in channelList"
-  :key="channel.key"
+  :key="channel.id"
  class="channel"
>
```

As the `.key` property doesn't exist on the channel object, we are currently passing `undefined` to `:key`, which is the same as not passing a key at all. Vue tries to reuse as much of the existing DOM as possible when doing updates, as we aren't giving it any hints on what to remove and what to keep, it ends up reusing the removed channel bubble. By adding the key, Vue then knows to invalidate and remove the node with that key, but it also means that it can keep all the nodes who's keys haven't changed.

src/renderer/components/ft-subscribe-button/ft-subscribe-button.scss
jasonhenriquez: Really great catch on the bug!! The vendor prefixes are the relic of me copying from elsewhere in the codebase; I'll remove all of these.

src/renderer/components/ft-subscribe-button/ft-subscribe-button.js
```diff
@@ -25,70 +25,99 @@ export default defineComponent({
...
+      return this.profileDisplayList.map((profile) => {
+        return profile?.name?.length > 0 ? Array.from(profile.name)[0].toUpperCase() : ''
+      })
+    },
+
     profileList: function () {
       return this.$store.getters.getProfileList
     },
 
+    /* sort by 'All Channels' -> active profile -> unsubscribed channels -> unsubscribed channels */
```
absidue, line 54: I think you meant to write subscribed in this comment, instead of unsubscribed twice :)

src/renderer/components/ft-subscribe-button/ft-subscribe-button.vue
```diff
@@ -1,12 +1,83 @@
...
+        tabindex="-1"
+        class="profileDropdown"
+      >
+        <ul
+          class="profileList"
+        >
+          <li
+            v-for="(profile, index) in profileDisplayList"
+            :id="'profile-' + index"
+            :key="'subscription-profile-' + index"
```
absidue, line 46: Could you please change the id's and key to use the profile `_id` instead of the index, also they key can be simplified to just use the `_id`, as it is just used by Vue inside this component and isn't added to the DOM.

jasonhenriquez: Yes, will do; I admittedly was just copying the pattern used by the `ft-profile-selector`. I was unsure of whether this could cause conflict if we had other elements in the DOM with the same `key` values (e.g., if we changed the `ft-profile-selector` accordingly), but I'm pretty sure that is not a major issue after checking up on the documentation.

absidue: For the `id` you should still use globally unique names, as those get placed into the DOM, but keys are only used for Vue processing in the scope of a component, so those don't have to have a unique prefix or anything like that.

jasonhenriquez: Hm... for some reason using the `_id` as key breaks `focusout`, even though it's populating properly... I am going to keep using the index + a specific prefix for the `id` and `key` (i.e., `subscription-profile-${index}` in light of this. I don't see any harm in it, and we do need unique prefixes as stated for the `id` if we're to have any other similar usage of `profile._id`s.

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
> All Channels and the currently active profile are always top of the list. Then goes unsubscribed profiles, then subscribed profiles.

This is not the case when u subscribe by clicking the subscribe button on a non All Channels profile


https://github.com/FreeTubeApp/FreeTube/assets/73130443/e41649c6-0c3f-4752-801a-341826350d65

jasonhenriquez:
> > All Channels and the currently active profile are always top of the list. Then goes unsubscribed profiles, then subscribed profiles.
> 
> This is not the case when u subscribe by clicking the subscribe button on a non All Channels profile

Not sure I see the issue. Currently active profile is shown after All Channels for convenience as well.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Apologies, i didn't read well :(

LGTM

efb4f5ff-1298-471a-8973-3d47447115dc:
Sorry i forgot to test this page

I dont think it makes sense that the dropdown expands right before the channel gets removed from the list. This behavior is good on every other instance the but on this page it seems weird to me

https://github.com/FreeTubeApp/FreeTube/assets/73130443/0da176e0-cf78-4422-b220-96e99938c096

jasonhenriquez:
That's fair - functionally inconsequential but of note - will add as chore.

PR #4242, opened 2023-10-27
add i18n-eslint-plugin

ChunkyProgrammer:
# Add i18n-eslint-plugin

## Pull Request Type
- [x] Dev enhancement

## Description
It is configured to find keys that don't exist and raw text instead of translated text in vue files

## Desktop
- **OS:** Windows
- **OS Version:** 10


.eslintrc.js
```diff
@@ -77,6 +78,18 @@ module.exports = {
     'unicorn/no-array-push-push': 'error',
     'unicorn/prefer-keyboard-event-key': 'error',
     'unicorn/prefer-regexp-test': 'error',
-    'unicorn/prefer-string-replace-all': 'error'
+    'unicorn/prefer-string-replace-all': 'error',
+    // '@intlify/vue-i18n/no-unused-keys': [
```
ChunkyProgrammer, line 82: @absidue this can be useful when removing unused keys in the future

jasonhenriquez:
It still misses things like the `More Options` in `ft-list-video`. Is this a shortcoming of the plugin?

ChunkyProgrammer:
> It still misses things like the `More Options` in `ft-list-video`. Is this a shortcoming of the plugin?

I'll see if i can configure it (it's not the html property title, it's a vue property which is why I think it's not being checked by default)
https://github.com/FreeTubeApp/FreeTube/blob/6d6fc7305909eb187b7aab37bf1aa4b63e7c5e04/src/renderer/components/ft-icon-button/ft-icon-button.js#L11

.eslintrc.js
absidue: Vue eslint plugins don't like that we have our templates and javascript in separate files, which is why they complain about false positives (tried turning on some of vue-eslint-plugin's, "no unused" rules in the past and they ended up with mostly false positives because of that, so I turned them off before I ever commited it).

ChunkyProgrammer: There doesn't seem to be too many false positives except when we access a translated value in a weird way (which we probably shouldn't be doing)

ex:
```js
    colorNames: function () {
      return this.colorValues.map(colorVal => {
        // add spaces before capital letters
        const colorName = colorVal.replaceAll(/([A-Z])/g, ' $1').trim()
        return this.$t(`Settings.Theme Settings.Main Color Theme.${colorName}`)
      })
    },
```

and
```js
      const contentType = this.$t('Age Restricted.Type.' + this.contentTypeString)
```

or if the key is in a different file like what we do for the router:
```js
    {
      path: '/settings/profile',
      meta: {
        title: 'Profile.Profile Settings'
      },
      component: ProfileSettings
    }
```

src/renderer/components/proxy-settings/proxy-settings.vue
```diff
@@ -60,16 +60,16 @@
...
-          {{ $t('Settings.Proxy Settings.Country') }}: {{ proxyCountry }}
+          {{ $t('Grammar.ColonText', {text: $t('Settings.Proxy Settings.Country')}) }} {{ proxyCountry }}
         </p>
         <p>
-          {{ $t('Settings.Proxy Settings.Region') }}: {{ proxyRegion }}
+          {{ $t('Grammar.ColonText', {text: $t('Settings.Proxy Settings.Region')}) }} {{ proxyRegion }}
         </p>
         <p>
-          {{ $t('Settings.Proxy Settings.City') }}: {{ proxyCity }}
+          {{ $t('Grammar.ColonText', {text: $t('Settings.Proxy Settings.City')}) }} {{ proxyCity }}
```
absidue, line 72: Please update the strings themselves to be `Ip: {ip}`, instead of using a generic `{text}:` construct.

PikachuEXE:
Please provide some examples on expected behaviour :S

ChunkyProgrammer:
> Please provide some examples on expected behaviour :S

The main ones are:
If a key doesn't exist then there will be a warning.
Ex: 
image

If a raw text is used then there is an error (this won't apply to attributes like class where it doesn't make sense to check)
Ex:
image
image

PikachuEXE:
Raw text in attr = OK
image

Raw text in template interpolation = OK
image

Missing Key = OK (I also tested adding new key afterward and warning disappeared)
image

PikachuEXE (commented):
The separator one seems unnecessary...
Is there no way to exclude some chars?

static/locales/en-US.yaml
```diff
@@ -928,3 +932,9 @@ Hashtag:
 Yes: Yes
 No: No
 Ok: Ok
+
+Grammar:
+  # adds a colon after a form label
+  ColonText: '{text}:'
+  # Index / Count
+  Index / Count: '{index} / {count}'
```
PikachuEXE, line 940: I think these are similar to the separators

ChunkyProgrammer: I'll remove Index / Count and I'll add a colon to the resouces that need one (French uses `  :  ` (space in front and at end) and armenian uses `՝  ` instead of `: `)

ChunkyProgrammer:
@absidue your suggestion on matrix makes sense but it seems to be working without making your suggested change
change language, color update

PikachuEXE (dismissed):

PikachuEXE (dismissed):

PikachuEXE (commented):
To avoid more conflict to https://github.com/FreeTubeApp/FreeTube/pull/4234 I am gonna approve this thing later

Also if you want checkmark translated you might want to name it something else (instead of "checkmark", I imagine a translator watching that key and got no idea how to "translate" it)

.eslintrc.js
```diff
@@ -77,6 +78,38 @@ module.exports = {
...
     'unicorn/prefer-regexp-test': 'error',
-    'unicorn/prefer-string-replace-all': 'error'
+    'unicorn/prefer-string-replace-all': 'error',
+    // '@intlify/vue-i18n/no-dynamic-keys': 'error',
+    // '@intlify/vue-i18n/no-unused-keys': [
+    //   'error',
+    //   {
+    //     extensions: ['.js', '.vue']
+    //   }
+    // ],
```
PikachuEXE, line 88: Might want to add some comment briefly explaining why they are added in commented form

ChunkyProgrammer:
Seems to be some weird issues with this PR, I'll look into it more and maybe reopen

PR #4255, opened 2023-10-30
Localize proxy test info when possible

ChunkyProgrammer:
# Localize proxy test info when possible

## Pull Request Type
- [x] Other - Localization

## Description
Currently we only get city, region and country name in English. This PR adds support for languages that are supported by the proxy test
image

## Testing 
- set locale to francais (French)
- set up proxy (ex: tor)
- click Test Button
- See info in French
image

## Desktop
- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/components/proxy-settings/proxy-settings.js
+++ b/src/renderer/components/proxy-settings/proxy-settings.js
@@ -29,7 +29,6 @@ export default defineComponent({
     return {
       isLoading: false,
       dataAvailable: false,
-      proxyTestUrl: 'https://ipwho.is/?output=json&fields=ip,country,city,region',
       proxyIp: '',
       proxyCountry: '',
       proxyRegion: '',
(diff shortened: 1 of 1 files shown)


src/renderer/components/proxy-settings/proxy-settings.js
```diff
@@ -64,6 +63,24 @@ export default defineComponent({
...
       return `${this.proxyProtocol}://${this.proxyHostname}:${this.proxyPort}`
+    },
+    lang: function() {
+      return this.$i18n.locale.replace('_', '-')
+    },
+    localeToUse: function() {
+      // locales found here: https://ipwhois.io/documentation
+      const supportedLangs = ['en', 'ru', 'de', 'es', 'pt-BR', 'fr', 'zh-CN', 'ja']
+      return supportedLangs.find(lang => {
+        return (this.lang === lang) || (this.lang.substring(0, 2) === lang.substring(0, 2))
```
jasonhenriquez, line 74: **nitpick:** we will have to split up this logic into two consecutive `find`s  if they ever end up adding support two locales with the same language code.

Aside from that, looks good.

jasonhenriquez, line 74:
```suggestion
      return supportedLangs.find(lang => this.lang === lang) ?? supportedLangs.find(lang => this.lang.substring(0, 2) === lang.substring(0, 2))
```

PikachuEXE:
LOL
image

src/renderer/components/proxy-settings/proxy-settings.js
```diff
@@ -64,6 +63,23 @@ export default defineComponent({
...
       return `${this.proxyProtocol}://${this.proxyHostname}:${this.proxyPort}`
+    },
+    lang: function() {
+      return this.$i18n.locale.replace('_', '-')
+    },
+    localeToUse: function() {
+      // locales found here: https://ipwhois.io/documentation
+      const supportedLangs = ['en', 'ru', 'de', 'es', 'pt-BR', 'fr', 'zh-CN', 'ja']
+      return supportedLangs.find(lang => this.lang === lang) ?? supportedLangs.find(lang => this.lang.substring(0, 2) === lang.substring(0, 2))
+      })
```
PikachuEXE, line 74:
```suggestion
```

PikachuEXE (changes requested):
If syntax error is fixed then it's OK
image

PikachuEXE (approved):

jasonhenriquez (approved):

absidue (approved):

PR #4261, opened 2023-10-31
IV: Fix HtmlEncoded search suggestions

ChunkyProgrammer:
# IV: Fix HtmlEncoded search suggestions

## Pull Request Type
- [x] Bugfix

## Description
The search suggestions for Invidious are html encoded so suggestions will be displayed like this when searching in Arabic + ( other languages):
image

This PR decodes the html so we can display the suggestions properly:
image

## Testing 
- set invidious as primary api
- copy some arabic text from ar.yaml and paste it into the search bar
- look at suggestions

## Desktop
- **OS:** Windows
- **OS Version:** 10


absidue:
This should really be fixed on the Invidious side, as YouTube doesn't return HTML entities which means Invidious is doing it.

ChunkyProgrammer:
Opened a PR for Invidious

PR #4264, opened 2023-10-31
Work around context menus in the devtools being displayed behind the window

absidue:
# Work around context menus in the devtools being displayed behind the window

## Pull Request Type

- [x] Bugfix

## Related issue
https://github.com/electron/electron/issues/38790

## Description
Currently Electron has a bug that makes the context menus in the dev tools show behind the devtools window, making them useless. This pull requests introduces the workaround suggested in the Electron issue, so that we get our context menus back.

## Screenshots 
The electron issue has screenshots.

## Testing 
Right click in the devtools window, if the menu shows up with this pull request it works :)

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -170,6 +170,10 @@ function runApp() {
   app.commandLine.appendSwitch('enable-file-cookies')
   app.commandLine.appendSwitch('ignore-gpu-blacklist')
 
+  // Work around for context menus in the devtools being displayed behind the window
+  // https://github.com/electron/electron/issues/38790
+  app.commandLine.appendSwitch('disable-features', 'WidgetLayering')
+
   // command line switches need to be added before the app ready event first
   // that means we can't use the normal settings system as that is asynchronous,
   // doing it synchronously ensures that we add it before the event fires


PikachuEXE:
Like this?
image

absidue:
Yes that is the correct behaviour, I'll send some before and after screenshots of how it looks on Windows 10 later.

absidue:
The pink line is the bottom border of the devtools window.
Before:
before

After:
after

jasonhenriquez (commented):
Can't replicate on Linux, but seems to be the proper fix. Will approve once we have another Windows developer sign off.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Im in the same boat as @jasonhenriquez

MarmadileManteater (approved):
I tested this change in 🪟 Windows.

lgtm

jasonhenriquez (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #2138, opened 2022-03-19
[Feature Request]: Add missing keyboard shortcuts


PAI5REECHO:
### Guidelines

- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a feature request that matches the one I want to file, without success.

### Problem Description

Several keyboard shortcuts that work on the YouTube website video player are missing: https://support.google.com/youtube/answer/7631406

### Proposed Solution

Add missing keybinds

### Alternatives Considered

None

### Issue Labels

ease of use improvement

vallode:
I compiled a list of the keyboard shortcuts provided for reference, @PAI5REECHO do you have any particular shortcuts in mind that you would prioritise?

Keyboard shortcut | Supported
-- | --
Spacebar | Yes :heavy_check_mark:
Play/Pause Media Key on keyboards | Yes :heavy_check_mark: (Tested by @smcclennon)
k | Yes :heavy_check_mark:
m | Yes :heavy_check_mark:
Stop Media Key on keyboards | Untested
Next/previous Track Media Key on keyboards | Yes merged in https://github.com/FreeTubeApp/FreeTube/pull/2239 :heavy_check_mark:
Left/Right arrow on the seek bar | Yes (Seeks 10 seconds instead of the 5 on Youtube) :white_check_mark:
j | Yes :heavy_check_mark:
l | Yes :heavy_check_mark:
. |Yes :heavy_check_mark:
, | Yes :heavy_check_mark:
\> | Yes (P also does the same) merged in #7487 :heavy_check_mark:
< | Yes (O also does the same) merged in #7487 :heavy_check_mark:
Home/End on the seek bar | Yes merged in #7487 :heavy_check_mark:
Up/Down arrow on the seek bar | Yes :heavy_check_mark:
Numbers 1 to 9 on the seek bar (not on the numeric pad) | Yes :heavy_check_mark:
Number 0 on the seek bar (not on the numeric pad) | Yes :heavy_check_mark:
/ | Yes (```ALT + D``` instead of ```/```) merged in #2140 :heavy_check_mark:
f | Yes :heavy_check_mark:
c | Yes :heavy_check_mark:
Shift+N | No (Should move to the next video. If you're using a playlist, will go to the next video of the playlist. If not using a playlist, it will move to the next YouTube suggested video) :heavy_check_mark:
[...]

PR #4282, opened 2023-11-05
Search input keyboard focus enhancements

sporeventexplosion:
# Search input keyboard focus enhancements

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Partially implements a feature requested in #2138

## Description

Select text in search input when using Ctrl-L/Meta-L or Alt-D to focus search input.
This is the behavior of the combined address/search box of both Chrome and Firefox. I think it makes sense here because the search input in the application is essentially both an address box and a search box.
    
Add new Slash key shortcut to focus input without selecting text.
This does not select the text in the input, which matches the behavior of the YouTube web frontend.

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -315,6 +315,14 @@ export default defineComponent({
       this.$refs.input.focus()
     },
 
+    isFocused() {
+      return document.activeElement === this.$refs.input
+    },
+
+    select() {
+      this.$refs.input.select()
+    },
+
     blur() {
       this.$refs.input.blur()
     },
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -310,7 +310,7 @@ export default defineComponent({
         switch (event.key) {
           case 'D':
           case 'd':
-            this.$refs.topNav.focusSearch()
+            this.$refs.topNav.focusSearch(true)
             break
         }
       }
@@ -322,8 +322,15 @@ export default defineComponent({
         case 'l':
           if ((process.platform !== 'darwin' && event.ctrlKey) ||
             (process.platform === 'darwin' && event.metaKey)) {
-            this.$refs.topNav.focusSearch()
+            this.$refs.topNav.focusSearch(true)
           }
+          break
+        case '/':
+          if (!this.$refs.topNav.isSearchFocused()) {
+            // prevent entering slash character into input
+            event.preventDefault()
+          }
+          this.$refs.topNav.focusSearch(false)
           break
       }
     },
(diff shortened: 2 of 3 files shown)


sporeventexplosion:
(note: the order of the comments is garbled. the above comments were posted during the first DST "repeated hour" for Pacific Time, which seems to have triggered a bug in GitHub's code)

absidue:
FreeTube used to have the search focus, but it was removed in #2108 solve #2019, so please don't add it back. The new keyboard shortcut could be considered, but I think it is important to keep in mind that typing the slash character may be easy on a US English keyboard layout, but for many other layouts that is not the case, so maybe a different key would be more suitable.

sporeventexplosion:
After much mucking around, I found a workaround for this behavior in Chromium: make the focus+select asynchronous by wrapping it in a `setTimeout(fn, 0)`. Now the problem is gone.

sporeventexplosion:
Would it make sense to make this an option? I don't think it makes much sense to not implement a feature that's standard in browsers just because it affects what I imagine to be a rather rare desktop environment configuration.

sporeventexplosion:
New auto-select behavior longer produces behavior that causes #2108

sporeventexplosion:
@absidue Is there still any desire to merge this?

PikachuEXE:
I can't get Alt-D to work on MacOS. (I tested my keyboard on https://keyboardchecker.com
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/altKey
image

Oh it seems a different issue
image
No idea how to fix at the moment, non blocking for me

Slash & Meta-L work

PikachuEXE (approved):

PikachuEXE:
Agree on removing slash as short cut, I can't type `/` anywhere in text box (focus always moved to search bar)

sporeventexplosion:
I agree, that should be separated form this PR.

sporeventexplosion:
Slash shortcut has been removed

PikachuEXE (dismissed):

src/renderer/components/ft-input/ft-input.js
```diff
@@ -315,6 +315,14 @@ export default defineComponent({
       this.$refs.input.focus()
     },
 
+    isFocused() {
+      return document.activeElement === this.$refs.input
+    },
+
```
absidue, line 321: Looks like this code is now unused.
```suggestion
```

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

PikachuEXE (approved):

absidue (approved):

PR #4288, opened 2023-11-06
fix: for long query throwing 413 error

### LINKED ISSUE #940: Search Box - Long query
**Behavior of the program**
After entering to Search Box a long query error messages are displayed.

**Expected behavior**
No results and no error messages. Probably the length of a search-query should be limited.

**To Reproduce**
1. Paste to a search box long query (we've tested with 150k symbols query, which is attached below)
2. Press Enter
3. See error messages

**Screenshots**
big query string

[150k query.txt](https://github.com/FreeTubeApp/FreeTube/files/5824298/150k.query.txt)

**Environment Information:**
 - OS: Windows
 - OS Version: Windows 10
 - FreeTube version: 0.11.1
 - Installation Method (If applicable): .exe
 - Primary API Used: [Local API, Invidious API]

**Additional context**
Our team (@Vovany4, @kgr0) has experienced this bug.

somanath-goudar:
# This PR fixes issue with long query for search box

## Pull Request Type
- [x] Bugfix

## Related issue
closes #940 

## Description
Findings:
1) There is a limit for the number of character a query can have.
2) If the query exceeds the limit either 413 error or a bad request is thrown
3) From my testing, I have found that the character limit is 100 characters
4) To fix this issue, I have limited the number of characters a user can enter inside the search box

## Screenshots
https://github.com/FreeTubeApp/FreeTube/assets/72937521/8288ead6-07d3-4fc1-9872-b817cd30c417


PikachuEXE (commented):
Please address the warnings from lint first

somanath-goudar:
@PikachuEXE Lint warnings are fixed. Please review

PrestonN:
Hey there! While your PR does fix the issue with searching, it unfortunately creates a different problem.

Our search bar handles a number of different functions outside of searching. The main one is that you can paste a YouTube URL into the search bar and FreeTube will parse that URL to take you directly to the video. Sometimes, this URL will be an Invidious URL. Other times, the URL will include playlist information and we'll show that as well once we've taken you to your video. It could also be a combination of the 2.

The TLDR is that limiting the search bar to 100 characters fixes the search problem, but it has the potential to break our other features as those might be parsing URLs that are longer than 100 characters. This URL was shared as an example.

https://invidious.einfachzocken.eu/watch?v=WLP_L7Mgz6M&index=1&list=PLOJU8YJjFwGOR740NdzTui2NxU-R96YW4

I would suggest taking a step back and digging through the search bar functionality in order to fix the core issue. We'll have to revert the limit in the search bar and only make changes to the functionality that handles search. I would instead show a toast message that states that the query is too long and refuse to handle the search.

If this is something that you're able to do in the next day or so, then feel free to make those changes to this PR. If you don't think you'll have time to make those changes any time soon, then please close this PR and create a new one once you have had time to work on this.

Thank you!

somanath-goudar:
Hello @PrestonN,

I now have a better understanding of the issue. I'm committed to resolving it. I'll work on finding a solution over the next 24 hours, and if I'm successful, I'll make the necessary commits to this pull request. However, if I encounter any challenges during this period, I may need to temporarily close the pull request while I investigate the problem further.

Issue #992, opened 2021-01-28
[Feature Request] Add Enable/Disable Keyboard Shortcuts to settings


hectorzg:

shadycloud:
I'm trying to think of how thiis would be done. Perhaps by nesting the keyboard shortcut code in src/renderer/components/ft-video-player/ft-video-player.js   inside of an if statement that references a slider button titled "Keyboard Shortcuts" in player settings... idk

shadycloud:
The more I think about it, "Enable Keyboard Shortcuts" seems redundant given the on/off nature of a slider button and the likely technical aptitude of anyone using this app. I no longer think "Enable" is a necessary part of any slider label.

jasonhenriquez:
It's more for preserving the "[verb] [object]" parallelism than conveying any unique information with the word, per se.

shadycloud:
Indeed. I see that now*

FelipeJz:
Disable all keyboard shortcuts? Some of them could be useful for some users or annoying for others. As i said in #5070, i just want to disable the numbers row, i think it would be more useful to separate the shortcuts in blocks so each user can disable whatever they want

luke-bravenboer:
I'm in the same boat, @FelipeJz - I use other keyboard hotkeys all the time, but sometimes I accidentally press a number key and suddenly lose where I was up to in a video

I arrived here because I came from your FR (#5070) as it said it was a duplicate of this and #251 - and I think we're both just looking at the wrong one, #251 is what we're actually wanting haha

FelipeJz:
> I'm in the same boat, @FelipeJz - I use other keyboard hotkeys all the time, but sometimes I accidentally press a number key and suddenly lose where I was up to in a video
> 
> I arrived here because I came from your FR (#5070) as it said it was a duplicate of this and #251 - and I think we're both just looking at the wrong one, #251 is what we're actually wanting haha

I guess we are all in the same boat then, a solution should be implemented to both change and disable the keyboard shortcuts 🤣 that should close both issues

mkalinski:
I think saying that this issue is a duplicate of #5070 misses an important point:

The 0-9 number row navigation is the only shortcut that can completely lose you the position in the video on one accidental keystroke. Every other keyboard shortcut is a minor annoyance at most if you accidentally trigger it, and can be easily undone, but not this one.

[...]

PR #4296, opened 2023-11-07
Add Enable/Disable Keyboard Shortcuts to settings

Hector118:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

- Closes https://github.com/FreeTubeApp/FreeTube/issues/992

## Description
Added a setting to the player settings section which toggles keyboard shortcuts on the video player.

## Desktop
- **OS:**  Ubuntu LTS Jammy Jellyfish
- **OS Version:** 22.04


src/renderer/store/modules/settings.js
```diff
@@ -226,6 +226,7 @@ const state = {
   hideLabelsSideBar: false,
   hideChapters: false,
   showDistractionFreeTitles: false,
+  keyboardShortcuts: false,
```
ChunkyProgrammer, line 229: Maybe this should be set to true as some users are already using keyboard shortcuts and will probably open an issue thinking that shortcuts are broken, not realizing they have to enable it

absidue: Definitely should be enabled by default, as the keyboard shortcuts have existed for a long time and most users want them.

A better option would be to name the setting `Disable Keyboard Shortcuts`, not only does that clearly indicate to the user that the default is to have the keyboard shortcuts enabled, but it would also make the code clearer as you would have `disableKeyboardShortcuts`, instead of the ambiguously named `keyboardShortcuts` value, that doesn't immediately tell you that it is a boolean.

absidue (commented):
As mentioned above, could you please rename the setting to `Disable Keyboard Shortcuts` and update the code accordingly.

> A better option would be to name the setting Disable Keyboard Shortcuts, not only does that clearly indicate to the user that the default is to have the keyboard shortcuts enabled, but it would also make the code clearer as you would have disableKeyboardShortcuts, instead of the ambiguously named keyboardShortcuts value, that doesn't immediately tell you that it is a boolean.

Hector118:
> As mentioned above, could you please rename the setting to `Disable Keyboard Shortcuts` and update the code accordingly.
> 
> > A better option would be to name the setting Disable Keyboard Shortcuts, not only does that clearly indicate to the user that the default is to have the keyboard shortcuts enabled, but it would also make the code clearer as you would have disableKeyboardShortcuts, instead of the ambiguously named keyboardShortcuts value, that doesn't immediately tell you that it is a boolean.

I have changed the name of the setting to "disable Keyboard Shortcuts" in the player settings section and in other files.

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -2087,8 +2090,8 @@ export default defineComponent({
       if ((event.ctrlKey || event.metaKey) && event.key.toLowerCase() === 'c') {
         return
       }
-
-      if (this.player !== null) {
+      // 'this.disableKeyboardShortcuts === true' address disabling keyboard shortcuts
+      if (this.player !== null && this.disableKeyboardShortcuts === true) {
```
ChunkyProgrammer, line 2094:
```suggestion
      if (this.player !== null && this.disableKeyboardShortcuts === false) {
```

src/renderer/store/modules/settings.js
```diff
@@ -226,6 +226,7 @@ const state = {
   hideLabelsSideBar: false,
   hideChapters: false,
   showDistractionFreeTitles: false,
+  disableKeyboardShortcuts: true,
```
ChunkyProgrammer, line 229:
```suggestion
  disableKeyboardShortcuts: false,
```

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -2087,8 +2090,8 @@ export default defineComponent({
       if ((event.ctrlKey || event.metaKey) && event.key.toLowerCase() === 'c') {
         return
       }
-
-      if (this.player !== null) {
+      // 'this.disableKeyboardShortcuts === true' address disabling keyboard shortcuts
```
ChunkyProgrammer, line 2093:
```suggestion
      // 'this.disableKeyboardShortcuts === false' address disabling keyboard shortcuts
```

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Hmm few shortcuts dont seem to be disabled. Should we even disable them?
 
- Focus searchbar `alt + d`
- Refresh `r`

Most of the bottom table shouldn't be touched because those are necessary but maybe these can also go 

- Command/Ctrl + , 	Navigate to Preferences
- Command + Y (MacOS)/Ctrl + H (Windows & Linux) 	Navigate to History
- Alt + Left Arrow 	Go back one page
- Alt + Right Arrow 	Go forward one page

All shortcuts can be found https://docs.freetubeapp.io/usage/keyboard-shortcuts/

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc This pull request currently only targets the player ones and puts the setting in the player settings.

efb4f5ff-1298-471a-8973-3d47447115dc:
Do we consider that enough for that issue to close?

jasonhenriquez:
I don't think we can justify the other interpretation as a feature we would want to implement, so that's fine in my book.

Hector118:
Made the changes that were suggested for disableKeyboardShortcuts.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Could u move the toggle to the bottom of the list.

Capture

Hector118:
> Could u move the toggle to the bottom of the list.
> 
> Capture

moved it to the bottom

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -343,6 +343,10 @@ export default defineComponent({
 
     proxyVideos: function () {
       return this.$store.getters.getProxyVideos
+    },
+
+    disableKeyboardShortcuts: function() {
+      return this.$store.getters.getDisableKeyboardShortcuts
     }
```
PikachuEXE, line 350:
```suggestion
    keyboardShortcutsEnabled: function() {
      return this.$store.getters.getEnableKeyboardShortcuts
    },
```

src/renderer/components/ft-video-player/ft-video-player.js
```diff
@@ -2087,8 +2090,8 @@ export default defineComponent({
       if ((event.ctrlKey || event.metaKey) && event.key.toLowerCase() === 'c') {
         return
       }
-
-      if (this.player !== null) {
+      // 'this.disableKeyboardShortcuts === false' address disabling keyboard shortcuts
+      if (this.player !== null && this.disableKeyboardShortcuts === false) {
```
PikachuEXE, line 2094:
```suggestion
      if (this.player !== null && this.keyboardShortcutsEnabled) {
```

src/renderer/components/player-settings/player-settings.js
```diff
@@ -321,6 +325,7 @@ export default defineComponent({
       'updateScreenshotFilenamePattern',
       'parseScreenshotCustomFileName',
       'updateCommentAutoLoadEnabled',
+      'updateDisableKeyboardShortcuts'
```
PikachuEXE, line 328:
```suggestion
      'updateDisableKeyboardShortcuts',
```

src/renderer/components/player-settings/player-settings.vue
```diff
@@ -87,6 +87,12 @@
           :default-value="enterFullscreenOnDisplayRotate"
           @change="updateEnterFullscreenOnDisplayRotate"
         />
+        <ft-toggle-switch
+          :label="$t('Settings.Player Settings.Disable Keyboard Shortcuts')"
```
PikachuEXE, line 91:
```suggestion
          :label="$t('Settings.Player Settings.Enable Keyboard Shortcuts')"
```

src/renderer/store/modules/settings.js
```diff
@@ -226,6 +226,7 @@ const state = {
   hideLabelsSideBar: false,
   hideChapters: false,
   showDistractionFreeTitles: false,
+  disableKeyboardShortcuts: false,
```
PikachuEXE, line 229:
```suggestion
  enableKeyboardShortcuts: true,
```

static/locales/en-US.yaml
```diff
@@ -257,6 +257,7 @@ Settings:
     Turn on Subtitles by Default: Turn on Subtitles by Default
     Autoplay Videos: Autoplay Videos
     Proxy Videos Through Invidious: Proxy Videos Through Invidious
+    Disable Keyboard Shortcuts: Disable Keyboard Shortcuts
```
PikachuEXE, line 260:
```suggestion
    Enable Keyboard Shortcuts: Enable Keyboard Shortcuts
```

src/renderer/components/player-settings/player-settings.vue
```diff
@@ -87,6 +87,12 @@
...
+          :compact="true"
+          :default-value="disableKeyboardShortcuts"
```
PikachuEXE, line 93:
```suggestion
          :default-value="enableKeyboardShortcuts"
```

src/renderer/components/player-settings/player-settings.vue
```diff
@@ -87,6 +87,12 @@
...
+          @change="updateDisableKeyboardShortcuts"
```
PikachuEXE, line 94:
```suggestion
          @change="updateEnableKeyboardShortcuts"
```

PikachuEXE (changes requested):
Besides the series of distraction free settings
I see no reason to prefix a setting with a negative term (e.g. disable)

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE wouldn't changing this to enable cause confusion for users?

PikachuEXE:
There are many enabled by default settings
It's more confusing when it's "on" (except those "hide" prefix settings which is already another cluster) but it means disabled
2023-11-29 15_02_14-Window
2023-11-29 15_03_46-Window

And yes even the "smooth scrolling" one should be prefixed with "Enable" IMO (not saying it should be changed in this PR)
See how Chrome/Firefox display setting for "smooth scrolling"

Chrome
image

Firefox
image

absidue:
I feel like in this context, using disable is okay, as I highly doubt most people will want to touch the setting, so making it clear that they are enabled by default seems like a good idea. We don't have any way of the user knowing what the default value is or whether they changed a setting from the default, which is how we end up with people turning on proxy videos through Invidious, complaining that stuff doesn't work, when the instance gets IP blocked and then blaming us for turning it on by default (which we don't).

Nobody has provided a good reason for why you would even want to disable a useful feature (the issue has no description and the person opening the pull request, could be the same person looking at the account name), that as far as I see it, has no downsides.

TL;DR This looks like a setting that only one person wants, so we should make it obvious to users that the default is enabled. Putting "Enable" in front of a setting that is enabled by default, makes it seem like an extra, that people might unknowingly disable, however in this case the setting exists to disable expected behaviour.

PikachuEXE:
Fine make the label and UI prefixed with disable but make the setting prefixed with enable

efb4f5ff-1298-471a-8973-3d47447115dc:
> Fine make the label and UI prefixed with disable but make the setting prefixed with enable

Is this the only blocker of the PR?

PikachuEXE:
Yup

PR #4338, opened 2023-11-15
Make video info section more concise

jasonhenriquez:
# Make video info section more concise

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
closes #4336 

## Description
- Puts title on its own row
- Puts video info section on its own row
- Puts subscribe button & video buttons on their own row
- Fixes bug of `videoOptions` buttons overflowing the video info section that occurred on some viewport widths

## Video 

https://github.com/FreeTubeApp/FreeTube/assets/84899178/18ce83a6-c7e9-4001-9fe5-d6e6ea4a3711

## Testing 
- Ensure that appearance is great at all screen sizes
- Ensure that the video options buttons no longer extend past the container when resizing viewport

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx
- **FreeTube version:** 0.19.1

PR #4350, opened 2023-11-18
Add back 3 Stylelint configs

jasonhenriquez:
# Add back 3 Stylelint configs

## Pull Request Type

- [x] Feature Implementation

## Related issue
[See chore](https://github.com/FreeTubeApp/FreeTube/projects/10#card-91110435)

## Description
- Adds back 3 Stylelint configs:
  - @double-great/stylelint-a11y
  - stylelint-config-sass
  - stylelint-config-standard

Changes from initial (intended) config:
- `selector-no-qualifying-type` now ignores attribute selectors
- Non-recommended rule `a11y/font-size-is-readable` is removed ([see discussion](https://github.com/FreeTubeApp/FreeTube/pull/4338#discussion_r1394907634))
- `[media-prefers-reduced-motion](https://github.com/double-great/stylelint-a11y/blob/main/src/rules/media-prefers-reduced-motion/README.md)` is now satisfied (but not added as a rule due to the limited capabilities of the Stylelint config)

## Testing 
This is a tough one to have clear testing requirements for. Many files were "changed," but pretty insubstantially so, and none should be changed functionally (excluding our new support for users with `prefers-reduced-motion`).

- Test enabling `prefers-reduced-motion` and validate that our transitions and animations are no longer present (can be simulated by removing the media query surrounding it on line 1062 of `themes.css`. Time: <5 min
- Smoke test (e.g., using settings, watching a video, and other basic features). Time: 10-20 min

## Desktop
- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

### DIFF
--- a/src/renderer/App.css
+++ b/src/renderer/App.css
@@ -1,9 +1,9 @@
 @font-face {
   font-family: Roboto;
-  src: url(assets/font/Roboto-Regular.ttf);
+  src: url("assets/font/Roboto-Regular.ttf");
 }
 
-#app {
+.app {
   display: flex;
   flex-wrap: wrap;
   font-family: 'Roboto', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
@@ -30,18 +30,18 @@
 .flexBox {
   display: block;
   user-select: unset;
-  -webkit-user-select: unset;
 }
 
-#changeLogText {
+.changeLogText {
   overflow-y: scroll;
   block-size: 40vh;
-  display: block
+  display: block;
 }
 
 .fade-enter-active, .fade-leave-active {
   transition: opacity .15s;
 }
+
 .fade-enter, .fade-leave-to /* .fade-leave-active below version 2.1.8 */ {
   opacity: 0;
 }
--- a/src/renderer/App.css
+++ b/src/renderer/App.css
@@ -6,7 +6,7 @@
 .app {
   display: flex;
   flex-wrap: wrap;
-  font-family: 'Roboto', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
+  font-family: Roboto, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
   block-size: 100%;
 }
 
@@ -46,7 +46,7 @@
   opacity: 0;
 }
 
-@media only screen and (max-width: 680px) {
+@media only screen and (width <= 680px) {
   .routerView {
     margin-block: 68px;
     margin-inline: 8px;
@@ -58,7 +58,6 @@
   }
 
   .flexBox {
-    margin-block-start: 60px;
-    margin-block-end: -75px;
+    margin-block: 60px -75px;
   }
 }
--- a/src/renderer/components/ft-loader/ft-loader.css
+++ b/src/renderer/components/ft-loader/ft-loader.css
@@ -53,23 +53,21 @@
     inset-block-start: 0;
     inset-inline-start: 0;
     background-color: var(--primary-color);
-
-    -webkit-animation: sk-bounce 2.0s infinite ease-in-out;
     animation: sk-bounce 2.0s infinite ease-in-out;
 }
 
 .double-bounce2 {
-    -webkit-animation-delay: -1.0s;
     animation-delay: -1.0s;
 }
 
-@-webkit-keyframes sk-bounce {
+@keyframes sk-bounce {
     0%,
     100% {
-        -webkit-transform: scale(0.0)
+        transform: scale(0.0)
     }
+
     50% {
-        -webkit-transform: scale(1.0)
+        transform: scale(1.0)
     }
 }
 
@@ -77,10 +75,9 @@
     0%,
     100% {
         transform: scale(0.0);
-        -webkit-transform: scale(0.0);
     }
+
     50% {
         transform: scale(1.0);
-        -webkit-transform: scale(1.0);
     }
 }
--- a/src/renderer/components/watch-video-chapters/watch-video-chapters.css
+++ b/src/renderer/components/watch-video-chapters/watch-video-chapters.css
@@ -3,8 +3,7 @@
 }
 
 .chaptersTitle {
-  margin-block-start: 10px;
-  margin-block-end: 0;
+  margin-block: 10px 0;
   cursor: pointer;
 }
 
@@ -35,11 +34,9 @@
 
 .chapter {
   display: grid;
-  grid-template-areas:
-    'thumbnail title'
-    'thumbnail timestamp';
-  grid-template-columns: auto 1fr;
-  grid-template-rows: min(auto, 2fr) 1fr;
+  grid-template:
+    'thumbnail title' 2fr
+    'thumbnail timestamp' 2fr / auto 1fr;
   column-gap: 10px;
   justify-items: start;
   cursor: pointer;
--- a/src/renderer/scss-partials/_ft-list-item.scss
+++ b/src/renderer/scss-partials/_ft-list-item.scss
@@ -54,6 +54,17 @@ $watched-transition-duration: 0.5s;
 .ft-list-item {
   padding: 6px;
 
+  .thumbnailImage {
+    @include is-sidebar-item {
+      block-size: 75px;
+    }
+
+    @include is-recommendation {
+      block-size: auto;
+      inline-size: 163px;
+    }
+  }
+
   &.watched {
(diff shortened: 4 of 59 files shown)


src/renderer/App.css
```diff
@@ -30,23 +30,23 @@
...
 .fade-enter-active, .fade-leave-active {
   transition: opacity .15s;
 }
+
 .fade-enter, .fade-leave-to /* .fade-leave-active below version 2.1.8 */ {
   opacity: 0;
 }
 
-@media only screen and (max-width: 680px) {
+@media only screen and (width <= 680px) {
```
PikachuEXE, line 49: This works coz we use electron = Chrome
https://caniuse.com/css-media-range-syntax

src/renderer/views/Channel/Channel.vue
```diff
@@ -369,6 +369,7 @@
         <ft-element-list
           v-if="!hideChannelCommunity && currentTab === 'community'"
           id="communityPanel"
```
PikachuEXE, line 371: Still referenced by `aria-controls` somewhere

PikachuEXE (commented):
Too big, can't review & approve now :P

src/renderer/views/Channel/Channel.vue
jasonhenriquez: We still have the ID. Just added it as a class as well

PikachuEXE: Yup I know. It's for other reviewers (coz diff won't show `aria-controls`

src/renderer/App.css
jasonhenriquez: It does work, but it's not the opinion of [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard). See [here](https://stylelint.io/user-guide/rules/media-feature-range-notation/) and [here](https://github.com/stylelint/stylelint-config-standard/blob/9d8d64daf469c6e7466594ceb003065556bcecba/index.js#L83)

src/renderer/components/ft-loader/ft-loader.css
```diff
@@ -53,34 +53,31 @@
...
     100% {
-        -webkit-transform: scale(0.0)
+        transform: scale(0.0)
     }
+
     50% {
-        -webkit-transform: scale(1.0)
+        transform: scale(1.0)
     }
 }
```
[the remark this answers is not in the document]
jasonhenriquez: Good catch!

src/renderer/App.css
absidue: Will this newer syntax break the media queries for FreeTubeCordova? That's where most of the users of the media queries will be.
Looks like you can enforce the old style instead of the new style by setting that stylelint setting to the other value.

jasonhenriquez: Looks like they're using [Electron v22](https://github.com/MarmadileManteater/FreeTubeCordova/blob/development/package.json), [so we should be good](https://www.electronjs.org/blog/electron-22-0)

jasonhenriquez: The linter is not smart enough to fix this rule violation the other way apparently... will have to do some better research to answer that question

jasonhenriquez: Based on Emma's response in the channel, I think we're good?

jasonhenriquez:
Requesting re-review. CC: @absidue @PikachuEXE @efb4f5ff-1298-471a-8973-3d47447115dc

absidue:
> Properly implementing the no-descending-specificity rule in this file seems to utterly destroy the styling. This would be its own entire initiative to unravel.

We really need to work on getting rid of unnecessarily nested CSS (not in this PR but at some point in the future). We'll need to keep the `.grid` and `.list` nesting, as we still want to have separate styles for them, but one level of nesting is still a massive improvement over going 6 levels deep. That should also make it a lot easier to fix `no-descending-specificity` problems, as rearranging stuff that is on the same level is much easier when nesting isn't in the way.

src/renderer/components/ft-list-playlist/ft-list-playlist.scss
```diff
@@ -1 +1 @@
-@use '../../scss-partials/_ft-list-item';
+@use '../../scss-partials/ft-list-item';
```
ChunkyProgrammer, line 1: Does this not cause issues? (The file name is still _ft-list-item)

This comment applies for all ft-list-*

jasonhenriquez: [Here's what I found on that](https://stackoverflow.com/a/27984975)

ChunkyProgrammer (dismissed):

PikachuEXE (commented):
These for now too many files

src/renderer/scss-partials/_ft-list-item.scss
```diff
@@ -86,42 +129,35 @@ $watched-transition-duration: 0.5s;
...
       padding: 2px;
       pointer-events: none;
     }
 
+    .videoWatched,
+    .live,
+    .upcoming {
+      text-transform: uppercase;
+      user-select: none;
+    }
```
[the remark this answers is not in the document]
jasonhenriquez: The way I did this last file was weird (a pull --theirs instead of just restarting from scratch). Resolved these oddities now.

src/renderer/components/ft-list-channel/ft-list-channel.scss
```diff
@@ -32,13 +35,6 @@
   }
 }
 
-.infoAndSubscribe {
-  display: flex;
-  flex-flow: row wrap;
-  justify-content: center;
-  inline-size: 100%;
-}
```
PikachuEXE, line 40: `.infoAndSubscribe` under `&.list` has no `display: flex;` & `inline-size: 100%;`
Expected?

PikachuEXE (commented):
Submit one when I spot one, way too many files to review (18/70 so far

src/renderer/components/ft-list-channel/ft-list-channel.scss
jasonhenriquez: This is a very old PR, so I can't remember for sure, but it seems like I groggily made changes to address the `no-descending-specificity` rule that I can't tell if they're intentional (and crafted to stop things from breaking) or are just mistakes. I'm going to comb through all of these again and check back when I'm done.

jasonhenriquez:
After combing through the files and doing more testing, `ft-list-channel` and `ft-tooltip` were affected by mistakes. This is now ready for review.

src/renderer/components/watch-video-chapters/watch-video-chapters.css
```diff
@@ -35,11 +34,9 @@
...
 .chapter {
   display: grid;
-  grid-template-areas:
-    'thumbnail title'
-    'thumbnail timestamp';
-  grid-template-columns: auto 1fr;
-  grid-template-rows: min(auto, 2fr) 1fr;
+  grid-template:
+    'thumbnail title' 2fr
+    'thumbnail timestamp' 2fr / auto 1fr;
```
PikachuEXE, line 39: Original value is `grid-template-rows: min(auto, 2fr) 1fr;`
Are the values right after change?

jasonhenriquez: Yes

src/renderer/themes.css
```diff
@@ -1041,7 +1042,7 @@ body[dir='rtl'] {
   --horizontal-directionality-coefficient: -1;
 }
 
-body[dir='rtl'] [data-prefix="fas"]:not([data-icon="magnifying-glass"]):not([data-icon="circle-question"]) {
+body[dir='rtl'] [data-prefix="fas"]:not([data-icon="magnifying-glass"], [data-icon="circle-question"]) {
```
PikachuEXE, line 1045: Is this correct?

src/renderer/components/watch-video-chapters/watch-video-chapters.css
PikachuEXE: Ya the current value seems invalid
image

src/renderer/themes.css
jasonhenriquez: Yes, just re-validated with Arabic and Hebrew. Peep the closing parenthesis location

PikachuEXE: Ah ok it's quite difficult to read with a long line & parenthesis...

PikachuEXE (approved):

ChunkyProgrammer (dismissed):

ChunkyProgrammer (dismissed):

PikachuEXE (dismissed):

jasonhenriquez:
This is currently passing for all files. Please get this one in next if possible!

PikachuEXE (approved):

ChunkyProgrammer (approved):

absidue (approved):

PR #4352, opened 2023-11-18
Show latest from channel

### LINKED ISSUE #546: Show only most recent video from a subscription
Invidious settings have an option showing only the most recent video from a subscription.  Showing only the most recent video is nice because then on your subscription page you have one thumbnail for each subscription and with hide watched videos turned on, as you watch videos the subscription page will show the most recent unwatched video from each subscription.

jasonhenriquez:
# Show latest from channel

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #546

## Description
Shows only the latest video from each subscribed channel in Subscriptions. Latest of each video tab that you're on, to be specific (Videos, Shorts, & Live, but not Community).

## Screenshots 
Screenshot_20231118_165810

## Testing 
- Test feature enabled with RSS feed enabled
- Test feature enabled without RSS feed disabled
- Test feature disabled

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

## Additional context
I wonder if it would be a neat feature later on to have these subscription settings show up on the subscription page. Not sure how to feel about it, but share your thoughts on that with your review if you have any.

### DIFF
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -26,6 +26,12 @@
           :compact="true"
           @change="updateHideWatchedSubs"
         />
+        <ft-toggle-switch
+          :label="$t('Settings.Subscription Settings.Only Show Latest Video from Channel')"
+          :default-value="onlyShowLatestFromChannel"
+          :compact="true"
+          @change="updateOnlyShowLatestFromChannel"
+        />
       </div>
     </div>
   </ft-settings-section>
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -337,6 +337,7 @@ Settings:
     Fetch Feeds from RSS: Fetch Feeds from RSS
     Manage Subscriptions: Manage Subscriptions
     Fetch Automatically: Fetch Feed Automatically
+    Only Show Latest Video from Channel: Only Show Latest Video from Channel
   Distraction Free Settings:
     Distraction Free Settings: Distraction Free Settings
     Sections:
--- a/src/renderer/components/subscription-settings/subscription-settings.js
+++ b/src/renderer/components/subscription-settings/subscription-settings.js
@@ -13,6 +13,9 @@ export default defineComponent({
     hideWatchedSubs: function () {
       return this.$store.getters.getHideWatchedSubs
     },
+    onlyShowLatestFromChannel: function () {
+      return this.$store.getters.getOnlyShowLatestFromChannel
+    },
     useRssFeeds: function () {
       return this.$store.getters.getUseRssFeeds
     },
@@ -24,7 +27,8 @@ export default defineComponent({
     ...mapActions([
       'updateHideWatchedSubs',
       'updateUseRssFeeds',
-      'updateFetchSubscriptionsAutomatically'
+      'updateFetchSubscriptionsAutomatically',
+      'updateOnlyShowLatestFromChannel'
     ])
   }
 })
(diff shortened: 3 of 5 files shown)


PikachuEXE:
As a user I have no idea what `latest video from each subscribed channel` means

jasonhenriquez:
> As a user I have no idea what `latest video from each subscribed channel` means

The label that's used in the code is `Only Show Latest Video from Channel`. [Label for this feature in Invidious is `Only show latest video from channel`](https://github.com/iv-org/invidious/blob/c5b87e3b5e5cc7f7f5c8baa7732bd6d81d8f910a/locales/en-US.json#L126). Are you saying the text is confusing or do you need context on the original feature request?

PikachuEXE:
Which means `Show Only The Latest Videos from Channels` (well not sure how to express **per channel** well)
The label text should make user easier to understand that it's **only one**, **latest**, **per channel**
Coz when I read it first I read it as latest videos from channel (For me it's quite easy to be confused by `Only Show Latest Video from Channel`, missing **only one** & **per channel**)

jasonhenriquez:
`Only show latest video from each subscribed channel`?

PikachuEXE:
`the latest`? Problem is that without `the` it reads similar to `Latest News` which seems multiple (per channel)

jasonhenriquez:
I think the singular `latest video` instead of `latest videos` makes that clear enough, no?

PikachuEXE:
Not so clear for me but I guess you can ask other people :P 
Again I personally have no idea what it does on first read (well user can try and found out, but better if it's expressed well from the beginning)
If others feel like it's easy enough to be understood then I won't reject it just for the label text

absidue:
What about "Only show latest video for each channel"?

src/renderer/components/subscription-settings/subscription-settings.vue
```diff
@@ -26,6 +26,12 @@
           :compact="true"
           @change="updateHideWatchedSubs"
         />
+        <ft-toggle-switch
+          :label="$t('Settings.Subscription Settings.Only Show Latest Video from Channel')"
```
PikachuEXE, line 30:
```suggestion
          :label="$t('Settings.Subscription Settings.Only Show Latest Video for Each Channel')"
```

static/locales/en-US.yaml
```diff
@@ -337,6 +337,7 @@ Settings:
     Fetch Feeds from RSS: Fetch Feeds from RSS
     Manage Subscriptions: Manage Subscriptions
     Fetch Automatically: Fetch Feed Automatically
+    Only Show Latest Video from Channel: Only Show Latest Video from Channel
```
PikachuEXE, line 340:
```suggestion
    Only Show Latest Video for Each Channel: Only Show Latest Video for Each Channel
```

PikachuEXE (commented):
I am OK with "for each channel"

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

absidue (commented):
I'm curious about the order of the filtering, currently it's done before we filter out premieres or currently live live streams, so theoretically you could end up with nothing being shown for a channel (filter picks the first item which is an upcoming premiere, upcoming premiere filter filters it out).

Would it be better to do this filtering after the content type filters, that way there is a higher chance that you will actually see something for all channels? (Should be as simple as moving the new code in the subscriptions helper file down 30ish lines).

absidue (approved):
:)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #4357, opened 2023-11-20
Display more profiles on the profile drop-down list

### LINKED ISSUE #2871: [Feature Request]: UI change: Move profile icons into column, similar to left column with channel icons and menu. (Instead of drop-down list for profile selection)
### Guidelines

### Problem Description

p

Currently the "Profile Select" Menu is a drop-down list, containing only 5 elements. If you have bigger number of profiles (e.x. 20) the process of switching them is inconvenient - you need to click on icon for drop-down list to appear and then do a lot of scrolling to find profile you want to switch for. 

**Channel profiles are very strong feature of FreeTube, but the inability to switch them in quick and convenient way is choking the potential of this feature.** 

I propose UI solution, which would allow user to switch profile quickly, with one click. 

### Proposed Solution

My proposition is to group profile icons in the column (on the right side) , in a very similar way to the left menu column. 
The column would have 3 states:
- minimal: with icons only
- wide: with icons and names
- hidden

The onclick behavior of the right-upper button (which now opens drop-down list) would be now to change state of the column. 

Concept screenshot (in minimal state):
(I've made it Paint, so the spacing between icons is unequal, sorry)
Profile UI

### Alternatives Considered

Alternative solution would be to just unfix the size of the drop-down list and adjust it's length to the number of elements. 
But then you would still need to make 2 clicks to switch profile, instead of one. 

### Issue Labels

display more information to user, ease of use improvement

### Additional Information

_No response_

CelularBat:
# Title
Current profile list has fixed size, so it can display only 5 profiles. If you have more profiles you have to scroll, if you have much more profiles, like 15+ scrolling for what you want to find in that little 5-items box is inconvenient.

This extends profile list size to fit the amount of profiles, but to no more than 80% of window size.  

## Pull Request Type

- [ x] Bugfix
- [ x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes  #2871

## Description
It adds 3 lines of .css

## Screenshots 
old:
p1
new:
p2

## Desktop

- **OS:**
- **OS Version: 2023**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-profile-selector/ft-profile-selector.css
+++ b/src/renderer/components/ft-profile-selector/ft-profile-selector.css
@@ -31,12 +31,15 @@
   padding: 5px;
   background-color: var(--card-bg-color);
   box-shadow: 0 0 4px var(--scrollbar-color-hover);
+  height: fit-content;
 }
 
 .profileWrapper {
   margin-block-start: 60px;
   block-size: 340px;
   overflow-y: auto;
+  height: fit-content;
+  max-height: 80vh;
 }
 
 .profile {


PikachuEXE:
I think people should just check those boxes after PR submitted to avoid strange description like this
image

Also this should be just a feature implementation (improvement) not bug fix

PikachuEXE (commented):
Doesn't work well with smaller height
image

CelularBat:
> Doesn't work well with smaller height image
I don't see much on your screen, because it's behind the console.

But I see the problem.
On smaller resolution like 1280x720 drop down list goes below the edge. But max-height: 70vh fixes it for this resolution. 

What's the minimum height it should work with?

PikachuEXE:
I would say use media query to have different `max-height`
I have no idea what's the minimum expected height though

CelularBat:
> I would say use media query to have different `max-height` I have no idea what's the minimum expected height though

The current solution is that position of the bottom edge of the list is hardcoded to 400px + 8px margin + 60px of upper bar, so it seems the smallest resolution it can handle and look "good" is 800x600 anyway. 

Current height of 400px is 66% of 800x600 display, so I think 60vh is a "safe" value, which wouldn't break small resolution displays, and would enhance display on bigger resolutions. Or I would make it compensate those hardcoded 60px of the upper bar.

So I will change it, do it in a way linter want me do, and try again.

PR #4359, opened 2023-11-20
Display more profiles on the profile drop-down list, v2 (fixed for low res, fixed linter) 

### LINKED ISSUE #2871: [Feature Request]: UI change: Move profile icons into column, similar to left column with channel icons and menu. (Instead of drop-down list for profile selection)

CelularBat:
Fixed: https://github.com/FreeTubeApp/FreeTube/pull/4357

Current profile list has fixed size, so it can display only 5 profiles. If you have more profiles you have to scroll, if you have much more profiles, like 15+ scrolling for what you want to find in that little 5-items box is inconvenient.

This extends profile list size to fit the amount of profiles.

- it scales with window size and resolution. 
- for resolution 800x600 and any lower it's exactly the same as it was before

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #2871

## Description
it changes 4 lines of .css

## Screenshots 
before:
p1
after (at 1920x1080)
p2
after (at 1920x1080 ,random window size)
p3
after (at 1280x720)
p4

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-profile-selector/ft-profile-selector.css
+++ b/src/renderer/components/ft-profile-selector/ft-profile-selector.css
@@ -27,7 +27,7 @@
   inset-block-start: 60px;
   inset-inline-end: 10px;
   min-inline-size: 250px;
-  block-size: 400px;
+  block-size: auto;
   padding: 5px;
   background-color: var(--card-bg-color);
   box-shadow: 0 0 4px var(--scrollbar-color-hover);
@@ -35,8 +35,10 @@
 
 .profileWrapper {
   margin-block-start: 60px;
-  block-size: 340px;
+  block-size: auto;
   overflow-y: auto;
+  max-block-size: calc(90vh - 100px);
+  min-block-size: 340px;
 }
 
 .profile {


PikachuEXE (approved):
Height is auto adjusted until at least 5 profiles are shown (like now)
Works great~

CelularBat:
> Great job @CelularBat! I do have two very minor nitpicks if that's okay:
> 
>     1. **nitpick:** Dropdown clips into the bottom bar on mobile, so you might need a [media query](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Using_media_queries) to handle that. Totally fine to pass on if this one is too much trouble to resolve.
>        Screenshot_20231120_075952
> 
>     2. **suggestion (non-blocking):** A code comment explaining the `max-block-size` will help if we ever decide to resize the bars at some point, or something like that.

1. Done. But some mobile testing would be nice, because I've tested it only on desktop. 
2. Comment added

src/renderer/components/ft-profile-selector/ft-profile-selector.css
```diff
@@ -27,16 +27,28 @@
...
+  max-block-size: calc(90vh - 100px); 
+  min-block-size: 340px;
+}
+
+/* Navbar changes position to horizontal with this media rule.
+Height adjust for profile list so it won't cover navbar.  */
+@media only screen and (max-width: 680px){
+	 .profileWrapper {
+		  max-block-size: calc(95vh - 180px);
+		}
```
PikachuEXE, line 51: Use spaces not tabs (and use editor respecting [`.editorconfig`](https://editorconfig.org/)
```suggestion
  .profileWrapper {
    max-block-size: calc(95vh - 180px);
  }
```

src/renderer/components/ft-profile-selector/ft-profile-selector.css
```diff
@@ -27,16 +27,28 @@
...
+  block-size: auto;
   padding: 5px;
   background-color: var(--card-bg-color);
   box-shadow: 0 0 4px var(--scrollbar-color-hover);
 }
 
 .profileWrapper {
   margin-block-start: 60px;
-  block-size: 340px;
+  block-size: auto;
```
PikachuEXE, line 38:
```suggestion
  block-size: auto;
```

src/renderer/components/ft-profile-selector/ft-profile-selector.css
```diff
@@ -27,16 +27,28 @@
...
   box-shadow: 0 0 4px var(--scrollbar-color-hover);
 }
 
 .profileWrapper {
   margin-block-start: 60px;
-  block-size: 340px;
+  block-size: auto;  
   overflow-y: auto;
+  /* profile list max height: 90% of window size - 100 px. It's scaled to be 340px on 800x600 resolution. 
+  Offset of 100px is to compensate for the fixed size of elements above the list, which takes more screen space on lower resolutions */
```
PikachuEXE, line 41:
```suggestion
  /*
  profile list max height: 90% of window size - 100 px. It's scaled to be 340px on 800x600 resolution. 
  Offset of 100px is to compensate for the fixed size of elements above the list, which takes more screen space on lower resolutions
  */
```

PikachuEXE (changes requested):
Small width seems fine
image
image

Only minor code style issues remain

jasonhenriquez (approved):
LGTM. Tested with four main cases:
1. Few profiles desktop view
2. Few profiles tablet & mobile view
3. Many profiles desktop view
4. Many profiles tablet & mobile view

Thanks for being so patient with our nitpicks! Contributing the first time to a new repo can be scary, but you've been very responsive and cooperative.

>     1. Done. But some testing on mobile would be nice, because I've tested it only on desktop.

Just so you know for the future, you can pretty accurately emulate the mobile experience using [Device Mode in Chrome DevTools](https://developer.chrome.com/docs/devtools/device-mode/).

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #4312, opened 2023-11-09
[Bug]: Search Returns No Values On Newly Updated Install


xcarnage:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

While using the search bar attempt a random search query. Results return empty. Clicking fetch more results still returns empty.

### Expected Behavior

I expected that search results would return information
![Screenshot (45)](https://github.com/FreeTubeApp/FreeTube/assets/23347336/7ac606e5-c7d0-4b37-b1e6-169c1b05f254)

### Issue Labels

content not loading

### FreeTube Version

v0.19.1 Beta

### Operating System Version

Windows Version 22h2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
Can you please switch to the console tab in the devtools and see what the red messages are, according to the status bar on the top right, there are 13 errors.

armakunizen:
I have the same problem on 0.19.1 beta. In the devtools (as you described) I get the following, hope this helps:

`renderer.js:2 VIDEOJS: WARN: using Vhs.BANDWIDTH_VARIANCE is UNSAFE be sure you know what you are doing
(anonymous) @ renderer.js:2
renderer.js:2 TypeError: Cannot read properties of undefined (reading 'name')
    at o$ (renderer.js:2:1384290)
    at renderer.js:2:1385432
    at renderer.js:2:1385439
    at Array.map (<anonymous>)
    at l$ (renderer.js:2:1384669)
    at renderer.js:2:1579971
    at async a.performSearchLocal (renderer.js:2:1579720)
performSearchLocal @ renderer.js:2
[...]

PR #4371, opened 2023-11-22
Local API: Support the author-less auto-generated album playlists in video lists

absidue:
# Local API: Support the author-less auto-generated album playlists in video lists

## Pull Request Type

- [x] Bugfix

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/4312#issuecomment-1812026760 (yes this only fixes the bug in that comment, because the original issues is a different bug, that needs to be fixed on the YouTube.js side)

## Description
This pull request adds support for auto-generated album playlists in video lists, including the search results. Those playlists don't have an author/channel, just a bit of text that says `Playlist`, so when FreeTube tries to get the author information, it errors.

This only fixes the issue on the FreeTube side, 

## Screenshots 
Here is a screenshot of the error, using the search query used in the comment on the issue `glauben sie ja nicht wen sie da vor sich haben`, as they only provided the hard to read log text instead of a screenshot:
screenshot of errors

This is what the playlist looks like when the parsing is fixed:
screenshot-of-the-parsed-album-playlist

## Testing 
Search for `glauben sie ja nicht wen sie da vor sich haben` and check that the playlist shows up (see screenshot above to see what it looks like).


src/renderer/helpers/api/local.js
```diff
@@ -482,9 +483,13 @@ export function parseLocalListPlaylist(playlist, author = undefined) {
       channelName = playlist.author.name
       channelId = playlist.author.id
     }
-  } else {
+  } else if (author) {
     channelName = author.name
     channelId = author.id
+  } else if (!isNullOrEmpty(playlist.author?.name)) {
```
jasonhenriquez, line 489: **question:** what is `isNullOrEmpty` doing differently from the default truthiness check? If it is doing anything, this should be before the `if (author)` check.

absidue: There is a difference between `playlist.author` and the `author` parameter.

The logic in this function is:
1. Check if the playlist has a valid author object, if yes use it
2. Check if a fallback author was passed into the function, if yes use that
3. Check if the playlist has a partial/incomplete author object, if yes use it

The important bit is that we only want to use the incomplete playlist author object, if the caller didn't provide a fallback, we want to give the fallback higher priority than incomplete data (the fallback is used ont the channel page for example, where the playlists don't come with author information in them).

jasonhenriquez: Ah, I'm sorry, I totally missed the `playlist.` part! That makes sense. Same Q as before, though, on why we have a `isNullOrEmpty` function. Is it something having to do with type coercion?

absidue: That function checks for nullish values and empty strings, it's useful because `0` counts as false as well, although I guess in this case that isn't a concern, so I can probably change it to just being `if (playlist.author?.name) {`, as empty strings are falsy but not nullish.

https://developer.mozilla.org/en-US/docs/Glossary/Nullish
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

absidue: Simplified the check now.

jasonhenriquez: Ah, ok, just was curious why we had it. That explanation would make sense, but looking at our existing uses, it seems to be "`!!` is yucky", which is fine lol

jasonhenriquez:
Am not finding it in the search results for some reason

Screenshot_20231122_131141

absidue:
Make sure you don't have any search filters set, as it disappears if you set the filter to playlists #justYouTubeThings.

jasonhenriquez:
Still nothin

Screenshot_20231122_131517

absidue:
Might have to scroll a bit, was below the fold for me, but you shouldn't need to press load more, you can also try quoting the query, I'll see if I can find any other queries that produce the same bug/need the same fix, as that's just the query that was provided in the issue.

absidue:
Probably going to be difficult to find other queries, as it just seems to be albums that YouTube wasn't able to map an artist channel to, because "Good Girl Gone Bad: The Remixes" for example works fine with the existing code, because it has a link to Rihanna's own YouTube channel.

ChunkyProgrammer:
This PR should fix the issue when using the invidious API https://github.com/iv-org/invidious/pull/3869

ChunkyProgrammer:
Searching `The Strxwberry Vault ` then fetching more might cause the issue too?

absidue:
Yes, that one works too, thank you. I'll add it to the testing section :)

jasonhenriquez (approved):

PikachuEXE:
Dunno why the playlist image won't download for me, but image loading is fine when clicked inside
image
image

Not working image URL (outside): https://i9.ytimg.com/s_p/OLAK5uy_nnZsC6XX9bXuAp8zvF-7N8Ucykb2dl2AE/maxresdefault.jpg?sqp=CICs-qoGir7X7AMGCP_a0aoG&rs=AOn4CLAl00IVUS7LwxPtK6KdG0OiwU0T_g&v=1700031871
Working image URL (in playlist view): https://i9.ytimg.com/s_p/OLAK5uy_nnZsC6XX9bXuAp8zvF-7N8Ucykb2dl2AE/maxresdefault.jpg?sqp=CICs-qoGir7X7AMICP_P9qoGEAE=&rs=AOn4CLCcY6M9nDuqgsQcY8YHX9E0670c1w&v=1700636671

efb4f5ff-1298-471a-8973-3d47447115dc:
Im noticing the same issue as @PikachuEXE is that something we want to address in here?

absidue:
I wonder if it's a VPN thing, not much we can do with YouTube giving us broken links, other than using a different one in the array or trying to fix them, but I don't think we can identify broken ones.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #3668, opened 2023-06-15
Update subscription view to be able to load videos from video cache per channel

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

A variant (?) of #3665

## Description

#3665 restores the "original" behaviour (well the code that was there but disabled somehow)
This is a different video caching strategy
Videos (and potentially other info like live streams and shorts) are cached per **channel** instead of profile
The difference is that every profile can potentially just load from cache if all the channels of selected profile have required content in cache, without the limitation of only caching one profile at a time

## Screenshots 

Similar to 3665 one so no repeating here

## Testing 

Preparation: 
- Create following profiles (other than default)
  - Profile A with **most** but not all channels
  - Profile B with only some channels from A
  - Profile C with some channels from A (1 is enough), some from the channel(s) **excluded** in A
  - Profile D with no channel

**Before starting each test please reopen window/app/refresh to ensure local state is reset**

A. Auto fetching disabled
A1. Loading from cache after all channel cached
- Switch to default profile (if not already)
- Ensure no video loaded
- Load videos from remote (via button, KB shortcut whatever)
- Ensure videos loaded
- Switch to profile A
- Ensure **correct** videos loaded, no network request made for fetching videos
- Switch to profile B
- Ensure **correct** videos loaded, no network request made for fetching videos
- Switch to profile C
- Ensure **correct** (no) videos loaded, no network request made for fetching videos
- Switch to default profile
- Ensure **correct** (all) videos loaded, no network request made for fetching videos

A2a. Refresh subscriptions effects - With default profile loaded first
- Switch to default profile (if not already)
- Load videos from remote
- Ensure videos loaded
- Switch to profile A
- Load videos from remote
- Switch to default profile
- Ensure **all** videos loaded, no network request made for fetching videos
  (due to only cache for channels in profile A refreshed, but cache for not refreshed channel are not purged)
- Switch to profile A
- Ensure **correct** videos loaded, no network request made for fetching videos
- Switch to profile B
- Ensure **correct** videos loaded, no network request made for fetching videos

A2b. Refresh subscriptions effects - Without default profile loaded first
- Switch to default profile (if not already)
- **DO NOT** load videos from remote
- Switch to profile A
- Load videos from remote
- Switch to default profile
- Ensure **NO** videos loaded
  (due to not all channels are cached)
- Switch to profile A
- Ensure **correct** videos loaded, no network request made for fetching videos
[...]

PR #4380, opened 2023-11-25
Show when Subscriptions / Trending / Most Popular were last updated

### LINKED ISSUE #1010: Show when subscriptions were last updated
I suggest showing the date and time of the latest subscriptions refresh time. See screenshot below:

last updated

jasonhenriquez:
# Show when Subscriptions / Trending / Most Popular were last updated

## Pull Request Type

- [x] Feature Implementation

## Related issue

closes #1010

## Description
- Updates refresh button to be `ft-refresh-widget` component with new full-width bar styling and the timestamp of when the data was fetched from (re-calculated each time page is loaded or data is updated)
- Adds `ft-refresh-widget` to Most Popular and Trending (as well as new titles - none existed)
- Separate timestamps being stored for Subscription Videos, Live Videos, Shorts, & Community Posts, Trending, and Most Popular

## Screenshots 
Screenshot_20231124_190539

## Testing 
- Test on desktop, tablet, and mobile views
- Test that buttons work properly on Subscription Videos, Live Videos, Shorts, & Community Posts, Trending, and Most Popular

## Desktop
- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

## Additional context
The values are stored in DB rather than local storage, which is not too valuable of a distinction at the moment, as we don't currently cache this data across different sessions. This means the main beneficiary of this distinction are users with "Fetch Feed Automatically", who get to see this information:

Screenshot_20231124_191630

With that said, I do think that small amount of benefit & other prospective future benefits are enough to not rewrite to update this code to do it through local storage. For example, it will be even more benefit to these users if we ever implement caching of these results across sessions. If we never do, that's fine too, but just worth noting.

Incidentally...  "Fetch Feed Automatically" being disabled has no bearing on "Most Popular" and "Trending," which doesn't sound right. But that sounds like a separate issue.


PikachuEXE:
Not tested yet, but I think the data entries for caching should be inside `src/renderer/store/modules/utils.js` instead of src/renderer/store/modules/settings.js

PikachuEXE:
Unexpected refresh button position initially
image

Also is not reasonable to show the last feed update time when nothing is shown (after opening a new window / reopening the app)
image

jasonhenriquez:
> Not tested yet, but I think the data entries for caching should be inside src/renderer/store/modules/utils.js instead of src/renderer/store/modules/settings.js

> Also is not reasonable to show the last feed update time when nothing is shown (after opening a new window / reopening the app)

See discussion in Additional Info: to reiterate, I'm thinking it might be extra useful to have these in long-term state if we decide we want to have long-term state caching for, say, subscriptions. That, and I do think it's still decently neat to have that info as a tidbit for users with that setting enabled. But am fine to change it if there is reasoned rebuttal given and/or agreement from others

PikachuEXE:
But only useful for those with `Fetch Feed Automatically` enabled (and in that case it would be generated anyway instead of from local db). But in my case I have that setting disabled and nothing is shown. So it's unreasonable to see last refresh time without the actual data.

Will re-test soon (I am on Windows now)

absidue:
> Incidentally, "Fetch Feed Automatically" being disabled has no bearing on "Most Popular" and "Trending," which doesn't sound right. But that sounds like a separate issue.

Considering that setting is in the "Subscription Settings", as a user I would be very surprised if it applied to Trending and Popular. Additionally it's not necessary to have that setting apply to anything but subscriptions, because you'll only visit trending and popular when you actually want to see content. The subscriptions page is the default page at startup (unless configured otherwise), so firstly people might want to change profile before fetching subscriptions and secondly they might want to select a tab other than videos first. Also don't forget the subscriptions page makes loads of requests, whereas the Trending and popular pages only make a few.

TL;DR, subscriptions is a special page so it makes sense that that is the only page with an option to disable automatically fetching content.

src/renderer/components/subscriptions-tab-ui/subscriptions-tab-ui.js
```diff
@@ -51,6 +55,10 @@ export default defineComponent({
     }
   },
   computed: {
+    currentLocale: function () {
+      return this.$i18n.locale.replace('_', '-')
+    },
+
```
absidue, line 61: Looks like this is unused? I'm guessing it's leftover from testing?
```suggestion
```

absidue (commented):
In addition to the other comment, it looks like you've added class="ref" in a couple of places, but can't see where it is referenced.

PikachuEXE:
Also now I remember https://github.com/FreeTubeApp/FreeTube/pull/3668
The "last data loading time" is much more complicated
Saving a single entry won't do the job in different profiles

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Every subscriptions tab got its own separate `Feed last updated`. To me it makes sense but maybe to a normal user not. Maybe we should specify it a bit more `Video feed last updated`, `Shorts feed last updated`, `Live feed last updated` and `Community feed last updated`

src/renderer/components/subscriptions-tab-ui/subscriptions-tab-ui.js
jasonhenriquez: Yes, thank you - as for the first one, that was a relic from before that also wasn't being used as far as I could see, but yeah, definitely need to interrogate those more rather than taking an overly cautious stance

jasonhenriquez:
> Also now I remember #3668 The "last data loading time" is much more complicated Saving a single entry won't do the job in different profiles

Can you take a screenshot of an existing case where the current logic is returning an inaccuracy? Had trouble grokking the proper invariant in the code to check if things had been updated from reading that thread

absidue:
Haven't tested this pull request yet but based on the code, you only save one last refreshed timestamp, so if you have multiple profiles that were refreshed at different times, this will only keep the latest across all profiles, which makes the timestamp useless because it doesn't match what is visible on screen when it uses the cached data.

absidue:
Also to make it even more complicated, we cache each channel separately, so if you have profiles with overlapping channels, you could be showing data that was cached at different times on the same page.

jasonhenriquez:
So I can see that the first suggestion is to store these data on a per-profile basis. For the second problem, there are cases where the first load will potentially be returning only cached data, in which case the label is inaccurate. And we need to be able to discern this case so that we do not update the timestamp when it occurs. Am I understanding you correctly?

absidue:
Here is an theoretical situation where you could end up with different refresh times in the current profile (untested as I haven't downloaded this pull request yet):

Assuming you have automatic subscription fetching disabled and this profile setup:
- All Channels (not relevant but including it for the sake of completeness)
    - Channel A
    - Channel B
    - Channel C
- Profile 1
    - Channel A
    - Channel B
-  Profile 2
    - Channel B
    - Channel C

0. As you have automatic fetching disabled the cache is currently empty
1. Switch to Profile 1
2. Fetch subscriptions
3. Wait 5 minutes
4. Switch to Profile 2
5. Fetch subscriptions
6. Switch to Profile 1
7. We are now in a situation where the refresh times are different for the channels in the currently active profile, for A it is 5 minutes ago, for B it is just now

jasonhenriquez:
Gotcha, I think that's encapsulated in my Edit 2. As a note, before I start implementing this, would it be more or less convenient/understandable to know whether a given profile Y obtained the latest of the currently visible subscriptions on X date (as is being proposed), or just to clarify in the label that this is the last timestamp of an update from any profile? Just trying to assess the relative loss in utility versus the time investment & potential risk of being more confusing, kinda, if you just wanted to know the last time you requested new subscription data from any profile.

PikachuEXE (commented):
It's strange to see shadow from the "bar" when has no spacing between itself and stuff below, add some spacing?
image

PikachuEXE:
Already found an issue
- Switch to profile A with 1 sub only
- Refresh
- Switch to profile with all sub & no videos shown (coz not all videos cached for all subs in profile)
- Still see last updated time

I don't expect to see any last updated time when there is no video shown
image

jasonhenriquez:
>     * Switch to profile with all sub & no videos shown (coz not all videos cached for all subs in profile)

Sorry, not sure if I understand. The intended behavior is to show no videos if it's not all of the subscriptions included? Why is that?

PikachuEXE:
See test case `A2b` in https://github.com/FreeTubeApp/FreeTube/pull/3668

jasonhenriquez:
> Why is that?

jasonhenriquez:
I notice as a result of that behavior, there's a weird case where subscribing to a new profile with "Fetch Feed Automatically" hides the existing feed (which happens to All Channels as well, naturally). Surely that can't be the desired behavior? Still added it in the latest commit because I'm not here to break practices, but it brings some odd behaviors if we want to have that as one.

src/renderer/components/subscriptions-community/subscriptions-community.js
```diff
@@ -69,19 +74,28 @@ export default defineComponent({
...
     this.isLoading = true
 
-    this.loadpostsFromCacheSometimes()
+    this.loadPostsFromCacheSometimes()
   },
   methods: {
-    loadpostsFromCacheSometimes() {
+    loadPostsFromCacheSometimes() {
       // This method is called on view visible
       if (this.postCacheForAllActiveProfileChannelsPresent) {
```
jasonhenriquez, line 88: Edit: To more explicitly clarify what I would believe to be the logical behavior here, I think we should have this if statement condition be modified to be `!this.updateFetchSubscriptionsAutomatically || this.postCacheForAllActiveProfileChannelsPresent`. And that I revert my last commit to move the minTimestamp logic outside of this if-statement as was before, as this case would no longer exist as is.

PikachuEXE:
Can the bar not disappear and re-appear on refresh? 
Feels weird

image
image

PikachuEXE:
Which one are you referring to?

jasonhenriquez:
> Sorry, not sure if I understand. The intended behavior is to show no videos if it's not all of the subscriptions included? Why is that?

> See test case A2b in https://github.com/FreeTubeApp/FreeTube/pull/3668

> I notice as a result of FreeTube having that behavior, there's a weird case where subscribing to a new profile with "Fetch Feed Automatically" disabled hides the existing feed (which happens to All Channels as well, naturally). (Testing on development branch; this is pre-existing behavior separate from this PR, to clarify.) Surely that can't be the desired behavior? Still accommodated it in the latest commit because I'm not here to break practices, but it brings some odd behaviors if we want to have that as one.

> Edit: To more explicitly clarify what I would believe to be the logical behavior here, I think we should have this if statement condition be modified to be !this.updateFetchSubscriptionsAutomatically || this.postCacheForAllActiveProfileChannelsPresent. And that I revert my last commit to move the minTimestamp logic outside of this if-statement as was before, as this case would no longer exist as is.

PikachuEXE:
`subscribing to a new profile with "Fetch Feed Automatically" disabled hides the existing feed`
is expected since the subscription cache (video/live/shorts/etc.) no longer contain stuff for all subscribed channels in active profile
If we choose to keep showing cached entries when new channel subscribed, there no way to know some entries are missing

PikachuEXE:
I would love to continue the discussion on behaviour change for subscription view but is it blocking this PR from being merged? (It's not merged yet)
I don't want the discussion here distracting other reviewers from reviewing it as is (i.e. no subscription behaviour change)

Maybe @jasonhenriquez you can make a draft PR in your own repo so we can test and discuss without disturbing others

(Unless this is a PR blocking issue)

jasonhenriquez:
I'm fine with this PR as is now.

PikachuEXE (dismissed):
Reviewed the behaviour, but not all the code (my eyes & brain tired)

PikachuEXE:
Re-tested and code reviewed except
The spacing between banner & new widget probably should be improved

image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE did you test all of the following cases:
https://github.com/FreeTubeApp/FreeTube/pull/3668
https://github.com/FreeTubeApp/FreeTube/pull/4380#issuecomment-1826378534
https://github.com/FreeTubeApp/FreeTube/pull/4380#issuecomment-1827035044

?

PikachuEXE:
Let me re-test
But notice that the PR now does not change any subscription related items fetching behaviour
i.e. Only need to test value displayed with different profiles with partial shared channels

- Create 2 profiles (using existing one is fine too)
- Add channel A&B to profile 1, B&C to profile 2
- Switch to profile 1, refresh, wait 2-3 minutes, switch to profile 2, refresh
- Ensure profile 1 show N minutes ago and profile 2 show a few moments ago (something like that not exact text)

Is this what you mean @efb4f5ff-1298-471a-8973-3d47447115dc anything missing/incorrect?

jasonhenriquez:
>  The spacing between banner & new widget probably should be improved

Addressed this now.
Screenshot_20240205_180411

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> Here is an theoretical situation where you could end up with different refresh times in the current profile (untested as I haven't downloaded this pull request yet):
> 
> Assuming you have automatic subscription fetching disabled and this profile setup:
> 
>     * All Channels (not relevant but including it for the sake of completeness)
>       
>       * Channel A
>       * Channel B
>       * Channel C
> 
>     * Profile 1
>       
>       * Channel A
>       * Channel B
> 
>     * Profile 2
>       
>       * Channel B
>       * Channel C
> 
> 
>     0. As you have automatic fetching disabled the cache is currently empty
> 
>     1. Switch to Profile 1
> 
>     2. Fetch subscriptions
> 
>     3. Wait 5 minutes
> 
>     4. Switch to Profile 2
> 
>     5. Fetch subscriptions
> 
>     6. Switch to Profile 1
> 
>     7. We are now in a situation where the refresh times are different for the channels in the currently active profile, for A it is 5 minutes ago, for B it is just now

@absidue i think this behavior is fine and does represent an accurate view of the situation or am i missing something here

PikachuEXE:
I thought https://github.com/jasonhenriquez/FreeTube/pull/2 is merged into this
But now I recheck and it seems not so I guess those updates should be added first

PikachuEXE:
Also might need to test for https://github.com/FreeTubeApp/FreeTube/pull/4667 since that one is likely merged before this

efb4f5ff-1298-471a-8973-3d47447115dc:
rebase needed for #4667

PikachuEXE:
While testing https://github.com/FreeTubeApp/FreeTube/pull/4667 case
If I have a profile with 1 channel only and I visit that channel directly (not refreshing sub)
I see videos in sub page but no time displayed

image

jasonhenriquez:
> I see videos in sub page but no time displayed

Keen eye and thanks for the heads up @PikachuEXE. Somehow the existence of that PR just flew over my head. Fixed now.

Screenshot_20240410_205100

src/renderer/store/modules/subscriptions.js
```diff
@@ -69,19 +69,21 @@ const actions = {
 }
 
 const mutations = {
-  updateVideoCacheByChannel(state, { channelId, videos }) {
+  updateVideoCacheByChannel(state, { channelId, videos, timestamp }) {
     const existingObject = state.videoCache[channelId]
     const newObject = existingObject ?? { videos: null }
     if (videos != null) { newObject.videos = videos }
+    if (timestamp != null) newObject.timestamp = timestamp
```
PikachuEXE, line 76: Not suggesting change to mutation but about `updateSubscriptionVideosCacheByChannel`
Why not just make `updateSubscriptionVideosCacheByChannel` pass `timestamp: timestamp: new Date()` into `updateVideoCacheByChannel` (There is no use case for passing non `new Date()` into `updateSubscriptionVideosCacheByChannel`)

Same for other methods

See example at https://github.com/FreeTubeApp/FreeTube/blob/v0.20.0-beta/src/renderer/store/modules/playlists.js#L63-L64

PikachuEXE (commented):
Test for https://github.com/FreeTubeApp/FreeTube/pull/4667 passed but...

src/renderer/store/modules/subscriptions.js
jasonhenriquez: I was thinking about making that the default value if none was passed in. I don't really have a preference now that it's already all implemented. Give this reply a thumbs up and I'll do it.

PikachuEXE: I am fine with default value approach (though in what use case a non current time is needed...?)
Why not just thumbs up? Coz that doesn't send a notification

jasonhenriquez: For the first Q, I dunno. We have similar usecases with functions like `updateLastCommunityRefreshTimestampByProfile` where we pass a `minTimestamp`, so you never know. For the second Q, I'm fine with checking the tab later; I just don't want anyone to feel like they're being demanded.

PikachuEXE (approved):

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -667,48 +664,7 @@ export default defineComponent({
...
-
-          if (timeUnit === 'month' && timeDiffFromNow >= 12) {
-            timeDiffFromNow /= 12
-            timeUnit = 'year'
-          }
-
-          // Using `Math.ceil` so that -1.x days ago displayed as 1 day ago
-          // Notice that the value is turned to negative to be displayed as "ago"
-          this.uploadedTime = new Intl.RelativeTimeFormat([this.currentLocale, 'en']).format(Math.ceil(-timeDiffFromNow), timeUnit)
+          this.uploadedTime = getRelativeTimeFromDate(new Date(this.data.published).toDateString())
```
absidue, line 667: Please revert the changes in this file, as it needs different logic for the relative date formatting, than the other places that you use the `getRelativeTimeFromDate` function.

The logic here handles weeks and specifically uses 30 day months (as mentioned in the code comment), because the timestamps that get passed in here come from the relative date calcuations in `calculatePublishedDate` which also uses 30 day months.

absidue (commented):
I haven't tested the code yet, this is just a quick review to prevent this pull request from introducing a regression.

src/renderer/components/ft-list-video/ft-list-video.js
jasonhenriquez: Thanks for catching this, this was poor diligence on my part.

jasonhenriquez: I had missed #4752 entirely, and this PR was using the older logic. It doesn't really seem like we'll need 31 day months anymore (not that we were doing that correctly before that fix, anyhow), but I kept it in as an optional parameter that is currently unused.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Pass on #4667 but still need answers to https://github.com/FreeTubeApp/FreeTube/pull/4380#pullrequestreview-1868385066

jasonhenriquez:
> I think this should result in a blank status but instead im still seeing fetched x minutes ago message

Fixed now.

> Profile A sits on 5minutes ago, if switch to All channels fetch and go back to A it will show the same as All channels, is this expected?

Yes. All Channels has all channels. If you fetched from All Channels, that includes A as a subset of all channels. Thus, the min timestamp displayed on each profile is equivalent to the All Channels timestamp, until you were to subscribe to an additional channel not currently included in All Channels.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Most Popular always shows blank status

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM, nice work on this one!

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Uh oh found a bug

https://github.com/FreeTubeApp/FreeTube/assets/73130443/77f9f41a-480f-4f4a-b1a1-c28657126e6f

manifest-9:
Is there an option to disable or hide the toolbar after it refreshes?

efb4f5ff-1298-471a-8973-3d47447115dc:
@manifest-9 No and IMO it shouldn't be hidden especially when something like https://github.com/FreeTubeApp/FreeTube/issues/4803 is going to be added at some point

PR #4391, opened 2023-11-26
Add support for viewing movie trailers with local api

ChunkyProgrammer:
# Add support for viewing movie trailers with local api

## Pull Request Type
- [x] Feature Implementation

## Related issue
Going to a video uploaded by `YouTube Movies & TV` would cause an error and nothing would be displayed.

## Description
This PR will play the preview for movies. 

## Testing
1) Go to https://www.youtube.com/watch?v=BfgE0HczUco (with local api)
2) search `Scott Pilgrim vs the World movie`, click on the video by `YouTube Movies & TV`
image
image

## Desktop
- **OS:** Windows
- **OS Version:** 10


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Sometimes it does load the trailer but sometimes i have the feeling that it is really trying to fetch to full movie, resulting in this error dump

Capture

ChunkyProgrammer:
> Sometimes it does load the trailer but sometimes i have the feeling that it is really trying to fetch to full movie, resulting in this error dump
> 
> Capture

What video did you get this error with?

efb4f5ff-1298-471a-8973-3d47447115dc:
Tested with the URL you provided in the PR body

ChunkyProgrammer:
@efb4f5ff-1298-471a-8973-3d47447115dc are you using a vpn or a proxy? I was able to reproduce with a vpn (the movie wasn't available in the region my vpn was in)

absidue:
@ChunkyProgrammer Sounds like you need to also check the playability status of the trailer, the current code just assumes it always playable.

efb4f5ff-1298-471a-8973-3d47447115dc:
Yes i was using a VPN

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

For anyone running into the same issues as me, use VPN server in Canada

jasonhenriquez:
**question:** To confirm, is this what we want to happen when there is an error displaying the content, due to VPN issues or something like that?

Screenshot_20231127_173414

jasonhenriquez (dismissed):
Can confirm that it works as intended (tested with Canadian VPN server).

efb4f5ff-1298-471a-8973-3d47447115dc:
> **question:** To confirm, is this what we want to happen when there is an error displaying the content, due to VPN issues or something like that?
> 
> Screenshot_20231127_173414

yup had the same issue on non Canadian server https://github.com/FreeTubeApp/FreeTube/pull/4391#pullrequestreview-1749445893

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc But is that the same as what happened before with unplayable videos or is that collection of errors a result of this pull request?

efb4f5ff-1298-471a-8973-3d47447115dc:
@absidue 

1. UNPLAYABLE error i also got before this PR
2. Vue warn error in mounted hook is new in this PR
3. typeError bitrate error i also got before this PR

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (dismissed):
Works with Canadian VPN server
But can't access in some other regions

absidue:
Might be best to always show the trailer if one exists, not just if there is an error, because free with ads movies exist and because they have DRM we can't show them either (Electron doesn't have Widevine out of the box, because it's closed source).

PikachuEXE (dismissed):

src/renderer/helpers/api/local.js
```diff
@@ -609,8 +609,8 @@ export function parseLocalListVideo(video) {
     author: video.author.name,
     authorId: video.author.id,
```
absidue, line 610: Movies don't have a channel ID on YouTube in search results, just a channel name, so this becomes `N/A` resulting in invalid channel links getting generated.

```suggestion
    video.author.id !== 'N/A' ? video.author.id : null,
```

You'll also have to update `ft-list-video` with a `v-else-if`, so that it shows the channel name without a link, if there is a channel name but no ID (both can be missing too, for example in shorts only playlists, which is why you can do just `v-else`).

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
What's changed since I last tested~

absidue:
The appearance and functionality of the movie items in search results (they had a broken channel link before, now it just show the text and the description will show up for them in the list layout).

PikachuEXE:
Got this error with Japan VPN searching `Scott Pilgrim vs the World movie`
image

src/renderer/helpers/api/local.js
```diff
@@ -537,22 +537,44 @@ export function parseLocalPlaylistVideo(video) {
...
+      title: movie.title.text,
+      author: movie.author.name,
+      authorId: movie.author.id !== 'N/A' ? movie.author.id : null,
+      description: movie.description_snippet?.text,
+      viewCount: null,
+      publishedText: null,
+      lengthSeconds: isNaN(movie.duration.seconds) ? '' : movie.duration.seconds,
+      liveNow: false,
+      isUpcoming: false,
+      premiereDate: null
```
absidue, line 559: This fixes the error
```suggestion
      lengthSeconds: isNaN(movie.duration.seconds) ? '' : movie.duration.seconds,
      liveNow: false,
      isUpcoming: false,
```

PikachuEXE (approved):
No more error seen when searching `Scott Pilgrim vs the World movie` (same VPN server)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
lgtm

Issue #1362, opened 2021-05-28
Way to dismiss videos or mark them as "Not Interested"


lima8zulu:
It would be great if there was a way to mark videos as "Not Interested" (to watch) or a way to dismiss or hide them. 

Currently I hide watched videos. If there's a video in my feed I don't want to watch I mark is as read to make it disappear, but it messes up with my watch history. It would be great if there was an option to just hide videos without marking them as watched.

lagapi:
I believe something to easily mark a video as watched or non interesting would be very useful. The same way we have a star button at the right of every thumbnail video, it could be added another button in the left part for this. Maybe a little circle like the read/unread button from thunderbird.

pseudomonas:
I don't want to hide watched videos (I like to re-watch) but sometimes there's a video with a thumbnail or title that I find distressing, and while on YouTube it's easy to hide, on Freetube it reqiuires workarounds. 

There might need to be some kind of interface to undo the hiding, though I guess this could be just a button in the settings that removes _all_ videos from being hidden, similar to deleting all history and marking all videos as unwatched?

efb4f5ff-1298-471a-8973-3d47447115dc:
Maybe a weirs suggestion but

What about adding the whole title of the video to the blacklist by clicking on a hide video button in the dropdown options

jasonhenriquez:
@efb4f5ff-1298-471a-8973-3d47447115dc I can see why you're saying that as that would be very easy to implement after #4202, but I am a bit worried about the imperfections of that solution w.r.t. unintuitive and undesirable byproducts of it (e.g., you block a video titled "a", and now you can't see most videos, or the title is a generic one word like "Music" and you now stop seeing videos with that in the title).

pseudomonas:
I think if I were implementing this, from looking a little at the code, I'd copy the "history"/"watched" mechanism so one could have a parallel to  `.historyEntryExists()` and  `.markAsWatched()` methods.  

Then one could hide videos on the `hidden` list as though one was hiding already-watched videos.

efb4f5ff-1298-471a-8973-3d47447115dc:
[...]

Issue #547, opened 2020-10-02
[Feature Request] Favorites + Watch Later + Dynamic Playlists/Watch Queue - not just one


deepspaceaxolotl:
I can say I really appreciate the direction Freetube is going in terms of customisability, it's quite heartening. On that topic, I would like to pitch in to keep the Watch Later button in the corner of videos/as an option under the player alongside that tab, as well as, rather than in place of, the upcoming on-the-fly playlists. I never used the Watch Later feature on YouTube, but have found it very useful on Freetube, especially with navigating in only one tab, and for keeping up with many subscriptions.

A separate Favourites tab would also be nice so I don't have to mix my 'bookmarked' videos and those I want to watch later, but just the functionality that's present in the old app would be enough for me to upgrade right away. (And exporting this list/these lists would be great! Another reason I sadly can't test out the new rewrite, losing my watch later videos.)

And of course, switches in the settings for showing or hiding those tabs (in conjunction with other distraction-free mode options) would be a great idea.

GilgusMaximus:
Thanks for the appreciation. First I'd like to say that I would appreciate it a lot if the issue templates are used in the future. Makes everything tidier.

1) The dynamic playlist was a proposal from #413 and #519. Its concrete implementation is yet to be discussed and probably 2 releases away with the playlists coming up just in the next release. Your idea only has differences at a few points. So your proposal will definitely be included in the discussion. 

2) With custom playlists in the next release you can make as many playlists as you want, including a favorites playlist. An additional tab wouldn't differ much from a normal playlist, However the idea of adding a favorite button so the video gets added to a favorites playlist could be a good compromise.

3) That depends on the concrete implementation of the UI. That is usually something Preston does.

heartlessglasses:
Oh, apologies! Am still a bit confused with the etiquette here, this wasn't really a bug report so I just went with a paragraph style... I'll keep it in mind for next time if it comes up.

Thank you for the response, and for linking those issues about custom playlists! Yeah, I'm liking the direction of flexibility you are heading in. I guess tabs aren't that big of a thing really, just would miss the button in the corner of video thumbnails as it's super easy to access, which is good for a click-and-forget feature like Watch Later. ^^;

nothub:
I have placed a bounty on this issue ticket via bountysource.com, I would really enjoy to have list that I can throw some videos into directly from the subscriptions overview and then watch them all at once.

[...]

PR #4394, opened 2023-11-27
Order watched recommended videos last

### LINKED ISSUE #397: Enhancement : Skip Play Next Video when already in History
Play next video is cool, but youtube give you ever and ever the same videos. It would be nice to discard already seen videos from auto play.

jasonhenriquez:
# Order watched recommended videos last

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #397

## Description
Orders watched videos last in the recommended videos list. Only sets ordering once when route is first loaded, not live during any "Mark as Watched" / "Mark as Unwatched" events (because this would be undesirable). This should functionally make infinite autoplay loops impossible.

## Testing 
- Enable autoplay and mark videos at the top of the rec list as watched
- Reload route and see that those watched videos are at the bottom of the rec list
- Let video autoplay and see that the video visibly at the top of the rec list is played

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -281,6 +281,12 @@ export default defineComponent({
           ?.filter((item) => item.type === 'CompactVideo')
           .map(parseLocalWatchNextVideo) ?? []
 
+        // place watched recommended videos last
+        this.recommendedVideos = [
+          ...this.recommendedVideos.filter((video) => !this.isRecommendedVideoWatched(video.videoId)),
+          ...this.recommendedVideos.filter((video) => this.isRecommendedVideoWatched(video.videoId))
+        ]
+
         if (this.showFamilyFriendlyOnly && !this.isFamilyFriendly) {
           this.isLoading = false
           this.handleVideoEnded()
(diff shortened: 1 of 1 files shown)


PikachuEXE (dismissed):

jasonhenriquez:
I was more referring to the physical ordering aspect, which would be jarring to have it jump out of view if you marked it as watched. Detaching the visible display order of the videos from how they would be autoplayed seems unintuitive. I think there's an interesting use case for manipulating the autoplay order on the go, but that would fall in line with either a Queue feature (#547) and/or hiding videos (#1362).

efb4f5ff-1298-471a-8973-3d47447115dc:
To me it feels a bit like unexpexted behavior when i mark the first video in the list as watched and still get it when the autoplay kicks in. I feeling could be wrong though.

jasonhenriquez:
> To me it feels a bit like unexpexted behavior when i mark the first video in the list as watched and still get it when the autoplay kicks in

I think that would make sense if it was a labeled feature like "Keep watched videos last," but as it stands, I think reordering the recommended videos list based on Mark as Watched is more jarring & not in line with user expectations of how things would work. Open to hearing other perspectives on this though!

PikachuEXE:
Right now I would say it's unnecessary to move entries marked as watched to the bottom
If they want to play non 1st item, just click on it

PikachuEXE:
So more like on video end
Pick 1st recommended **not marked as watched** video to autoplay? (fallback to 1st video if all watched)

efb4f5ff-1298-471a-8973-3d47447115dc:
Yup exactly like that. If for some reason i see a video on top of the recommended videos list but i already seen it. The user marks it as watched. When its time for the autoplay to go to the next video it should skip that one on the top i just marked as watched and pick the 2nd one in line

jasonhenriquez:
I don't think that's intuitive behavior unless you're expecting that to happen & behave like that already, which most people wouldn't be.

efb4f5ff-1298-471a-8973-3d47447115dc:
Good point!

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue:
Could you please use a local/temporary variable and only assign the final value to `this`, instead of assigning and then overwriting the value in `this`?

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #4405, opened 2023-11-29
Fix style for the Select component

### LINKED ISSUE #1804: Dropdowns arent working properly
**Behavior of the program**
When u click on some dropdowns they wont open.

**Expected behavior**
Dropdowns open properly like all other dropdowns

**To Reproduce**
1. Go to settings page
2. Click on General Settings to fold out
3. Click on Preferred API Backend or Thumbnail Preference

**Link to video**

https://user-images.githubusercontent.com/73130443/136675760-9affbb3b-632c-401b-bad0-28d7f786100e.mp4

**Screenshots**
None.

**Environment Information (please complete the following information):**
OS: [Windows 10]
OS Version: [Windows 10 20H2]
FreeTube version: [v0.15.0-nightly-1266 Beta]
Installation Method (If applicable): [Portable]
Primary API Used: [Local API]

**Additional context**
None.

pkrasicki:
# Fix style for the Select component

## Pull Request Type
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Fixes #1804

## Description
Fixes the width of dropdowns and changes their style a little bit.

## Screenshots 
light theme

dark theme

## Testing 
Testing dropdowns manually on settings page, channel view and in comment section with different window sizes.

## Desktop

- **OS:** Any
- **OS Version:** Any
- **FreeTube version:** 0.19.1 dev

## Additional context
None.

### DIFF
--- a/src/renderer/components/ft-select/ft-select.css
+++ b/src/renderer/components/ft-select/ft-select.css
@@ -24,8 +24,6 @@
 .select {
   position: relative;
   inline-size: 200px;
-  padding-block: 0 10px;
-  padding-inline: 0 10px;
   margin-block-start: 30px;
 }
 
@@ -37,13 +35,13 @@
 .select-text {
   position: relative;
   font-family: inherit;
-  background-color: transparent;
+  background-color: var(--search-bar-color);
   color: var(--primary-text-color);
-  inline-size: 240px;
-  padding-block: 10px;
-  padding-inline: 0 10px;
-  font-size: 18px;
-  border-radius: 0;
+  inline-size: 100%;
+  block-size: 45px;
+  padding-inline-start: 1rem;
+  font-size: 16px;
+  border-radius: 5px;
   border: none;
 }
 
@@ -62,13 +60,13 @@
   appearance: none;
   -webkit-appearance:none;
   text-overflow: ellipsis;
-  padding-inline-end: 1.1rem;
+  padding-inline-end: 1.5rem;
 }
 
 .iconSelect {
   position: absolute;
   inset-block-start: 10px;
-  inset-inline-end: 15px;
+  inset-inline-end: 10px;
   /* Styling the down arrow */
   padding: 0;
   content: '';
@@ -80,8 +78,8 @@
 
 .selectTooltip {
   position: absolute;
-  inset-block-start: -20px;
-  inset-inline-end: 17px;
+  inset-block-start: -22px;
+  inset-inline-end: 0;
 }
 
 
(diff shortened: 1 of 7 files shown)


jasonhenriquez:
I'll check this out. FYI, you can run `npm run lint-fix` to resolve the linting issues.

jasonhenriquez (commented):
**praise:** I think it's a positive change. It is more visually prominent, which some may not like, but I think that's ultimately a better choice for accessibility. It was sometimes difficult to see how wide the dropdown & its click area was before. I also did not encounter the original bug, so that looks good as well.

**issue (pre-existing):** This isn't a new issue with the `ft-select` component, but I do see that this particular one is unclickable on its bottom half (it's at the top of the comments section for a given video). This is most likely related to its placement rather than the component itself, but take a jab at this if you don't mind.
Screenshot_20231129_075853

**chore (blocking):** Just run the `npm run lint-fix` command (may have to do a soft git reset first) to resolve the linting issues.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Noticed that the tooltip placement moved from just above the dropdown arrow. 

ft

Capture

efb4f5ff-1298-471a-8973-3d47447115dc:
Maybe im a bit biased because i got used to that alignment but to me it looks more visually pleasing. If this is the better way to do it than we should definitely keep this. Lets wait for others to have a say in it.

pkrasicki:
Btw, I think it should be in the same position for dropdowns and input fields.

jasonhenriquez:
I think both look fine, although after looking at it for too long, the difference in right alignment is actually quite vexing to me. I'd probably prefer changing it, and see if making the left-aligned label have the same margin as the right makes things better or worse (probably worse, if I were to wager). I think the differing placement of the tooltips compared to `ft-input` is fine and makes sense based on width if the item below is addressed, but you can experiment with that.

**nitpick (pre-existing):** These should probably have a `max-inline-size` to prevent this type of stretching behavior. Ideally, this would be configurable as a `prop` to allow overrides.

Screenshot_20231130_073016

jasonhenriquez:
Hi @pkrasicki, are you still good to implement that final suggestion?

pkrasicki:
@jasonhenriquez done.

moved tooltip

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez (approved):

PikachuEXE:
Issue spotted but already exists in `dev` branch (i.e. not introduced by this PR)

at width ~901px
image

So the only diff is the select got a box like style?

jasonhenriquez:
It also fixes the click area to match its visual area. Like when trying to click between the area between the bottom border and the arrow here.
Screenshot_20240102_185221

PikachuEXE (approved):

Issue #2768, opened 2022-10-25
[Bug]: Video meta file removal is unreliable again


ghost:
### Guidelines

- [ ] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for open and closed issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.

### Describe the bug

Some video meta files aren't removed with the video meta file removal option checked.
Same as #1562. It was fixed but is happening again, though rarely, since a few versions.

### Expected Behavior

All video meta files are always removed.

### Issue Labels

inconsistent behavior

### FreeTube Version

0.17.1-nightly-2189

### Operating System Version

Windows 11 22H2

### Installation Method

Portable

### Primary API used

Local API

### Additional Information

See #1562.

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
I'm guessing that you close FreeTube while you are on the video page?

PR #4423, opened 2023-12-04
Various small watch page changes

absidue:
# Various small watch page changes

## Pull Request Type

- [x] Other - Refactor

## Related issue
#2768

## Description
This pull request contains a few changes that I made while doing the player migration, which I decided to stick in a separate pull request so that the main one can be slighly smaller.

The changes:
- Use the existing `getProxyUrl` function to transform local streaming URLs to proxied ones for the locally generated Invidious DASH manifest, reduces duplicate code and it does a few more checks, so it should be safer/more reliable.
- Use `.split()` instead of a regex match to get the mimetype for the download labels
- Simplify the `checkIfTimestamp` function and fix `NaN` bug, the try-catch was unnecessary as `parseInt()` returns `NaN `instead of throwing an error and `this.$route.query` is always defined. When there are no query parameters `this.$route.query` is just an empty object, not `undefined`.
- Create a data URI for the local API DASH manifest instead of writing it to the disk, we already do that for the locally generated Invidious DASH manifest and it's worked fine so far :). Unfortunately we can't switch the storyboard's over to a data URI yet, as we haven't done a videojs-vtt-thumbnails release in a long time, so my change to support data URIs over there hasn't been released yet (although the switch is already done in the player migration)

## Testing 
- Watch a video with DASH with proxying enabled and disabled
- Click the download button and check that the mime types are still displayed the same
- If you can get timestamp stuff to work feel free to try it, but it only ever works randomly for me because it currently relies on a hardcoded timeout

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -9,13 +9,15 @@ function getCurrentInstance() {
 }
 
 export function getProxyUrl(uri) {
+  const currentInstance = getCurrentInstance()
+
   const url = new URL(uri)
   const { origin } = url
-  if (!url.searchParams.has('host') && origin !== getCurrentInstance()) {
+  if (!url.searchParams.has('host') && origin !== currentInstance) {
     // invidious requires host param to be filled with the origin of the stream
     url.searchParams.append('host', origin.replace('https://', ''))
   }
-  return url.toString().replace(origin, getCurrentInstance())
+  return url.toString().replace(origin, currentInstance)
 }
 
 export function invidiousAPICall({ resource, id = '', params = {}, doLogError = true, subResource = '' }) {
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE:
How to test timestamp thing?

Others work fine

absidue:
If you want to check that the new code in `checkIfTimestamp` works, you could add `console.log(this.timestamp)` in `mounted` after the call to `checkIfTimestamp()`.

If you want to test if the timestamp actually seeks the video, well you'll just have to keep trying it over and over again and hope that it randomly works on one of your tries, that's the same before and after this pull request.

PikachuEXE (approved):
Oh now I realize it's just URL like
https://youtu.be/oeO-WTskmQw?t=162

It works like before

PR #4428, opened 2023-12-07
Vue3 Migration

ChunkyProgrammer:
# Vue3 Migration

## Pull Request Type
- [x] Other - migrate to vue3

## Description
This PR migrates FreeTube to Vue3. From what I can tell, this mostly works but I'd like to do some more investigation and testing (hence the draft PR). I might close this PR if I don't have enough time to fix it but the existing code should be useful for anyone who wishes to take a stab at the vue 3 migration

upgrade vue to 3.3.10
upgrade vue-i18n to 9.8.0
upgrade vue-observe-visibility to 2.0.0-alpha.1
upgrade vue-router to 4.2.5
upgrade vuex to 4.1.0

## Screenshots 
image

## Testing 
- daily drive, if an error occurs that you cant reproduce in the development branch then report it

## Desktop
- **OS:** Windows
- **OS Version:** 10


src/renderer/router/index.js
```diff
@@ -15,9 +14,8 @@ import Channel from '../views/Channel/Channel.vue'
 import Watch from '../views/Watch/Watch.vue'
 import Hashtag from '../views/Hashtag/Hashtag.vue'
 
-Vue.use(Router)
-
-const router = new Router({
+const router = createRouter({
+  history: createWebHistory(),
```
absidue, line 18: We are currently using hash history (default in the old vue-router https://v3.router.vuejs.org/api/#mode), so you should do the same here.

src/renderer/router/index.js
```diff
@@ -15,9 +14,8 @@ import Channel from '../views/Channel/Channel.vue'
 import Watch from '../views/Watch/Watch.vue'
 import Hashtag from '../views/Hashtag/Hashtag.vue'
 
-Vue.use(Router)
-
-const router = new Router({
+const router = createRouter({
```
absidue, line 17: According to the docs scrollBehavior changed from using `x` and `y` to using `left` and `top`, so this should probably use those here too: https://router.vuejs.org/guide/migration/#scrollBehavior-changes

sorry, i've looked at that migration page a few too many times 🙈

absidue (commented):
Haven't done any testing or a proper code review yet.

_scripts/webpack.renderer.config.js
```diff
@@ -134,8 +136,6 @@ const config = {
   ],
   resolve: {
     alias: {
```
absidue, line 138: Looks like webpack is picking the wrong export for the electron renderer build, it's picking the correct one for the web build though, guessing the combined web+node environment confuses it.
very-wrong-vue
wrong-vue




```suggestion
    alias: {
      vue$: 'vue/dist/vue.runtime.esm-bundler.js',

```

absidue:
Refreshing subscriptions errors in `formatNumber`.

absidue:
The build warnings can be fixed by removing the `set` and `del` imports and replacing them with direct object modification, as Vue 3 can properly detect property changes.

`set(x, y, z)` can be replaced by `x[y] = z`

`del(x, y)` can be replaced by `delete x[y]`

_scripts/webpack.main.config.js
```diff
@@ -36,6 +36,8 @@ const config = {
   },
   plugins: [
     new webpack.DefinePlugin({
+      '__VUE_PROD_DEVTOOLS__': true,
+      '__VUE_OPTIONS_API__': true,
```
PikachuEXE, line 40: 1

PikachuEXE (commented):
Notes for these new constants (global feature flags)
https://github.com/vuejs/core/blob/v3.3.11/packages/vue/README.md#bundler-build-feature-flags

src/renderer/App.vue
```diff
@@ -32,18 +32,17 @@
...
+        ref="router"
+        v-slot="{ Component }"
+        class="routerView"
       >
-        <!-- <keep-alive> -->
-        <RouterView
-          ref="router"
-          class="routerView"
-        />
+        <transition>
```
absidue, line 41:
```suggestion
        <transition
          mode="in-out"
          name="fade"
        >
```

absidue (commented):
To add back the fade transition, you will need to rename the CSS class in the App.css file as mentioned here: https://v3-migration.vuejs.org/breaking-changes/transition.html#migration-strategy

As I can't leave comments on unchanged files, here is the code diff for the App.ccs file:

```diff
- .fade-enter, .fade-leave-to /* .fade-leave-active below version 2.1.8 */ {
+ .fade-enter-from, .fade-leave-to {
  opacity: 0;
}
```

_scripts/webpack.renderer.config.js
```diff
@@ -115,6 +115,8 @@ const config = {
   plugins: [
     processLocalesPlugin,
     new webpack.DefinePlugin({
+      '__VUE_PROD_DEVTOOLS__': true,
```
absidue, line 118: Should this only be set to `true` for dev builds like we did before?

ChunkyProgrammer:
This PR is mostly complete as there may be some small bugs but code review is welcome

src/renderer/store/modules/utils.js
```diff
@@ -771,12 +770,12 @@ const mutations = {
     if (!sameVideo) {
       // setting properties directly doesn't trigger watchers in Vue 2,
       // so we need to use Vue's set function
```
absidue, line 772: This comment can be removed now, as it's about Vue 2.
```suggestion
```

src/renderer/i18n/index.js
```diff
@@ -31,20 +32,17 @@ export async function loadLocale(locale) {
...
     }
   } else {
     const url = createWebURL(`/static/locales/${locale}.json`)
 
     const response = await fetch(url)
-    const data = await response.json()
-    i18n.setLocaleMessage(locale, data)
+    data = await response.json()
   }
+  i18n.global.setLocaleMessage(locale, data)
```
absidue, line 45: Could you please revert this changes or add an early return, as the current code will attempt to set the local messages, even if an error occurred while reading them from the file system, which the previous code didn't.

src/renderer/main.js
```diff
@@ -1,11 +1,12 @@
 // import the styles
```
absidue, line 1: This comment can probably be removed as it doesn't seem to be relevant to anything happening near the top of this file anymore.
```suggestion
```

absidue (commented):
Could you please change this line to use `null` or `undefined` instead of `false`, refer to: https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html

https://github.com/FreeTubeApp/FreeTube/blob/695055dbd8f1a01b23558d2c00c2939f75639817/src/renderer/components/ft-list-lazy-wrapper/ft-list-lazy-wrapper.vue#L4

ChunkyProgrammer:
This PR would be a good starting point to look at when creating a Vue migration PR but this is not something that I would want to maintain at this time so I will be closing this

jasonhenriquez:
Thanks for the effort you did put into this @ChunkyProgrammer, as it is very good. Whenever you have the time to answer, would you be comfortable sharing the remaining (potential) challenges with a Vue 3 Migration as you understand it that was not covered in this PR? I guess I'm wondering if having this for v0.23 would be a feasible landmark in our roadmap.

PR #4431, opened 2023-12-08
split electron-builder config to a separate file from build script

selfisekai:
# Title

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

allows to use [electron-tasje](https://codeberg.org/selfisekai/electron_tasje) instead of _scripts/build.js (`tasje --config ./_scripts/ebuilder.js pack`).

## Testing 

manually

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

typical distro policies disallow packaging prebuilt binaries, in this case of electron (e.g. https://docs.fedoraproject.org/en-US/packaging-guidelines/what-can-be-packaged/#prebuilt-binaries-or-libraries). tasje takes an electron-builder config and packages the app without downloading anything


absidue:
Just a warning that if you aren't going to use the pre-bundled electron, make sure that you lock the version in the distro dependency to the same major version the FreeTube uses. Both AUR and NixOS made the mistake of not ensuring that the version is the same and then ended up with completely broken FreeTube installs (crashing and sitting on a blank window forever).

_scripts/ebuilder.js
absidue, line 1: Could you please add the word config to the name of this file, as `ebuilder` doesn't indicate it's purpose very well on it's own.

selfisekai:
in Alpine Linux I'm the maintainer of the electron package, every app gets rebuilt with a major version upgrade of electron

absidue:
Rebuilding alone is not enough, you also need to make sure that the version of Electron in the `package.json` matches up with the one you are running it on.

So unless you are planning on doing extensive testing on FreeTube every time you upgrade Electron, please lock the version. The fact that you describe your own tasje tool in it's README as `absolutely terrible for app developers` does not inspire me with confi that we won't have to deal with another AUR or NixOS situation again, because when stuff breaks, people come to us first.

selfisekai:
> Rebuilding alone is not enough, you also need to make sure that the version of Electron in the package.json matches up with the one you are running it on and make sure that the app code doesn't use any removed APIs.

the buildscripts are also adjusted to use our node headers, packaged together with the electron. this has failed build-time, delaying electron updates (like now), but so far never on runtime

absidue:
Please either lock the version or do extensive testing before releasing Electron upgrades, because when you break FreeTube people come to us first, not you. I'm sure that both the AUR and NixOS maintainers had the same confidence that you have, but they still ended up breaking stuff with upgrading Electron to version that FreeTube didn't support.

absidue (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):
`yarn build:arm64` works fine

PR #4370, opened 2023-11-22
Support extracting about information from YouTube's new about popup

absidue:
# Support extracting about information from YouTube's new about popup

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description
Draft until https://github.com/LuanRT/YouTube.js/pull/537 is merged and released

This pull request adds support for extracting the about information from YouTube's new popup, I also added support for showing the number of videos that the channel has. Auto-generated game channels like the Minecraft one below, still use the old about popup tab.

This doesn't fix the missing information on the about tab when Invidious is selected as the backend, as that requires changes on their side, to parse and include it in their API response.

## Testing 
Gibi ASMR (has the new about popup)
https://www.youtube.com/channel/UCE6acMV3m35znLcf0JGNn7Q

Minecraft - Topic (has the old about tab)
https://www.youtube.com/channel/UCQvWX73GQygcwXOTSf_VDVg

PR #4464, opened 2023-12-16
Add invidious fallback for likes

MitchelPaulin:
# Add invidious fallback for likes

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
youtubei.js is just not returning video likes for me, this however is not treated as an error, so I was just getting likes of 0. Invidious however was giving me video likes. I added a fallback where if the likes are not found I ask Invidious for them. I also moved the fetch likes code to the bottom of the try/catch in case something else above throws, since this would also fetch likes from invidious, so it avoids having to make two fetches. 

I also removed any reference to youtube dislikes form the file since its no longer relevant 

## Desktop

- **OS: POP_OS 22.04
- **FreeTube version:** 0.19.1"

## Additional context
Is anyone else having issues with the likes not loading on any videos? Maybe youtube is trying to block likes for people that are not logged in? I've attempted to update youtubei and tried with/without my VPN and its always undefined. If this is just a me issue it might not be worth merging.

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -354,22 +354,6 @@ export default defineComponent({
             break
         }
 
-        if (this.hideVideoLikesAndDislikes) {
-          this.videoLikeCount = null
-        } else {
-          try {
-            if (!isNaN(result.basic_info.like_count)) {
-              this.videoLikeCount = result.basic_info.like_count
-            } else {
-              console.warn('Had to fallback to invidious to get video likes')
-              this.getLikesInvidious()
-            }
-          } catch {
-            this.videoLikeCount = 0
-            console.error('Could not load video likes')
-          }
-        }
-
         this.isLive = !!result.basic_info.is_live
         this.isUpcoming = !!result.basic_info.is_upcoming
         this.isLiveContent = !!result.basic_info.is_live_content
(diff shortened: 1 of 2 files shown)


jasonhenriquez:
**suggestion (blocking):** Check `this.backendPreference === 'local' && this.backendFallback` is true before trying to load data from Invidious.

**question:** Does this ever happen the other way around, where we would need to add the corresponding Local fallback if Invidious fails? Or is this a weird bug specific to the Local API? If it is a weird bug with the Local API, is this bandaid really getting at the problem?

I feel like I might have been seeing this recently as well, although I can't exactly remember. If anyone can replicate this occurring and/or show this fallback in action (with non-mocked data), that would be nice for testing it.

absidue:
I would rather have my YouTube.js pull request merged that parses YouTube's new layout, as that's the proper fix and how we usually fix the local API, instead of adding an extra request to Invidious for every single video that you watch.

MitchelPaulin:
> I would rather have my YouTube.js pull request merged that parses YouTube's new layout, as that's the proper fix and how we usually fix the local API, instead of adding an extra request to Invidious for every single video that you watch.

That would be preferable for sure. 

Could you link it? I was taking a look at youtubei.js and couldn't see any open issue or PR addressing this, unless youtube.js is something else

MitchelPaulin:
Oh it's a PR for freetube not to youtbei.js https://github.com/FreeTubeApp/FreeTube/pull/4370/files. I guess this is fixed in 8.0.0.

I guess my only question then is error handling. In the case we can't find something like likes, should that throw and fall into the invidious logic? Seems like that would be better rather than having missing info like it currently does

absidue:
> > I would rather have my YouTube.js pull request merged that parses YouTube's new layout, as that's the proper fix and how we usually fix the local API, instead of adding an extra request to Invidious for every single video that you watch.
> 
> 
> 
> That would be preferable for sure. 
> 
> 
> 
> Could you link it? I was taking a look at youtubei.js and couldn't see any open issue or PR addressing this, unless youtube.js is something else 

I'm sorry for not linking it, I assumed that the name of it made it rather obvious.
https://github.com/LuanRT/YouTube.js/pull/557

PR #4491, opened 2023-12-26
Change icon color of FT logo 

### LINKED ISSUE #4020: [Feature Request]: Change icon color of FT logo on About page based on set theme

AnshuMishra01:
Change icon color of FT logo

Pull Request Type

 Feature Implementation
Related issue

closes #4020

Description
Changed icon color of FT logo on based on set theme

Available Methods :-

As far as I can see, you have three choices:

1) I used an online tool to vectorize our existing logo. You can add it to _icons (our current icon SVG is using a raster image, which doesn't work for this), add it inline to the template, and use the fill CSS property to dynamically change the color to the --primary-text-color for that theme. logoBlack

2) You can create different versions of this logo icon by changing the fill="#ffffff" to the primary text color for the given theme and use those.

3) You can use an editor like GIMP to create logo images for the themes missing them (with the fill color of that theme's --primary-text-color) (probably more effort than two).

Implemented 2 option :)


absidue:
Please remove the yarn error log and revert the addition of `yarn` as a dependency.

absidue:
I do like the idea of dynamically changing the colour :)

AnshuMishra01:
ok Sure

AnshuMishra01:
Its done :)

absidue:
Ideally the logo on the about page should match the one in the navigation bar. Unfortunately this current approach, forces the logo to always be one single colour, so you loose the blue and red colors, which makes this pull request a branding change.

PR #4515, opened 2024-01-03
Add toggle to suppress sending additional args to external players

### LINKED ISSUE #4493: [Feature Request]: Prevent options being passed to external player's command
### Guidelines

### Problem Description

Currently Freetube passes options like `--speed=…`  or `--start=…` to external player (mpv in this case). It would be nice to have the ability to ignore them and just pass the video URL.

### Proposed Solution

I currently use the following wrapper script to `mpv` to achieve this, and point external player in settings to it:

```
#!/usr/bin/bash

set -e
set -uo pipefail

# mpv with non-option arguments only (workaround to ignore playback-affecting options)
literal=0
arguments=
for arg in "$@"; do
    if [[ $literal == 1 ]]; then
        arguments+=" $arg"
    elif [[ $arg == -- ]]; then
        literal=1
    elif [[ $arg =~ ^- ]]; then
        :
    else
        arguments+=" $arg"
    fi
done

eval exec mpv"$arguments"
```

### Alternatives Considered

none

### Issue Labels

### Additional Information

_No response_

trostboot:
# Title

Add toggle to suppress sending additional args to external players

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #4493 

## Description

Adds a toggle setting in the external players sections that lets users suppress any args being sent to the external player aside from the video URL.
Also adds corresponding logic to skip processing of args if this toggle is enabled (this made the diff a mess even though it's a minimal change, tried different orders but git no likey).

## Screenshots 

ignoreArgs

## Testing 

Tested on Windows using SMPlayer and mpv with the toggle on and off. Seems to work as expected.
Testing with more combinations is always appreciated.
[Testing build](https://github.com/trostboot/FreeTube/actions/runs/7401284680).
## Desktop

- **OS:** Windows
- **OS Version:** 11 22H2

## Additional context

I'm open to a better name for the setting if anyone thinks it's not clear enough.

### DIFF
--- a/src/renderer/components/external-player-settings/external-player-settings.js
+++ b/src/renderer/components/external-player-settings/external-player-settings.js
@@ -37,6 +37,9 @@ export default defineComponent({
     externalPlayerIgnoreWarnings: function () {
       return this.$store.getters.getExternalPlayerIgnoreWarnings
     },
+    externalPlayerIgnoreArgs: function () {
+      return this.$store.getters.getExternalPlayerIgnoreArgs
+    },
     externalPlayerCustomArgs: function () {
       return this.$store.getters.getExternalPlayerCustomArgs
     },
@@ -58,6 +61,7 @@ export default defineComponent({
       'updateExternalPlayer',
       'updateExternalPlayerExecutable',
       'updateExternalPlayerIgnoreWarnings',
+      'updateExternalPlayerIgnoreArgs',
       'updateExternalPlayerCustomArgs'
     ])
   }
--- a/src/renderer/components/external-player-settings/external-player-settings.vue
+++ b/src/renderer/components/external-player-settings/external-player-settings.vue
@@ -21,6 +21,14 @@
         :tooltip="$t('Tooltips.External Player Settings.Ignore Warnings')"
         @change="updateExternalPlayerIgnoreWarnings"
       />
+      <ft-toggle-switch
+        :label="$t('Settings.External Player Settings.Ignore Arguments')"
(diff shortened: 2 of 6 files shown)


PikachuEXE (approved):
I have no external player, can only review code

absidue:
Not sure about the naming, as it sounds like it will also ignore the users custom arguments, which is probably undesirable.

trostboot:
That's exactly what it does/what was requested in the linked issue. It also disables the custom args field if enabled to make that clear.

absidue:
If that is the case, then you should add a visual indicator to the custom arguments field, so that it is clear that it is disabled/ignored too. That being said, from a user perspective I would expect that to just disable passing any extra information. If I have gone out of my way to add custom arguments, then I would want them to be used.

trostboot:
It will grey out when disabled, sorry, I haven't updated the screenshot in the initial post, I added the logic to disable it afterwards.

I can see where you're coming from. What would you suggest, then? Pass on custom arguments anyway and rely on the user to clear the custom argument field if they don't them passed on for whatever reason, temporarily or otherwise?

trostboot:
Changed it so custom arguments are still passed on if there are any and changed the name and tooltip accordingly.

ignoreArgs2

absidue (approved):
Hint for other reviewers: Turning on the "Hide Whitespace" option makes the diff significantly more readable.

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #3536, opened 2023-05-13
[Bug]: Spamming Star save video icon saves duplicate videos in playlist page


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Click very fast to save video
2. Go to playlist page
3. See duplicate videos

https://github.com/FreeTubeApp/FreeTube/assets/73130443/3f1570c7-fcd8-4ebc-813f-386b8133a0d1

https://github.com/FreeTubeApp/FreeTube/assets/73130443/f2ed7088-4b36-4632-8433-8ee71d1e58dc

### Expected Behavior

No duplicates

### Issue Labels

race condition, visual bug

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/a4d45b5fa8a8a1726808c9eebe307643f8dde9c9

### Operating System Version

Win10 22H2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Issue #1762, opened 2021-09-29
Feature Request:  Add setting to have the Playlist/Favorite Star in different colors.


cocoonkid:
Feature Request:  Add setting to have the Playlist/Favorite Star in different colors. (Cyan, Blue, Red, Green or colorpicker)

I wear yellow glasses to extend screentime and this is killing me :-)

efb4f5ff-1298-471a-8973-3d47447115dc:
Note this will probably gets implemented by the new design from @dragosnfy

cocoonkid:
commenting to unstale.

efb4f5ff-1298-471a-8973-3d47447115dc:
Wasn't necessary because i already removed the label

cocoonkid:
sorry :-) It still showed the warning for me so I went ahead.

PR #4517, opened 2024-01-04
Add back favorites shortcut

jasonhenriquez:
# Add back favorites shortcut

## Pull Request Type

- [x] Feature Implementation

## Related issue
favorites shortcut removed in #4234
reopens #3536 and #1762

## Description
- Adds back favorites shortcut
  - Fixes broken internal functions (`removeVideos` and `deleteVideoIdsByPlaylistId`) to do so
  - Shortcut does not show on Favorites playlist page (to avoid redundancy)
  - Minor visual improvement: video overlay icons show on `:has(:focus-visible)` instead of `:focus-within`
- Removes Watch Later playlist (as discussed in Matrix developer chat, there is no reason to make Favorites legacy, thus there is no need for a separate "Watch Later" playlist)
- Makes Favorites playlist protected (prevents deletion of playlist, which would be problematic)

## Screenshots 
Screenshot_20240103_175611
Screenshot_20240103_180431
Screenshot_20240103_175507

## Testing 
- Test adding/removing from Favorites using shortcut
- Test removing from Favorites using shortcut removes all duplicates as well
- Test added/removed videos appear properly after restarting FreeTube
- Visual improvement test: Unfavorite a video in Subscriptions using a mouse click, stop hovering over the video, and see the Save to Favorites & Add to Playlist buttons disappear from view. See that they still show when unfavoriting using keyboard navigation & the Space key.

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx

### DIFF
--- a/src/datastores/handlers/base.js
+++ b/src/datastores/handlers/base.js
@@ -147,7 +147,7 @@ class Playlists {
   static deleteVideoIdsByPlaylistId(_id, videoIds) {
     return db.playlists.updateAsync(
       { _id },
-      { $pull: { videos: { $in: videoIds } } },
+      { $pull: { videos: { videoId: { $in: videoIds } } } },
       { upsert: true }
     )
   }
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -144,17 +144,22 @@ export default defineComponent({
       return this.$route.name === 'history'
     },
 
-    inFavoritesPlaylist: function () {
-      return false
-      // const index = this.favoritesPlaylist.videos.findIndex((video) => {
-      //   return video.videoId === this.id
-      // })
+    watchLaterPlaylist: function () {
+      return this.$store.getters.getWatchLater
+    },
 
-      // return index !== -1
+    inWatchLaterPlaylist: function () {
+      return this.watchLaterPlaylist.videos.some((video) =>
+        video.videoId === this.id
+      )
     },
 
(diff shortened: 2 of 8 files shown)


PikachuEXE:
I can't test this much due to missing playlist
image

That's one of the reasons I allow user to pick the playlist they want to use (no special treatment = most backward compatible)

PikachuEXE:
I am a user who don't need that playlist and the feature (quick bookmark)
As mentioned before I want to be able to disabled the feature and more importantly remove the useless empty playlist
Making it protected is exactly what I don't want

jasonhenriquez:
[See comment under other PR](https://github.com/FreeTubeApp/FreeTube/pull/4518#issuecomment-1876174833) - don't care how it's resolved, but "missing" items need to be addressed. If you don't have a protected Favorites playlist, you need to find some other mechanism for selecting a default bookmarked playlist.

src/datastores/handlers/base.js
```diff
@@ -147,7 +147,7 @@ class Playlists {
   static deleteVideoIdsByPlaylistId(_id, videoIds) {
     return db.playlists.updateAsync(
       { _id },
-      { $pull: { videos: { $in: videoIds } } },
+      { $pull: { videos: { videoId: { $in: videoIds } } } },
```
PikachuEXE, line 150: Is this a fix?

jasonhenriquez: Yes, it is. But this function is not in use in your PR, it seems. You can add this fix as well.

Issue #961, opened 2021-01-20
Hardware acceleration


marcih:
I don't know what media player back-end FreeTube uses so I don't know what it would take to implement this, but I would like to see hardware-accelerated video playback added. This is especially desirable on portable devices like the PinePhone or even simply laptops, which cannot afford to have their CPU's highly taxed for extended periods of time due to thermals and battery usage.

PrestonN:
Unfortunately, I cannot do this.

FreeTube is built using Electron, which is a Chromium base for creating desktop applications. Video acceleration is not supported for Electron, which means that I cannot support it for FreeTube.

The best work around that I've seen would be allowing FreeTube to redirect a video over to a player such as VLC or MPV so that we can take advantage of their hardware acceleration features, which is something that I'm hoping to do at some point.

For now, the best luck I've had with watching videos on the Pinephone has been switching to the legacy formats and setting your default quality to 360p. Videos play pretty well with those settings, though obviously it isn't the highest quality.

Apologies for the inconvenience.

marcih:
I just realized, what about VAAPI? Is support for that mainlined into Electron or does it have to be patched in? Or how about having [mpv as backend](https://github.com/Kagami/mpv.js) for the media player as an option?

As for using legacy format, that seems to work better than DASH: the PinePhone can smoothly play 720p videos on legacy whereas 720p on DASH stutters.

PrestonN:
I can look into it, however implementing that library would be a huge undertaking to do. I also do not plan on focusing on Pinephone optimizations right now so this isn't a priority.

I will say that if it allows proper video acceleration within Electron, then I would definitely like to consider including it at some point.

fabianski7:
From what I found, it is possible to build the electron using the flag `use_vaapi = true`  to enable hardware acceleration support, and then run it with `--enable-accelerated-video`.  have you tried this @PrestonN?

https://github.com/electron/electron/issues/18942
https://github.com/electron/electron/pull/19736

No one can blame freetube for not having this, because chromium and firefox do not have support enabled by default, but at least it is possible to activate them manually and it helps a lot when it is necessary to load a video that requires more of the hardware, such as a 1080p60

PrestonN:
[...]

PR #4532, opened 2024-01-08
Fix hardware acceleration flags

sp1cyf0x:
# Fix hardware acceleration flags

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/961, already closed but issue still present.

## Screenshots 

Here is the Before-After if the pull request is implemented as flags.

Before:
before

After:
after

## Testing 

This pull request has NOT been tested.
## Desktop

- **OS: Arch Linux**
- **OS Version: Rolling release 2024-01-08**
- **FreeTube version: https://github.com/FreeTubeApp/FreeTube/actions/runs/7440373756**

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -166,9 +166,11 @@ function runApp() {
   let mainWindow
   let startupUrl
 
-  app.commandLine.appendSwitch('enable-accelerated-video-decode')
-  app.commandLine.appendSwitch('enable-file-cookies')
-  app.commandLine.appendSwitch('ignore-gpu-blacklist')
+  // Enable hardware acceleration via VA-API
+  // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/vaapi.md
+  app.commandLine.appendSwitch('use-gl', 'angle')
+  app.commandLine.appendSwitch('use-angle', 'gl')
+  app.commandLine.appendSwitch('enable-features', 'VaapiVideoDecodeLinuxGL')
 
   // Work around for context menus in the devtools being displayed behind the window
   // https://github.com/electron/electron/issues/38790


sp1cyf0x:
```
app.commandLine.appendSwitch('use-gl', 'angle')
app.commandLine.appendSwitch('use-angle', 'gl')
```
These two lines are not needed when testing but I include it because Chromium Docs said so.

You can remove them if it they are the default behaviour or already implemented somewhere else.

absidue:
Did you test on Windows and macOS too before removing the existing flags?

If not please only add the new ones and stick them inside an if, so they only apply on Linux where they are needed.

sp1cyf0x:
> Did you test on Windows and macOS too before removing the existing flags?

Those flags are all deprecated. They do nothing.

`#ignore-gpu-blacklist` is now `#ignore-gpu-blocklist` but it is not really FreeTube's job to decide if we should bypass the blocklist. User can make that option themselves or ask remove their hardware from the blocklist.

You can add it back if you still decide to keep it.

The flag doesn't seem to cause any issue on Windows.

I can not test on MacOS since I do not own the hardware.

sp1cyf0x:
> If not please only add the new ones and stick them inside an if, so they only apply on Linux where they are needed.

Electron doesn't seem to have a native way to detect OS, there are examples/modules on the Internet but as I'm not very familiar with Electron, I will not attempt to do it.

If you see it necessary to have such function, you can instruct me on how it should be done or find someone to do it.

src/main/index.js
```diff
@@ -166,9 +166,9 @@ function runApp() {
   let mainWindow
   let startupUrl
 
-  app.commandLine.appendSwitch('enable-accelerated-video-decode')
-  app.commandLine.appendSwitch('enable-file-cookies')
-  app.commandLine.appendSwitch('ignore-gpu-blacklist')
+  // Enable hardware acceleration via VA-API
+  // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/vaapi.md
+  app.commandLine.appendSwitch('enable-features', 'VaapiVideoDecodeLinuxGL')
```
absidue, line 171: As requested, here is the code with the operating system check (based on the Windows and macOS ones in the same file).
```suggestion
  if (process.platform === 'linux') {
    // Enable hardware acceleration via VA-API
    // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/vaapi.md
    app.commandLine.appendSwitch('enable-features', 'VaapiVideoDecodeLinuxGL')
  }
```

sp1cyf0x: > As requested, here is the code with the operating system check (based on the Windows and macOS ones in the same file).

`Win32` and `Darwin`? No wonder I couldn't find it...

I found [the documentation](https://nodejs.org/api/process.html#processplatform) for it, your code is correct.

absidue: Just a tip for the future, if ever decide to contribute to FreeTube or another open source project again: Being rude to the maintainers of projects that you want to contribute to is not a great move.

sp1cyf0x: > Just a tip for the future, if ever decide to contribute to FreeTube or another open source project again: Being rude to the maintainers of projects that you want to contribute to is not a great move.

I don't think I was that rude. But yeah, I will try to be more polite.

jasonhenriquez (approved):
LGTM (tested with `intel-gpu-tools` like in the OP).
Screenshot_20240114_093543

absidue (approved):

ChunkyProgrammer (approved):

PR #4543, opened 2024-01-14
Support YouTube using PageHeader on user channels not just auto-generated ones

absidue:
# Support YouTube using PageHeader on user channels not just auto-generated ones

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description
**This is in draft status as it depends on the following YouTube.js pull requests: https://github.com/LuanRT/YouTube.js/pull/577 and https://github.com/LuanRT/YouTube.js/pull/581**

This pull request fixes the `TypeError: Cannot read properties of undefined (reading 'name')` error messages that you may have seen poping up on the subscriptions page and on the channel pages. There are other places in FreeTube that make requests to the channel page, however I decided not to change anything there yet, as that would cause conflicts with https://github.com/FreeTubeApp/FreeTube/pull/4340.

YouTube seems to have an A/B test going on to use the `PageHeader` node on user channels, previously it was only used on system channels like `@gaming`. Interestingly though, when you do have the A/B test, you can only view smaller channels (less than 100k subscribers), larger ones and ones that are verified (both normal and artist) always return internal server errors (HTTP 500 status code), which we can't do anything about, because our requests are seemingly identical for smaller channels as they are for larger channels, so it seems very unlikely that it is an issue on our end.

This pull request extracts most of the information we need, the only thing missing is the channel banner, which YouTube doesn't currently provide, so you'll just see the default one.

## Screenshots 
error-message

## Testing 

In `src/renderer/helpers/api/local.js` add `visitor_data: 'CgtwUmJqXzhJdVFCSijM_oStBg%3D%3D',` to the `InnerTube.create` call in the `createInnertube` function e.g. between line 46 and 47.

Afterwards refresh your subscriptions and visit various smaller channels, the smaller ones should would apart from missing the banner and larger ones will return HTTP 500 errors. That smaller ones work, is what you should be testing for.

```diff
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -44,6 +44,7 @@ async function createInnertube(options = { withPlayer: false, location: undefine
 
   return await Innertube.create({
     retrieve_player: !!options.withPlayer,
+    visitor_data: 'CgtwUmJqXzhJdVFCSijM_oStBg%3D%3D',
     location: options.location,
     enable_safety_mode: !!options.safetyMode,
     client_type: options.clientType,
```

## Desktop

- **OS:** Windows
- **OS Version:** 10


jsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "vueCompilerOptions": {
     "target": 2.7
+  },
+  "compilerOptions": {
+    "strictNullChecks": true
```
absidue, line 6: This makes vscode actually show `undefined` and `null` in type definitions e.g. auto-complete and the popup that appears when you hover over functions.

PikachuEXE:
Got conflict but bot not announcing?

efb4f5ff-1298-471a-8973-3d47447115dc:
is it possible to provide a build so i can daily drive this for a day or two

PikachuEXE:
https://github.com/PikachuEXE/FreeTube/actions/runs/7719513038
Only got dev & this PR

PikachuEXE:
Oh you mean Up Next
I don't know why I see no error in dev console

PikachuEXE:
It seems the upstream library not returning any up next videos (even in development branch)
Maybe YT changed something?

image

absidue:
@curather Looks like YouTube isn't showing them on the website either, so nothing we can do other than wait for YouTube to fix their isssue.

absidue:
@curather That's just a placebo, YouTube randomly returns the empty feed. This pull request has no change that affects the watch page, even the YouTube.js doesn't affect it.
Anyway that issue is off-topic for this pull request.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

package.json
```diff
@@ -77,7 +77,7 @@
     "vue-observe-visibility": "^1.0.0",
     "vue-router": "^3.6.5",
     "vuex": "^3.6.2",
-    "youtubei.js": "^8.2.0"
+    "youtubei.js": "^9.0.1"
```
PikachuEXE, line 80: Though might as well update to `9.0.2`

absidue: Bumped it now

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #4544, opened 2024-01-14
Remove existence checks and fallbacks for files that always exist

absidue:
# Remove existence checks and fallbacks for files that always exist

## Pull Request Type

- [x] Other - Performance Improvement

## Description
Currently we check if the invidious fallback instances file, the external player data file and the geolocation files exist on the filesystem before loading them, however as they are in the source code repository and bundled with FreeTube, they will always exist unless someone goes out of the way to remove them from the FreeTube bundle or accidentally deletes them during development. So I think it's fine to remove the checks and fallbacks, because if they are actually missing, you'll actually get an error message, so that you know about the problem, instead of the silent fallback that we currently have.

For the geolocation files we do actually need to check if one exists in the display language, so that we can fallback to the `en-US` one if it doesn't, to keep that working I decided to copy the behaviour of the web build, which embeds a list of known files at build time, allowing use to just check the array instead of doing the runtime file system check every time.

## Testing 
1. Change the display language and check that the list of regions for trending still gets updated
2. Check that opening an external player still works

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/webpack.renderer.config.js
+++ b/_scripts/webpack.renderer.config.js
@@ -1,5 +1,5 @@
 const path = require('path')
-const { readFileSync } = require('fs')
+const { readFileSync, readdirSync } = require('fs')
 const webpack = require('webpack')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const VueLoaderPlugin = require('vue-loader/lib/plugin')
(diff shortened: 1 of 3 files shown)


ChunkyProgrammer (approved):

PikachuEXE (commented):
Checked (1) but I can't check (2)

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
In the same boat as Pika

absidue:
Actually all you need to do is check that the external players list is still populated in the settings, you don't actually meed to try it (if reading the file were to fail the list would be empty anyway).

PikachuEXE (approved):
yup didn't read code

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4565, opened 2024-01-17
Update most paginated places to auto load next page (except comments)

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/discussions/4521

## Description

Make following places to auto load next page
- every subscription tab
- channel tabs with pagination
- hashtag page
- history page
- search result page
- user playlists page

Comments already handled by https://github.com/FreeTubeApp/FreeTube/pull/3352 so not handled by this new option

## Screenshots 

image

## Testing 

Test places above with pagination and scroll down to ensure next page auto loaded **only if** option enabled

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


jasonhenriquez:
This would make sense to be next to Comment Auto Load. Which by the way, should not be under Player Settings, which we currently have it at for some reason. So maybe move that too.

PikachuEXE:
Update comment section to auto load next page when this new option is enabled

I don't see why `Comment Auto Load` should be in `General` section when the only place it affects is comment section in watch page which is next to player (strictly speaking the `Up Next` & playlist components are also outside the actual video player and we place options like `Autoplay Playlists` & `Play Next Video` in the same section)

absidue:
Instead of duplicating the code across all components, it would probably be better to create a dedicated component for it and then reuse that in the other components. When it becomes visible it can emit an event and then the "parent" can use that to call it's own `increaseLimit` or `loadMore` function.

PikachuEXE:
@absidue Done (except comment section with custom logic

efb4f5ff-1298-471a-8973-3d47447115dc:
Everything works like expected. I did notice that this doesnt work for creator playlists but i assume that is because of YT. 

If yes on previous statement then we have change messaging in tooltip

PikachuEXE:
I need an example of creator playlists
Also please suggest a new message (I think there might be more than 1 type of special playlists that might not work with this new feature in the future...)

efb4f5ff-1298-471a-8973-3d47447115dc:
Sorry for the delay, its not working for me on these playlists

- https://youtube.com/playlist?list=PL8mG-RkN2uTw7PhlnAr4pZZz2QubIbujH
- https://youtube.com/playlist?list=PL950974742C0D24CC
- https://youtube.com/playlist?list=PLBsP89CPrMeNm71T5gYC6jebm9vPbLBiP

Also wanted to chime in on the earlier topic of the placement of Comment Auto Load. Why not merge the Auto comment load and this setting into one and put it under General Settings as Infinite Scroll?

PikachuEXE:
Coz `Comment Auto Load` also loads the comment section, not just next page

PikachuEXE:
Implemented auto load for single playlist page (remote & local playlists)

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
I think merging the feature should happen but is out of scope here, LGTM

PikachuEXE:
Hello~?

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

jasonhenriquez:
> This does NOT handle video comment auto load

Looking back at this, are we certain that maintaining this level of granularity is worth the cost of adding another setting to the pile (compared to combining these into one setting)? See this discussion [here](https://github.com/FreeTubeApp/FreeTube/issues/4594#issuecomment-2054101358):

> Adding more controls & optionality is something we would prefer not to do. We would prefer the solution that had the fewest amount of additional controls or settings added because we want to be conscientious of cluttering up the UI (more than it already is, but we're working on that).
    As a corollary of 1, that means making development choices that:
    a. have good-for-most-everyone default behaviors, and
    b. (when 2a does not suffice) lean into our existing settings and controls without having to add additional ones.

PikachuEXE:
So what's suggested is: replace `commentAutoLoadEnabled` with this new setting?

jasonhenriquez:
I think one could argue for the technical implementation of either replacing the existing setting or adding this as a new behavior of the `commentAutoLoadEnabled` variable. The latter is maybe less preferable until I finally get an answer on what is actually desired for #4342. But yes, behaviorally, replacing the Comment Auto Load setting with this one.

PikachuEXE:
Removing commentAutoLoadEnabled is fine for me
As long as we inform users on release note

PikachuEXE:
Setting `commentAutoLoadEnabled` removed, tooltip for new setting updated

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

static/locales/en-US.yaml
```diff
@@ -251,6 +251,9 @@ Settings:
     Fallback to Non-Preferred Backend on Failure: Fallback to Non-Preferred Backend
       on Failure
     Enable Search Suggestions: Enable Search Suggestions
+    Auto Load Next Page:
+      Label: Auto Load Next Page
+      Tooltip: This also handles video comment auto load
```
jasonhenriquez, line 256:
```suggestion
      Tooltip: Load comments and additional page results without having to press "Load More" buttons.
```

PikachuEXE: "page results" seems like a strange term...
`Load additional pages and comments without clicks.`? (Can swap `without clicks` with `automatically` not sure which one better

jasonhenriquez: I'm good with that (the "automatically" version)

PikachuEXE: Updated

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez (approved):

absidue (approved):

Issue #4578, opened 2024-01-21
[Bug]: Huge user playlists (in the 1000s and 10,000s range) take a long time to load and display, leaving the Playlist view does so as well, and so does viewing the next video in the playlist


Gorrrg:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Have a significantly huge playlist, mine has over 17,000 items
2. View that playlist
3. Wait and wait

And...
1. Click a video from that playlist
2. Wait and wait and wait (takes as long as loading the playlist to leave the playlist)

And..
1. Watch a video in that playlist to the end
2. Wait and wait and wait
3. Next video plays

### Expected Behavior

I've recently switched to a newer snapshot (v3847) and it has the new playlist system. Now my Favorites were converted to a playlist called Favorites. On the older version this single Playlist was displaying instantly, and now with the new system, every single video in the playlist seems to get processed. And I just expect the top most items showing up as fast as they used to. I also often switch between Subscription view and Playlist view and this waiting time just breaks my workflow, distracting me.

What's even worse is that if I want to watch a video by clicking on it from that playlist it again takes this huge amount of time until the Video view is loaded, the entire playlist gets processed again. And when I've finished watching one video loading the next also takes seconds. Making the entire thing the opposite of enjoyable to use.

Of course, I could just create a new Favorites playlist, but eventually I end up with hundreds of playlists. And maybe I want to keep those videos in that playlist so it's easier to find videos I have watched half a year or a year ago.

I wonder if those playlists are saved as a JSON file? Maybe it's time to switch to a database format?

### Issue Labels

usability issue

### FreeTube Version

v0.19.1-nightly-3847 Beta

### Operating System Version

Windows 11 Pro, 23H2

### Installation Method

.exe

[...]

PR #4593, opened 2024-01-23
Update single playlst view for user playlist to have pagination

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Maybe this makes https://github.com/FreeTubeApp/FreeTube/issues/4578 better

## Description

Add pagination to single playlist view for user playlists

## Screenshots 

image

## Testing 

- Add some playlists with 100+ videos (e.g. copy some playlists from https://www.youtube.com/channel/UCXuqSBlHAE6Xw-yeJA0Tunw, sort by Last Video Added)
- Ensure single playlist view got videos paginated
- Ensure pagination limit NOT remembered
- Ensure user playlist can be copied without warning even not all videos loaded
- Ensure this issue is fixed: `When moving a video in a playlist up or down, the animation for the videos swapping plays twice`
- What else

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

PR #4597, opened 2024-01-25
Playlist performance improvements

absidue:
# Playlist performance improvements

## Pull Request Type

- [x] Bugfix
- [x] Performance Improvement

## Related issue
#4578
Mostly solves the issue, however we will need to do some more involved refactoring to fix the watch page entirely.

## Description
This pull request addresses some of the performance problems concerning gigantic user playlists (over 17k items) that were brought up in the linked issue. It doesn't completely solve the issue, especially on the watch page, as we are still rendering at least one `<div>` per playlist item, so that we can figure out when to load the items (there is a reason YouTube only displays a maximum of 200 videos on their watch page), we also don't unload items when they disappear from view again. Fixing the watch page properly will require switching to a virtual scrolling approach, so that we really only have as many HTML elements in the DOM as we really need, which is out of scope of this pull request.

Here is the list of changes that this pull request makes:
1. Add pagination to the playlist page for user playlists, extra pages can be loaded with the load more button.
2. Fixes the load more loading icon being displayed off-screen (it's now used for both local API and user playlists, we don't support pagination for Invidious).
3. Avoids caching users playlists when switching from the playlist page to the watch page, as user playlists are already in the store so it's just extra work. The caching is still in place for remote playlists, as it saves making some network requests on the watch page.
4. Various minor changes to the playlists store, in the worst case (every possible fixable issue is present) these shave about 30 milliseconds off, in the best case (all checks succeed, no repairs to the data necessary) they add a negligible overhead (less than a millisecond), so they seem worth it.
5. Makes the `ft-list-video` element slightly cheaper to render:
    - Only creates the description element when there is actually a description (user playlists don't have one, the code in the store explicitly removes them)
    - Removes an unnecessary template wrapper
    - Replaces the `watched` property with the `historyEntryExists` computed property, as they served the same purpose.
6. Fix the video number not getting lazy loaded, it will now load at the same time as the rest of the video item.

The final result is that the playlist page loads quickly, as it only has to render 100 items initially, however if you press load more enough times, things do slow down, because we don't unload items once they are loaded. The watch page is still slow to load, it did go from 16.24 to 14.88 seconds with the test playlist on my computer, but that is still slow. Doing a performance benchmark shows that the majority of time is spent in Vue's internals, as it struggles with the sheer number of component instances and DOM elements (16200+ of each).

## Testing 
1. Backup the playlists in your dev environment (Almost the same directory as usual, replace the last path segment (`FreeTube`) with `Electron` https://docs.freetubeapp.io/usage/data-location/)
2. Download this database that contains a test playlist with 16200 items inside it.
3. Rename it to `playlists.db`
4. Place it in the dev data location
5. Test the playlist page and watch page for that playlist

The playlist was created by copying https://www.youtube.com/playlist?list=PLxYGDfxfyiF85ZwSBchVX7datipeOJhtH multiple times inside itself.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -4,7 +4,7 @@ import debounce from 'lodash.debounce'
 import FtLoader from '../../components/ft-loader/ft-loader.vue'
 import FtCard from '../../components/ft-card/ft-card.vue'
 import PlaylistInfo from '../../components/playlist-info/playlist-info.vue'
-import FtListVideoLazy from '../../components/ft-list-video-lazy/ft-list-video-lazy.vue'
+import FtListVideoNumbered from '../../components/ft-list-video-numbered/ft-list-video-numbered.vue'
 import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
 import FtButton from '../../components/ft-button/ft-button.vue'
 import {
@@ -21,7 +21,7 @@ export default defineComponent({
     'ft-loader': FtLoader,
     'ft-card': FtCard,
     'playlist-info': PlaylistInfo,
-    'ft-list-video-lazy': FtListVideoLazy,
+    'ft-list-video-numbered': FtListVideoNumbered,
     'ft-flex-box': FtFlexBox,
     'ft-button': FtButton
   },
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -54,6 +54,7 @@ export default defineComponent({
       channelId: '',
       infoSource: 'local',
       playlistItems: [],
+      userPlaylistVisibleLimit: 100,
       continuationData: null,
       isLoadingMore: false,
       getPlaylistInfoDebounce: function() {},
@@ -102,7 +103,11 @@ export default defineComponent({
     },
 
     moreVideoDataAvailable() {
-      return this.continuationData !== null
+      if (this.isUserPlaylistRequested) {
+        return this.userPlaylistVisibleLimit < this.videoCount
+      } else {
+        return this.continuationData !== null
+      }
     },
 
     isUserPlaylistRequested: function () {
@@ -117,6 +122,18 @@ export default defineComponent({
 
       return this.selectedUserPlaylist?._id !== this.quickBookmarkPlaylistId
     },
+
+    visiblePlaylistItems: function () {
+      if (this.isUserPlaylistRequested) {
+        if (this.userPlaylistVisibleLimit < this.videoCount) {
+          return this.playlistItems.slice(0, this.userPlaylistVisibleLimit)
+        } else {
+          return this.playlistItems
+        }
+      } else {
+        return this.playlistItems
+      }
+    }
   },
   watch: {
     $route () {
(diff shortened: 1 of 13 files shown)


PikachuEXE:
Not necessary need to be fixed in this PR
But I do notice the watch page playlist still load many playlist items
Perhaps that's issue making the rendering slow
image

PikachuEXE:
`When moving a video in a playlist up or down, the animation for the videos swapping plays twice` is fixed in [`f1d700d` (#4593)](https://github.com/FreeTubeApp/FreeTube/pull/4593/commits/f1d700d98008e7d887c8c4aa4d3f3d5f3d11f3d3)

Since we probably only merge this PR instead of that I think it's better to also include that fix as well

src/renderer/views/Playlist/Playlist.js
```diff
@@ -117,6 +122,18 @@ export default defineComponent({
...
+      if (this.isUserPlaylistRequested) {
+        if (this.userPlaylistVisibleLimit < this.videoCount) {
+          return this.playlistItems.slice(0, this.userPlaylistVisibleLimit)
+        } else {
+          return this.playlistItems
+        }
+      } else {
+        return this.playlistItems
+      }
+    }
```
PikachuEXE, line 136: Early return
```suggestion
      if (!this.isUserPlaylistRequested) {
        return this.playlistItems
      }

      if (this.userPlaylistVisibleLimit < this.videoCount) {
        return this.playlistItems.slice(0, this.userPlaylistVisibleLimit)
      } else {
        return this.playlistItems
      }
    }
```

PikachuEXE (commented):
I don't like too many nesting levels :)

absidue:
> Not necessary need to be fixed in this PR
But I do notice the watch page playlist still load many playlist items
Perhaps that's issue making the rendering slow

Yes, I'm sorry that I didn't explain that well enough in the pull request description. That is out of scope of this pull request though, as it will require a complete rethink of how we do lazy loading (see section about virtual scrolling). Our current lazy loading implementation throughout the app, requires you to have one div per item (video, channel, playlist) to exist at all times, so that you can detect when it becomes visible (gets scrolled into view) and trigger the loading of it's children.

A better but more complicated to implement approach would have a parent element be in charge of the entire list, making sure that there are only ever enough items in the DOM, as are actually visible items. That probably involves listening to scroll events and reacting based on that or something. I haven't looked into how other virtual scrolling implementations work, so I not quite sure how we would do it.

absidue:
> `When moving a video in a playlist up or down, the animation for the videos swapping plays twice` is fixed in [`f1d700d` (#4593)](https://github.com/FreeTubeApp/FreeTube/pull/4593/commits/f1d700d98008e7d887c8c4aa4d3f3d5f3d11f3d3)
> 
> Since we probably only merge this PR instead of that I think it's better to also include that fix as well

I can't seem to be able to reproduce that problem on this pull request, happy to try fixing it if you can provide steps to reproduce it.

absidue:
> I am not talking about the no. of divs But the elements that should be invisible are rendered as visible (my fault for not expanding one of them when taking the screenshot) But I can take another look after this is merged

As far as I can tell this is working as expected, all divs with the `placeholder` class don't have any children. Once an item has been loaded it stays loaded, we don't unload them.

PikachuEXE (approved):
No conflict good enough

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Very unrelated to this PR but, big number likes to hide behind thumbnail


https://github.com/FreeTubeApp/FreeTube/assets/73130443/46c1bd26-1b2c-43fd-8ec0-24198e02ea29

ChunkyProgrammer (approved):

PR #4664, opened 2024-02-11
Fix playlists database import and export not using the actual database format

absidue:
# Fix playlists DB import and export not using the actual DB format

## Pull Request Type

- [x] Bugfix

## Description
Currently the playlists export generates a JSON array with the playlists and the playlists import expects said JSON array, however that doesn't match the actual playlists database format, so you can't import a real playlists database file. If we exported it as JSON file that would be fine, but we explicitly label it as a database file and expect a database file to be picked during the import.

This pull request changes the import and export to match the real database format (each playlist is a JSON object on it's own line, with a final new line at the end of the file), so that we are consistent across the board.

While yes this is technically a breaking change compared to previous nightly builds, it's not a breaking change compared to the previous (0.19.1) release. It corrects incorrect behaviour and people cannot expect stability across nightly builds, as they are by definition development/bleeding edge builds, which we've never recommended normal users use for every day use.

## Testing 
1. Create multiple playlists
2. Close FreeTube
3. Backup your playlists database file (https://docs.freetubeapp.io/usage/data-location/, `yarn dev` uses a folder called `Electron` instead)
4. Open FreeTube again and click `Remove All Playlists` in the `Privacy Settings`
5. Import your backed up playlists database
6. Check that your playlists show up correctly in the UI
7. Export your playlists
8. `Remove All Playlists` in the `Privacy Settings` again
9. Import the playlists database that you just exported
10. Check that your playlists show up correctly in the UI

If both the original import (a real database written by nedb) and the secondary one (the database created by the export) were successful (no errors and playlists show up), this pull request works as expected.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 30248d6 (nightlies)

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -882,7 +882,10 @@ export default defineComponent({
         showToast(`${message}: ${err}`)
         return
       }
-      const playlists = JSON.parse(data)
+      data = data.split('\n')
+      data.pop()
+
+      const playlists = data.map(playlistJson => JSON.parse(playlistJson))
 
       const requiredKeys = [
         'playlistName',
@@ -1015,7 +1018,11 @@ export default defineComponent({
         ]
       }
 
-      await this.promptAndWriteToFile(options, JSON.stringify(this.allPlaylists), 'All playlists has been successfully exported')
+      const playlistsDb = this.allPlaylists.map(playlist => {
+        return JSON.stringify(playlist)
+      }).join('\n') + '\n'// a trailing line is expected
+
+      await this.promptAndWriteToFile(options, playlistsDb, 'All playlists has been successfully exported')
     },
 
     exportPlaylistsForOlderVersionsSometimes: function () {


PikachuEXE:
I checked the file for history & the logic at 0.19.1
Using JSON as input/output is done **since the day playlist import/export was introduced**

https://github.com/FreeTubeApp/FreeTube/commit/d21a7f1c245c92df8e3eadb0a401fee9daf36848#diff-3d237cfb3b1c9b7a1d14cc731643319d4e50982c43202b626bc6de4d25376269R1225

So this is a breaking change for **all previous FT versions**

absidue:
I'll add back support for the old format in the import, for the sake of backwards compatibility so that you can import both a real FreeTube database and the JSON array format (if you want to export your playlists for 0.19.1 or below you need to enable the backwards compatibility switch before exporting anyway, so the new export doesn't need support for both).

absidue:
It now supports importing the proper database format and the old array format. The export is only for the proper database format, anyone that wants to export their playlists to 0.19.1 or below or to nightlies before the playlists PR, needs to use the `Export Playlists For Older FreeTube Versions ` switch anyway, which isn't affected by this pull request.

That makes this only a breaking change for the nightlies between this pull request and the playlists pull request, which is completely fine.

PikachuEXE (approved):
Export: tested both export formats with text editor
Import: tested 3 formats, 2 above, 1 with older format but with json

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Probably a dump q based on the convo im reading but still going to ask it, is this expected?

https://github.com/FreeTubeApp/FreeTube/assets/73130443/3877d959-d223-4294-8ac8-c02952cf3afc

PikachuEXE:
> Probably a dump q based on the convo im reading but still going to ask it, is this expected? 

If you are expecting default generated playlists, they are only generated after app restart (or new window open?)
Otherwise not so sure what are you referring to in the video about expected or not.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Dismiss my previous q😬

ChunkyProgrammer (approved):

Issue #4152, opened 2023-10-14
[Feature Request]: Copy Subscription across windows when manual loading


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

1. Disable automatically loading the Subscriptions page.
2. Click Load Subscriptions in window 1.
3. Open a new window 2.
4. window 2 wants you to click to load the subscriptions again, while window 1 shows the subscriptions that have already been loaded.

### Proposed Solution

Window 2 should show the subscriptions that were already loaded in window 1.

### Alternatives Considered

None.

### Issue Labels

ease of use improvement, improvement to existing feature

### Additional Information

Original issue was opened by @toby63 as a bug report in #3447 but as this isn't a bug this got converted to a future request

PR #4667, opened 2024-02-11
Update subscription cache when visiting a channel

absidue:
# Update subscription cache when visiting a channel

## Pull Request Type

- [x] Feature Implementation

## Description
Currently when you visit a channel, we fetch the latest information from YouTube and display it to the user, we also update the channel name and thumbnail in the subscriptions database. What we don't do yet is update the subscriptions cache with the videos, shorts, live and community post content that we got.

This pull request changes the channel page, so that it updates the subscription cache with the information on the channel page. For the videos, live and community tabs, we just overwrite the current cache for that channel, as what we just received is the last known state (same as what a refresh on the subscriptions page would do). For the shorts tab we unfortunately cannot do that, as the shorts channel tab doesn't have published dates (we need them for sorting on the subscriptions page), on the subscriptions page we always use RSS for the shorts tab, instead this pull request will just update the title, channel name and view count of the existing items in the cache.

You might be wondering what the point of this is, as we don't syncronise the subscription cache across windows (each window has it's own one https://github.com/FreeTubeApp/FreeTube/issues/4152) and don't persist it on disk, so they disappear between app launches. This is basically a "free" subscription refresh for the visited channel as we already have the information without making any additional API requests. It will mostly benefit users that use the same window for a long time, but will help if we do implement synchronisation across windows in the future or disk persistence, as it would also allow us to reduce API requests (e.g. this channel was refreshed less than 1 minute ago, no need to fetch it again).

## Testing 
As waiting for any of your subscribed channels to upload a new video would be a slow and unpredictable test case, here are two that allow you to skip the waiting.

**Please note that all steps for a given test case have to be performed in the same window, as we don't synchronise the subscription cache between windows yet**

### Test case 1
This test case will require you to have DeArrow disabled and either not have any watch history or for you to pick a channel that has videos that you haven't watched.

1. Enable `Fetch feeds from RSS` in the `Subscription Settings` section
2. Refresh your subscriptions
3. Visit one of the channels that you are subscribed to
4. Go back to the subscriptions page
5. Notice that video durations have shown up for the videos on that channel (if you have DeArrow enabled or have watched any of those videos, FreeTube would be showing the duration from there instead of nothing, so you wouldn't notice the difference).

### Test case 2
1. Disabled `Fetch feed automatically` in the `Subscription Settings` section
2. Create a new profile
3. Subscribe to a channel that yet from the search results or watch page (don't visit the channel page yet).
4. Visit the subscriptions page and notice that it is empty
5. Visit the channel
6. Go back to the subscriptions page and notice that the videos, live and community tabs should now be populated (shorts won't be as the shorts channel tab doesn't have published dates, so we can only update the metadata in the existing cache, as we need the published dates for sorting)

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/store/modules/subscriptions.js
+++ b/src/renderer/store/modules/subscriptions.js
@@ -54,6 +54,10 @@ const actions = {
     commit('updateShortsCacheByChannel', payload)
   },
 
+  updateSubscriptionShortsCacheWithChannelPageShorts: ({ commit }, payload) => {
+    commit('updateShortsCacheWithChannelPageShorts', payload)
+  },
+
   updateSubscriptionLiveCacheByChannel: ({ commit }, payload) => {
     commit('updateLiveCacheByChannel', payload)
   },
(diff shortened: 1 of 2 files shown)


PikachuEXE (approved):
In addition I also tested:
- Create/choose profile with >1 profile (auto fetch sub disabled)
- Confirm no sub shown
- Visit 1 channel
- Confirm no sub shown
- Visit remaining channels
- Confirm sub (& others) shown

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4723, opened 2024-02-27
Make video thumbnails have certain height before image loading starts to avoid layout shifts

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Similar to https://github.com/FreeTubeApp/FreeTube/pull/4713

## Description

As title
The constant aspect ratio is only applied before image loading finished (no idea how to detect image loading start) in case the image got different aspect ratio

## Screenshots 

Image loading not started yet
image

Image loaded started
image

## Testing 

- Turn off `Replace HTTP Cache` (so that renderer does the image loading)
- In dev tool change network speed to slow 3G
- Find some random playlist to load (user playlist might be faster)
- Ensure thumbnail element got height before image loading starts

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.scss
+++ b/src/renderer/components/ft-list-video/ft-list-video.scss
@@ -3,3 +3,8 @@
 .thumbnailLink:hover {
   outline: 3px solid var(--side-nav-hover-color);
 }
+
+.thumbnailImage.thumbnailLoading {
+  // Makes img element sized correctly before image loading starts
+  aspect-ratio: 16 / 9;
+}
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -105,6 +105,7 @@ export default defineComponent({
       hideViews: false,
       addToPlaylistPromptCloseCallback: null,
       debounceGetDeArrowThumbnail: null,
+      thumbnailLoaded: false,
     }
   },
   computed: {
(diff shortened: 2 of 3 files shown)


src/renderer/components/ft-list-video/ft-list-video.scss
```diff
@@ -3,3 +3,8 @@
 .thumbnailLink:hover {
   outline: 3px solid var(--side-nav-hover-color);
 }
+
+.thumbnailImage.thumbnailLoading {
+  // Makes img element sized correctly before image loading starts
+  aspect-ratio: 16 / 9;
```
absidue, line 9: According to the [aspect-ratio docs on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio#fallback_to_natural_aspect_ratio), you can do this for images, which will make it use 16/9 until the image is downloaded and after that use the image's aspect ratio.

That would allow you to remove the javascript and the thumbnail loading class.
```suggestion
  aspect-ratio: 16/9 auto;
```

PikachuEXE: And it's done

absidue (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4725, opened 2024-02-27
Make chapter section scroll to current chapter when scroll into view port when expanded

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Currently the chapter section only scroll to current chapter when expanded
With this PR it scroll to current chapter when scrolled into view port (becomes visible) which is useful especially on videos with many chapters
Users can still scroll around chapters when chapter section is visible

## Screenshots 

Start a video
image

Change current timestamp
image

Scroll back down
image

## Testing 

- Find a video with chapters (plenty in [LTT](https://www.youtube.com/channel/UCXuqSBlHAE6Xw-yeJA0Tunw))
- Expand chapter section
- Scroll up (and maybe resize window) until chapter section hidden
- Change current timestamp (far away from current one would be easier for testing)
- Scroll down and ensure current chapter at the new timestamp is scrolled to

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/watch-video-chapters/watch-video-chapters.vue
+++ b/src/renderer/components/watch-video-chapters/watch-video-chapters.vue
@@ -1,5 +1,12 @@
 <template>
   <ft-card class="videoChapters">
+    <div
+      v-observe-visibility="observeVisibilityOptions"
+    >
+      <!--
+        Dummy element to be observed by Intersection Observer
+      -->
+    </div>
     <h3
       class="chaptersTitle"
       tabindex="0"
--- a/src/renderer/components/watch-video-chapters/watch-video-chapters.js
+++ b/src/renderer/components/watch-video-chapters/watch-video-chapters.js
@@ -33,7 +33,27 @@ export default defineComponent({
 
     compact: function () {
       return !this.chapters[0].thumbnail
-    }
+    },
+
+    observeVisibilityOptions() {
+      return {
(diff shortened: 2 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/components/watch-video-chapters/watch-video-chapters.vue
```diff
@@ -1,5 +1,12 @@
 <template>
   <ft-card class="videoChapters">
+    <div
+      v-observe-visibility="observeVisibilityOptions"
+    >
+      <!--
+        Dummy element to be observed by Intersection Observer
+      -->
+    </div>
```
absidue, line 9: Instead of creating a new element, you can probably just add the directive to the `<div>` with `v-show="showChapters"` on it, that way it won't fire when the chapters aren't visible and you won't need the dummy element.

PikachuEXE: Done~

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

ChunkyProgrammer (approved):

PR #4736, opened 2024-03-02
Split view count and published date into two lines on small displays

MarmadileManteater:
# Split view count and published date into two lines on small displays

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
The view count tends to split in a way that looks awkward on small displays. The number part of the view count ends up on a different line than the text part of the view count, and it makes it a little difficult to read. This PR addresses this by splitting the published date and view count onto different lines on small displays so that the view count is less likely to wrap in this way.

## Screenshots 
|before|after|
|--|--|
|before-2|after-2|

## Testing 
1. Make sure you are using the local API (iv seems to return 0 for likes ATM)
2. Find a video with a high view count and a high like count (ex: https://youtu.be/K_CbgLpvH9E)
3. Reduce the screen width to `<= 460px` 
4. Ensure the view count does not split the text part of the line from the number part

## Desktop

 - OS: Windows 10
- **FreeTube version:** 65b7ba1f88f2f5595a8284de2ebc90a8563b2da2

## Additional context

I picked 460px because view counts and like counts can vary. The example video in the screenshots has plenty of space at screen width 460px, but not all videos do. That is why the width was picked.

### DIFF
--- a/src/renderer/components/watch-video-info/watch-video-info.scss
+++ b/src/renderer/components/watch-video-info/watch-video-info.scss
@@ -70,31 +70,43 @@
 .videoMetrics {
   font-size: 14px;
   color: var(--tertiary-text-color);
-
   .likeSection {
-  color: var(--tertiary-text-color);
-  display: flex;
-  flex-direction: column;
-  margin-inline-start: auto;
-  margin-block-start: 4px;
-  max-inline-size: 210px;
-  text-align: end;
-
-  @media screen and (max-width: 680px) {
-    margin-inline-start: 0;
-    text-align: start;
-  }
-
-  .likeBar {
-    border-radius: 4px;
-    block-size: 8px;
-    margin-block-end: 4px;
-  }
-
-  .likeCount {
-    margin-inline-end: 0;
+    color: var(--tertiary-text-color);
     display: flex;
-    gap: 3px;
+    flex-direction: column;
+    margin-inline-start: auto;
+    margin-block-start: 4px;
+    max-inline-size: 210px;
+    text-align: end;
+
+    @media screen and (max-width: 680px) {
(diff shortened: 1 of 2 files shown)


src/renderer/components/watch-video-info/watch-video-info.scss
```diff
@@ -70,31 +70,43 @@
 .videoMetrics {
   font-size: 14px;
   color: var(--tertiary-text-color);
-
```
PikachuEXE, line 73: Why removing the blank line

src/renderer/components/watch-video-info/watch-video-info.scss
```diff
@@ -70,31 +70,43 @@
...
+      display: flex;
+      gap: 3px;
+    }
+  }
+  .datePublishedAndViewCount {
+    @media only screen and (max-width: 460px) {
+      display: flex;
+      justify-content: left;
+      flex-direction: column;
+      .seperator {
```
PikachuEXE, line 104: Ditto
```suggestion
      flex-direction: column;

      .seperator {
```

src/renderer/components/watch-video-info/watch-video-info.scss
MarmadileManteater, line 73:
```

```

MarmadileManteater: i mistakenly removed it, but it should be back now

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #4247, opened 2023-10-29
[Feature Request]: remove the text in search / go to URL


wiiguy:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

whenever i have searched for something and then search something again i have to manually remove the text in the search bar before i can search for something again

### Proposed Solution

is it possible to remove the text that is in it after the search ?

### Alternatives Considered

maybe make the text auto select after we click the search whenever there is text in it ?

### Issue Labels

ease of use improvement, new feature

### Additional Information

.

efb4f5ff-1298-471a-8973-3d47447115dc:
Wont implement, search is on par with YT UX and u can always just click the clear input button

https://github.com/FreeTubeApp/FreeTube/assets/73130443/aa1e5d53-cf61-4d7e-bb16-b61915b413d0

PR #4747, opened 2024-03-05
Auto select all text when search bar is focused

### LINKED ISSUE #4746: [Feature Request]: Select all text when focusing the search bar
### Guidelines

### Problem Description

Currently it doesn't select all text when you focus the search bar via Ctrl+L, only if you tab out and in.
Most of my keyboard-driven use case is as follows:
1. Copy a YouTube link from somewhere.
2. Go to FreeTube and press Ctrl+L to focus the search bar.
3. Press Backspace (slow) or Ctrl+Backspace (faster) to clear the previous link or search query.
4. If there's multiple words, I have to press Ctrl+Backspace for every word.
5. Only then can I paste the link to watch the video.

### Proposed Solution

It would be nice if it auto selects all text when you focus the search bar so that it's only one backspace away from clearing.

### Alternatives Considered

Another solution would be to use Ctrl+U to clear all text even if it's not selected. This is implemented in YouTube and pretty much every HTML input I've encountered but doesn't seem to work on FreeTube.

### Issue Labels

ease of use improvement, new feature, new keyboard shortcut

### Additional Information

https://github.com/FreeTubeApp/FreeTube/assets/12945959/ebe5f7b3-281d-4c41-ad53-3ba873be2d6a

9mdv:
# Auto select all text when search bar is focused and implement Ctrl+U keyboard shortcut.

## Pull Request Type
- [x] Feature Implementation

## Related issue
Closes #4746 and somewhat related to #4247

## Description

This is my attempt at implementing the auto select feature. When focusing the search bar via Ctrl+L, it will auto select all text for easy deletion. The Ctrl+U is a convenience shortcut that will clear all the text inside the search bar which is pretty standard on all HTML input textbox as well as on YouTube itself.

## Desktop
- **OS:** Linux
- **OS Version:** 6.6.20 LTS
- **FreeTube version:** 0.19.2

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -253,6 +253,10 @@ export default defineComponent({
         return
       }
 
+      if (event.getModifierState('Control') && event.key === 'u') {
+        this.handleClearTextClick()
+      }
+
       if (this.visibleDataList.length === 0) { return }
 
       this.searchState.showOptions = true
@@ -316,6 +320,7 @@ export default defineComponent({
 
     focus() {
       this.$refs.input.focus()
+      this.select()
     },
 
     select() {
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -253,8 +253,14 @@ export default defineComponent({
         return
       }
 
-      if (event.getModifierState('Control') && event.key === 'u') {
+      // Press Ctrl+u (Windows/Linux) or Cmd/Meta+u (MacOS) inside the search box to clear the text
+      const isCapslockOff = !event.getModifierState('Capslock')
+      const isCtrlOrCmdPressed = (process.platform !== 'darwin' && event.ctrlKey) ||
+        (process.platform === 'darwin' && event.metaKey)
+
+      if (isCapslockOff && isCtrlOrCmdPressed && event.key === 'u') {
         this.handleClearTextClick()
+        return
       }
 
       if (this.visibleDataList.length === 0) { return }
@@ -320,7 +326,6 @@ export default defineComponent({
 
     focus() {
       this.$refs.input.focus()
-      this.select()
     },
 
     select() {


src/renderer/components/ft-input/ft-input.js
```diff
@@ -253,6 +253,10 @@ export default defineComponent({
         return
       }
 
+      if (event.getModifierState('Control') && event.key === 'u') {
```
PikachuEXE, line 256: This does not work on MacOS

See https://github.com/FreeTubeApp/FreeTube/blob/v0.19.2-beta/src/renderer/App.js#L321-L327 for current implementation

Also it should detect both cases `u` and `U` in case of cap lock or something else

And also I think it should have 
```js
        // Early return
        return
```

PikachuEXE (changes requested):

absidue:
I suspect that CTRL+U selecting text in web browsers is a Linux specific thing, as using that combination in any web browser on Windows views the source code.
Therefore it might make sense to only have that shortcut work on Linux if at all.

This pull request will also reintroduce problems for your fellow Linux users, as selecting text causes Linux to copy it for some people.

ArturWagnerBusiness:
Just wanted to let everyone here know that CTRL+L already selects the search bar and highlight everything for me.
I am running `0f2ef668dcd5c56740cae9de856dbd47920206d9` 29th Feb 2023 on Manjaro ~20 days out of date.
Maybe the selection happens differently on other systems?
Also as absidue mentioned, CTRL+U does not select any text on Manjaro Firefox 123.0, however, it opens source code of the page. On the other side, CTRL+L already selects everything in the search bar once pressed anywhere on the page. So CTRL+U might not be a Linux specific thing maybe but a browser specific?

src/renderer/components/ft-input/ft-input.js
9mdv: Can't test on a Mac but should probably work now, if you press Command+u inside the search box it should clear the text if there's any.

9mdv:
> I suspect that CTRL+U selecting text in web browsers is a Linux specific thing, as using that combination in any web browser on Windows views the source code. Therefore it might make sense to only have that shortcut work on Linux if at all.

Ctrl+u is for **_clearing_** the text inside the search box and will only trigger there, so not really a global shortcut. With this enabled, it would be easier to use with a keyboard: Ctrl+L to focus the search box -> Ctrl+u to clear the old text -> Ctrl+v to paste the new text or link.

> This pull request will also reintroduce problems for your fellow Linux users, as selecting text causes Linux to copy it for some people.

You're right, this will probably pollute some people's clipboard if their system is set to copy the [primary](https://wiki.archlinux.org/title/Clipboard#Selections) selection. Probably better to put it behind a setting, via a toggle.

9mdv:
Closing this PR as I've found an alternative with the same 3 keystrokes: Ctrl+L A V.

PR #4750, opened 2024-03-06
Improve touch controls for dash quality selector

MarmadileManteater:
# Improve touch controls for dash quality selector

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
MarmadileManteater/FreeTubeCordova#239

## Description
The dash quality selector does not respond well to touch input unlike the legacy and audio quality selectors. The `handleClick` event is being called, but the way that is being called on touch input causes `e.target` to reflect a different element than the one actually tapped on. This causes the quality selector to seem like it is ignoring user input when, it is just treating most taps as if they were on the current quality label (which causes the click handler to return immediately). 

This PR addresses this by adding a `touchstart` event to the button element which simply calls `handleClick` with the appropriate target. Due to the fact that the quality selector is typically hidden and shown by a `hover` psuedo class, code was also added to the touchstart event to handle flipping the `vjs-lock-showing` class.

## Screenshots 
|before|after|
|--|--|
|before|after|

## Testing 
1. Ensure `dash` is the preferred format
2. Open a video
3. Ensure the dash quality selector hasn't changed at all on mouse input
4. Ensure the dash quality selector behaves the same way with touch input as mouse input

## Desktop

 - OS: Windows 10
- **FreeTube version:** 672803d30c3fcac392da50e152203b582d192500

## Additional context
To be clear, it is possible to make the dash quality selector "work" (kind of) as is, but it requires "hovering" the button by hold tapping before making a quick shorter tap. Simply tapping once doesn't do anything most of the time.

### DIFF
--- a/src/renderer/components/ft-video-player/ft-video-player.js
+++ b/src/renderer/components/ft-video-player/ft-video-player.js
@@ -1817,7 +1817,20 @@ export default defineComponent({
 
           // For default auto, it may select a resolution before generating the quality buttons
           button.querySelector('#vjs-current-quality').innerText = defaultIsAuto ? autoQualityLabel : currentQualityLabel
-
+          button.addEventListener('touchstart', (e) => {
+            // make it easier to toggle the vjs-menu on touch (hover css is inconsistent w/ touch)
+            if (!e.target.classList.contains('quality-item') && !e.target.classList.contains('vjs-menu-item-text')) {
+              const vjsMenu = button.querySelector('.vjs-menu')
+              let vjsMenuClass = vjsMenu.getAttribute('class')
+              if (vjsMenuClass.indexOf(' vjs-lock-showing') === -1) {
+                vjsMenuClass += ' vjs-lock-showing'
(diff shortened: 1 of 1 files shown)


PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

So idk if this is intentional but shouldnt the speed behave the same as the quality menu? If i tap on speed it will just move up instead of opening the menu for it, like for the qualities

https://github.com/FreeTubeApp/FreeTube/assets/73130443/f777b305-cbc2-41db-81d0-f2ea6de03417

MarmadileManteater:
I think it makes more sense to behave like the other quality controls (legacy & audio). Personally, I find the adjusting playback speed a little clunky. Its acceptable, but I don't think I would want that behavior for the quality selector.

ChunkyProgrammer (approved):

Issue #3736, opened 2023-07-08
[Feature Request]: keep relative timestamps ("Minutes ago") in Subscriptions up-to-date


aleksejrs:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The Subscriptions page shows relative times like "3 minutes ago", and those times are not updated when I leave the page/profile and come back, so I don't know if the videos are still recent, or if I need to refresh.

### Proposed Solution

Show the times relative to now, not to when the page was last updated.

### Alternatives Considered

Show absolute times.
Show the last refresh time.

### Issue Labels

ease of use improvement, improvement to existing feature

### Additional Information

v0.18.0-nightly-3087 Beta

absidue:
YouTube only gives us the relative timestamps (unless you enable the setting to use the RSS feeds, but then you loose out on video durations and live status information), so can only display the information we get from them, just translated to the display language you have chosen in FreeTube.
To keep them accurate we would either need to refresh the subscriptions every time you visit the page, which is not only wasteful but would also likely get you ratelimited by YouTube, or do some inaccurate calculations (why inaccurate? because `5 hours ago` could be anywhere between exactly 5 hours ago or 5 hours 59 minutes and 59 seconds ago, so instead of displaying an outdated timestamp, we would end up displaying misinformation if we did calculations).

aleksejrs:
Then show the last refresh time (or whatever related is available)?

absidue:
already requested in: https://github.com/FreeTubeApp/FreeTube/issues/1010

As that makes this a duplicate of the issue linked above, I'll close this as a duplicate unless you have anything else?

absidue:
duplicate of #1010

absidue:
Reopened this, as I ended up implementing it in the linked pull request. The next time this issue is closed, it will be because that pull request got merged :).

PR #4752, opened 2024-03-08
Fix handling of video published date in video lists

### LINKED ISSUE #2579: [Bug]: "Invalid date" after manually marking video watched from Subscription tab
### Guidelines

### Describe the bug

1. go to subscription tab
2. click menu button for any video
3. mark as watched
4. go to history tab
5. release date of video is listed as `invalid date`

### Expected Behavior

1. go to subscription tab
2. click menu button for any video
3. mark as watched
4. go to history tab
5. release date of video is listed as a valid date e.g `9/16/2022`

This behavior is seen if is at step 2+3 you click into the video, instead of manually marking it as watched  

### Issue Labels

text/string issue

### FreeTube Version

v0.17.1 Beta

### Operating System Version

Debian 11.5

### Installation Method

.deb

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

absidue:
# Fix handling of video published date in video lists

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
Fixes https://github.com/FreeTubeApp/FreeTube/discussions/4541 (not sure if GitHub automatically closes discussions with pull requests but I'm listing it here anyway)
Fixes #2579
Implements #3736

## Description
This started as an investigation into #4541 (FreeTube showing published dates in German, despite the display language being set to English) and in the process of fixing that, I ended up solving #2579 and implementing #3736.

The issue with the translated published strings was because we were using the `publishedText` property in Invidious' API response, which it turns out is a translatable string on the Invidious side, so when an instance like `invidious.nerdvpn.de` decides to randomly return German translations for no apparent reason, FreeTube bugs out and shows that to the user. The solution to that is using the `published` property instead, which contains Invidious' parsing of the `publishedText` into a unix timestamp in milliseconds.

As I was switching to the `published` field for Invidious anyway, I decided I might as well do it for the local API too, that means parsing the relative date strings during the local API parsing, so that `ft-list-video` only has to deal with unix timestamps. Converting it back to a relative timestamp is then done in `ft-list-video` based on the current date and time. One major benefit of doing that, is that when we save videos in the subscription cache, they are no longer saved with `10 minutes ago`, instead they have the unix timestamp, so instead of only being able to show `10 minutes` ago when it is restored from the cache 3 hours later, it will now show `3 hours ago`.
Please note that as these timestamps are calculated based on relative timestamps, they might not match up with YouTube's, as they have access to the absolute timestamp in their databases. However showing a potentially slightly off relative timestamp is a lot better than still showing `10 minutes ago` 3 hours after the video was published.

The mark as watched invalid date issue was caused by us saving the text `"10 minutes ago"` in the watch history, but as we save the unix timestamp on the watch page, the code that read the date from the history was expecting a unix timestamp, which `"10 mins ago"` most certainly is not. The solution to that was quite easy now that the rest of the `ft-list-video` component was already using unix timestamps.

## Testing 
1. Test various video lists (e.g. subscriptions, search, trending) with both the local and Invidious API and make sure that published dates show up correctly.
2. Mark a video as watched from the subscription page (3 dots menu -> Mark as Watched) with the local or Invidious API without RSS and check the history tab to confirm that it shows an actual date and time instead of `Invalid date`
3. Refresh your subscriptions without RSS, wait a while and then come back later, switch to a different page (the timestamps don't update live while you are on the same page, as that would be quite expensive to do every second) and then back to the subscriptions page and it should show updated timestamps without you refreshing the subscriptions.

**This pull request doesn't change how published dates are handled for community posts, but I will switch that to the `published` field in a follow up pull request, so we get the same updated timestamps when we restore it from the cache.**

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -7,7 +7,6 @@ import {
   formatNumber,
   openExternalLink,
   showToast,
-  toLocalePublicationString,
   toDistractionFreeTitle,
   deepCopy
 } from '../../helpers/utils'
@@ -98,7 +97,7 @@ export default defineComponent({
       duration: '',
       description: '',
       watchProgress: 0,
-      publishedText: '',
+      published: undefined,
       isLive: false,
       isUpcoming: false,
       isPremium: false,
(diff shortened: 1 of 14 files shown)


PikachuEXE:
Date format seems changed for some reason

Before this PR (history page)
image

This PR
image

absidue:
I had based it on the existing formatting code, which seems to only switch to months when it's above 31 days and doesn't support weeks, so I'll need to change that.

absidue:
@PikachuEXE @efb4f5ff-1298-471a-8973-3d47447115dc Those issues should be fixed now :).

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Marked watched date reflects date of vid being published when viewing it in History

https://github.com/FreeTubeApp/FreeTube/assets/73130443/4f49b87d-d4b8-49ec-9791-f38b99cd15bb

Edit: Most popular tab has interesting publish dates

Capture

absidue:
> Marked watched date reflects date of vid being published when viewing it in History

That's the same as if you watch it on the watch page.

efb4f5ff-1298-471a-8973-3d47447115dc:
Wow im not sure i ever noticed that. To me that doesnt make sense but that is probably out of scope of this PR

absidue:
Fixed now (Invidious gives all videos on the popular page the type `shortVideo`, even though they are clearly normal videos).

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
I still see date format change
image

absidue:
It respects the users chosen locale now, so as you presumably selected `en-US` as your display language, you get month/day/year.

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #4763, opened 2024-03-13
Add i18n-eslint-plugin

ChunkyProgrammer:
# Add i18n-eslint-plugin

## Pull Request Type
- [x] Dev enhancement

## Description
It is configured to find keys that don't exist and raw text instead of translated text in vue files

## Examples of new behavior
If a key doesn't exist then there will be a warning.
Ex: 
image

If a raw text is used then there is an error (this won't apply to attributes like class where it doesn't make sense to check)
Ex:
image
image

This PR is an updated version of: https://github.com/FreeTubeApp/FreeTube/pull/4242

## Desktop
- **OS:** Linux Mint
- **OS Version:** 21.3
- **FreeTube version:** 0.19.1 (nightly)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,11 +47,12 @@ module.exports = {
     'plugin:vue/recommended',
     'standard',
     'plugin:jsonc/recommended-with-json',
-    'plugin:vuejs-accessibility/recommended'
+    'plugin:vuejs-accessibility/recommended',
+    'plugin:@intlify/vue-i18n/recommended'
   ],
 
   // https://eslint.org/docs/user-guide/configuring#configuring-plugins
-  plugins: ['vue', 'vuejs-accessibility', 'n', 'unicorn'],
+  plugins: ['vue', 'vuejs-accessibility', 'n', 'unicorn', '@intlify/vue-i18n'],
 
   rules: {
     'space-before-function-paren': 'off',
@@ -77,6 +78,39 @@ module.exports = {
     'unicorn/no-array-push-push': 'error',
     'unicorn/prefer-keyboard-event-key': 'error',
     'unicorn/prefer-regexp-test': 'error',
-    'unicorn/prefer-string-replace-all': 'error'
+    'unicorn/prefer-string-replace-all': 'error',
+    '@intlify/vue-i18n/no-dynamic-keys': 'error',
+    // TODO: enable at a later date. currently disabled to prevent massive conflicts for initial PR
+    // '@intlify/vue-i18n/no-unused-keys': [
+    //   'error',
+    //   {
+    //     extensions: ['.js', '.vue', 'yaml']
+    //   }
+    // ],
+    '@intlify/vue-i18n/no-duplicate-keys-in-locale': 'error',
+    '@intlify/vue-i18n/no-raw-text': [
+      'error',
+      {
+        attributes: {
+          '/.+/': [
+            'title',
+            'aria-label',
+            'aria-placeholder',
+            'aria-roledescription',
+            'aria-valuetext',
+            'tooltip',
+            'message'
+          ],
+          input: ['placeholder', 'value'],
+          img: ['alt']
+        },
+        ignoreText: ['-', '•', '/']
+      }
+    ],
+  },
+  settings: {
+    'vue-i18n': {
+      localeDir: './static/locales/*.yaml'
+    }
   }
 }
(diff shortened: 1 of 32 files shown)


src/renderer/views/SubscribedChannels/SubscribedChannels.js
absidue, line 1: Looks like these changes were only done for testing purposes and were accidentally committed?

ChunkyProgrammer: Oh oops, yeah I was getting a weird webpack error due go a circular dependency (as i was originally importing i18n in strings.js but i switched to just passing the i18n object to the function)

src/renderer/views/SubscribedChannels/SubscribedChannels.js
```diff
@@ -2,7 +2,7 @@ import { defineComponent } from 'vue'
 import { mapActions } from 'vuex'
 import FtCard from '../../components/ft-card/ft-card.vue'
 import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
-import FtInput from '../../components/ft-input/ft-input.vue'
+// import FtInput from '../../components/ft-input/ft-input.vue'
```
ChunkyProgrammer, line 5:
```suggestion
import FtInput from '../../components/ft-input/ft-input.vue'
```

src/renderer/views/SubscribedChannels/SubscribedChannels.js
```diff
@@ -12,7 +12,7 @@ export default defineComponent({
   components: {
     'ft-card': FtCard,
     'ft-flex-box': FtFlexBox,
-    'ft-input': FtInput,
+    // 'ft-input': FtInput,
```
ChunkyProgrammer, line 15:
```suggestion
    'ft-input': FtInput,
```

absidue:
Instead of making every call to get the colors have the overhead of translating the color names, even if you only need the values, could you please change the colors back to the object they were in before and instead add a `getColorNames()` (feel free to pick a different name, that's just one I came up with on the spot) function that returns the translated color names.

ChunkyProgrammer:
Conflict resolved itself 😳

.eslintrc.js
```diff
@@ -77,6 +78,39 @@ module.exports = {
...
+          input: ['placeholder', 'value'],
+          img: ['alt']
+        },
+        ignoreText: ['-', '•', '/', 'YouTube', 'Invidious', 'FreeTube']
+      }
+    ],
+  },
+  settings: {
+    'vue-i18n': {
+      localeDir: './static/locales/*.yaml'
```
absidue, line 113: Could you please only include the files for the locales that are allowed to be used in FreeTube, that should hopefully be slightly faster and will prevent potential linting issues from cropping up for files that we don't use.

You could do it in the same way as we do for the WatchExternalFiles plugin in `webpack.renderer.config.js`.

.eslintrc.js
absidue, line 1: https://eslint-plugin-vue-i18n.intlify.dev/started.html#configuration-eslintrc mentions that if you don't specify `messageSyntaxVersion`, it will result in messages being parsed twice. Could you please set that, so that we avoid that extra parsing.

src/renderer/store/modules/utils.js
```diff
@@ -262,7 +257,7 @@ const actions = {
       }
 
       if (parsedString !== replaceFilenameForbiddenChars(parsedString)) {
-        reject(new Error('Forbidden Characters')) // use message as translation key
+        reject(new Error(i18n.t('Settings.Player Settings.Screenshot.Error.Forbidden Characters'))) // use message as translation key
```
absidue, line 260: This comment is outdated now and could be removed.
```suggestion
        reject(new Error(i18n.t('Settings.Player Settings.Screenshot.Error.Forbidden Characters')))
```

src/renderer/store/modules/utils.js
```diff
@@ -611,15 +606,13 @@ const actions = {
     externalPlayerMap.unshift(playerNone)
 
     const externalPlayerNames = externalPlayerMap.map((entry) => { return entry.name })
-    const externalPlayerNameTranslationKeys = externalPlayerMap.map((entry) => { return entry.nameTranslationKey })
     const externalPlayerValues = externalPlayerMap.map((entry) => { return entry.value })
```
absidue, line 609: Could you please remove `nameTranslationKey` on line 601 as well (I can't comment on unchanged lines).

PikachuEXE (commented):
Tested a bit and seems fine
But I am waiting for comments from others first in case I miss something

absidue (approved):

PikachuEXE (approved):

jasonhenriquez (approved):

PR #868, opened 2020-12-12
Added config variable for portable settings store

rddim:
---
Added config variable for portable settings store
---

**Important note**
Backup your settings from `%AppData%\Roaming\FreeTube`

**Pull Request Type**

- [ ] Bugfix
- [x] Feature Implementation

**Related issue**
Fixes #130 Fixes #746

**Description**
This PR add a config variable `FREETUBE_APPDATA_DIR` for portable settings store in a custom folder.

**Testing (for code that is not small enough to be easily understandable)**

1. Backup your settings from `%AppData%\Roaming\FreeTube` and rename/remove the `FreeTube` folder
2. Windows users should use the zipped portable version, not the portable exe
3. In the folder where the `FreeTube.exe` lives:
   * create a shortcut and add `%comspec% /c set FREETUBE_APPDATA_DIR=%CD%\Data&& FreeTube.exe` to the `Target` field.
OR
   * create a batch file containing:
```bat
set FREETUBE_APPDATA_DIR=%CD%\Data
FreeTube.exe
```
4. Run the shortcut or the batch file. In both case the `cmd.exe` will be running (don't close it) and the `Data` folder will be created in the same folder as `FreeTube.exe`

**Any ramifications remaining**
1. The empty folder `%AppData%\Roaming\FreeTube\Dictionaries` is created
2. The registry key for the browser extension is added to `HKCU\Software\Classes\freetube`

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: 7 (build 7601.0)
 - FreeTube version: 0.9.3 nightly

**Additional context**
1. The code is taken from https://github.com/getferdi/ferdi/blob/develop/src/index.js#L13:L23
2. The issues with the empty folder and the registry key can be solved with the portable launcher

PR #4765, opened 2024-03-14
Store portable apps settings in same directory as the EXE

ChunkyProgrammer:
# Store portable apps settings in same directory as the EXE

## Pull Request Type
- [x] Bugfix

## Related issue
Thanks @rddim for the initial PR https://github.com/FreeTubeApp/FreeTube/pull/868

## Description
This PR should fix the issue with the portable exe storing the data in AppData (thus not making it very portable)

## Testing 
- Build the App on windows
- run the exe file (not setup.exe)
- notice local directory is made
- change settings notice that settings.db file is updated

Build should be available here: https://github.com/ChunkyProgrammer/FreeTube/actions/runs/8283749172/job/22667673906

## Desktop
- **OS:** Linux Mint
- **OS Version:** 21.3
- **FreeTube version:** 0.19.1 (nightly)

## Additional context
I coded this on Linux but did some testing on Windows

### DIFF
--- a/src/datastores/index.js
+++ b/src/datastores/index.js
@@ -7,6 +7,13 @@ if (process.env.IS_ELECTRON_MAIN) {
   const { join } = require('path')
   // this code only runs in the electron main process, so hopefully using sync fs code here should be fine 😬
   const { statSync, realpathSync } = require('fs')
+  const path = require('path')
+
+  if (process.platform === 'win32' && process.env.PORTABLE_EXECUTABLE_DIR != null) {
+    app.setPath('appData', process.env.PORTABLE_EXECUTABLE_DIR, `${app.name}AppData`)
+    app.setPath('userData', path.join(app.getPath('appData'), `${app.name}AppData`))
+  }
+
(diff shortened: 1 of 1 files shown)


BobbyMcJefferson:
Not sure why this issue was closed. 

It still saves it's data in AppData/Roaming/FreeTube and not in the .exe folder.

PikachuEXE:
Ya I have no idea too

https://portableapps.com/about/what_is_a_portable_app#definition

A portable app is a computer program that doesn't need to be installed into Windows like traditional apps so you can carry it with you on a portable device or synced cloud folder and use on any Windows computer. When your USB flash drive, portable hard drive, or other portable device is plugged in or your cloud drive is synced, you have access to your software and personal data just as you would on your own PC. And when you unplug the device, none of your personal data is left behind.

I think to clarify FT might want to provide both portable & normal version (even though it's installed via zip)
Coz currently it's named portable but behaves like an installed app (by archive extraction) which is confusing

PR #4810, opened 2024-03-25
Move usingElectron from computed into data

absidue:
# Move usingElectron from computed into data

## Pull Request Type

- [x] Code cleanup

## Description
As that value is set at build time and never changes at runtime, we don't need any of the benefits of a computed property (avoiding doing costly operations more often than necessary), so it's better suited in `data`.

## Testing 
Check that the external player settings still show up in the Electron build (`yarn run dev`), check that they are still correctly hidden in the web build (`yarn run dev:web`).

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 8fa8e1fd99bf45adff5364f56ede2b6831e0f53e

### DIFF
--- a/src/renderer/components/player-settings/player-settings.js
+++ b/src/renderer/components/player-settings/player-settings.js
@@ -26,6 +26,7 @@ export default defineComponent({
   },
   data: function () {
     return {
+      usingElectron: process.env.IS_ELECTRON,
       formatValues: [
         'dash',
         'legacy',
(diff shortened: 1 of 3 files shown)


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
External Player Settings isnt listed when i ran `yarn run dev:web`

Capture

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #4561, opened 2024-01-16
Update add video ot playlist prompt to show presence count when adding one video

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/PikachuEXE/FreeTube/issues/55

## Description

When adding **one** video, display no. of times the video is added in playlists
If not added nothing extra would be displayed

## Screenshots 

image
image

## Testing 

- Create Playlist (A) & (B) (if needed)
- Add a video to A once, B twice (or more)
- Try to add the same video again to see what's displayed

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

I got text for adding 2+ videos ready but probably easier reviewing this first

PR #4818, opened 2024-03-27
Update add video to playlist prompt to show presence count when adding 2+ videos

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Follow up of https://github.com/FreeTubeApp/FreeTube/pull/4561

## Description

When adding **2+** videos, display no. of videos already added in individual playlists
If not already added nothing extra would be displayed

## Screenshots 

image

## Testing 

- Create Playlist (A) & (B) (if needed)
- Add a video to A & B, 2nd video to A, 3rd video to B
- Try to copy playlist A/B

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-playlist-selector/ft-playlist-selector.js
+++ b/src/renderer/components/ft-playlist-selector/ft-playlist-selector.js
@@ -90,8 +90,9 @@ export default defineComponent({
     multiVideoPresenceCountInPlaylistText() {
       if (this.multiVideoPresenceCountInPlaylist == null) { return null }
 
-      return this.$tc('User Playlists.AddVideoPrompt.{videoCount} Videos Already Added', this.multiVideoPresenceCountInPlaylist, {
+      return this.$tc('User Playlists.AddVideoPrompt.{videoCount}/{totalVideoCount} Videos Already Added', this.multiVideoPresenceCountInPlaylist, {
         videoCount: this.multiVideoPresenceCountInPlaylist,
+        totalVideoCount: this.toBeAddedToPlaylistVideoList.length,
       })
     },
     videoPresenceCountInPlaylistText() {
--- a/src/renderer/components/ft-playlist-selector/ft-playlist-selector.js
+++ b/src/renderer/components/ft-playlist-selector/ft-playlist-selector.js
@@ -75,10 +75,32 @@ export default defineComponent({
         count: this.loneVideoPresenceCountInPlaylist,
       })
     },
+    multiVideoPresenceCountInPlaylist() {
+      if (this.toBeAddedToPlaylistVideoList.length < 2) { return null }
+
+      // Count of to be added videos already present in this playlist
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm im wondering how useful this would be because if A & B are a a bit larger than a few videos it would be really difficult for the user to know what videos are already in that playlist

PikachuEXE:
Out of scope for me at least for this PR
When there are many entries I don't see a good way to represent the diff anyway IMO

efb4f5ff-1298-471a-8973-3d47447115dc:
Is it really out of scope? 

I mean if we take #4561 as example it is very clear to the user that Video A has been added to Playlist B and C a certain amount, doesnt matter how big or small that playlist is. If we look at this PR the user only knows that some Videos ABCDEF are also in Playlist Y and Z but has no idea which one of them are really in there. This creates a find the needle in the haystack kinda situation

PikachuEXE:
#4561 only works when adding **one** video
Nothing is displayed right now for copying a playlist = adding 2+ videos

efb4f5ff-1298-471a-8973-3d47447115dc:
My opinion on this hasnt changed. Not sure how useful this is.

PikachuEXE:
There is no way we display too much info especially when the no. of videos to be added/copied is large like 100
The new info is to help users avoid easy to spot mistakes (at least they know there is and how many videos already in target playlist(s) and that might be desirable or not for them)

jasonhenriquez:
Thinking about it a bit more, I think the reason why I'm having trouble with this one is because we allow duplicates. If we didn't have duplicates, and duplicates were automatically purged, it would be neat information to know how many of the videos I'm adding to a playlist will actually be added vs. are already there. As it stands, adding, e.g., 30/50 duplicate videos into another playlist is just a bad time, as you'd probably want to go back and manually get rid of the duplicates.

jasonhenriquez:
> I did see a few Youtube playlists have duplicate videos already

I wouldn't think removing support for duplicates in user playlists would affect this, no? Unless the point is that we should heed following their lead on this front. My response to that is that we have far more copying and multi-adding mechanisms than YT does, that are made much more inconvenient by the support for duplicates. The only circumstance I would be okay with keeping duplicate support is if adding duplicate entries was a very intentional act, and could not be done accidentally.

PikachuEXE:
> But I would imagine the ratio of "no" to "yes" answers would be at least 10:1.

Open a discussion and consult enough people in this case
Also for music playlists it's quite possible to add duplicate videos
So at most disable adding duplicates by default (if proven somehow someone wants that mode

PikachuEXE:
Probably replaced by https://github.com/FreeTubeApp/FreeTube/pull/5044

PR #4885, opened 2024-04-06
Use more generic icon for `Quick bookmark` feature

MarmadileManteater:
# Use more generic icon for `Quick bookmark` feature

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description
The quick bookmark feature currently uses a `clock` as its icon. This makes sense for if the quick bookmark playlist is set to Watch Later, but it can be changed to different playlists such as Favourites. Because of this, I believe it makes sense for the icon for quick bookmark to be more generic. I believe the `bookmark` icon is a great candidate because it is already used in reference to the Playlists view. 

## Screenshots 

|before|after|
|--|--|
|image|image|

|before|after|
|--|--|
|image|image|

## Desktop

 - OS: Windows 10
 - OS Version: Pro Version 22H2 Installed on ‎4/‎3/‎2022 OS build 19045.4170 Experience Windows Feature Experience Pack 120.2212.4180.0
- **FreeTube version:** 8b16bd5d6d10e3a23db4ef30d3094068d72b8656


ChunkyProgrammer (approved):

jasonhenriquez:
**suggestion:** Another concern that was mentioned was the color not having enough of a contrast with most themes to be visually remarkable. Thoughts on (re-)sprucing it?

MarmadileManteater:
> **suggestion:** Another concern that was mentioned was the color not having enough of a contrast with most themes to be visually remarkable. Thoughts on (re-)sprucing it?

It could be the primary theme colour:
image

image

image

efb4f5ff-1298-471a-8973-3d47447115dc:
Would the secondary theme color confusing?

MarmadileManteater:
> Would the secondary theme color confusing?

I don't think it would be confusing. That would make it the same colour as the buttons it is next to on the watch page, except inverted as far as the icon and button are concerned.

image

However, I personally think the primary colour pops a little more next to the other secondary colour buttons.

absidue (dismissed):

PikachuEXE:
I doubt using primary color is a good idea though
I wonder why all videos in my history were bookmarked already
It's not the primary action (like an OK button on a confirmation prompt) so I see no reason to use a different color on the icon 
Also it looks strange in my color scheme at least

image

MarmadileManteater:
> I doubt using primary color is a good idea though I wonder why all videos in my history were bookmarked already It's not the primary action (like an OK button on a confirmation prompt) so I see no reason to use a different color on the icon Also it looks strange in my color scheme at least
> 
> image

I admit it does look like the video is already bookmarked especially considering the fact that, previously, a lack / presence of colour was used to denote the state of that button. 

I like how it looks without colour, but I was trying to make it pop more.

jasonhenriquez:
I liked that the old system of theme-specific favorites icon colors ensured a higher contrast with the given theme regardless of primary/secondary color choices. Using the primary/secondary colors for this is tough because it puts it in visual competition with the other primary/secondary-colored controls we have throughout the app. So 1. that, 2. increasing the actual icon size (as done here), and 3. (in the case of the video overlay version) increasing the opacity would resolve the problem with this feature that users are currently having in my opinion.

MarmadileManteater:
> I liked that the old system of theme-specific favorites icon colors ensured a higher contrast with the given theme regardless of primary/secondary color choices. Using the primary/secondary colors for this is tough because it puts it in visual competition with the other primary/secondary-colored controls we have throughout the app. So 1. that, 2. increasing the actual icon size (as done here), and 3. (in the case of the video overlay version) increasing the opacity would resolve the problem with this feature that users are currently having in my opinion.

what do you think of this?
image

PikachuEXE:
I think there is no need to change the icon size, the opacity change is good enough
image

Also it would introduce a visual issue in playlist
image

jasonhenriquez:
It looks really good now! My final suggestions would be 1) to create a new opacity just for this case rather than overriding the value of `$thumbnail-overlay-opacity` for its other use cases (e.g., in watched styling) and 2) to try out the size increase in `watch-video-info` too.

jasonhenriquez (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
great teamwork! :)

PR #4892, opened 2024-04-07
Add fallback to "es" when strings are missing in "es_AR" and "es-MX"

absidue:
# Add fallback to "es" when strings are missing in "es_AR" and "es-MX"

## Pull Request Type

- [x] Feature Implementation

## Description
As discussed in the Matrix chat, the `español (Argentina)` and `español (MX)` locales are currently 44% and 66% translated respectively, whereas the `español (España)` locale is 100% translated. To improve the user experience for the users of the `español (Argentina)` and `español (MX)` locales, this pull request adds a fallback to `español (España)` for missing strings (it will still fallback to `en-US` if the strings are also missing from `es`).

https://hosted.weblate.org/projects/free-tube/#languages

This means that the fallback strategies look like this now:

| Locale | Fallback strategy |
| --- | --- |
| español (Argentina) | `es_AR -> es -> en-US` |
| español (MX) | `es-MX -> es -> en-US` |
| all other locales | `{locale} -> en-US` |

## Testing 
Set the display language to `español (Argentina)` and `español (MX)` and check that the logs in the console say that it is falling back to `es` instead of `en-US`.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/i18n/index.js
+++ b/src/renderer/i18n/index.js
@@ -8,7 +8,16 @@ Vue.use(VueI18n)
 
 const i18n = new VueI18n({
   locale: 'en-US',
-  fallbackLocale: { default: 'en-US' }
+  fallbackLocale: {
+    // https://kazupon.github.io/vue-i18n/guide/fallback.html#explicit-fallback-with-decision-maps
+
+    // es_AR -> es -> en-US
+    es_AR: ['es'],
+    // es-MX -> es -> en-US
+    'es-MX': ['es'],
+    // any -> en-US
+    default: ['en-US'],
+  }
 })
 
 export async function loadLocale(locale) {
@@ -18,6 +27,7 @@ export async function loadLocale(locale) {
   }
   if (!activeLocales.includes(locale)) {
     console.error(`Unable to load unknown locale: "${locale}"`)
+    return
   }
 
   // locales are only compressed in our production Electron builds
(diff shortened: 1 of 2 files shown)


MarmadileManteater:
Do you think it would be beneficial to do the same thing for Portuguese? (since it is another language with multiple regional versions present in freetube)

PikachuEXE:
> check that the logs in the console say that it is falling back to es instead of en-US

No idea how to do this

absidue:
> Do you think it would be beneficial to do the same thing for Portuguese? (since it is another language with multiple regional versions present in freetube)

It doesn't seem to be suffering from the same lack of translations that the Spanish ones do, but I can add the fallback just in case.
- `pt`: 99%
- `pt-BR`: 100%
- `pt-PT`: 96%

absidue:
> > check that the logs in the console say that it is falling back to es instead of en-US
> 
> No idea how to do this

Open the dev tools, switch to the console tab and look at the fall back warnings from vue-i18n.

absidue:
@MarmadileManteater Added them now :)

PikachuEXE (approved):
I turn warnings off by default, too spammy

image

MarmadileManteater (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
I dont see warnings for Português (BR) fallback to pt

ChunkyProgrammer:
> I dont see warnings for Português (BR) fallback to pt

Pt-br is 100% translated so that makes sense

ChunkyProgrammer (approved):

Issue #4857, opened 2024-04-04
[Bug]: search in playlist doesn't work for channel names


abdelrahman-y:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

I can't search for channel names in playlist 

### Expected Behavior

it used to provide videos of channels names given in search box

### Issue Labels

content not loading

### FreeTube Version

v0.20.0-beta, 54b49dc

### Operating System Version

windows 11 22631

### Installation Method

.zip

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

v0.19.2

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
This has never been implemented so this isnt a bug.

vF360:
> This has never been implemented for playlists so this isnt a bug.

Why wasn't it implemented? Seems like an odd thing to not include, given it was available in previous versions.

efb4f5ff-1298-471a-8973-3d47447115dc:
Please create a feature request issue for this.

PR #4919, opened 2024-04-10
Add back author search in video playlist searching

jasonhenriquez:
# Add back author search in video playlist searching

## Pull Request Type

- [x] Feature Implementation

## Related issue
N/A (marked as bug in #4857)

## Description
Reuses previous paradigm (the paradigm currently in place in History.js). This was the pre-existing behavior before the Playlist PR. See https://github.com/FreeTubeApp/FreeTube/blob/5c8d49bf51dde8ffd8c8d291dce3e5ff27b0d9fb/src/renderer/views/UserPlaylists/UserPlaylists.js#L98-L103.

## Screenshots 
Screenshot_20240410_063611

## Testing 
- Search for video title (variegate casing, should be insensitive)
- Search for author (variegate casing, should be insensitive)

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW
- **FreeTube version:** Why do we even ask this one for PRs?

### DIFF
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -142,7 +142,11 @@ export default defineComponent({
       if (this.processedVideoSearchQuery === '') { return this.playlistItems }
 
       return this.playlistItems.filter((v) => {
-        return v.title.toLowerCase().includes(this.processedVideoSearchQuery)
+        if (typeof (v.title) !== 'string' || typeof (v.author) !== 'string') {
+          return false
+        } else {
+          return v.title.toLowerCase().includes(this.processedVideoSearchQuery) || v.author.toLowerCase().includes(this.processedVideoSearchQuery)
+        }
       })
     },
     visiblePlaylistItems: function () {


ChunkyProgrammer (dismissed):

MarmadileManteater (dismissed):

absidue (commented):
Instead of bailing out when either of them are missing it would be better to check them individually e.g. with this pseudo-code:

```
if has title and match in title
    return true

if has author and match in author
    return true

return false
```

ChunkyProgrammer (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Probably going to suggest something controversial here. 

I think i like to see a Filter on the main playlist page that enables Search by Author so i instantly know in what playlists the author is in in and if i click on that playlist that search shows me the same query as the one i just on the main playlist page. Just like how it works with the other filter

https://github.com/FreeTubeApp/FreeTube/assets/73130443/7484a798-a590-4f9b-a0e7-6bbd6f1bb1c9

jasonhenriquez:
It sounds interesting, but I don't want to add it as a feature unless a user asks for it. I don't have a use case for that personally.

PikachuEXE:
Users should be able to search for author in user playlists view too when `Playlists with matching videos` is on
Otherwise it's inconsistent since they can search for video title there now

Places to change: search `if (this.doSearchPlaylistsWithMatchingVideos) {`

jasonhenriquez:
My bad, I misread the suggestion as adding a new filter toggle for video author search specifically. Added now.

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> My bad, I misread the suggestion as adding a new filter toggle for video author search specifically. Added now.

Well i actually did mean that but Pika suggestions is way better

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4921, opened 2024-04-10
Sort videos within playlist

### LINKED ISSUE #4870: [Feature Request]: Sort videos within playlist
### Guidelines

### Problem Description

Sorting videos within the playlist isnt possible yet

### Proposed Solution

add the sort by dropdown when u go into one of your local playlist and add the following sorting options

- A-Z (Video title)
- Z-A (Video title)
- A-Z (Channel name)
- Z-A (Channel name)
- Custom (Custom sorting will be set when user moves a video up/down in the list)
- Sorting by upload date of video (Recently uploaded / Earliest uploaded)
- Sort by views (Most viewed / Least Viewed)
- Sort by video length (Shortest something? / Longest something) 
- Sort by Recently added / Earliest added

### Alternatives Considered

None.

### Issue Labels

new feature

### Additional Information

Question asked in https://github.com/PikachuEXE/FreeTube/issues/54#issuecomment-1724931583

> But the sorting shouldn't affect the stored order right? (or the playing order)
Otherwise it will be way more complicated.

Answered in https://github.com/PikachuEXE/FreeTube/issues/54#issuecomment-1724938258

> I agree but i think Custom will fulfill that if we use it by default

> To clarify some more, when a playlist gets created and u add videos to it the default filter will always be Custom. If the user selects another filter that filter will be applied but if it is not to their liking they should be able to switch back to Custom and see their previously set Custom list. This way the user doesnt have to change the whole list again to their liking

kommunarr:
# Sort videos within playlist

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4870

## Description

- Implements sorting types:
- [x] A-Z / Z-A (Video title)
- [x] A-Z / Z-A (Channel name)
- [x] Custom (Custom sorting will be set when user moves a video up/down in the list)
- [x] Sort by Recently added / Earliest added

Implementation details:
- Only for user playlists, not remote playlists.
- I didn't implement Sort By Views or Video Length because those seem quite unimportant. I can implement these if someone actually wants these features.
- The default sort preference is "Latest added first". If the user sets their sort preference in a user playlist, it applies to all of their user playlists by default.
- The order adjustment buttons (i.e., move up/down) are not present for sort orders of Custom. Rationale: we want "Custom" orderings to persist even when the sorting preference is changed to something else. 

Unrelated mini-fix:
- Adds `locale` property that was being referenced (but was actually missing) in `src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js`

## Screenshots 
Screenshot_20240410_213458

## Testing 
- Test all orders work properly on a remote playlist (tested with [this one](https://youtube.com/playlist?list=PLqdOVNygCKQKz5fsGoRN5uqwS3WnO_JY6))
- Test all orders work properly on a local playlist
- Test "move/up down" buttons show only for Custom sort orders, and that they continue to work properly

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## More info
I have a PR coming out after this is merged for handling sorting in `watch-video-playlist`. I split this up to ease the testing criteria and the relative unimportance of the other sorting types compared to the big "Added" and "Custom" ones.


absidue:
> Technically, only each individual "batch" of data is sorted for remote playlists, due to obvious technical limitations. I think this is fine.

I would prefer to not have sorting for remote playlists at all, if we can't do it properly.

absidue:
Also personally I think that there are too many sorting options, it feels a lot like a "because we can" and not because anyone would ever actually use it (all the people that complained just want a way to reverse playlists).

I think we just need 2:
- Default, respects the order in the playlist, move buttons are shown
- Reversed (just reverses the playlist, otherwise respects the order in the playlist, move buttons shown but move in the correct direction in the UI, so opposite in code)

ChunkyProgrammer:
I think these are good options but the select box definitely seems to have a lot of options for a `sort by`. I wonder if there should be a separate control (ex: checkbox) for ascending/descending? That way the amount of items in the select box will be halved and reversing the playlist would take one click.

jasonhenriquez:
I'm not opposed to that. Thoughts? @absidue @efb4f5ff-1298-471a-8973-3d47447115dc @PikachuEXE

PikachuEXE:
Disagree
Not every sort order must have a "reverse" one (weak argument I know
Having 2 UI components just to get a sort order seems unnecessary, non-standard UI/UX
You can reorder the options if you want

PikachuEXE:
Can sort by component be placed on the right like most other places?
image
image
image

static/locales/en-US.yaml
```diff
@@ -871,6 +870,20 @@ Playlist:
...
+  Sort By:
+    Sort By: Sort By
+    Custom: Custom
+    CustomDescending: Custom (descending)
+    Default: Default
+    DefaultDescending: Default (descending)
+    DateAddedNewest: Date added (newest)
+    DateAddedOldest: Date added (oldest)
+    DatePublishedNewest: Date published (newest)
+    DatePublishedOldest: Date published (oldest)
```
PikachuEXE, line 882: Also it's not date but time but I am fine with the key
```suggestion
    DateAddedNewest: Latest added first
    DateAddedOldest: Earliest added first
    DatePublishedNewest: Latest published first
    DatePublishedOldest: Earliest published first
```

static/locales/en-US.yaml
```diff
@@ -871,6 +870,20 @@ Playlist:
   View: View
   Views: Views
   Last Updated On: Last Updated On
+  Sort By:
+    Sort By: Sort By
+    Custom: Custom
+    CustomDescending: Custom (descending)
```
PikachuEXE, line 876: `Custom` already sound a bit weird to me
`Custom (descending)` makes it even more weird

I guess this means "user specified" order
`User-defined`? (Suggest a better if you can

PikachuEXE (changes requested):

jasonhenriquez:
I'm seeing some different perspectives on this from everyone. I think the option I'm leaning to the most is removing "Custom (Descending)", implementing @PikachuEXE's position (& most likely label) changes, and making "added descending" the default. With "Custom (Descending)" gone we're are at 7 playlist sort preferences, compared to YT's 5, which isn't what I would call excessive. I think it's more that the current ordering & wording of the labels could stand to be adjusted to reduce cognitive complexity. Let me know what y'all think.

jasonhenriquez:
Here's how it looks like now:

Screenshot_20240410_213458

PikachuEXE:
Why isn't custom the default sort order?

jasonhenriquez:
Rationale being that "Latest added first" is the previous behavior before the Playlist PR, and seems to be the most requested ordering from people. Custom ordering seems to be relatively less leveraged on average and thus less suited to be the default behavior.

PikachuEXE:
I am fine having default option changed with the last used sort order is saved

But we should disable order changing UI unless the sort order is "custom"
image

(YT just persists the sort order when a sort order option picked, it sucks)

jasonhenriquez:
> But we should disable order changing UI unless the sort order is "custom"

One step ahead of you:
https://github.com/FreeTubeApp/FreeTube/blob/9db7e481a43711da4e55e8b5e7a5a65b906d9228/src/renderer/views/Playlist/Playlist.vue#L76-L77

PikachuEXE (approved):

src/renderer/views/Playlist/Playlist.js
```diff
@@ -15,6 +16,19 @@ import {
 import { extractNumberFromString, setPublishedTimestampsInvidious, showToast } from '../../helpers/utils'
 import { invidiousGetPlaylistInfo, youtubeImageUrlToInvidious } from '../../helpers/api/invidious'
 
+const SORT_BY_VALUES = {
+  // TODO: store video.published for user playlists
+  // DatePublishedNewest: 'date_published_newest',
+  // DatePublishedOldest: 'date_published_oldest',
```
absidue, line 22: If you want to implement this, you'll have to keep in mind that shorts don't have a published date on the shorts channel tab or in shorts only playlists (not returned in youtube's response, thats's why we always use RSS on the shorts tab on the subscriptions page), so if someone were to add a video to a playlist from there, it would make it impossible to sort by published. For the watch history we just don't store the published date if it isn't available, but we also don't have any sorting functionality that would depend on it.

jasonhenriquez: Thanks for this context absidue. Just removedd the code for these.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Only show sort by drowdown when there are > 1 vids in list

efb4f5ff-1298-471a-8973-3d47447115dc:
Seeing some weird behavior when moving vids up and down in the list when custom is set

https://github.com/FreeTubeApp/FreeTube/assets/73130443/4ecd1138-8083-46ab-9619-90d097cad0ab

jasonhenriquez:
Do a `yarn`. As for the animation oddity, I don't know if it's related to this PR or just the way the animation is, but I'll look into it

efb4f5ff-1298-471a-8973-3d47447115dc:
I almost agree with everything on the current implementation. However i would want to talk some more about the things i see differently:

1. User chooses to sort playlist Author (Z-A) -> now they select Custom and the sorting reverses back. Why should the sort change? Maybe they want to make changes in their custom list based on the previous selected sort (maybe im being stupid here bc im testing this with playlists that aren't heavily modified in the custom order by me).
2.  > If the user sets their sort preference in a user playlist, it applies to all of their user playlists by default. Why does it apply globally and not individually?

jasonhenriquez:
It's a fight between customizability and ease of use. If everyone wants to see all of their playlists in the same order (more common use case, IMO), they would have to make N*2 clicks to do that under that system, versus 2 clicks under this one. And then you can talk about having a second setting for setting a default for all versus just for one playlist, but that's just adding too much stuff.

efb4f5ff-1298-471a-8973-3d47447115dc:
Only waiting for weird visual behavior to be solved before i approve

jasonhenriquez:
This issue seems to be happening in the `development` branch as well. Still investigating it, but seems like the "move" animation" used to only happen once "to", but now it's happening "to" & "fro".

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
I think this fully closes the issue. It doesn't implement a few options i suggested in there but that's fine as they don't seem that important

PikachuEXE (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -107,7 +107,9 @@ export default defineComponent({
     newPlaylistDefaultProperties: function () {
       return this.$store.getters.getNewPlaylistDefaultProperties
     },
-
+    locale: function () {
```
ChunkyProgrammer, line 110: Is this used?

jasonhenriquez: > Adds locale property that was being referenced (but was actually missing) in src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
https://github.com/FreeTubeApp/FreeTube/blob/8fd5055c4b68fb39a7268c161105f49061f44dd7/src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js#L54-L90

ChunkyProgrammer: Oops, missed that comment.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):

Issue #4924, opened 2024-04-10
[Feature Request]: Hide Searchbar and Search filter in Add to Playlist when there are < 2 playlists 


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

On the Playlist page we hide, Search, Sorting and Search filter if there 0 or 1 playlist. In the add to playlist popup window this is only true for Sorting but not for Search and Search filters.

### Proposed Solution

Hide Searchbar and Search filter in Add to Playlist when there are < 2 playlists 

### Alternatives Considered

N/A

### Issue Labels

visual improvement

ShivamK2002:
I want to contribute on this @kommunarr @efb4f5ff-1298-471a-8973-3d47447115dc

efb4f5ff-1298-471a-8973-3d47447115dc:
@anurag2787 go for it :)

efb4f5ff-1298-471a-8973-3d47447115dc:
> ### Additional Information

> Search is also available for the Channel List and History page when there is one channel/video in there

I think it would be good if you also implement this in the same PR

anurag2787:
> > ### Additional Information
> 
> > Search is also available for the Channel List and History page when there is one channel/video in there
> 
> I think it would be good if you also implement this in the same PR

Sure! I will work on implementing this for the Channel List and History page as well in the same PR

PR #4929, opened 2024-04-11
Improve video playlist search UX

jasonhenriquez:
# Improve video playlist search UX

## Pull Request Type

- [x] Feature Implementation

## Related issue
None

## Description
Clicking a button to enable/disable search has felt pretty arbitrary and cumbersome in my experience. This PR:
1. Removes the playlist video search button & makes the search bar appear alongside the other `playlist-info` controls by default, where `playlistInVideoSearchMode` is now true when the search bar is non-empty.
2. Focuses the playlist search bar on Ctrl+F. This is an experimental change testing how people like the convenience of Ctrl+F for a route with a prominent search bar. I, personally, have been pressing it in my personal usage on this page thinking it would focus this bar, to minor frustration.

## Screenshots 

Screenshot_20240411_091333

## Testing 
- (REG) Ensure search filter is populated when searching for a video from the User Playlists page
- Ensure search filter is active/inactive based on if trimmed search input is empty
- Try using the icons with search filter active (I thought this was a reason why we had this implemented like this, but it turns out all of the playlist options work just the same)

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
> Removes the playlist video search button & makes the search bar appear alongside the other playlist-info controls by default, where playlistInVideoSearchMode is now true when the search bar is non-empty.

Please Hide search when there is 0 or 1 vids in the playlist. I see no reason for it to be there.

> Focuses the playlist search bar on Ctrl+F. This is an experimental change testing how people like the convenience of Ctrl+F for a route with a prominent search bar. I, personally, have been pressing it in my personal usage on this page thinking it would focus this bar, to minor frustration.

Please list it in https://docs.freetubeapp.io/usage/keyboard-shortcuts/ after PR is merged.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
This PR does not close #4924. That is about the add to playlist window.

PikachuEXE (dismissed):
No necessary in this PR
Maybe also add Ctrl+F to playlists view to be consistent?

efb4f5ff-1298-471a-8973-3d47447115dc:
> This is an experimental change testing how people like the convenience of Ctrl+F for a route with a prominent search bar.

Thinking about this again and i think this should belong in the Experimental Settings with a toggle

jasonhenriquez:
> Thinking about this again and i think this should belong in the Experimental Settings with a toggle

Honest q: why? This is something people will either type in or not type in.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -259,6 +259,12 @@ export default defineComponent({
 
     this.updateQueryDebounce = debounce(this.updateQuery, 500)
   },
+  mounted: async function () {
```
absidue, line 262: This function doesn't run any async code and adding an event listener is a quick enough operation that we can do it synchronously.
```suggestion
  mounted: function () {
```

absidue (commented):
Could you please update the CSS for the `searchInputsRow` class in `ft-playlist-info.scss` and remove the grid stuff, as that causes a wider margin on the right side than the left side, as there is a blank colum with an 8px gap.

src/renderer/components/playlist-info/playlist-info.js
jasonhenriquez: This is copied verbatim from another usage; removed it from both instances now.

absidue: Thank you for removing it in both places.

PikachuEXE:
@absidue 's comment about `searchInputsRow` addressed?

jasonhenriquez:
I somehow missed that comment. So many self-facepalms since coming back lol. Lemme check

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
Either no margin at all or have margin on both sides
Also I think there should be spacing between the input box and the buttons above (8px?)
image

SC after applying my suggested changes
image

jasonhenriquez:
@PikachuEXE Thank you for concealing my secret that I put `inline` instead of `block` and forgot to hit save 🤦‍♂️

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #4933, opened 2024-04-12
Set max height for storyboards depending on window width

MarmadileManteater:
# Set max height for storyboards depending on window width

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [X] Other (visual issue)

## Description
The current implementation picks the largest storyboard possible. However, the size of these storyboards can end up being a bit unwieldy especially on small displays. This PR addresses this by setting the storyboards to a max height of 90px when the screen is narrow `<=500px`.

## Screenshots 
|before|after|
|--|--|
|image|image|

## Testing 
**testing that the storyboard is a different size on smaller displays**
1. Ensure `Local API` is your backend preference
2. Open a video with large storyboards (ex: https://youtu.be/qcH2wgRLiV8)
3. Reduce the window width to below `500px`
4. Ensure the storyboard used is smaller than before
 
**testing that the resize listener is being removed properly**
1. Repeat steps 1 and 2 above
2. Open the inspect element to the `Sources` tab
3. Expand the "global listeners" to see all resize events bound to the window
4. Navigate away from the video (to another video or to another view)
5. Refresh the global listeners in the sources tab
6. See that there are no active resize listeners linked to `Watch.js`

## Desktop

- **OS:** Windows 10
- **OS Version:** 22H2 (OS Build 19045.4170)
- **FreeTube version:** af2913592ee0e989dba8f5909ff7909f9bc42516

## Additional context
I had considered simply setting the max-height to 90px without a resize listener (since that is already used as the max height for invidious storyboards). That might be a valid option. I'm not a big fan of storing the callback in the `data`. I would love a way to easily bind an anonymous function to the window `resize` that I can clean up later if anyone can think of one. I would simply define it as a method, but it requires data from the `result` object which is only in scope within the `getVideoInformationLocal` function. That is why I used a method which returns a callback instead of a method as a callback.

### DIFF
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -126,6 +126,7 @@ export default defineComponent({
       commentsEnabled: true,
 
       onMountedRun: false,
+      storyboardResizeCallback: null
     }
   },
   computed: {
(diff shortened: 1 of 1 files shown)


absidue:
Could we just check window.innerWidth once when we set `this.storyboardSrc`, without the resize listener? I'm not a big fan of messing around in player internals.

MarmadileManteater:
> Testing cases still the same after latest commit?

They have a slightly different order now (and there are less of them). I just updated it

absidue (approved):

ChunkyProgrammer (approved):

jasonhenriquez (approved):

PR #4935, opened 2024-04-12
Fix imports of the electron "module" that are missing an Electron guard

absidue:
# Fix imports of the electron "module" that are missing an Electron guard

## Pull Request Type

- [x] Build improvement

## Related issue
Noticed them while working on #4931

## Description
Currently some of our imports of the `electron` module are correctly guarded by `IS_ELECTRON` checks, others unfortunately are not which means the web webpack config (and the downstream Android one) need to tell webpack to ignore the `electron` imports. This pull request adds the guards in the places that there were missing.

I recommend turning on GitHub's `Hide Whitespace` setting while reviewing this pull request, especially for the `src/renderer/store/modules/settings.js` file.

## Testing 
Run `yarn run pack:web`, if it doesn't complain about not being able to bundle Electron, I didn't forget to add a guard.

## Desktop

- **OS:** Windows
- **OS Version:** 10


ChunkyProgrammer (approved):
Should we add a `yarn run pack:web` to the linter ci too so we can make sure we dont merge things that break the web builds? (Even though we dont officially support the web builds) https://github.com/FreeTubeApp/FreeTube/blob/e65946294e487fcf8fbeb06175acb53f2b280df7/.github/workflows/linter.yml#L29

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4936, opened 2024-04-12
Local API: Fix extracting the published date and view count in playlists

absidue:
# Local API: Fix extracting the published date and view count in playlists

## Pull Request Type

- [x] Bugfix

## Description
When a video has it's view count hidden inside a playlist, we don't extract the published date, this pull request fixes that. While fixing that I also noticed that the extraction of the view count for videos that do have one, is broken too, due to YouTube changing the accessibility text. The view count used to be at the end of the accessibility text, now it has the following structure: `{title} by {uploader} {view count} views {relative published date} {written out duration e.g. "22 minutes"}` and if the view count is missing `{title} by {uploader} {relative published date} {written out duration e.g. "22 minutes"}`.

The Invidious API doesn't return the view count or the published date, presumably because extracting them is a bit ugly.

## Testing 
View the following playlists with the local API.
Some vidoes are missing their view counts: https://youtube.com/playlist?list=PLBsP89CPrMeO7uztAu6YxSB10cRMpjgiY
Live: https://youtube.com/playlist?list=PL6NdkXsPL07Il2hEQGcLI4dg_LTg7xA2L

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -697,14 +697,16 @@ export function parseLocalPlaylistVideo(video) {
     // the accessiblity label contains the full view count
     // the video info only contains the short view count
     if (video_.accessibility_label) {
-      const match = video_.accessibility_label.match(/([\d,.]+|no) views?$/i)
+      // the `.*\s+` at the start of the regex, ensures we match the last occurence
+      // just in case the video title also contains that pattern
+      const match = video_.accessibility_label.match(/.*\s+([\d,.]+|no)\s+views?/)
 
       if (match) {
         const count = match[1]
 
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):

jasonhenriquez (approved):
LGTM.

Unrelated to this PR, but in a similar vein, videos in search results with 0 views show up as "NaN views".

Screenshot_20240413_093708

absidue:
Ah good catch, I'll fix that in a separate PR :)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #4943, opened 2024-04-13
Fix/ Improve accessibility of playlist icon buttons

### LINKED ISSUE #4940: [Bug]: Playlist action icon accessibility issues
### Guidelines

### Describe the bug

1. Open a video inside of a playlist and hit the Shuffle, Loop, Reverse, or Pause button. 

- See that color is the only thing indicating whether the control is active or not. This makes color the only way to distinguish between these states of activity, thus introducing the inaccessibility anti-pattern of [color dependence](https://guides.18f.gov/accessibility/color/).`
- See that `aria-pressed` is not set to true.

Screenshot_20240412_190412

### Expected Behavior

- Add a separate visual indication to the control when it is active and/or inactive. Example from YT: the icon's font-weight increases.
- Add `aria-pressed: true` when a toggle is active.

Screenshot_20240412_191818

### Issue Labels

accessibility issue, usability issue

### FreeTube Version

v0.20.0

### Operating System Version

OpenSUSE TW

### Installation Method

.zip

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

sossost:
# Title
Improve accessibility of playlist icon buttons 

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #4940 

## Description
The aria-pressed attribute has been added to playlist icon buttons.
In addition, the line thickness when the icon is deactivated has been adjusted to be slightly thinner, and when activated, the line thickness has been adjusted to be slightly thicker, so that it is possible to tell whether the icon is activated through not only the color but also the thickness of the icon.

## Screenshots 
before before

after after

## Desktop

- **OS: macOs**
- **OS Version: Sonoma 14.4.1**
- **FreeTube version: v0.20.0 Beta**


src/renderer/components/watch-video-playlist/watch-video-playlist.css
```diff
@@ -51,16 +51,22 @@
...
+  stroke: var(--bg-color);
 }
 
 .playlistIcon:hover {
   background-color: var(--side-nav-hover-color);
   color: var(--side-nav-hover-text-color);
+  stroke-width: 10;
+  stroke: var(--side-nav-hover-text-color);
   transition: background 0.2s ease-in;
 }
```
jasonhenriquez, line 64:
```suggestion
.playlistIcon:hover {
  background-color: var(--side-nav-hover-color);
  color: var(--side-nav-hover-text-color);
  transition: background 0.2s ease-in;
}
```

This looks great @sossost, thanks for doing this so quickly! My only suggestion is that we keep the inactive hover state the same, as otherwise this makes it seem like it's active.

sossost: Thanks for the feedback, @jasonhenriquez! I appreciate your quick review. I agree with keeping the inactive hover state unchanged to avoid confusion about its activity status.
In this case, need to change the stroke color to the hover background color without deleting the style from the hover properties.
I used stroke color to make the icons thin without changing them.
스크린샷 2024-04-14 오전 1 24 13
스크린샷 2024-04-14 오전 1 23 57


I'll make the necessary adjustments and update the pull request shortly.

jasonhenriquez: Thanks for the update! However, It's now not keeping the wider stoke width on active hover.

sossost: Thanks for pointing that out! I missed updating the stroke width for the active hover state. I'll add a rule for `.playlistIconActive:hover` to ensure the wider stroke width and active color are maintained during hover. I'll make this change and update the pull request shortly.

jasonhenriquez: Thanks @sossost! This is working perfectly now. My last thought is that while we're here, can you make the `.playlistIconActive:hover`'s `stroke` & `color` be `--accent-color-hover`? I think this was another pre-existing bug I didn't even notice.

sossost: Thanks, @ jasonhenriquez! I'm glad to hear it's working well now. I agree with your suggestion to update the `.playlistIconActive:hover` stroke and color to `--accent-color-hover`. It sounds like a good catch on a pre-existing issue. I'll make these changes and push the update soon.

jasonhenriquez (approved):
Awesome stuff. This looks much better to use and is now far more accessible to our disabled users as well. LGTM

sossost:
@jasonhenriquez Thank you so much for taking the time to review and for your positive feedback! I'm happy to hear that the changes improve usability and accessibility. I love working together to improve the experience for all our users.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Thanks for addressing this @sossost :)

ChunkyProgrammer (approved):

PR #4947, opened 2024-04-13
Fix user playlist reverse button behavior

jasonhenriquez:
# Fix user playlist reverse button behavior

## Pull Request Type

- [x] Bugfix

## Related issue
None

## Description

Co-author: @absidue (identified where issue was occurring)

User playlists that are reversed revert to their normal ordering after the next video is loaded. This PR fixes that by making our `parseUserPlaylist` function maintain a reversed order. It also removes a `watch` property that @absidue identified as not relevant.

## Testing 
- Reverse a playlist and then navigate to another video in the playlist. See that the playlist is still reversed

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
CC: @PikachuEXE if we're missing a purpose for this `watch` property. Its removal isn't strictly required for this fix.

Note: we're doing `x = x.toReversed()` instead `x.reverse()` to avoid a warning for mutating Vuex data.

### DIFF
--- a/src/renderer/components/watch-video-playlist/watch-video-playlist.js
+++ b/src/renderer/components/watch-video-playlist/watch-video-playlist.js
@@ -166,13 +166,6 @@ export default defineComponent({
       // Re-fetch from local store when current user playlist updated
       this.parseUserPlaylist(this.selectedUserPlaylist, { allowPlayingVideoRemoval: true })
     },
-    playlistItemId (newId, _oldId) {
-      // Playing online video
-      if (newId == null) { return }
-
-      // Re-fetch from local store when different item played
-      this.parseUserPlaylist(this.selectedUserPlaylist, { allowPlayingVideoRemoval: true })
-    },
     videoId: function (newId, oldId) {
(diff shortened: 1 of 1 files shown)


PikachuEXE:
Not related but found that no progress bar displayed when playlist reversed (remote + local)
Expected?
image

PikachuEXE (approved):
About removal of `playlistItemId` in `watch`
Seems fine for playlist changes I made when playing the playlist
Should be fine to remove

jasonhenriquez:
> Not related but found that no progress bar displayed when playlist reversed (remote + local)
Expected?

Apparently? I'll have to check to see if there's a rationale.
https://github.com/FreeTubeApp/FreeTube/blob/372bc1bdd104f18d8c80ff6853603f40f1cec026/src/renderer/components/watch-video-playlist/watch-video-playlist.vue#L43-L49

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #4949, opened 2024-04-14
User playlists as grid

### LINKED ISSUE #4708: [Feature Request]: Option to use grid display in playlists
### Guidelines

### Problem Description

I actually miss the previous UI where I was simply using the save videos as watch later. I'm ok with the new way of having a star button that simply saves to a default playlist, but unfortunately, since I still use this playlist as watch later, it got quite big and is quite difficult to navigate as a list of videos instead of a grid as it used to be.

### Proposed Solution

Introduce an option to display playlists using the grid view that already exist within the application.

### Alternatives Considered

-

### Issue Labels

### Additional Information

Ideally, it would also be nice to be able to sort the list by date added, at least, in default and reverse order, but I guess it would require to open another issue here. 😉

jasonhenriquez:
# User playlists as grid

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4708 

## Description
- Implements grid display option for user playlists
- Attaches this behavior to the existing grid versus list Video View Type setting
- Remote playlists will still use the list view, as has been the case

## Screenshots 
Screenshot_20240414_084944

## Testing 
- Check editing the name and description of the user playlist to varying lengths
- Check on mobile, tablet, & desktop views
- Check remote playlists are not affected

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
Implements the "fixed top bar" as a mixin that I would like to reuse for #4380, either adding it in this PR or into the other depending on which gets merged first.

Waiting on #4921 and #4929 to get merged first.

### DIFF
--- a/src/renderer/scss-partials/_ft-list-item.scss
+++ b/src/renderer/scss-partials/_ft-list-item.scss
@@ -1,6 +1,35 @@
 $thumbnail-overlay-opacity: 0.85;
 $watched-transition-duration: 0.5s;
 
+@mixin is-side-nav-open {
+  @at-root {
+    .sideNavOpen#{&} {
+      @content;
+    }
+  }
+}
+
+@mixin fixed-top-bar {
+  position: fixed;
+  inset-block-start: 60px;
+  inset-inline-end: 0;
+  inline-size: calc(100% - 80px);
+  margin-inline: 0;
+  padding-inline: 10px;
+
+
+  @include is-side-nav-open {
+    inline-size: calc(100% - 200px);
+  }
+
+  @media only screen and (max-width: 680px) {
+    inline-size: 100%;
+    @include is-side-nav-open {
+      inline-size: 100%;
+    }
+  }
+}
+
 @mixin is-result {
   @at-root {
     .result#{&} {
--- a/src/renderer/views/Playlist/Playlist.scss
+++ b/src/renderer/views/Playlist/Playlist.scss
@@ -1,34 +1,76 @@
-.routerView {
+@use '../../scss-partials/ft-list-item';
+
+.playlistItemsCard {
   display: flex;
+  flex-direction: column;
+  grid-gap: 10px;
+  margin: 0;
+  padding: 10px;
 }
 
-.playlistInfo {
-  background-color: var(--card-bg-color);
+.playlistInfoContainer {
   box-sizing: border-box;
-  block-size: calc(100vh - 132px);
-  margin-inline-end: 1em;
   padding: 10px;
-  position: sticky;
-  inset-block-start: 96px;
   /* This is needed to make prompt always above video entries */
   /* Value being too high would block search suggestions */
   z-index: 1;
-  inline-size: 30%;
-
   &.promptOpen {
     // Otherwise sidebar would be above the prompt
     z-index: 200;
   }
 }
 
-.playlistItems {
-  display: flex;
-  flex-direction: column;
-  grid-gap: 10px;
-  margin: 0;
-  padding: 10px;
-  inline-size: 60%;
+.playlistInfo {
+  background-color: var(--card-bg-color);
 }
+
+.routerView {
+  display: flex;
+
+  &.grid {
+    flex-direction: column;
+    margin-block-start: 200px;
+
+    .playlistItemsCard {
+      inline-size: 85%;
+      margin-block: 0px 60px;
+      margin-inline: auto;
+    }
+
+    .playlistInfoContainer {
+      @include ft-list-item.fixed-top-bar;
+      padding-block: 0;
+    }
+
+    .playlistInfo {
+      inline-size: 85%;
+      margin-inline: auto;
+      padding-block: 10px;
+      padding-inline: 16px;
+      box-shadow: 0 2px 1px 0 var(--primary-shadow-color);
+      box-sizing: border-box;
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+    }
+  }
+
+  &.list {
+    .playlistItemsCard {
+      inline-size: 60%;
+    }
+
+    .playlistInfoContainer {
+      block-size: calc(100vh - 132px);
+      inline-size: 30%;
+      inset-block-start: 96px;
+      margin-inline-end: 1em;
+      position: sticky;
+      inline-size: 60%;
+    }
+  }
+}
+
 
 .playlistItem {
   display: grid;
@@ -74,18 +116,20 @@
 @media only screen and (max-width: 850px) {
   .routerView {
     flex-direction: column;
+
+    &.list {
+      .playlistInfo {
+        box-sizing: border-box;
+        position: relative;
+        inset-block-start: 0;
+        z-index: 1;
+        block-size: auto;
+        inline-size: 100%;
+      }
+    }
   }
 
-  .playlistInfo {
-    box-sizing: border-box;
-    position: relative;
-    inset-block-start: 0;
-    z-index: 1;
-    block-size: auto;
-    inline-size: 100%;
-  }
-
-  .playlistItems {
+  .playlistItemsCard {
     box-sizing: border-box;
     inline-size: 100%;
   }
@@ -93,4 +137,16 @@
 
 .message {
   color: var(--tertiary-text-color);
+}
+
+@media only screen and (max-width: 680px) {
+  .routerView.grid {
+    margin-block-start: 218px;
+    .playlistInfoContainer {
+      position: absolute;
+    }
+    .playlistInfo {
+      inline-size: 90%;
+    }
+  }
 }
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -102,6 +102,10 @@ export default defineComponent({
       return this.$store.getters.getBaseTheme
     },
 
+    isSideNavOpen: function () {
+      return this.$store.getters.getIsSideNavOpen
+    },
+
     mainColor: function () {
       return this.$store.getters.getMainColor
     },
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -4,7 +4,8 @@
     id="app"
     :class="{
       hideOutlines: outlinesHidden,
-      isLocaleRightToLeft: isLocaleRightToLeft
+      isLocaleRightToLeft: isLocaleRightToLeft,
+      isSideNavOpen: isSideNavOpen
     }"
   >
     <top-nav ref="topNav" />
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -115,6 +115,9 @@ export default defineComponent({
     listType: function () {
       return this.$store.getters.getListType
     },
+    userPlaylistListType: function () {
+      return this.$store.getters.getUserPlaylistListType
+    },
     thumbnailPreference: function () {
       return this.blurThumbnails ? 'blur' : this.$store.getters.getThumbnailPreference
     },
@@ -253,6 +256,7 @@ export default defineComponent({
       'updateLandingPage',
       'updateRegion',
       'updateListType',
+      'updateUserPlaylistListType',
       'updateThumbnailPreference',
       'updateForceLocalBackendForLegacy',
       'updateCurrentLocale',
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -115,9 +115,6 @@ export default defineComponent({
     listType: function () {
       return this.$store.getters.getListType
     },
-    userPlaylistListType: function () {
-      return this.$store.getters.getUserPlaylistListType
-    },
     thumbnailPreference: function () {
       return this.blurThumbnails ? 'blur' : this.$store.getters.getThumbnailPreference
     },
@@ -256,7 +253,6 @@ export default defineComponent({
       'updateLandingPage',
       'updateRegion',
       'updateListType',
-      'updateUserPlaylistListType',
       'updateThumbnailPreference',
       'updateForceLocalBackendForLegacy',
       'updateCurrentLocale',
--- a/src/renderer/components/general-settings/general-settings.vue
+++ b/src/renderer/components/general-settings/general-settings.vue
@@ -57,6 +57,13 @@
         @change="updateListType"
       />
       <ft-select
+        :placeholder="$t('Settings.General Settings.User Playlist View Type')"
+        :value="userPlaylistListType"
+        :select-names="viewTypeNames"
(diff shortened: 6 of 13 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
> Waiting on #4921 and #4929 to get merged first.

They're both merged :)

efb4f5ff-1298-471a-8973-3d47447115dc:
Ready for review but PR is still in draft?

jasonhenriquez:
You got to give me a minute, I'm multitasking lol

efb4f5ff-1298-471a-8973-3d47447115dc:
Srry just excited to test :)

efb4f5ff-1298-471a-8973-3d47447115dc:
Search results playing hide and seek now

https://github.com/FreeTubeApp/FreeTube/assets/73130443/c67eb382-f538-4fb6-aa06-4ed5a3c6d67d

Seems to be some overlap with the top bar

Capture

Grid is enabled -> go to remote playlist -> when fetching the playlist notice a weird rectangular artifact

https://github.com/FreeTubeApp/FreeTube/assets/73130443/2505dcf4-eb5f-42e2-8277-29bf0e3d3978

When quick bookmark is enabled you will see that all the videos in the list receive the checkmark. This doesnt happen in list mode because it doesnt make sense to do that because you can just remove the vids by clicking the delete icon on the thumbnail

How it works atm in list mode

https://github.com/FreeTubeApp/FreeTube/assets/73130443/576e4ac5-59e7-419b-85c8-8b2dd3b1d63e

How it works in this PR on grid mode

https://github.com/FreeTubeApp/FreeTube/assets/73130443/c58d3af2-56e3-42a3-956b-8a81ff347156

Todo: 

icons on thumbnails are fixed instead of showing on hover
video numbering is missing on grid mode
sorting custom doesnt make me change the order of the videos by hovering on the thumbnail and clicking the arrow icon

Capture2

jasonhenriquez:
I couldn't get video numbering to look good on grid view. If anyone has any suggestions for it, please share, but I'm going to leave that one by the wayside for the time being.

jasonhenriquez:
> Search results playing hide and seek now

Emma's fix will cover this one too once it's rebased thankfully

jasonhenriquez:
> Grid is enabled -> go to remote playlist -> when fetching the playlist notice a weird rectangular artifact

> icons on thumbnails are fixed instead of showing on hover

> When quick bookmark is enabled you will see that all the videos in the list receive the checkmark. This doesnt happen in list mode because it doesnt make sense to do that because you can just remove the vids by clicking the delete icon on the thumbnail


> sorting custom doesnt make me change the order of the videos by hovering on the thumbnail and clicking the arrow icon

Clarification: Do we want to show the arrow icons as left and right arrows? Do we show these icons fixed as well, then? I also need to add back the "remove" button

efb4f5ff-1298-471a-8973-3d47447115dc:
> Clarification: Do we want to show the arrow icons as left and right arrows? 

Left to right makes more sense so i say yes.

> Do we show these icons fixed as well, then? 

Yes these will also be fixed see latest nightly build playlists for reference

src/renderer/views/Playlist/Playlist.scss
```diff
@@ -81,23 +128,40 @@
...
 }
 
 .message {
   color: var(--tertiary-text-color);
 }
+
+@media only screen and (width <= 680px) {
+  .routerView{
+    &.grid {
+      margin-block-start: 230px;
```
MarmadileManteater, line 156: I realize this specific value (as well as the one above) is set because the position of the fixed-top-bar is either `fixed` or `absolute`, and it won't push the body of the page down below it. However, I feel it looks weird in the edge cases of either more than expected or less than expected content. 

image

image

MarmadileManteater:
I noticed that the playlist description truncates with an ellipsis, but the title does not.

image

src/renderer/views/Playlist/Playlist.scss
jasonhenriquez: This is unfortunately hard to fix generically. I can truncate the title with ellipses and add a styling case for no videos.

PikachuEXE (approved):
Probably off-topic
The setting `Video View Type` also controls user playlists view and maybe we should update the label text

jasonhenriquez:
@efb4f5ff-1298-471a-8973-3d47447115dc Do you have any more changes requested, or is the tag safe to remove?

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -72,7 +72,7 @@
         <ft-icon-button
           v-if="inUserPlaylist && canMoveVideoUp"
           :title="$t('User Playlists.Move Video Up')"
-          :icon="['fas', 'arrow-up']"
+          :icon="((listType === 'list' || forceListType === 'list') && forceListType !== 'grid') ? ['fas', 'arrow-up'] : ['fas', 'arrow-left']"
```
ChunkyProgrammer, line 75: Maybe make the condition a computed property since it's used twice? `((listType === 'list' || forceListType === 'list') && forceListType !== 'grid')`

jasonhenriquez: Five times, actually! (Good call-out)

PikachuEXE:
Grid view in narrow screen:
image

MarmadileManteater:
I personally don't agree the list view is generically better on mobile because of long video titles.

jasonhenriquez:
Fair point, sounds like we need more work to pick and choose which rules we want to apply from each to any new centralized mobile video view type, and/or if we're feeling ambitious, make a broader set of changes to the `ft-list-video` mobile view (e.g., something like the current YT mobile app video UX). Anyway, I'll create a separate issue for this broader proposal.

efb4f5ff-1298-471a-8973-3d47447115dc:
We still need numbering for the videos, suggestions are welcome :)

Moving and removing have no real animation anymore

https://github.com/FreeTubeApp/FreeTube/assets/73130443/f98afa5c-b0cd-4cc1-bc63-7c465a6210df

Before:

https://github.com/FreeTubeApp/FreeTube/assets/73130443/dc1200d6-f550-4c32-afea-092b517b2325

jasonhenriquez:
From [my earlier reply](https://github.com/FreeTubeApp/FreeTube/pull/4949#issuecomment-2065451174):

> I couldn't get video numbering to look good on grid view. If anyone has any suggestions for it, please share, but I'm going to leave that one by the wayside for the time being.

For the lack of animations, this is because we're not using `transition-group` for these. I think it's a bit much for horizontal movements, especially when it could be both vertical and horizontal when you're switching the position of elements that are currently on two separate grid rows.

efb4f5ff-1298-471a-8973-3d47447115dc:
Could you maybe provide some screencaps on your tries to add the numbering so i can understand why it would degrade the design

jasonhenriquez:
I didn't save anything because I couldn't get it to look good. I can give it another try if you think this is an important feature, but I'm not optimistic that I can do it without increasing the size of each video grid element and/or making the UI a bit busier, which both end up being more costly than beneficial.

efb4f5ff-1298-471a-8973-3d47447115dc:
Well at this moment i think its important but that is only bc i didnt see how bad it looked like

jasonhenriquez:
Something like this is the best I got. I'm not a fan of it. There aren't many places you can put the number, and it's hard to fit a number there in a way that looks good without making it bigger.

Screenshot_20240425_211740

jasonhenriquez:
One viable way without affecting space would be to have a section on the right/left of the thumbnail that shows the proper playlist index, similar to the `playlistCountContainer` on playlist results. But I'm still not too huge of a fan of it. I can take a go at it or pass up on it. @efb4f5ff-1298-471a-8973-3d47447115dc

efb4f5ff-1298-471a-8973-3d47447115dc:
Thanks for providing the image. I now have a better understanding of why this degrades the current UX. The only thing i can think about if we want to include numbering is inside the thumbnail on the top left side.

Lets wait on the others to chime in. I dont want to make a fuss about this if the rest thinks the PR in its current state is fine.

PikachuEXE:
Top left is preoccupied
Bottom left?
image

jasonhenriquez:
The weird thing about keeping it in the video grid is that it has to escape the `watched` styling, but that makes it stick out visually

jasonhenriquez:
@efb4f5ff-1298-471a-8973-3d47447115dc Any final suggestions or concerns before moving forward on reviewing this one? I can create a separate issue for adding numbering to `ft-element-list`s if we'd prefer to keep that tracked

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
No suggestions or found anything weird. LGTM

Edit: Yeah maybe we should keep it tracked in an issue

PikachuEXE:
Reducing width to 745px looks like this
Maybe it should force list view on that width?
image

jasonhenriquez:
I want to use 680px because it's our current main mobile/tablet breakpoint; exact point of one video wide depends on whether sidebar is open, potential future features like #5007, etc.

PikachuEXE (approved):

jasonhenriquez:
@MarmadileManteater @ChunkyProgrammer @absidue (if uncomfortable being the third review, you can leave a non-binding review).

src/renderer/scss-partials/_ft-list-item.scss
```diff
@@ -2,6 +2,36 @@
...
+
+  @media only screen and (width <= 680px) {
+    inline-size: 100%;
+
+    @include is-side-nav-open {
+      inline-size: 100%;
+    }
+  }
+}
+
```
absidue, line 34: As far as I can tell this is only used in the `Playlist.scss` and `ft-refresh-widget.scss` files, so it probably makes more sense to place it in a new partial file, as it seems unrelated to the rest of the stuff in this file.

absidue:
I'm not a fan of having such a large UI element pinned to the top of the screen while I scroll. I'm sure it's okay for people that use FreeTube with the window maximized or fullscreened, but as it covers an entire row of videos, on smaller window sizes it covers 2 out of the 3 rows that are visible in the grids on other pages.

Instead of pinning the everything, I think it would be better to only pin the essentials e.g. the title and the search bar, that way it still achieves the goal I think you are aiming for (keeping the essentials easily accessible at all times) but would also solve my gripe about it taking up so much of the screen. Happy to discuss what elements should be considered essential, if you don't agree with the two I chose.

jasonhenriquez:
I'm open to the idea, but could you explain why it's a bigger problem for grid view than list view? 30% width x 100% height versus 85% width x 218px height. Only problematic default device view I saw was the Nest Hub, where it has a decent width but small height. I could either add a breakpoint for small height but normal width, or have just the icons and search bar stay as you scroll.

absidue:
I'm not comparing grid view to list view, I'm comparing the grid views throughout the rest of the app, to the grid view introduced in this pull request, which has one row less than the rest if them. However if you do want me to compare them, with the list view people don't expect information density, but they do with the grid view.

jasonhenriquez:
Updated now to force list view if width is below a certain threshold as well. Let me know what you think.

absidue:
Looking at your most recent commit, I think you've completely misunderstoood the point I am trying to make. I don't want you to force the list view in more situations, I want you to fix the grid view.

With the list view you don't miss out on any information by having the side bar there, as it fills what would otherwise be unused space. Adding a pinned top bar to the grid view requires reducing the amount of content you would otherwise see, as the view scrolls vertically, so always blocking vertical space for things like the description, which most people would happily scroll back to the top to read, has a negative impact on the user experience, not a positive one like the side bar in the list view does.

jasonhenriquez:
Please let me know if you have any good ideas on how to implement that, I'm having trouble with it at the moment.

absidue:
There is probably no pure CSS solution for it but using an intersection observer (`v-observe-visibility` with an offset) to toggle a boolean variable in `data` to hide and show elements with `v-show` should definitely be possible.

jasonhenriquez:
Thanks for the help. Before I get started, does this look okay? It's 123px tall versus the current 199-218px. It occludes most of a thumbnail, whereas the current occludes a thumbnail, title, and author.

Screenshot_20240519_150724

absidue (commented):
Like it a lot better now, thank you. Just one small thing, we probably don't need the `PLAYLIST_HEIGHT_FORCE_LIST_THRESHOLD` stuff anymore now that the bar shrinks when you scroll.

jasonhenriquez:
I kept that because it's at a width where only one row of content is displayed. Any smaller, and the top bar occludes the main content too much (in the edge case of someone with a wide but short window).

absidue:
In that case we can leave it in, if it were to caus issues for actual mobile users, it's not a hard thing to remove later on. I just thought I would mention it as it was added in response to one of my comments but didn't actually solve the issue (that was solved by collapsing the pinned content on scroll) and didn't seem particularly useful.

absidue (approved):

jasonhenriquez:
That's fair. I kept it in a lower breakpoint than the original because I noticed the main content was inaccessible at a small enough height, but I never actually stated that explicitly, so I think that's fair to call out for sure.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

PikachuEXE:
Not blocking 
1. I found it weird that when I scroll down I see playlist action buttons (which I think not quite useful **when scrolled down** unlike search bar) but they can stay there is 2nd point below is resolved
2. I expect to see the playlist title (not description or other stuff) 

If it's too difficult to implement then it's fine to have this merged first

image

jasonhenriquez:
I think it's doable, but the reason I prefer showing the action buttons is because you know what playlist you're looking at when you first land on it and most likely don't need a reminder, but you could gain from being able to do actions from any scroll level without having to scroll back up.

PikachuEXE:
With multi-window I found it not quite the case for `you know what playlist you're looking at` (only works on first view)
Unless the window title shows part of the playlist name (which it doesn't right now) I would be a bit lost
As for action buttons I am fine for them to remain visible

jasonhenriquez:
Updating the window title to match the current playlist sounds like a good idea. I can do that in the same PR or a separate one, whichever you prefer.

efb4f5ff-1298-471a-8973-3d47447115dc:
Lets not push this back any further. Jason already created a separate PR so lets merge this :)

PikachuEXE (approved):
Potential merge conflict incoming~

efb4f5ff-1298-471a-8973-3d47447115dc:
smh why i didnt test for this......

https://github.com/FreeTubeApp/FreeTube/assets/73130443/8ac836f2-5a85-40ba-8e78-06e52fe925dc

PR #4231, opened 2023-10-25
List sorting & display settings

kommunarr:
# List sorting & display settings

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
#4096, [comment](https://github.com/FreeTubeApp/FreeTube/issues/1739#issuecomment-925846285)

## Description
- Implements radio group support for `ft-icon-button` dropdown
- Implements `ft-button` option for `ft-icon-button` base button appearance
- Implements three buttons in Theme Settings for fitting three of our features with new capabilities:
  1) Settings Sections,
     - Use default sort (default), sort A to Z, sort Z to A
   2) Side Nav subscription bubbles, and
   3) the `ft-profile-selector` dropdown:
     - Sort A to Z, sort Z to A
     - Display type: list or grid view
       - (grid-specific): Number of items per grid row
       - (grid-specific): Show or hide titles
   

## Video 

This video demos everything listed in the Testing section:

https://github.com/FreeTubeApp/FreeTube/assets/84899178/84b554c7-e7ed-4b73-b8e2-3b1b5f373701

Edit: After minor updates to settings placement and changing the settings section sort:
Screenshot_20231026_144250

## Testing 
- Test radio group functionality of the `ft-icon-button` dropdown:
  - Working up/left/down/right arrow navigation
  -  `Tab` / `Shift` + `Tab` brings you out of the radio group to the next / previous tabbable element
  - `Spacebar` checks the focused checkbox
- Check that the expanded side nav, settings section, and profile selector dropdown behave normally with `settings.js` defaults
- Check that side nav subscriptions look good with every combination of `grid`, `itemsPerGridRow`, and `showTitles` and respects the selected sorting order
- Check that the profile selector looks good with every combination of `grid`, `itemsPerGridRow`, and `showTitles` and respects the selected sorting order
- Check that the settings sections respect the selected sorting order

## Desktop

- **OS:** OpenSUSE Tumbleweed
- **OS Version:** 2023xxxx
- **FreeTube version:** 0.19.1

## Additional context
Remaining related discussions: #1739, #987, #800, parts of #4096, #2871

The new radio group support in the `ft-icon-button` dropdown was just created as a helpful tool for bringing about what I was envisioning, but I can see it coming in handy again for creating related settings in a space-compressed and low-cognitive-complexity way. See discussion under #2295.

Among other reasons, I also created this when pondering the design of this issue that I intend to work on: #1615.

[...]

PR #4950, opened 2024-04-14
Add icon button visual labels

### LINKED ISSUE #4259: [Feature Request]: Make label visible for most icon buttons
### Guidelines

### Problem Description

Most `ft-icon-button`s in FreeTube have no visible label and rely on the `title` attribute (& the scant meaning in the icons themselves) to convey meaning. Here are the issues with this:
1. They're not for everybody: buttons without visible labels preclude most mobile users and impair many disabled users from being able to actually see what the labels are <sup>1 . To quote Jacob Michelini from his article "Unlabeled Icons: Sacrificing Usability for Aesthetics":
> For low-vision users, many icons contain tiny details that make them indistinguishable. For screen-reader users, they are totally unreadable unless tagged properly. For users with fine motor impairment, the tiny unlabeled icon you chose to just barely fit the UI is a nightmare to tap accurately. It’s also a nightmare for all users when they try to operate your app with one thumb while carrying groceries on the bus.
2. It's a bad user experience<sup> 1 </sup><sup>2 </sup><sup>3 </sup>: it's not immediately intuitive what all of these buttons do just by looking at their icon, and hovering over them each time you want to check is not fun or fast. We have gotten dozens of GitHub issues raised about "missing" functionality on sharing videos, downloading videos, et cetera — and I don't blame the people who make them.

To quote the Nielsen Norman Group: 
> And always include a visible text label. As Bruce Tognazzini once said, “a word is worth a thousand pictures.”

Sources:
1. https://uxplanet.org/unlabeled-icons-sacrificing-usability-to-look-pretty-415408a0e115
2. https://axesslab.com/icons-ruining-interfaces/
3. https://www.nngroup.com/articles/icon-usability/

### Proposed Solution

Add labels to almost all buttons in FreeTube that are missing them. Make this the default behavior for `ft-icon-button`s. Exceptions should be rare and only in cases of universal intelligibility. Examples of buttons that could meet this definition: `Save video`, `More Options`. Here's a heuristic by Nielsen Norman on this:
> Use the 5-second rule: if it takes you more than 5 seconds to think of an appropriate icon for something, it is unlikely that an icon can effectively communicate that meaning.

### Alternatives Considered

None other suffice.

### Issue Labels

ease of use improvement

### Additional Information

_No response_

jasonhenriquez:
# Add icon button visual labels

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4259 

## Description
I've had this sitting on the backburner since November. This PR makes many of our `ft-icon-button`s, particularly in `watch-video-info` and `playlist-info`, have visual labels. This makes it far easier to understand many of our non-obvious icon actions. This problem is less apparent to us contributors of the project who are by now very familiar with these buttons' usage, but I expect this will be very useful to most everyone else.

## Screenshots 
Screenshot_20240414_171252
Screenshot_20240414_171231

## Testing 
- Confirm that icon button labels have not been added in places where they shouldn't be
- Confirm that there are no visual oddities with these labels in other languages (e.g., Russian, German, Arabic)
- Test the User Playlist list view and the video view on desktop, tablet, and mobile device sizes

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
- We're currently trying to display all of the words we can fit in. We can change this to an implementation where we have a "short version" of the label for this case, that defaults to the first word of the label if none is set (through Western-centric checking for whitespace). This would look cleaner & more similar to the YT implementation, but on the other hand, I'm not sure it yields enough of a benefit compared to the current implementation.
- I have this PR as a draft because I imagine there will be some discussion about this. I don't intend to start any controversy, so I will keep this drafted until it's revised to a point where we're sufficiently on the same page as a team to move forward with it.


absidue:
My biggest concern is narrow screens and mobile devices, they are already struggling to display our current desktop centric UI, so don't have the space to have buttons that are at least 3 times wider (in some languages they could be even larger) than they currently are, so maybe on narrow screens we should hide the labels, to make the buttons look the same as they do before this pull request?

jasonhenriquez:
> Maybe on narrow screens we should hide the labels, to make the buttons look the same as they do before this pull request?

I hesitate to go for this solution because mobile users also greatly benefit from labels for icon buttons, and apps like YT Mobile still have icon labels even given the space considerations because of it. If you're willing to try the YT app on a mobile phone in English or German, you can see what I mean. When it wraps over on the video info page, it becomes a horizontal scroll region. We do the same in this PR, minus the same "prettiness" of the invisible scrollbar. Ours is even notably less wide by a decent margin compared to YT's on German, although that is mostly because we're currently showing the Add to Playlist, Quick Bookmark, Open in External Video Player, and Share buttons as icons still, which I'm still on the fence about (at least for the Share button).

Screenshot_20240415_072645
Screenshot_20240415_073157

jasonhenriquez:
Some other notable aspects are that YT is able to fit in the full word "Herunterladen", where as we only fit 9 / 13 characters in. I could try messing around with the icon size and button radius to fit more content in. 

More differences: YT tries to have single-word labels. The icons all have a minimum size, but if they need more space, they're given it instead of ever cutting the label short like we do (that's in fact why they can fit "herunterladen"). Contrast with us, who have multi-word labels on icon buttons all of equal size (for visual uniformity), and we also break up a single word across two lines mid-character if it can't fit on the first line. I think there are trade-offs of each of our implementations, but ultimately, I do think ours is better suited for the scenario of a trying to maximize utilization of the existing labels for users of all languages. It does look somewhat uglier due to those reasons though, as well as the presence of ellipses. I am still on the fence about a middle ground solution to this; see discussion in the OP:
> We're currently trying to display all of the words we can fit in to a comfortable size. We can change this to an implementation where we have a "short version" of the label for this case, that defaults to the first word of the label if none is set (through Western-centric checking for whitespace). This would look cleaner & more similar to the YT implementation, but on the other hand, I'm not sure it yields enough of a benefit compared to the current implementation.

efb4f5ff-1298-471a-8973-3d47447115dc:
I like that the buttons will have visual labels and i think we shouldnt hide it on narrow screens. I dont have an opinion on showing the whole word or cutting it off but im leaning towards showing the whole word the best we can if if its starting to get ugly, cut it off

PikachuEXE:
If there is an option to disable it then it's fine for me
The screen looks too busy with all the labels present and longer label text = good luck

efb4f5ff-1298-471a-8973-3d47447115dc:
> If there is an option to disable it then it's fine for me

Maybe merge option Hide Labels in Sidebar with this one to prevent another toggle

jasonhenriquez:
I like Piped's icon buttons. As in, just regular buttons with an icon attached, something I added in #4231. Thoughts on that?

Screenshot_20240423_154840

efb4f5ff-1298-471a-8973-3d47447115dc:
That also looks good. I have no real preference so choose the easiest option :)

kommunarr:

PR #4960, opened 2024-04-15
Show error message when search is over 100 characters

### LINKED ISSUE #940: Search Box - Long query

dkshxd:
# Show error message when search is over 100 characters

## Pull Request Type
- [X] Bugfix

## Related issue
Closes  #940

## Description
Show error message when search is over 100 characters. Also logs the error to console. The event handler on the search box is suppressed as well when over 100 characters as this can also trigger the exception screen.

## Screenshots 

100-char-limit

## Testing 
* Has been tested
* Tested cases
1. Normal search - less 100 characters
2. Search over 100 characters
3. Trigger the event handler on the search box by typing inside search box when under and over 100 characters
4. Empty search

## Desktop

- OS: Windows 
- OS Version: 10
- FreeTube version: 0.20 Beta

### DIFF
--- a/static/locales/it.yaml
+++ b/static/locales/it.yaml
@@ -276,6 +276,9 @@ Settings:
       Ask Before Opening Link: Chiedi prima di aprire il link
       Open Link: Apri link
       External Link Handling: Gestione dei link esterni
+    Auto Load Next Page:
+      Label: Carica automaticamente la pagina successiva
+      Tooltip: Carica automaticamente pagine e commenti aggiuntivi.
   Theme Settings:
     Theme Settings: 'Impostazioni del tema'
     Match Top Bar with Main Color: 'Abbina la barra superiore al colore principale'
@@ -932,6 +935,15 @@ Playlist:
 #* Published
 #& Views
   Playlist: Playlist
+  Sort By:
+    Sort By: Ordina per
+    Custom: Personalizzato
+    VideoTitleDescending: Titolo (Z-A)
+    AuthorDescending: Autore (Z-A)
+    AuthorAscending: Autore (A-Z)
+    DateAddedNewest: Prima gli aggiunti più nuovi
+    DateAddedOldest: Prima gli aggiunti più vecchi
+    VideoTitleAscending: Titolo (A-Z)
 Toggle Theatre Mode: 'Attiva/disattiva la modalità teatro'
 Change Format:
   Change Media Formats: 'Cambia formato video'


PikachuEXE:
For other viewers: 100 chars limit is from a reported test result https://github.com/FreeTubeApp/FreeTube/issues/940#issuecomment-1794697557

src/renderer/views/Search/Search.js
```diff
@@ -90,26 +91,33 @@ export default defineComponent({
...
+              this.performSearchLocal(payload)
+              break
+            case 'invidious':
+              this.performSearchInvidious(payload, { resetSearchPage: true })
+              break
+          }
         }
+      } else {
+        console.warn('Search character limit is: ', searchCharLimit)
+        showToast(i18n.t('Search character limit', { searchCharacterLimit: searchCharLimit }))
```
PikachuEXE, line 120: This few lines should be before the main code and perform an early return
No point making the remaining existing code nested when it can be avoided
```js
if (payload.query.length > searchCharLimit) {
  console.warn('Search character limit is: ', searchCharLimit)
  showToast(i18n.t('Search character limit', { searchCharacterLimit: searchCharLimit }))
  return
}
```

src/renderer/helpers/api/local.js
```diff
@@ -68,7 +69,14 @@ export async function getLocalSearchSuggestions(query) {
...
   }
 
-  return await searchSuggestionsSession.getSearchSuggestions(query)
+  const searchCharLimit = store.getters.getSearchCharacterLimit
+
+  if (query.length <= searchCharLimit) {
+    return await searchSuggestionsSession.getSearchSuggestions(query)
+  } else {
+    // There's an event handler on the search input so avoid displaying an exception
+    console.error('Query is over ' + searchCharLimit + ' characters')
```
PikachuEXE, line 78: Similar to another file, these lines should be added to the top of function with early return to avoid calling `createInnertube` unnecessarily

PikachuEXE (changes requested):

src/renderer/views/Search/Search.js
dkshxd: I've made the changes requested

src/renderer/helpers/api/local.js
dkshxd: I've made the changes requested

PikachuEXE: @absidue I guess we should throw an error here instead of returning? I am not quite sure

dkshxd: The problem is this code is called when typing inside the search bar by the event handler. Due it's usage you might see it quite a few times. The assumption I'm making is they are going to hit the search button eventually which will then show the proper error message popup (Inside Search.js).

src/renderer/views/Search/Search.js
```diff
@@ -90,6 +91,14 @@ export default defineComponent({
   },
   methods: {
     checkSearchCache: function (payload) {
+      const searchCharLimit = this.$store.getters.getSearchCharacterLimit
+
+      if (payload.query.length > searchCharLimit) {
+        console.warn('Search character limit is: ', searchCharLimit)
+        showToast(i18n.t('Search character limit', { searchCharacterLimit: searchCharLimit }))
```
ChunkyProgrammer, line 98: You dont need to import i18n in this file
```suggestion
        showToast(this.$t('Search character limit', { searchCharacterLimit: searchCharLimit }))
```

PR #4967, opened 2024-04-17
Replace actions with nodejs 16

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

Saw deprecation messages about some actions using nodeJS 16
image
This PR replaces those actions with one that uses nodeJS 20

See deprecation messages at one of the builds: https://github.com/FreeTubeApp/FreeTube/actions/runs/8713342604

## Screenshots 

From https://github.com/PikachuEXE/FreeTube/actions/runs/8714715984/job/23905412969
image
image

## Testing 

I have NOT tested `Set Version Number Variable` with specific branch name yet
If this is needed let me know

Also release workflow is NOT tested

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,16 +77,14 @@ jobs:
           } else if (${{ env.IS_RC }}) {
             return "${{ env.VERSION_NUMBER_RC }}"
           } else {
-           return "${{env.VERSION_NUMBER }}"
+            return "${{env.VERSION_NUMBER }}"
           }
         # script: if ${{ env.IS_DEV }} then echo "::set-output name=VERSION_NUMBER::${{ env.VERSION_NUMBER_NIGHTLY }}" else echo "::set-output name=VERSION_NUMBER::${{ env.VERSION_NUMBER }}" fi
 
     - name: Update package.json version
-      uses: jossef/action-set-json-field@v2.1
+      uses: jaywcjlove/github-action-package@main
       with:
-        file: package.json
-        field: version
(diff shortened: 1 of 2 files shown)


absidue (approved):
I also confirmed that the version number inside the packages is correct on the linked workflow run.

.github/workflows/build.yml
```diff
@@ -57,20 +57,18 @@ jobs:
     - run: yarn run ci
     - run: yarn run lint
     - name: Get Version Number
-      uses: jozsefsallai/node-package-version@v1.0.4
-      with:
-        path: 'package.json'
-        follow-symlinks: false
+      id: getPackageInfo
+      uses: jaywcjlove/github-action-package@main
```
ChunkyProgrammer, line 61: Is it possible to link to a specific version of this action? I imagine this might cause issues if they ever decide to do a breaking change

PikachuEXE: If you look at it's last major version bump it's only updating nodeJS
So while it's possible breaking changes would be introduced, I bet on he is only updating nodeJS version next major version bump
https://github.com/jaywcjlove/github-action-package/releases/tag/v2.0.0
 (I see most actions don't get updated than updated too much, i.e. include many breaking changes

image

ChunkyProgrammer (approved):

jasonhenriquez (approved):

PR #4970, opened 2024-04-17
Intuitive input bindings

jasonhenriquez:
# Intuitive input bindings

## Pull Request Type

- [x] Feature Implementation

## Related issue
None

## Description
- Because we don't use forms in most of FreeTube, `ft-input`s that should have an `onsubmit` action fire on <kbd>Enter</kbd> are not actually doing anything on that command, thus leading to some frustrating behavior. This configures <kbd>Enter</kbd> on any `ft-input`s with a corresponding submission action to trigger that action.
- We have many prominent search bars, and <kbd>Ctrl</kbd>+<kbd>F</kbd> is a universal standard for initiating a "find". This adds an event listener to focus these search bars on their applicable routes when <kbd>Ctrl</kbd>+<kbd>F</kbd> is pressed.

## Testing 
- Test <kbd>Ctrl</kbd>+<kbd>F</kbd> focuses search bars on all routes with a prominent one (History, UserPlaylists, SubscribedChannels, User Playlist, Add to User Playlist Prompt).
- Test 'Enter' properly submits all forms (Profile Edit, User Playlist Edit, Test Proxy, Set Password).

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
Q: Should this be added to our list of keyboard shortcuts?
A: IMO, these are pretty universal/expected behaviors that users will discover by just trying more often than not. I can maybe budge on the <kbd>Ctrl</kbd>+<kbd>F</kbd> one since that's a bit more specific, and existing users may have tried it before & mentally discounted it.

### DIFF
--- a/src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
+++ b/src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
@@ -10,6 +10,7 @@ import FtSelect from '../../components/ft-select/ft-select.vue'
 import FtToggleSwitch from '../../components/ft-toggle-switch/ft-toggle-switch.vue'
 import {
   showToast,
+  ctrlFHandler
 } from '../../helpers/utils'
 
 const SORT_BY_VALUES = {
@@ -44,6 +45,8 @@ export default defineComponent({
       lastShownAt: Date.now(),
       lastActiveElement: null,
       sortBy: SORT_BY_VALUES.LatestUpdatedFirst,
+      keyboardShortcutHandler: (e) => ctrlFHandler(e, this.$refs.searchBar),
+
     }
   },
   computed: {
@@ -198,12 +201,13 @@ export default defineComponent({
   },
   mounted: function () {
     this.lastActiveElement = document.activeElement
-
     this.updateQueryDebounce = debounce(this.updateQuery, 500)
     // User might want to search first if they have many playlists
     this.$refs.searchBar.focus()
+    document.addEventListener('keydown', this.keyboardShortcutHandler)
   },
   beforeDestroy() {
+    document.removeEventListener('keydown', this.keyboardShortcutHandler)
     this.lastActiveElement?.focus()
   },
   methods: {
--- a/src/renderer/views/UserPlaylists/UserPlaylists.js
+++ b/src/renderer/views/UserPlaylists/UserPlaylists.js
@@ -12,6 +12,7 @@ import FtInput from '../../components/ft-input/ft-input.vue'
 import FtIconButton from '../../components/ft-icon-button/ft-icon-button.vue'
 import FtToggleSwitch from '../../components/ft-toggle-switch/ft-toggle-switch.vue'
 import FtAutoLoadNextPageWrapper from '../../components/ft-auto-load-next-page-wrapper/ft-auto-load-next-page-wrapper.vue'
+import { ctrlFHandler } from '../../helpers/utils'
 
 const SORT_BY_VALUES = {
   NameAscending: 'name_ascending',
@@ -52,6 +53,7 @@ export default defineComponent({
       doSearchPlaylistsWithMatchingVideos: false,
       activeData: [],
       sortBy: SORT_BY_VALUES.LatestPlayedFirst,
+      keyboardShortcutHandler: (e) => ctrlFHandler(e, this.$refs.searchBar),
     }
   },
   computed: {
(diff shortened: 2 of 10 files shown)


src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -44,6 +45,8 @@ export default defineComponent({
       lastShownAt: Date.now(),
       lastActiveElement: null,
       sortBy: SORT_BY_VALUES.LatestUpdatedFirst,
+      keyboardShortcutHandler: (e) => ctrlFHandler(e, this.$refs.searchBar),
```
absidue, line 48: Could you please move all of these into the methods sections, seems more fitting there :)

jasonhenriquez: Sorry, they were not working as lambdas in the `methods` section (was causing an issue with `$refs` being `undefined`, fixed now

absidue: Yep, you can't use lambdas in `methods` because they inherit their parent's `this` as they don't have their own, so Vue can't overwrite it to point to the component.

> Arrow functions differ in their handling of this: they inherit this from the parent scope at the time they are defined. This behavior makes arrow functions particularly useful for callbacks and preserving context. However, arrow functions do not have their own this binding. Therefore, their this value cannot be set by bind(), apply() or call() methods, nor does it point to the current object in object methods.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

src/renderer/views/UserPlaylists/UserPlaylists.js
```diff
@@ -242,6 +247,10 @@ export default defineComponent({
       })
     },
 
+    keyboardShortcutHandler: function (event) {
+      return ctrlFHandler(event, this.$refs.searchBar)
```
absidue, line 251: I don't think the return here is needed (ditto for all the other ctrlFHandler usages).

jasonhenriquez: Just tested and can confirm. I'll remove those if we think that is confusing.

absidue (commented):
While testing I noticed that the CTRL+F keyboard shortcut doesn't work on channel pages. Could you please add the keyboard shortcut there too. Everything else looks good :).

absidue:
> Q: Should this be added to our list of keyboard shortcuts?
A: IMO, these are pretty universal/expected behaviors that users will discover by just trying more often than not. I can maybe budge on the Ctrl+F one since that's a bit more specific, and existing users may have tried it before & mentally discounted it.

I definitely think the <kbd>CTRL</kbd>+<kbd>F</kbd> keyboard shortcut should be added to the docs.

absidue (dismissed):

PikachuEXE (dismissed):
I feel weird having enter linked to test proxy
But it's a small issue so whatever

efb4f5ff-1298-471a-8973-3d47447115dc:
After running `yarn run dev` and waiting for the window to start, it will eventually start and close itself immediately and this is only happening on this PR

Skip to 35 sec

https://github.com/FreeTubeApp/FreeTube/assets/73130443/1ee64317-95a3-4190-9837-389342e6cf51

jasonhenriquez:
Cannot replicate. Maybe try switching to `development`, deleting your local of my branch, and pulling it from my remote.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/components/ft-select/ft-select.vue
```diff
@@ -28,7 +28,7 @@
     <label
       class="select-label"
       :for="sanitizedId ?? sanitizedPlaceholder"
-      :hidden="disabled"
+      v-if="!disabled"
```
PikachuEXE, line 31: Say hi to warning

jasonhenriquez: ? Cannot replicate. Can you take a screenshot?

PikachuEXE: https://github.com/FreeTubeApp/FreeTube/pull/4970/files#diff-04846a18b5ef86834f2cbee911c63a2f616a0e4f68d8f269ffaaa3e5051ad8b3R31

2024-04-26 10_47_56-Window

jasonhenriquez: :facepalm: I thought you mean in the dev console. My bad lol

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #4980, opened 2024-04-18
Set refresh widget z-index to below dropdown search bar

MarmadileManteater:
# Set refresh widget z-index to below dropdown search bar

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
The topNav is currently lower than the z-index of the ft-refresh-widget. This causes the search to be below the refresh widget when it drops down. This PR addresses this by setting the widget's z-index to `3`. It also hides the box shadow on the topNav when there is a refresh widget present in the routerView. I added that because it was previously hidden by the ft-refresh-widget.

## Screenshots 
|before|after|
|--|--|
|before|after|

## Testing 
1. Ensure the window is narrow `<=680px`
3. Check that the search is visible when expanded
3. Check that the refresh widget is still above watch video progress

## Desktop

- **OS:** Windows 10
- **OS Version:** 22H2 (OS Build 19045.4291)
- **FreeTube version:** b69ecd12a56b231bc8277df33e0909a3886a65e4

### DIFF
--- a/src/renderer/components/ft-refresh-widget/ft-refresh-widget.css
+++ b/src/renderer/components/ft-refresh-widget/ft-refresh-widget.css
@@ -13,7 +13,7 @@
   align-items: center;
   gap: 5px;
   justify-content: flex-end;
-  z-index: 4;
+  z-index: 3;
 }
 
 .floatingRefreshSection:has(.lastRefreshTimestamp + .refreshButton) {
--- a/src/renderer/components/top-nav/top-nav.scss
+++ b/src/renderer/components/top-nav/top-nav.scss
@@ -21,6 +21,10 @@
   inline-size: 100%;
   z-index: 4;
 
+  &:has(+ .sideNav + .routerView .floatingRefreshSection) {
+    box-shadow: none;
+  }
+
   @media only screen and (width >= 961px) {
     display: grid;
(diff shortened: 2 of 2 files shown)


PikachuEXE (approved):

absidue (approved):

jasonhenriquez (approved):

jasonhenriquez:
Just making sure, do we want the `box-shadow` between the `refresh-widget` and the top nav to stay gone on desktop and/or tablet view?

Screenshot_20240422_183752

MarmadileManteater:
I personally think it looks better without the extra box shadow on all window widths, but I would interested to hear what other people have to say about it.

PikachuEXE:
I am fine coz my theme settings make it look like this
image

jasonhenriquez:
I think I'm fine with it, the only thing irking me is the top left section between the top nav and side nav is inconsistent now depending on the route. Would prefer any solution that standardizes it.

jasonhenriquez:
^ boosting this just so we don't forget. It's a bit more annoying to add/remove the `box-shadow` for only the sliver between the top nav and side nav, which suggests adding it back is the more expedient choice.

PR #4987, opened 2024-04-20
Serve files from a custom protocol instead of using file://

absidue:
# Serve files from a custom protocol instead of using file://

## Pull Request Type

- [x] Feature Implementation

## Description
This pull request implements Electron security recommendation 18: ["Avoid usage of the `file://` protocol and prefer usage of custom protocols"](https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols). Long story short is that the `file://` protocol in Electron allows you to access any file on the disk through the URL, by using a custom protocol we can limit that to just the app files. I chose to use an allow list approach, so that it is limited to the files that we bundle with FreeTube that the renderer process needs to work (the `main.js` file gets filtered out for example, as that is only needed by the main process). It also removes one way of the renderer process knowing where on the machine FreeTube is installed.

Please note this only applies to production builds, not in development mode (it would break hot module replacement/live updates if we did something similar in dev mode).

This pull request alone won't block reading any file on the disk, we do have `nodeIntegration` turned on at the moment, so you can just load the `fs` module and do whatever you want with the file system, but it is a step in the right direction. Also I'm sure we can all agree that reviewing smaller pull requests is much easier than a gigantic one that tries to change everything in one go.

## Testing 
As this change only applies to production builds, here is one for you to test: https://github.com/absidue/FreeTube/actions/runs/8764629053

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/injectAllowedPaths.mjs
+++ b/_scripts/injectAllowedPaths.mjs
@@ -1,0 +1,50 @@
+/**
+ * Injects the paths that the renderer process is allowed to read into the main.js file,
+ * by replacing __FREETUBE_ALLOWED_PATHS__ with an array of strings with the paths.
+ *
+ * This allows the main process to validate the paths which the renderer process accesses,
+ * to ensure that it cannot access other files on the disk, without the users permission (e.g. file picker).
+ */
+import { closeSync, ftruncateSync, openSync, readFileSync, readdirSync, writeSync } from 'fs'
+import { dirname, join, relative, resolve } from 'path'
+import { fileURLToPath } from 'url'
+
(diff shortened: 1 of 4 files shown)


PikachuEXE:
What do I test with that custom build though?

absidue:
Sorry about that, I've just updated the testing section.

PikachuEXE (approved):
Also tested optional

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez (approved):

PR #4988, opened 2024-04-20
Use webpack dev server's open option

absidue:
# Use webpack dev server's open option

## Pull Request Type

- [x] Other - Cleanup

## Description
Was looking through the webpack dev server docs for something else and noticed that it has an option to launch the browser, which means we don't need to do it ourselves, which simplifies the code a bit.

## Testing 
Run `yarn dev:web`, your default browser should be launched the same as it was before this change.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/dev-runner.js
+++ b/_scripts/dev-runner.js
@@ -130,7 +130,7 @@ function startRenderer(callback) {
   })
 }
 
-function startWeb (callback) {
+function startWeb () {
   const compiler = webpack(webConfig)
   const { name } = compiler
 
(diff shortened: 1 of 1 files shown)


jasonhenriquez:
This doesn't work for me on Linux, but the previous logic didn't either. `xdg-open` works, so I don't think it's a problem with my configuration. Will let others go ahead on this one.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM, tested in ubuntu vm

Issue #312, opened 2019-07-27
Subscribe to playlists


davidoskky:
I've seen you intend to show the playlists in #195; I've also seen the #82 proposing to filter the playlists out of the subscriptions.
I would instead prefer to subscribe for playlists, since generally I only want to see a part of what is being uploaded.
I don't know if a separate subscribe button would be appropriate or if it would be best to have some overall filtering options that would allow doing something like that.

PrestonN:
I'd probably consider this as a low priority for now, but it seems like a neat idea I could probably do down the line.  I don't know when I'll work on this but I'll update here whenever I do.

jkms:
+1

There are a few really big channels where I'm only interested in one particular, regularly updated playlist

esotericDisciple:
+1+
...What could be more important, besides tabbed browsing (which also doubles as playlist saving/bookmarking)

esotericDisciple:
Also would like a progress bar/page on this github for when the playlists page and the star button will actually be ready, even if it is a placebo like ol]Windows' copy-progress bar...

GilgusMaximus:
If you want to see what comes up next, please take a look into the projects page of Freetube here on Github.
The currently planned version v0.10.0 can be found [here](https://github.com/FreeTubeApp/FreeTube/projects/4) and will not include this feature. These features all come together in one big release because all need the same underlying system that still has to be developed. 

Because we are currently in a development break, due to the pretty stressful pre-rewrite release weeks and also personal responsibilities, we are on a bit of a hold right now. 

We won't provide a progress bar for anything that is not planned for the next feature release and even there is usually not a progress bar available, if it is not something that requires help and is visible in the PR section.

esotericDisciple:
Boo!

eguiraud:
Hi,
is "subscribe to playlists" the same as "save/favorite/star a playlist"? Asking because #1170 was closed as a duplicate of this issue, but they don't seem exactly the same thing to me. My usecase: I am studying from a long playlist of ~100 videos (finished, no more videos will be added, so no need to subscribe to playlist updates) and it would be nice to be able to bookmark/star it rather than having to search for it every time.

(amazing work on the app btw, I don't think I could have switched away from youtube otherwise -- thanks!)

GilgusMaximus:
This will in the end depend on how playlists and subbing will be implemented. Looking at it though, your suggestions falls in between subscription of playlists and #547. Don't worry, we will definitely implement this as well.

[...]

Issue #4594, opened 2024-01-23
[Feature Request]: Inline user playlists in sidebar on desktop view


jasonhenriquez:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The flow of accessing favorites is more jarring for desktop view with the inclusion of user playlists, now having to click the Playlist tab, wait for the `/userPlaylists` route to load, then click my desired playlist to get where I want to be. It would be desirable if I didn't have to do two actions to access my playlist of choice (as was the case before).

### Proposed Solution

Inline the user playlists in the side nav for the desktop view, much like the YT implementation. This makes this for a more streamlined UX flow that matches how I used it before.
![Screenshot_20240123_133911](https://github.com/FreeTubeApp/FreeTube/assets/84899178/2011510d-6024-4941-86cf-a45fb05546dc)

One major question is what we do with the Playlists tab on the desktop view. I say hide it because it then no longer serves a real purpose other than easy access to the "Create New Playlist" button. We can probably add an "Add New Playlist" button as a sidenav option to compensate.

Another major question is if we add a "Show more" / "Show less" button like YT. I say yes to this as well, because users might have a significant number of playlists.

A third major question is if we show a playlist that exists outside of the "Show more". Yes: the quick bookmark target.

Nothing should be changed at all on the mobile view.

### Alternatives Considered

- Keep Playlists tab on desktop view below all playlists in lieu of an inline "Create new playlist" button.
- Place items in top nav instead of side nav (bad idea - too short/valuable of real estate, and isn't where users recognize it to be).
- No "Show more" / "Show less" button in MVP (it's not going to be the worst thing ever, but some users will be pretty severely impacted. I see no reason to rush this out the gate).

### Issue Labels

ease of use improvement, improvement to existing feature

absidue:
I would like to add two things onto this:
[...]

Issue #1492, opened 2021-07-02
Keyword search history


KyleFN:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Oftentimes I search then start watching a video, following related videos and then later forget what I was even searching for in the first place. Or I later in the day want to look back on what I was searching for earlier.

**Describe the solution you'd like to see implemented**
Left-clicking on the search box should show a dropdown of the last 10 previously searched keywords.

**Describe alternatives you've considered**
Watch history gives an idea, but isn't as accurate as the actual search terms I was looking for. 

**Screenshots**
N/A

**Additional context**
N/A

efb4f5ff-1298-471a-8973-3d47447115dc:
This is a good idea. What also works (if u searched a few times within the same session) is smashing CTRL + Z a few times.

GilgusMaximus:
This will need some discussions on the actual implementation

efb4f5ff-1298-471a-8973-3d47447115dc:
Woule be nice if multiple profiles have their own independent search history

tassoman:
Would be a nice feature, also related to ability of mark "favourite" some of the keywords, so that can't be _forgotten_ then a query trigger a on click

Issue #1426, opened 2021-06-18
Suggestion for importing Chrome/Firefox bookmarks into FreeTube


RagglenLove:
**Is your feature request related to a usage problem (not a bug)? Please describe.**

**Describe the solution you'd like to see implemented**
I have many YouTube links to videos and channels on my Firefox bookmarks. So, I’d like a way to import my "bookmarks.html" into FreeTube, where FreeTube will only grab links that are YouTube, like youtube.com or youtu.be.

So for example, links that are channels, like https://youtube.com/user/achannelthatsawesome or https://www.youtube.com/channel/UC1QdcIOh4l6l3a4udgU9bvw should go to the sidebar, while links that are videos, like https://youtube.com/watch?v=otCDg8kmVs4 or https://youtu.be/ZCuqECjmaFc should go to "Your Playlists".

**Describe alternatives you've considered**

**Screenshots**

**Additional context**

GilgusMaximus:
This can be done once user playlists are available.

RagglenLove:
> This can be done once user playlists are available.

I thought user playlists were already available of as version 0.13.1?

![Screenshot_2021-06-20_18-44-02](https://user-images.githubusercontent.com/45503845/122671503-77420300-d1b6-11eb-8711-ef299fb98a9b.png)

GilgusMaximus:
Favorites are available. But the goal is to have as many playlists as one like. The favorites is just a temporary solution to fulfill the need to somehow store videos one likes

exekutive:
I am really surprised that there is no way to import bookmarks into FT yet.  At least give us a way to import youtube video URL lists in plain text. This would be a game changing feature.

efb4f5ff-1298-471a-8973-3d47447115dc:
Closing as we don't want to have to maintain support for allot of browsers for a niche feature like this

PR #5003, opened 2024-04-23
Page bookmarks

### LINKED ISSUE #1505: Ability to save searches
Sometimes I like to search for videos and playlists on broader topics (e.g.: Countries, Music Genres, Games etc.) and I would like to have a list of these searches.

Probably:
- a menu item called "Searches" or "Search Favorites" (or something that sounds better): This would include the list of searches
- a button on the "Search Results" page to add the search to the list

Unlike a recent history of searches, this would only include handpicked favorites.

-

**Screenshots**

-

**Additional context**

-

### LINKED ISSUE #4040: [Feature Request]: Subscribe to search results
### Guidelines

### Problem Description

At times it is interesting to follow some keywords or searches. For example I am interested in some scientists that give talks and would be cool to get to subscribe to a search result based on their names.

### Proposed Solution

Subscribe to search results.

### Alternatives Considered

I am unsure what to add here.

### Issue Labels

new feature

### Additional Information

_No response_

kommunarr:
# Page bookmarks

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #1505
closes #4040
#312
partial alternative for #4594
provides setup for #1492

## Description
  - The ability to bookmark routes (not just keyword searches) through a button on the top nav
  - Be able to set a named alias for those bookmarks through a modal popup
  - When the current route is saved, it shows as starred in the top nav
  - Clicking on the top nav for a bookmarked route opens a modal allowing you to edit the name and/or remove the bookmark
  - Searching for a substring of a bookmark link or alias in the search bar shows the corresponding bookmarks at the top of the search results, with visual & ARIA identifiers
    - Navigating to a page bookmark clears the input
  - Default route name is chosen based off of the `document.title`, playlist name, or search query depending on route
  - Delete all bookmarks setting in Privacy Settings
  - Thanks to the full path of routes being stored, including parameters, "Search" bookmarks are restored with the same search filter settings they were made with
  - If the current route is bookmarked, that bookmark does not show in the search results while still on that route
  - The user is warned with an inline message, but not barred from doing so, when their current page bookmark name open in the modal is a duplicate of an existing page bookmark name
  - The search results for a bookmarked result show an icon corresponding to the route type
  - The search results are now limited to 15 max results, prioritizing page bookmarks first
  - Search results are now each visually constrained to one line/row, due to much longer results being possible now
  - Removing a page bookmarked user playlist also removes the page bookmark automatically (note: only applies to the main Playlist page(s) being page bookmarked)
 
## Screenshots 
Screenshot_20240611_163520
Screenshot_20240611_163627
Screenshot_20240611_180201

## Testing 
- Test that bookmarks & bookmark removals are remembered after reloading
- Test that bookmark name updates are remembered and reflected after reloading
- Test that bookmarks are set to reasonable defaults for all types of page routes
- Test bookmark deletion in Privacy Settings
- Test that bookmarks appear when matching at the top of the search results
- Test that navigating bookmark search results with the Up/Down, Right/Left, and Enter keys behaves as expected
- Test that bookmarks appear and function properly even when `General Settings > Enable Search Settings` is disabled
- Test that bookmark icon enabled/disabled states are discernible with `Theme Settings > Match Top Bar with Main Color` enabled (secondary color theme is used as fill)
- Test that page bookmarking a user playlist page and deleting that playlist also removes the page bookmark
- Test that page bookmarking multiple user playlists' pages and then removing all playlists in Privacy Settings removes all of the page bookmarks for the now-deleted user playlist pages
- Test that page bookmark with a duplicate name shows a warning

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

### Future PRs ### 
- Import/export bookmarks (somewhat related to #1426)
- (Needs refinement / discussion) Hub for viewing and editing all bookmarks
- (Questionable / needs discussion) Truncate search results such that they never wrap to a new line (this is apparent if you want to create exceptionally long playlist names as an edge case, but this wrapping of long search results onto new lines is already existing & encounterable behavior)
- (Questionable / needs discussion) Hide page bookmarks setting (separate from Enable Search Suggestions toggle)
- (Even more questionable / needs discussion) Have the set Search filter parameters update when taking in a saved Search Results page with given search parameters (more confusing than not, in my opinion)
- Search history (#1492): We can easily fit generic search history entries into this paradigm by creating `search-history.db` entries with `isBookmark: false` set. **Edit:** probably should just leverage the existing search cache as the data source & existing access methods, but we can reuse the UI patterns added in this PR. I can revert https://github.com/FreeTubeApp/FreeTube/pull/5003/commits/392fa75a871c949edbda42b57eab9e80fef191bb, the commit I added specifically to better accommodate the possibility of a generic permanent search history through the `search-history.db`, if desired. It depends on whether we want to support a permanent search history and/or one with time-based expiry rules going forward.

### Aspects open for discussion ###
  - Related: I keep `spellcheck` on for the bookmark name insertion `ft-input` because we do it elsewhere for most of our inputs, even though it's a bit ugly. This one is debatable, but if I change it for this one, I'm also inclined to change it for other instances as well (e.g., User Playlist name insertion, profile name insertion).
- I put the Bookmark icon next to the search bar and filter button because it helps make it visually clear that it's related to them.


efb4f5ff-1298-471a-8973-3d47447115dc:
This feature differs quite a bit from what I imagine it to be. Lets take the following artistic masterpiece as a starting point (this applies to #1505 and #1492):

Screenshot (331)

ONLY when the user clicks on an empty search bar (imagine the search bar being empty in the pic above):
- The user will see the bookmarked searches listed on top, marked with a colored star. 
- The user also sees the recent searches they made right beneath the bookmarked searches and can be removed by hitting the X. 

When the user starts to type into the search bar:
- Bookmarked and recent searches will not be listed.
- Only search suggestion will be listed.
- User can only bookmark from the search suggestions list and can be added to the bookmarked list (see picture below)

Screenshot (332)

Debatable:
- Bookmarking searches with filters applied to it. I think the user should stay in control of their bookmarks just with search suggestions. Apply the filter before/after you hit search

I dont see the need for:
- Bookmarking routes
- The popup to give the bookmark an alias

jasonhenriquez:
> ONLY when the user clicks on an empty search bar (imagine the search bar being empty in the pic above):
    The user also sees the recent searches they made right beneath the bookmarked searches and can be removed by hitting the X.

Addressing #1492 is out of scope for this PR, IMO, unless we want to make it even bigger.

> When the user starts to type into the search bar: Bookmarked and recent searches will not be listed.

I disagree on this one. This sets a hard limit on the number of bookmarks, makes searching bookmarks harder, and is different from most browser (and in the case of YT for search history) behavior that people are most familiar with, where those are put at the top of the results.

> User can only bookmark from the search suggestions list and can be added to the bookmarked list (see picture below)

I disagree with this one for the reason that it's a prohibitive UX. You don't often think to save a page before you actually click on it, which is going to lead to a good deal of frustration. You probably want it to be somewhere visually constant so that you can bookmark a page whenever you want to. See my discussion from the OP:

> Main routes like /about are bookmarkable, even though they're equally accessible through the side nav. This might seem unnecessary, but having controls on the top nav phase in and out of being active seems like a poor & confusing choice in comparison. Having it be omnipresent like all of our other top nav icons, and communicating that any route works for it, makes the new feature more apparent and easier to understand/use. 

> I dont see the need for:
    Bookmarking routes

See answers to the above on why I disagree. In short, the cognitive model of browser bookmarking is one that users know about and are more effortlessly familiar with, but the model you're proposing is one that users will have to put in some work to learning and understanding the behaviors you've laid out, as it's a novel (at least to my knowledge) system and ruleset. 

> I dont see the need for:
    The popup to give the bookmark an alias

I can see the argument for this one. I would be open to having the modal close action for this one be to save rather than cancel (similar to Firefox), but I do worry that's not as intuitive given that differs from our other modals. Alternatively, we could have bookmarking not open the modal by default, but a toast pops up presenting you the option to edit it (otherwise, you can just click the icon again).

efb4f5ff-1298-471a-8973-3d47447115dc:
> > ONLY when the user clicks on an empty search bar (imagine the search bar being empty in the pic above):
> > The user also sees the recent searches they made right beneath the bookmarked searches and can be removed by hitting the X.
> 
> Addressing #1492 is out of scope for this PR, IMO, unless we want to make it even bigger.

Included that one just to make the examples easier to understand

> > When the user starts to type into the search bar: Bookmarked and recent searches will not be listed.
> 
> I disagree on this one. This sets a hard limit on the number of bookmarks, makes searching bookmarks harder, and is different from most browser (and in the case of YT for search history) behavior that people are most familiar with, where those are put at the top of the results.

Ah i didnt know. How does YT search history work? Could you maybe expand some more on the how browsers tackles bookmarks and how we do a similar job of it (i havent used browser bookmarks for a long time)

> > User can only bookmark from the search suggestions list and can be added to the bookmarked list (see picture below)
> 
> I disagree with this one for the reason that it's a prohibitive UX. You don't often think to save a page before you actually click on it, which is going to lead to a good deal of frustration. You probably want it to be somewhere visually constant so that you can bookmark a page whenever you want to. See my discussion from the OP:
> > Main routes like /about are bookmarkable, even though they're equally accessible through the side nav. This might seem unnecessary, but having controls on the top nav phase in and out of being active seems like a poor & confusing choice in comparison. Having it be omnipresent like all of our other top nav icons, and communicating that any route works for it, makes the new feature more apparent and easier to understand/use.

Ah ok I think I understand

> > I dont see the need for:
> > Bookmarking routes
> 
> See answers to the above on why I disagree. In short, the cognitive model of browser bookmarking is one that users know about and are more effortlessly familiar with, but the model you're proposing is one that users will have to put in some work to learning and understanding the behaviors you've laid out, as it's a novel (at least to my knowledge) system and ruleset.

👍 im fine with it but please provided some examples of the familiar behaviors like requested above.

> > I dont see the need for:
> > The popup to give the bookmark an alias
> 
> I can see the argument for this one. I would be open to having the modal close action for this one be to save rather than cancel (similar to Firefox), but I do worry that's not as intuitive given that differs from our other modals. Alternatively, we could have bookmarking not open the modal by default, but a toast pops up presenting you the option to edit it (otherwise, you can just click the icon again).

After testing some more i withdraw this. I think its very valuable to search for something with a filter applied and bookmark it with an alias so you exactly know what it is

jasonhenriquez:
> How does YT search history work? 

Screenshot_20240501_065600

> Could you maybe expand some more on the how browsers tackles bookmarks and how we do a similar job of it (i havent used browser bookmarks for a long time)

Some behaviors vary based off of the specific browser and what settings you have enabled, but at least in every major browser, bookmarks and search history are prioritized over regular results by default. [See discussion here on Chromium's behavior for example.](https://superuser.com/a/1550589) As to which takes precedence over which, I think it varies. In Firefox, judging from a few minutes of testing,  bookmarks and search history are equal, with relevance, closeness of order, and time of search seeming to be the apparent factors of sorting in order or priority descending.

> im fine with it but please provided some examples of the familiar behaviors like requested above.

This is grabbed from Firefox:

Screenshot_20240501_070302

jasonhenriquez:
> ONLY when the user clicks on an empty search bar (imagine the search bar being empty in the pic above):
The user will see the bookmarked searches listed on top, marked with a colored star.
The user also sees the recent searches they made right beneath the bookmarked searches and can be removed by hitting the X.

I disagree on the first one. I think we should only do this for search results, because this matches browser / YT behavior better.

jasonhenriquez:
[Here are the builds](https://github.com/jasonhenriquez/FreeTube/actions/runs/8977181830) in case you're interested in testing it out for your regular use. I've admittedly been pretty eager to incorporate this into my personal use.

jasonhenriquez:
There was feedback provided in the Matrix chat from one developer that this might constitute creep. I see two ways of thinking about it:
- I think I can see that bookmarking arbitrary pages can be much less useful, so there's a very reasonable argument to be made for restricting this feature to specific routes (page searches, and arguably user playlist pages). This would probably require some refactor to the UX & have it located in the page content, rather than in the top nav.
  - Pros:
    - No support for "useless" use cases
    - Aligns with solutioning system of: there is always one way to solve any problem, and that's all that's needed
  - Cons:
    - Although this solution seems less wasteful, I have no idea what the mental model is for "saving searches" that isn't "bookmarking." What it will be is not a thing that most anyone has come into contact with, and thus it's most likely going to be less intuitive to understand & use than the current system.
    - We could theoretically keep the existing UX but solely just have it disappear / be disabled on most routes, but then you run into another mental model issue of "bookmark icon that is seemingly arbitrarily absent / disabled a lot of the time." Even if we could find a way to make it non-confusing, such a change to the existing behavior seems very likely to cause more frustration than any noticeable benefit.

- On the other hand, FreeTube is basically a super browser for navigating the certain websites that we cover, and there isn't much risk for confusion or other negative externalities if we lean a bit further into those strong mental models. Having a good & accessible UX means that users don't have to try hard to learn how to use your app. And so oftentimes, if users want to save a given page to go back to, sometimes they will want to bookmark it. The same way that even though YouTube has user playlists, and the ability to subscribe to channels, and so on, people still bookmark certain user playlists, and certain channels, and so on, because it's a more comfortable & familiar way for them to engage with the app, that probably has subtler conveniences to the flow of it (e.g., it's all under one system, I have a naming schema I use that makes accessing pages lightning fast, and so on).
  - Pros:
    - Aligns with the solutioning system of: whatever is easiest to use, we're all for it, even if it means more than one way that a problem could be solved.
    - Aligns with the conventional mental model of browser page bookmarks, making it much easier for users to solve many common problem types, both for new use cases and even existing ones.
  - Cons:
    - Provides theoretical support for "useless" use cases (e.g., bookmarking the Subscriptions page).
    - One could make an argument that it creates an opposing system to "Quick Bookmarks," thus confusing users.
      - I will say, I don't find this one very convincing. Saving videos through that system is more accurately termed "favoriting" and "adding to a playlist" than "bookmarking". They're located in very different places spatially & behaviorally, letting us very easily occupy the two spaces of YouTube's playlist/favoriting mental model in the side nav and the browser's page bookmarking mental model by the search bar in the top nav. 
    - Deviates from the herd of the other similar apps, where this isn't a thing.
      - For similar reasons to those given in my answer above, I think this is an example of what makes FreeTube unique. People using web applications like Piped and Invidious are most definitely leveraging their browser functionalities of page bookmarks and so on in their personal use instead of solely leveraging the capabilities provided by those websites directly. The benefit of FreeTube is that we are a whole application designed around your viewing experience. We're your browser to search and bookmark whatever pages you want. We're your vault of playlists. We're your private viewing history, with no strict need to have strong trust or commitment to any given host.

And so, barring some major missing cons to the second option that I'm most definitely missing (as I am oft to do), I think the main issue isn't about what problems are caused by this PR in its current state at a user level, but rather the philosophical problem of how we want to let users solve problems. I lean on the side of "whatever is easiest to use, we're all for it, even if it means more than one way that a problem could be solved" because it is - when done right (and not in the vein of "infinite feature creep, keep adding things to keep the developer brain happy") - the best UX. I can see why the other often opposing philosophy of removing any possible chaff is so appealing from an optimization perspective (Z is possible through way X, so requesting us to build Y is just putting more effort onto developers), and as a heuristic (if the minimum problem needed to be solved is X, adding more could just be adding more complexity & effort without a commensurate return in utility yielded). 

But in this particular case, I think this is more clearly in favor of the former solutioning method because any other mental model or more limited form of the feature is surprisingly asking more of our users than its current state. I'm certainly willing to hear other opinions on it, especially if anyone else wants to [try out the build](https://github.com/jasonhenriquez/FreeTube/actions/runs/9362748649) for their own personal use and get beyond just the theoretical side of the discussion. Especially from non-developer-y people, because I do know we devs tend to get in our heads & worlds separate from what's actually going on.

PikachuEXE:
A few comments after trying it out briefly
- Lacking cancel/close button on Edit Bookmark modal (especially for mobile)
- Lacking "page type" in default bookmark name (if the bookmark entries can be shown with icons/whatever indicating the page types this might not be a big issue)
- Deleting local playlist should auto delete related bookmark entry
- Being able to create bookmark entries with duplicate name is unexpected for me

jasonhenriquez:
> Lacking cancel/close button on Edit Bookmark modal (especially for mobile)

From the Web side, the idea is that the `Save` button closes if no changes, similar to how it works in browsers like Chrome and FF (except in Chrome, the button is labeled `Done`). This is maybe a bit different feeling than those because it's a modal and not a context menu (which I also will add to the Future PRs section now). In the meantime, would a label of `Done` better communicate that?

For mobile, I honestly didn't think about it needing something different, and I'm glad you pointed it out. Interestingly, it seems like most mobile web browsers I'm testing with don't actually show a modal by default at all. Firefox Mobile saves the bookmark with a toast in one click, then lets you either re-click the icon to remove, press the toast to Edit, or go to a Bookmark Settings to edit it after you find it visually in the list. What about mobile one-click save, toast on save that opens the Modal, and re-click on a bookmarked video to open the modal?

> Lacking "page type" in default bookmark name (if the bookmark entries can be shown with icons/whatever indicating the page types this might not be a big issue)

This one is interesting. I'm guessing you're referring to the custom routes where we show the playlist name or video title and not the `translateWindowTitle` default title routes. Yeah, I'm on board with a visual indicator of the page corresponding to each route in lieu of the magnifying glass icon.

> Deleting local playlist should auto delete related bookmark entry

This one is interesting as well. I could see it being convenient for the use case of "get rid of this video everywhere", having them linked like this, but I'm not sure if that would ever be undesirable. Maybe I'll add this one with an undo toast.

>  Being able to create bookmark entries with duplicate name is unexpected for me

This is in line with how browsers act, so I don't think it would be surprising / a bad UX as a user necessarily. I'll revisit this one when we have the route-specific icon, because in such a case having multiple same-named bookmarks would still be understandable if they're different pages for the same channel (e.g., their UULF playlist, their channel page, etc.,) or different gradients of the same content (e.g., a "GOAT music" video, a "GOAT music" playlist).

PikachuEXE:
I am fine with `Save` only if when there are changes (new bookmark/edit with changes)
But having `Done` constantly seems easier & consistent
From a Chromium base browser:
image

To have mobile one-click save the default title must be good enough and it doesn't seem to be right now (also consider usually web pages got more unique titles unlike inside FT)
Also in mobile it's easy to mis-click, so I am not sure if one-click should be enabled **in the first release version**

> I could see it being convenient for the use case of "get rid of this video everywhere"

I am talking about local playlist (not videos inside local playlist). Having a bookmark entry remained after local playlist deleted = can visit a page with deleted local playlist = unexpected for me
No idea what "get rid of this video everywhere" means

> This is in line with how browsers act

It's possible to have the same page/URL bookmarked in several folders (or even duplicate entries in the same folder)
However browsers have a bookmark management page and we don't have
I don't know how bad this can be after page type displayed but at least we should show a warning (with count maybe) when trying to create/edit an entry while entries with the same name exist

src/datastores/handlers/base.js
```diff
@@ -167,12 +167,39 @@ class Playlists {
   }
 }
 
+class SearchHistory {
```
PikachuEXE, line 170: Probably should rename all related code to page bookmark?

PikachuEXE (commented):
Playlists page & single playlist page (remote & local) bookmark entries got different icons?
image
image

Could use better default titles, e.g. playlist to have channel name when exists like https://github.com/FreeTubeApp/FreeTube/pull/5226, same for videos

src/datastores/handlers/base.js
jasonhenriquez: The intention is that we could have persistent keyword search history entry logic here in the future (see `Future PRs` in OP), and page bookmarks are a type of search history.

jasonhenriquez:
> Playlists page & single playlist page (remote & local) bookmark entries got different icons?

Yes, see [likely implementation for the inline user playlist icon](https://github.com/FreeTubeApp/FreeTube/issues/4594#issuecomment-2115253335) for the motivation on that.

> Could use better default titles, e.g. playlist to have channel name when exists like https://github.com/FreeTubeApp/FreeTube/pull/5226, same for videos

If we don't like the current titles we're using, could we modify the `document.title` for those routes accordingly in a separate PR instead of having one-off overrides throughout the codebase?

src/renderer/components/page-bookmark-prompt/page-bookmark-prompt.js
```diff
@@ -24,6 +24,16 @@ export default defineComponent({
...
     },
+    pageBookmarks: function () {
+      return this.$store.getters.getPageBookmarks
+    },
+    duplicateNameCount: function () {
+      const currentBookmarkAdjustment = this.name === this.pageBookmark?.name ? -1 : 0
+      return currentBookmarkAdjustment + this.pageBookmarks.filter((pageBookmark) => pageBookmark.name === this.name).length
+    },
+    duplicateNameMessage: function () {
+      return this.$tc('Page Bookmark["There is {count} other bookmark with the same name."]', this.duplicateNameCount, { count: this.duplicateNameCount })
```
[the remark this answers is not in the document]
jasonhenriquez: Eh, I think having the key match the EN-US content as close as possible is best, even for minor things like punctuation, which help communicate tone

PikachuEXE:
Right below that comment there is:

> I wouldn't use fa-bars for the playlists button, as it is visually very similar to the icon used for the channels one.

https://github.com/FreeTubeApp/FreeTube/issues/4594#issuecomment-2115281762

I find it hard to distinguish one from another too
image

> If we don't like the current titles we're using, could we modify the document.title for those routes accordingly in a separate PR instead of having one-off overrides throughout the codebase?

Let me do that later

jasonhenriquez:
Hm, any suggestions? What if we made the Channels tab [user-check](https://fontawesome.com/icons/user-check?f=classic&s=solid)?

PikachuEXE:
It might help but playlist (at least user playlist) should still use `bookmark`, no idea about remote playlist

jasonhenriquez:
@PikachuEXE How about this? It's a bit odd since we're using filled versus unfilled to signify Quick Bookmark icon versus non-Quick Bookmark, but that's in a separate enough context, so I think it should be understandable here.

Screenshot_20240611_212959

PikachuEXE:
Then the problem being not being able to remember solid/normal icons means local or remote playlist
It's still better than before 
Still pending on me to get the title updated before re-testing (coz that might help

efb4f5ff-1298-471a-8973-3d47447115dc:
Are the testing steps still the same? I dont want to overlook anything

PikachuEXE:
Pending https://github.com/FreeTubeApp/FreeTube/pull/5273 then change icon
But I don't know what else's left

kommunarr:
Looks like the main testing criteria are all accounted for. Arguably you can also test having lots of bookmarks and search results (outcome: search results only reach as far as window size allows, as is the case in v21).

kommunarr:
I can resolve the merge conflicts if we're all on the same page of this feature being implemented. I don't see any public concerns mentioned in this thread, and I have [this unchallenged rebuttal](https://github.com/FreeTubeApp/FreeTube/pull/5003#issuecomment-2146772725) to one contributor's former concerns.

efb4f5ff-1298-471a-8973-3d47447115dc:
I think its safe to resolve conflicts

kommunarr:
> > closes #312
> 
> This does not close this issue. Subscribing to a playlist is like having a copy of it in your playlists page but every time you come back to it, it looks up the main playlists and update if there are changes.

I see, so this feature is defined by its presence on the User Playlists page? I will remove that it closes it, although admittedly not sure if there will be much incentive to prioritize it once this PR is merged to be fair.

> > I keep the  - FreeTube part in much of the default bookmark names, even though it's a bit ugly, because it helps to further visually (& search-wise) demarcate bookmarks, and is thus a good default.
> 
> I dont see any reason to include the `- FreeTube`. Could you maybe elaborate some more on this? Users already know they are using FT so why include it in the bookmark name by default?

My main point was that we don’t have a visual / semantic identifier for “default name” versus “custom name”, and that might be relevant in certain cases (e.g., is “Wildest Video Ever” what I labeled a wild video because I thought it was like that, or is that just the title?) Although looking back at it in retrospect, maybe that’s not an important piece of information most of the time, or if it is important, maybe we can just put quotes around default video titles (we may already doing that, idk, I’m on my phone and just going off memory). So I think I will just remove it.

> 
> > I put the Bookmark icon next to the search bar and filter button because it helps make it visually clear that it's related to them.
> 
> Seems fine to me.
> 
> > Main routes like /about are bookmarkable, even though they're equally accessible through the side nav. This might seem unnecessary, but having controls on the top nav phase in and out of being active seems like a poor & confusing choice in comparison. Having it be omnipresent like all of our other top nav icons, and communicating that any route works for it, makes the new feature more apparent and easier to understand/use.
> 
> Now that i had a closer look at this PR i think that we shouldnt do this. There is no reason for users to bookmark pages like these.

Looking back on it, I’m fine with disabling the button on certain predefined pages that will not ever want to be bookmarked by 90+% of people (as per our judgment). It doesn’t really matter much, but maybe it saves somebody a misclick every now and then, or prevents  someone from getting dependent on the feature in an uninentional way that we don’t want to support going forward. My intent when writing this is that I just didn’t want us to arbitrarily make it that, for example, you can’t bookmark playlists because we don’t want you to do that.

kommunarr:
> Settings page is still bookmarkable

See comment
https://github.com/FreeTubeApp/FreeTube/blob/2941e159a56dedebfcb0379b0ec5cabb658f8c8c/src/renderer/components/top-nav/top-nav.js#L56

I can comment it out and put it as a TODO if we think that's important to keep disabled in the interim - I do remember @MarmadileManteater saying she was working on creating linkable settings sections.


> So this just came up in my mind, why not show all the saved bookmarks in a separate page instead of all in the searchbar?

See description. I don't think this is crucial for the MVP:
> (Needs refinement / discussion) Hub for viewing and editing all bookmarks


>I cant bookmark community posts

Whoops, good catch!


> Is this intended behavior? Only seeing the bookmarks if a character matches one of the characters in the bookmark name? If so how will the user know that they hit the max nr of bookmarks?

Yes, it's like search results in a browser search bar. Not sure I understand your concern if you could clarify.


>  I can press enter if the menu is empty and it will still save the bookmark

Excellent catch, on it


> Something here looks off. I think its maybe the buttons that make the searchbar appear its not in center? Maybe its the other way around?

Hm, maybe needs a more constrained width. Will look into it

> Doesnt seem to be in the original spot anymore. Searchbar also moved. Is this an intentional change?

This got messed up at the last second and I somehow missed it. Thanks!

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc Addressed all of your comments, with verbal-only responses to points 1, 2, & 4.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (dismissed):
Navigation history dropdown icon next?

src/renderer/helpers/strings.js
```diff
@@ -56,6 +56,46 @@ export function translateWindowTitle(title) {
...
+    case '/post':
+      return ['fas', 'message']
+    case '/playlist': {
+      const solidOrRegular = route.includes('?playlistType=user') ? 'fas' : 'far'
+      return [solidOrRegular, 'bookmark']
+    } case '/watch':
+      return ['fas', 'play']
+    default:
+      return null
+  }
```
absidue, line 96: I haven't tested this pull request yet but this list includes a lot of routes that there is no point bookmarking, as they have icons in the side bar.

I haven't looked into the latest changes on this pull request yet, but last time I looked at it I was planning to veto it entirely, as it was so far away from what it should have been, just saving search results like people asked. I had stopped myself just before posting the message as you were already demotivated enough and I didn't want you abandoning FreeTube, again.

absidue (changes requested):

src/renderer/helpers/strings.js
absidue: I have just seen your justifications above about you wanting to follow web browser behaviour. FreeTube is a desktop app, not a web browser, it makes sense to be able to bookmark any page in a web browser as you can open anything on the internet, there is no point in doing that in FreeTube and should frankly be considered broken behaviour.

absidue:
After catching up on the thread, it looks like some of the problems that made me want to veto this pull request in the first place still exist (thankfully you conceded that this is not the right solution for saving playlists, so that's one thing at least) and that you are unwilling to budge on those remaining things and I am unwilling to budge from my position of not wanting to merge a pull request with major flaws. So it seems like we are at an impasse.

src/renderer/helpers/strings.js
kommunarr: The reason I kept icons for all routes is because I intend to use these for navigation history. [See comment here.](https://github.com/FreeTubeApp/FreeTube/pull/5227#issuecomment-2515391099)

kommunarr: > [I]t makes sense to be able to bookmark any page in a web browser [...] there is no point in doing that in FreeTube

This is not the current behavior in this PR, and this indicates to me that you have not actually tested it in its current state before reaching your conclusions. From what I can gather, from a purely functional perspective, your opinion is that the array below should just include `Search Results` and `Hashtag`, at which point some other tweaks to the implementation & UX could be made as well if we're 100% certain are the only two we'll ever be concerned about.
https://github.com/FreeTubeApp/FreeTube/blob/5ddfbcc1c439a1229bdb3313f8ef3ceb9d73cab4/src/renderer/components/top-nav/top-nav.js#L50-L58

My disagreement, that is seemingly shared by at least a few other team members, is that it doesn't present a detriment to the user experience to also allow users to save other relevant content like posts, channels, videos, playlists, and linkable settings sections / settings searches (TBD) for easy use if they so prefer. I'll provide counterarguments to three concerns to this approach now.

1. **Some people will not use this feature.** This is, in absence of additional harms applied to those users in exchange (a separate concern I will address later on), a neutral and acceptable state of affairs.
2. **Adding a new control will be confusing or otherwise materially harm the user experience by merit of it being present.** This is a valid concern insofar as the existence of any new control is something that can risk contributing to clutter or comprehension problems. This is where I believe therso far away from what it should have been, just saving search resulte was a misunderstanding with regards to my rebuttal I provided to this a while back:
> I have just seen your justifications above about you wanting to follow web browser behaviour [...] there is no point in doing that in FreeTube and should frankly be considered broken behaviour.

The claim I was making was not "if web browsers do it, let's copy what they do." Rather, I was addressing the comprehension concern that users would have comprehension issues learning what this feature is and how to use it, such that it would negatively contribute to their user experience. To clarify, this means that this is very easy to learn how to use and interact with as a control compared to any alternative access method that has been suggested so far. 

To bring it back to:
> It's so far away from what it should have been, just saving search results

What does the mental model look like for "save an arbitrary search so I can go back to it later?" and the corresponding subproblems? The supposedly more modest alternative is actually more foreign and harder to learn & use to most users! Whereas this is something that anyone knows how to use because we're appropriating the same mental model.

3. (**Philosophical claim) There is not a strict need for this, either because it has low salience or that there are / will be alternative access methods for these five additional types of content such that presenting this new access method is problematic.** Please correct me if I'm wrong, but I believe this is the main thing that you're hinting at with this quote:
> [T]here is no point in doing that in FreeTube and should frankly be considered broken behaviour.

This is a philosophical disagreement that I don't think you will readily change your mind on, but I do think I can at least try to address some relevant factors here.
  i. I do believe there is a bit of a personal bias in that you do not intend to use this feature yourself. This in itself fine, as I am also biased as someone who wants to use this feature, so I don't think holding that bias is inherently a bad thing.
  ii. Related to i, you think that the user base at large would feel the same. This one neither of us can discuss definitively because you and I both have dead zones for how we use the app & our priorities versus the user base at large. From a democratic perspective, let's say that we cancel each other out so far, thus leaving the rest of the team to be the deciders of ii.
  iii. Related to i and ii but not exactly, I do believe that part of this disagreement is that if we have other access methods for these forms of content (e.g., subscribing to a channel, saving a video to a playlist), adding a new one for them is buggy, wasteful, and/or a faux pas. This is something I addressed earlier, but what I disagree with in this philosophical claim is that it originates from a place of "if the users just interacted with the app the way we told them to, they wouldn't need this additional trash." Even if everything is possible through a rigid access method, that doesn't make it the only way that users can or should go about doing it. I'll link to [my earlier reply](https://github.com/FreeTubeApp/FreeTube/pull/5003#issuecomment-2146772725) where I addressed this in more depth, but needless to say, I do think there is room for serious disagreement with this philosophical claim.

As it stands, from the inventory above, my assessment is that most of your disagreement is represented by philosophical disagreement in point number 3 and its subclaims i-iii. Please correct me if I'm wrong on that, as I'm sure there might be more to your perspective, but you have not elucidated it yet, so I am going off of the available information you have provided. If I am correct that it is mainly philosophical concerns, I do think from a democratic perspective we should leave it to the rest of the team to determine whether to merge this or not, with you and me only providing advisory opinions in terms of the philosophical disagreements.

absidue (commented):
Incomplete code review.

I'm still opposed to adding multiple ways to do the same things, we already have more powerful ways of d
saving channels and videos, via subscribing and adding to a playlist respectively. I also see no point in saving a link to the settings, we already have the settings themselves linked in the sidebar and when you visit the setting's page we have the side bar that you implemented that lets you jump to specific sections. If you want to search for settings lets implement a proper settings search on the settings page, not this.

kommunarr:
> I'm still opposed to adding multiple ways to do the same things, we already have more powerful ways of
saving channels and videos, via subscribing and adding to a playlist respectively.

I don't know if we'll ever reach full agreement on this philosophical claim, and I think that's okay in itself. If I could provide a bit more reasoning on why I disagree with this for practical reasons, here's a bit of it:

While I obviously don't have concrete metrics on this, anecdotally speaking, there is a sizable contingent of YT / Piped / etc. users who are bookmarkers, i.e., they use their browser's bookmarking feature for saving and organizing videos or relevant pages. While some of this can be attributed to beliefs that it's relatively more private, the more pertinent explanation for a lot of those users is "I just like doing it more that way." In reality, that explanation likely has layers of "YT's UX is more confusing for me," "I'm familiar with the mental model of saving things in bookmarks," "YT's UX takes more time or searching in comparison to the way that I use it." 

Now, one answer to these users from FreeTube could be that they're dinosaurs, and they just have to give that up if they want to use our app, as we already have access methods for any kind of page. The problem is that these folks, as much as we may not agree with them in full, actually have some merit to the friction they feel. For example, that a method exists for doing something != it's the best experience doing that something. One could say "oh, you can already save things through X way" - but it's not just about saving: it's finding, retrieving, accessing, and archiving in the common language & access pattern of pages. 

If there's a channel I'm in the process of binging, I could spend a half minute scrolling through my subscribed channels to find it in, or go to the Channels tab and search the name, or search their name in the search bar and try to find it in the results, but not all of these are perfect & intuitive to most users, especially if it's not something they've consciously thought about (note: rule one of UX design: don't make me think](https://www.interaction-design.org/literature/article/don-t-make-me-think-key-learning-points-for-ux-design-for-the-web)). If there's one particular video I'm thinking of that I want to watch, I could of course a) toggle the "Playlist with Matching Videos" setting, search a keyword, click on that playlist, and click on the particular video that matches that keyboard, or b) put what I remember of the title in the search bar and hope to find it in the results, but again, I think it's fair to say that's not the best or most intuitive solution for most people — especially given that they are not as familiar with the app as you and me. If there's a particular remote playlist or UULF-channel-as-playlist I like to watch, we don't currently have any good solutions, but we could theoretically decide to emulate the idea of saving remote playlists like YouTube has as a feature — I'm not fully sold on it as an idea / experience, but even if/when we do implement it, it will be more or less the same functionality — bookmarking a link to a certain kind of page that you have to go to "certain kind of page" section (Playlists) to get to.

The main point I'm getting at here with these — alongside a few other kinds of saveable content that we don't have any existing means to easily save & return to — is that there are theoretically "most powerful" ways to save and organize a few types of content in FreeTube, but the mental models and access methods are anything but homogeneous, thus introducing a steeper learning curve to getting to know & use the app (harming the UX for the majority of users). Even for power users, the speed, convenience, & level of cognition required for finding, retrieving, accessing, & archiving each different kind of content are harmed by their heterogeneity. I posit that this is why many YouTube users choose — either partially or to a great extent — their browser's bookmarking functionality in lieu of YouTube's features for everything. I further assert that people who think FreeTube's UX can initially and/or occasionally be subtly frustrating to use is at least somewhat attributable to this unfilled usability gap.

> I also see no point in saving a link to the settings, we already have the settings themselves linked in the sidebar and when you visit the setting's page we have the side bar that you implemented that lets you jump to specific sections. If you want to search for settings lets implement a proper settings search on the settings page, not this.

I do agree more with this particular part based on considering your reasoning and will remove that for now. The initial idea was that if we have linkable settings searches, rather than the flow of "click settings tab -> focus secondary search bar -> search formal settings name" for more regularly modified settings, it would instead be "focus primary search bar -> search bookmark name / have it queued up already -> click entry" and it takes you there. But parsing it out, the effort & complexity difference between those flows might be pretty marginal in practice. I will also work on resolving your other mentioned points and migrate this new component to the Composition API.

kommunarr:
If we don't have adequate consensus on the "separate system" of data idea, which is the main rebuttal to this that I find merit in, we can make refactors to the system to have it, e.g., be a "save page as corresponding type" button instead (e.g., on a video, it represents adding / is it added to your Quick Bookmark playlist, on a channel, it represents subscribing / is it subscribed to by your current profile). I can see the issue with having it be "separate from" playlists, etc., but I still think the heterogeneity of access methods and the difficulty of searchability of saved content are salient and worthwhile problems to solve if nothing else.

absidue:
This pull request in it's current state doesn't actually improve the searchability of saved content, it actually makes it worse as it introduces a second way to save various things and introduces a search mechanism for that specific way of saving content, so nothing changes about the searchability of the existing things. So you would have two ways of saving things that work completely differently, so I would argue that it actually does the opposite of what you seem to be trying to achive, as you would then have to also remember where you saved what.

Additionally reworking this pull request to do that would still not improve the situation, as you would have a magic button that does completely different things on different pages, which would just lead to more confusion.

I still don't understand how the prominent subscribe and quick bookmark buttons are difficult to find or use, it makes even less sense with the context that you literally designed both of them. I also still don't understand why you are so set on taking a web browser content and forcing it into something that is not a web browser, it feels a lot like you've been doing something one way for a long time in a separate program and instead of being flexible and adapting to the way that FreeTube does things (using the UI components that you literally designed), you want everyone else to adapt to your way of doing things even if that makes it worse for everyone else.

ChunkyProgrammer (commented):
Sorry for the late review, I can see this feature being useful for FreeTube power users.

From what I can tell, browsers don't add bookmarks to search suggestions when searching (at least by default) and only include search history/history in there. Is the plan to remove bookmarks from the search suggestions when search history is added and there's a separate page to view all bookmarks?

absidue:
I've updated my comment to remove some of the more fiery parts, in the hopes that you'll consider the other parts of my message.

PikachuEXE:
I can think of >=2 things worth saving right now (which can't be saved): 
- online playlist
- search criteria (keywords included) (which indirect saves hashtag, channel, not sure about channel with search filter

I agree saved search criteria can be shown in search suggestion (if implemented well). But can also be shown in a separate page for better view (see criteria details, management etc.). Online playlist bookmark have no place in search criteria by default IMO. (Just create another page for it)

I saw stuff like community post which I don't use so I don't comment on it

kommunarr:
Following some private discussion with some teammates, the main course of action is likely going to be implementing #1492 first, partially by adopting the search history DB machinery added here. After that, I wrote this set of options in terms of the team deciding the best course of action for page bookmarks:

A. Keeping the PR concept mostly as-is
  - Pros: see aforementioned points [here](https://github.com/FreeTubeApp/FreeTube/pull/5003#issuecomment-2543145432)
  - Cons: See previously discussed points of it being a separate system of data not adequately leveraging the more powerful systems of playlists and subscriptions

B. Breaking down aspects of the page bookmarking PR to potentially implement piecemeal as independent chunks:
- Aspect 1 (already fully implemented): icons per route (also needs alt text, I believe, as noted by @absidue )
- Aspect 2 (partially implemented, but will need some adjustment): user playlists, current profile- subscribed channel names, and Quick Bookmark videos show up in search bar results based on most similar 1-4 matches to query
  - Pros: greater searchability
  - Cons: if user profile has very high amount of subscriptions and/or videos in their Quick Bookmark playlist (high 4 figures and up), the suggestions will take a moment to properly populate, or we cap it to the first, e.g., 5000 entries (actual number determined in practice based on performance testing)
- Aspect 3a: saving search results and posts through i) the bookmarking UI added in the PR or ii) some alternative UI
- Aspect 3b: have the bookmarking UI added in this PR be used as profile subscribe button for channels and save to Quick Bookmark playlist for videos
  - Pros: alongside 3ai, maintains consistency of the bookmarking UI saving content on given page
  - Cons: polymorphic behavior could be confusing, low utility in and of itself

src/renderer/components/ft-input/ft-input.js
```diff
@@ -2,7 +2,9 @@ import { defineComponent } from 'vue'
 import { mapActions } from 'vuex'
 
 import FtTooltip from '../ft-tooltip/ft-tooltip.vue'
-import { isKeyboardEventKeyPrintableChar, isNullOrEmpty } from '../../helpers/strings'
+import { getIconForRoute, isKeyboardEventKeyPrintableChar, isNullOrEmpty } from '../../helpers/strings'
+
+const MAX_VISIBLE_LIST_ITEMS = 15
```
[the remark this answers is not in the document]
kommunarr: Good catch, will look back into fixing this and your other comments once we can reach a cleaner team consensus on broader course of action as discussed [here](https://github.com/FreeTubeApp/FreeTube/pull/5003#issuecomment-2552238802).

kommunarr:

Issue #4387, opened 2023-11-26
[Feature Request]: Use tabs on settings page


pkrasicki:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

I think the current design of the settings page using expandable sections is not the most intuitive and could be improved on desktop and mobile.

![before](https://github.com/FreeTubeApp/FreeTube/assets/49721209/4b3221ce-95ed-4f23-aa62-afd6ce6a4fd6)

### Proposed Solution

I propose to use tabs instead. On desktop:

![after2](https://github.com/FreeTubeApp/FreeTube/assets/49721209/0113ece1-6abd-4048-a325-563e58461c5a)

If tabs don't fit on the screen, a horizontal scrollbar would be shown.

On mobile tabs would be shown vertically:

![after-mobile-all](https://github.com/FreeTubeApp/FreeTube/assets/49721209/b1b4be1e-5c70-45f7-971d-2682e6e8abd6)

Clicking a tab would take the user to a different screen. An extra arrow button was added to allow the user to go back to categories:

![after-mobile-general](https://github.com/FreeTubeApp/FreeTube/assets/49721209/caa16bce-1e7d-49e3-ad4b-f6cdf987a8ce)

### Alternatives Considered

Vertical tabs like in GNOME settings (see the screenshot below).

### Issue Labels

visual improvement

### Additional Information

Toggle switches are usually located on the right side, so maybe that should be changed too. Examples:

GNOME settings
![switches-gnome](https://github.com/FreeTubeApp/FreeTube/assets/49721209/d86e5911-dd06-4c68-91a3-49b5b673324f)

Android settings:
![switches-android](https://github.com/FreeTubeApp/FreeTube/assets/49721209/bf0b8e72-de02-46ab-9fc2-315b0e3ad5b9)

iOS settings:
![switches-ios](https://github.com/FreeTubeApp/FreeTube/assets/49721209/5b39fa26-ae7b-4295-9ddf-f8138cc1c14a)
 settings:

[...]

PR #5010, opened 2024-04-24
Implement settings section sorting option

jasonhenriquez:
# Implement settings section sorting option

## Pull Request Type

- [x] Feature Implementation

## Related issue
None

## Description
Implement programmatic list of the settings sections, which makes it easy to programmatically reference them in dynamic ways that will be useful in the future (ex: #4387). This makes something like sorting them very easy. 

Setting an alphabetical ordering is a good choice for the users who forget the arbitrary section ordering and need to find a particular section (as opposed to us die-hard contributors who remember them by heart). This will be more apparent with the sections collapsed, and/or any menu-type component added for #4387.

This code was salvaged from #4231. Only A-Z sorting is implemented, as was determined in the discussion under that PR. 

> [!NOTE]  
Changing the language does not refresh the sort if the setting is enabled first until the route is reloaded. This is an easy way to prevent the sections from jumping sporadically (& losing your place) when changing the language.

## Screenshots 
Screenshot_20240423_224504

## Testing 
- Sort works in arbitrary language (see note above: make sure to reload page)
- Sort does not re-trigger when changing to a language with a different alphabetical sort order

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

### DIFF
--- a/src/renderer/views/Settings/Settings.js
+++ b/src/renderer/views/Settings/Settings.js
@@ -40,16 +40,103 @@ export default defineComponent({
   data: function () {
     return {
       usingElectron: process.env.IS_ELECTRON,
-      unlocked: false
+      unlocked: false,
+      settingsComponentsData: [
+        {
+          type: 'general-settings',
+          title: this.$t('Settings.General Settings.General Settings')
+        },
+        {
+          type: 'theme-settings',
+          title: this.$t('Settings.Theme Settings.Theme Settings')
+        },
+        {
+          type: 'player-settings',
+          title: this.$t('Settings.Player Settings.Player Settings')
+        },
+        {
+          type: 'external-player-settings',
+          title: this.$t('Settings.External Player Settings.External Player Settings'),
+          usingElectron: true
+        },
+        {
+          type: 'subscription-settings',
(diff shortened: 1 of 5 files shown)


src/renderer/views/Settings/Settings.js
```diff
@@ -40,17 +40,104 @@ export default defineComponent({
...
       return this.$store.getters.getAllSettingsSectionsExpandedByDefault
     },
+
+    settingsSectionSortEnabled: function () {
+      return this.$store.getters.getSettingsSectionSortEnabled
+    },
+
+    settingsSectionComponents: function () {
+      let settingsSections
+      if (!this.usingElectron) {
```
absidue, line 127: We only needed `usingElectron` inside of the Vue template, as you can only access things in there that have been specifically exposed to it.
As this check has now been moved into a JavaScript file, you can use the build flag directly. Please remove the `usingElectron` entry from the top of `data` when you do this change, as it is no longer needed.
```suggestion
      if (!process.env.IS_ELECTRON) {
```

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5011, opened 2024-04-24
Fetch invidious instances from file then use api

ChunkyProgrammer:
# Fetch invidious instances from file then use API

## Pull Request Type
- [x] Bugfix

## Description
Instead of waiting for the Invidious API endpoint for FreeTube to start, this PR will default to loading the included file of invidious instances and then update the instance list if it is successful

## Testing 
- add console.logs in the `fetchInvidiousInstancesFromFile` and `fetchInvidiousInstances` actions to ensure that the api is being updated properly

## Desktop

- **OS:** Linux Mint
- **OS Version:** 21.3
- **FreeTube version:** 0.20.0 - nightly

### DIFF
--- a/src/renderer/store/modules/invidious.js
+++ b/src/renderer/store/modules/invidious.js
@@ -16,14 +16,26 @@ const getters = {
 }
 
 const actions = {
+  async fetchInvidiousInstancesFromFile({ commit }) {
+    const url = createWebURL('/static/invidious-instances.json')
+
+    const fileData = await (await fetch(url)).json()
+    const instances = fileData.filter(e => {
+      return process.env.SUPPORTS_LOCAL_API || e.cors
+    }).map(e => {
+      return e.url
+    })
+
+    commit('setInvidiousInstancesList', instances)
+  },
+
+  /// fetch invidious instances from site and overwrite static file.
   async fetchInvidiousInstances({ commit }) {
     const requestUrl = 'https://api.invidious.io/instances.json'
-
-    let instances = []
     try {
       const response = await fetchWithTimeout(15_000, requestUrl)
       const json = await response.json()
-      instances = json.filter((instance) => {
+      const instances = json.filter((instance) => {
         return !(instance[0].includes('.onion') ||
           instance[0].includes('.i2p') ||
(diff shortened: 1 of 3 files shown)


jasonhenriquez (commented):
`https://invidious.flokinet.to/` is removed even though it still appears in [the list of official instances](https://docs.invidious.io/instances/). Rationale?

ChunkyProgrammer:
> `https://invidious.flokinet.to/` is removed even though it still appears in [the list of official instances](https://docs.invidious.io/instances/). Rationale?

I just used the script to generate the list:
Looks like it's showing as API disabled (https://api.invidious.io/)
image

jasonhenriquez:
Thanks for the info @ChunkyProgrammer, I was wondering. Do you have a link to the script?

ChunkyProgrammer:
@jasonhenriquez yes, it's [this script](https://github.com/FreeTubeApp/FreeTube/blob/development/_scripts/getInstances.js) which can be ran usimg `yarn run get-instances`

jasonhenriquez (approved):
Tested and is working as expected. LGTM, although could theoretically be more optimal if we could trigger the API call & file read simultaneously, then have logic where the file read cannot overwrite the API call as long as `instances.length !== 0`. I'm used to using `rxjs` for that; not sure if there's a Vue-y way.

PikachuEXE (dismissed):

src/renderer/store/modules/invidious.js
```diff
@@ -16,42 +16,48 @@ const getters = {
...
           instance[0].includes('.i2p') ||
           !instance[1].api ||
           (!process.env.SUPPORTS_LOCAL_API && !instance[1].cors))
       }).map((instance) => {
         return instance[1].uri.replace(/\/$/, '')
       })
+
+      if (instances.length !== 0) {
+        commit('setInvidiousInstancesList', instances)
+        return true
```
absidue, line 49: Looks like the return values from this function are unused, so they can probably be removed?

absidue (approved):

PikachuEXE (approved):

Issue #5008, opened 2024-04-23
[Feature Request]: Remember the Playlists "Sort By" selection between sessions


profucius:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Currently, the Sort By dropdown selection box forgets the chosen selection between sessions (close Freetube and then launch again). It always defaults to "Recently Played". I want it to remember which choice was last selected between sessions.

### Proposed Solution

Make it remember the last chosen selection. 

You can implement this similarly to this: In the beta build 4242 (at least), within a playlist there is another "Sort By" dropdown selection box. This selection _does_ remember the last chosen selection between sessions.

### Alternatives Considered

No alternatives.

### Issue Labels

improvement to existing feature

### Additional Information

![image](https://github.com/FreeTubeApp/FreeTube/assets/18254607/45d25c2a-a005-4088-b6d8-d027b9441eaf)

jasonhenriquez:
Inviting discussion from other contributors, but I think this is a reasonable change. If someone changed that sort, they probably would want to keep it. This is especially relevant for #4594, where we wouldn't want the sort order of the inline playlists reverting if it was changed in a given session.

efb4f5ff-1298-471a-8973-3d47447115dc:
The add to playlist window also has a sort by that works in the same way as the OP described. Should we also change that? Im not sure

jasonhenriquez:
> The add to playlist window also has a sort by that works in the same way as the OP described. Should we also change that?

I honestly think if we make the Playlist search UX more compact, we do not need either the Add Video to Playlist search bar or sort bar. So I think it's fine that we keep it the same for now.

jasonhenriquez:
> Would it also be possible to also remember the sorting of it when you played a playlist. Like when you sort latest added first when you play the playlist it would also be the same

If you're talking about having the user playlist sort order being reflected in the box on the /Watch route, this is a separate issue that I have a soon-to-be-released PR for.

susansli:
Hi! I would like to do this issue.

[...]

PR #5013, opened 2024-04-24
Respect playlist sort order in `watch-video-playlist`

jasonhenriquez:
# Respect playlist sort order in watch-video-playlist

## Pull Request Type

- [x] Feature Implementation

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/5008#issuecomment-2074563978

## Description
Implements sort order for a given user playlist when opening it with watch-video-playlist. This ensures 

## Screenshots 
Screenshot_20240424_112212

## Testing 
- Test sort order is reflected in `watch-video-playlist` for arbitrary sort
- Insert inserted elements are shown in the proper place

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
Playlist sort is computed when the user playlist in `watch-video-playlist` is first loaded and when it is updated with additions / removals during that session. While this is not necessarily having an apparent performance effect in my testing, I would imagine it's at least a potential concern w.r.t. sum power draw across our user base.

As it stands, the O(nlogn) sorting operation is executed for each playlist load, or upon each addition/removal to a playlist while it is currently being watched. In the future, 1) we may want to consider adding an ordering property to playlistItems for the custom sort order instead of relying on its default order, 2) thus allowing the "sort" action to be one-and-done per sort preference change by storing that playlist in its sort preference, and 3) using binary search for each subsequent insertion into a playlist. This would remove the O(nlogn) operation upon each playlist load, and it would reduce the addition cost to O(klogn), where k is the number of elements being added.
- Alternatively, instead of implementing 1, we can 1b) just remove "Custom". although we'll still have to implement 2.
- Note that 2 means we'll have to update all playlists upon each sort, unless we implement it in a way that delays the sort change on a given playlist until that playlist is actually accessed (e.g., having a property on the DB object called "sortOrder" that we check if it lines up with the userPlaylistSortOrder on DB retrieval).
- We can make "Added First" the new default logical stored ordering with 1 or 1b, thus improving performance and minimizing processing power even more.
- By modifying the base ordering, the playlist thumbnail will update as per the user's preferences instead of the currently disparate logical order, which is a more preferable state of affairs.


src/renderer/helpers/playlists.js
```diff
@@ -0,0 +1,41 @@
...
+    case SORT_BY_VALUES.DateAddedOldest:
+      return a.timeAdded - b.timeAdded
+    case SORT_BY_VALUES.VideoTitleAscending:
+      return a.title.localeCompare(b.title, locale)
+    case SORT_BY_VALUES.VideoTitleDescending:
+      return b.title.localeCompare(a.title, locale)
+    case SORT_BY_VALUES.AuthorAscending:
+      return a.author.localeCompare(b.author, locale)
+    case SORT_BY_VALUES.AuthorDescending:
+      return b.author.localeCompare(a.author, locale)
```
absidue, line 36: As you already mentioned potential performance problems, using `Intl.Collator` instead of `localeCompare` should help with that a bit.

> When comparing large numbers of strings, such as in sorting large arrays, it is better to create an `Intl.Collator` object and use the function provided by its `compare()` method.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

```suggestion

  let collator

  if (
    sortOrder === SORT_BY_VALUES.VideoTitleAscending ||
    sortOrder === SORT_BY_VALUES.VideoTitleDescending ||
    sortOrder === SORT_BY_VALUES.AuthorAscending ||
    sortOrder === SORT_BY_VALUES.AuthorDescending
  ) {
    collator = new Intl.Collator([locale, 'en'])
  }

  return playlistItems.toSorted((a, b) => {
    const first = !reversed ? a : b
    const second = !reversed ? b : a
    return compareTwoPlaylistItems(first, second, sortOrder, collator)
  })
}

function compareTwoPlaylistItems(a, b, sortOrder, collator) {
  switch (sortOrder) {
    case SORT_BY_VALUES.DateAddedNewest:
      return b.timeAdded - a.timeAdded
    case SORT_BY_VALUES.DateAddedOldest:
      return a.timeAdded - b.timeAdded
    case SORT_BY_VALUES.VideoTitleAscending:
      return collator.compare(a.title, b.title)
    case SORT_BY_VALUES.VideoTitleDescending:
      return collator.compare(b.title, a.title)
    case SORT_BY_VALUES.AuthorAscending:
      return collator.compare(a.author, b.author)
    case SORT_BY_VALUES.AuthorDescending:
      return collator.compare(b.author, a.author)
```

jasonhenriquez:
I'm still feeling shaky about the energy expenditure aspect of it, so I'm leaning towards shelving it this release and working on the performance suggestions.

jasonhenriquez:
Update 5/23:
- Nedb doesn't let you insert at a specific place. So then I thought that maybe you can do the sorting in `grabAllPlaylists`, but you'd also have to do it every time the locale changed as well if it's set to one of the alphabetical sorting options. I suppose that both changing your language frequently and using an alphabetical sorting option are pretty rare, so I think this is still a pretty clear net benefit.
  - Delaying applying the sort until the playlist is accessed is weird too, as the thumbnails on the UserPlaylists page would be misrepresentative, unless we want to also just load all of the sorts there as well.
- Has anyone ever looked at how YT handles moving playlist videos? It's very weird, I'm surprised it is the way it is. I'm still trying to wrap my head around it. It seems like whenever you change the sort order of the playlist, all of your changes are wiped. When you change the sort order and click "Play All," it opens & plays it in your suggested sort order. But when you move the videos around while the playlist is being watched in a different sort order, the next time you open the playlist page, it maintains the position of those manual moved video entries, but with the default `Date added (newest)` as the sort!

absidue:
I think it's best to keep the current solution of sorting on demand, rather than storing the sorted playlist. It will also make adding a video to a playlist more expensive, as we'll have to sort the playlist every time you add a video to the playlist, regardless of what page you are on, instead of just if you are on the playlist or watch page.

We can skip the first sort on the watch page by using the already sorted playlist from the playlists page, by using the cached playlist trick that we use for remote playlists (currently not done for the user playlists as we fetch them from the playlists store every time), where we pass through the videos from the playlist page to the watch page, by saving them to the utils store. You will still need to fetch it from the store and sort it if the user accesses the watch page directly (e.g. middle clicking on a video on the playlists page, which would open a new window right to the watch page).

jasonhenriquez:
I'm seeing merit to what you're saying, although:

> It will also make adding a video to a playlist more expensive, as we'll have to sort the playlist every time you add a video to the playlist, regardless of what page you are on, instead of just if you are on the playlist or watch page.

This isn't exactly true, to be pedantic. If you have the sorting applied in `grabAllPlaylists`, on sort change, and (for specifically the alphabetical sort options) on locale change, then you can do a binary insertion or deletion in the mutation logic at O(logn) cost, which is different from sorting the playlist. The main struggle I'm having is that when watching a playlist and adding or removing videos to it, it will require the playlist to be re-sorted in full each time, which is the full O(nlogn). 

> We can skip the first sort on the watch page by using the already sorted playlist from the playlists page, by using the cached playlist trick that we use for remote playlists (currently not done for the user playlists as we fetch them from the playlists store every time)

This is definitely the right choice if we're to go this path. Thanks for this suggestion.

jasonhenriquez:
Update for those not privy to the private discussion: We will be addressing these performance concerns separately. Only additional requested change for this PR is that we send the cached playlist with sort order enabled whenever available to avoid an unnecessary re-sort whenever pulling up the cached user playlist.

jasonhenriquez:
Gonna wait till #5158 is merged first.

jasonhenriquez:
One other unrelated note (concerning future work) with doing sorting on the fly is that we won't be able to implement placing the proper playlist thumbnail corresponding to the video at the top of the sort order on the User Playlist page as easily.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Insert inserted elements are shown in the proper place

What do you mean with this? User is on the watch page and adds video to the playlist they are currently watching?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/helpers/playlists.js
```diff
@@ -0,0 +1,52 @@
...
+
+  let collator
+
+  if (
+    sortOrder === SORT_BY_VALUES.VideoTitleAscending ||
+    sortOrder === SORT_BY_VALUES.VideoTitleDescending ||
+    sortOrder === SORT_BY_VALUES.AuthorAscending ||
+    sortOrder === SORT_BY_VALUES.AuthorDescending
+  ) {
+    collator = new Intl.Collator([locale, 'en'])
```
PikachuEXE, line 24: (Optional) I know nothing about `Intl.Collator`
But I would like to know if using a constant object be possible (map of `locale` to `Collator` objects lazily created)
(No idea how much it helps, if too complicated might not worth it

PikachuEXE (approved):
Seems to be working well

src/renderer/helpers/playlists.js
jasonhenriquez: I'll mark it as a chore, although I'm skeptical if the access time would be considerably less than the creation time of this

absidue (approved):

PR #5022, opened 2024-04-25
Only approve web API permission requests for permissions that FreeTube needs

absidue:
# Only approve web API permission requests for permissions that FreeTube needs

## Pull Request Type

- [x] Other - Security Improvement

## Description
In a web browser when a webpage wants to use an API that could provide access to sensitive information or allows potentially unwanted actions, the web browser prompts the user for permission, Electron however defaults to approving all permission requests. This pull request implements [Electron security guideline 5: "Handle session permission requests from remote content"](https://www.electronjs.org/docs/latest/tutorial/security#5-handle-session-permission-requests-from-remote-content) by only approving permissions requests for the two permissions that FreeTube actually uses, other permission requests get rejected. I also decided to reject all permission requests for non-FreeTube URLs, that's a future proofing measure in case we ever show a remote page in an iframe or something.

FreeTube only needs the following permissions:
- `fullscreen`: So that the video player can enter full screen
- `clipboard-sanitized-write`: To allow the user to copy video URLs and error messages

### Tests to check that functionality that relies on permissions still works
* Test that copying a link from the 3 dots menu on a video in a video still works (that the link gets written to your clipboard, you can just paste into a text editor or something right afterwards to confirm it)
* Test that the video player can still be put into fullscreen

### Test to check that unwanted permission requests are rejected
* Run `await navigator.mediaDevices.getUserMedia({ audio: true })` (requesting an audio stream from the microphone) in the dev tools, it should throw `Uncaught DOMException: Permission denied`

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -286,6 +286,32 @@ function runApp() {
       })
     }
 
+    // Electron defaults to approving all permission checks and permission requests.
+    // FreeTube only needs a few permissions, so we reject requests for other permissions
+    // and reject all requests on non-FreeTube URLs.
+    //
+    // FreeTube needs the following permissions:
+    // - "fullscreen": So that the video player can enter full screen
+    // - "clipboard-sanitized-write": To allow the user to copy video URLs and error messages
+
+    session.defaultSession.setPermissionCheckHandler((webContents, permission, requestingOrigin) => {
+      if (!isFreeTubeUrl(requestingOrigin)) {
+        return false
+      }
+
+      return permission === 'fullscreen' || permission === 'clipboard-sanitized-write'
+    })
+
+    session.defaultSession.setPermissionRequestHandler((webContents, permission, callback) => {
+      if (!isFreeTubeUrl(webContents.getURL())) {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

jasonhenriquez:
**question:** This is more my ignorance and not an actual concern, but why does opening external links work when `openExternal` is not configured?

absidue:
FreeTube intercepts all clicks on `a` elements, if it's a YouTube link it gets opened in FreeTube, otherwise depending on the user's settings either nothing happens, the user gets prompted or it gets opened by passing it to the main process via IPC call, which opens the link.

https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/App.js#L345-L390

https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/helpers/utils.js#L328-L339

jasonhenriquez (approved):

PR #5023, opened 2024-04-25
Fix handling of emojis with ZWJ sequences in profile initials

absidue:
# Fix handling of emojis with ZWJ sequences in profile initials

## Pull Request Type

- [x] Bugfix

## Description
At the moment we use `Array.from` to get the first character in the profile name to use as the initial in the bubble, while that correctly handles basic emojis, it doesn't handle flag or family emojis as well as ones with skin tone modifiers very well, splitting them up into their individual characters instead of keeping them as one unit. This pull request fixes that by using `Intl.Segmenter` instead, which does locale aware splitting and also splits by user-perceived character, so it keeps the emojis intact.

### Beforehand
before

### Afterwards
after

## Testing 
Test profile names with some complex emojis in them:
https://emojipedia.org/flags
https://emojipedia.org/emoji-zwj-sequence
https://emojipedia.org/people#families-couples

Please note that Windows doesn't show flags correctly, that's not an issue with FreeTube but an issue with Windows.
> Emoji flags are supported on all major platforms except Windows, which displays two-letter country codes instead of emoji flag images. Windows has never supported any national flag emoji design.

## Desktop

- **OS:** Windows
- **OS Version:** 10


jasonhenriquez (approved):
LGTM. TIL!

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Off topic, I would like to see it displayed as

image
https://emojipedia.org/woman-technologist

PR #5029, opened 2024-04-27
Settings menu & UX improvements

### LINKED ISSUE #4387: [Feature Request]: Use tabs on settings page

jasonhenriquez:
# Settings menu

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4387

## Description
- Implements a menu in settings for more easily navigating the sections
- Currently scrolled-to section is marked as active
- CHANGE: Removes "Settings expanded by default" setting and `<details>` / `<summary>` tags (now inapplicable to current design)
- CHANGE: General Settings is now always the top setting, regardless of sorting preference (read comment on e9c527a5d399416334aead62215cd201f84f0de8)
- CHANGE: Other settings sections style changes, including rounded corners, external section titles, and removal of `<hr>` tags (for a sleeker aesthetic)
- CHANGE: Removes experimental removing of underlines on links pre-hover with the Hot Pink theme (breaks menu link styling, and was not ever noticeably beneficial)

## Screenshots 
Screenshot_20240426_220558

Testing
- Confirm that Settings Menu works intuitively
- Confirm that General Settings is always at the top of the settings, even after sorting
- Confirm that "long" version of settings label is displayed in the Settings Menu if no "short" label exists (i.e., as of now, with any non-`EN-US` locale)
- Test tablet & mobile views

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

### Remaining action items
- Use SponsorBlock icon instead of generic `shield` icon for `SponsorBlock Settings` (this icon is not supported by `font-awesome-icon`)
- [Needs further thought/discussion/refinement] Find way to integrate Settings Menu with the mobile view (currently, we're just hiding it)

The above two I can see leading into minor quibbles on the how and what, and are somewhat arguable if they're needed at all, so I am strongly considering saving those for a second PR.

I'm inviting thoughts on this question and the overall design before undrafting this one, as this design does have some minor differences with the original proposed in the issue.


PikachuEXE:
1. Not a breaking issue but why rounded corners
FT has square "back plates" in other places

image
image

2. (I am not a mobile user so I am guessing here)
The old UI is faster for user to get to lower sections (find section, open, do stuff)
The new one requires plenty of scan time due to all expanded settings

jasonhenriquez:
> Not a breaking issue but why rounded corners

**Short explanation:** This was admittedly a subjective design choice that I think makes for a sleeker look here. 

**Longer explanation:** In design, you want to maximize the alignment of elements, such that if you were to draw a straight horizontal or vertical line, you will have page elements line up as much as possible. This makes visual scanning easier. And so if we have straight edges, we would be inclined to align each settings section title along the edge exactly. But having the headings non-indented also makes their importance in the visual hierarchy less clear. So rounded edges here hit a sweet spot where there's an indentation for the titles still, and also alignment with the top edge. 

> The old UI is faster for user to get to lower sections (find section, open, do stuff)
The new one requires plenty of scan time due to all expanded settings

Are you referring to the removal of the "Settings expanded by default" setting? That's a fair note. If Emma and/or other mobile-savvy folks think mobile navigation will be harder without that setting and without having some form of accessing the new menu, I can try to get a mobile version of the Settings menu included in this PR rather than in a later one.

PikachuEXE:
I am fine with rounded edges due to your explanation

General better for desktop (me) but ya mobile side might be well...

jasonhenriquez:
@efb4f5ff-1298-471a-8973-3d47447115dc This was something that took a lot of work to hit a sweet spot for. The main problems we want to solve are 1) that the active Menu section appears to be the current section you're on, 2) that clicking on a section makes it actually show as the active section, and 3) that clicking on a section actually takes you to that section. The problem with choosing the top of the viewport below the top nav is that it it doesn't work for 1, as the last section that's just peaking through the top of the screen often appears as the active one, seeming pretty odd / not useful. I tried with other solutions like the exact midpoint of the viewport being the "current section line", which has the problem of either 2 or 3. I decided on using the top 1/4th of the viewport height as the "section line", which ends up being a decent compromise for all 3. Let me know if you have any better solution in mind, but this is the best I could find.

src/renderer/components/ft-settings-menu/ft-settings-menu.vue
```diff
@@ -0,0 +1,30 @@
...
+    class="settingsMenu"
+  >
+    <h2 class="header">
+      {{ $t('Settings.Settings') }}
+    </h2>
+    <a
+      v-for="(settingsSection) in settingsSections"
+      :id="settingsSection.type + '-link'"
+      :key="settingsSection.type + '-link'"
+      :href="'#' + settingsSection.type"
```
absidue, line 12: As the in-app navigation uses the URL hash too, it might be better if we find a different way to scroll to the sections to avoid potential navigation issues.

In release builds: `app://bundle/index.html#/subscriptions`
In dev mode: `http://localhost:9080/index.html#/subscriptions`

src/renderer/views/Settings/Settings.vue
```diff
@@ -1,34 +1,33 @@
...
+            :label="$t('Settings.Sort Settings Sections (A-Z)')"
+            :default-value="settingsSectionSortEnabled"
+            :compact="false"
+            @change="updateSettingsSectionSortEnabled"
+          />
+        </div>
+        <div class="settingsSections">
+          <template
+            v-for="(settingsComponent) in settingsSectionComponents"
+            ref=""
```
absidue, line 20: No point adding an empty ref property.
```suggestion
```

src/renderer/components/ft-settings-menu/ft-settings-menu.js
```diff
@@ -0,0 +1,16 @@
...
+  props: {
+    settingsSections: {
+      type: Array,
+      required: true
+    },
+  },
+  methods: {
+    getTitleForSection: function(settingsSection) {
+      return settingsSection.shortTitle !== '' ? settingsSection.shortTitle : settingsSection.title
+    }
```
absidue, line 14: If you handle the fallback in the Settings component instead of here, you won't need any fallback logic here and can get rid of this method, referencing `settingsSection.shortTitle` directly in the template.

absidue:
> [Needs further thought/discussion/refinement] Find way to integrate Settings Menu with the mobile view (currently, we're just hiding it)

For mobile I think a good approach would be to fake paginate the settings. When you open the settings page all you would see is the menu (would of course have to be redesigned a bit so it looks good taking up the whole screen) and then when you click on one of the items, it hides the menu and shows just that section, as well as a back button that takes you back to the overview. The toggling could be handled by the menu component emitting an event when a section is clicked, allowing the Settings component to "navigate" to the section.

jasonhenriquez:
Updated now with your suggestions.
Screenshot_20240515_001051
Screenshot_20240515_001205

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM

MarmadileManteater:
I realise this was discussed above, but I want to talk about it because it bothers me: rounded corners. I understand that there may be well founded reasons for this design. However, my counterpoint would be that this design doesn't exist in a vacuum. Next to the other views, this one is very different. There is nothing inherently wrong with the rounded corners cardless look, but it doesn't look anything like the rest of the app.

What if the `settingsSection` had `ft-card` directly under it instead?
image

jasonhenriquez:
Thanks for the feedback, that's fair. Does this look okay now?

Screenshot_20240517_181820

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez:
I'm personally not as much visually a fan of the wrapping. Maybe I can make it a bit wider to accommodate these edge cases. My hope was/is that the short label translations can circumvent the issue after enough time.

PikachuEXE:
Maybe use a smaller font-size? 16px:

image
image
image
image

PikachuEXE (dismissed):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/constants.js
```diff
@@ -74,6 +74,9 @@ const SyncEvents = {
   }
 }
 
+// Settings
+const ACTIVE_CLASS_NAME = 'active'
+
```
absidue, line 79: As this is only used in the `Settings.js` file, you might as well just specify it in there, saves you the comment and the import.

src/renderer/views/Settings/Settings.vue
```diff
@@ -1,34 +1,49 @@
 <template>
-  <div>
+  <div class="settings settingsPage">
```
absidue, line 2: As these two hardcoded classes are just used for this element, could they be consolidated into a single class?

src/renderer/views/Settings/Settings.vue
```diff
@@ -1,34 +1,49 @@
...
+          />
+        </div>
+        <div class="settingsSections">
+          <template
+            v-for="(settingsComponent) in settingsSectionComponents"
+          >
+            <component
+              :is="settingsComponent.type"
+              :ref="settingsComponent.type"
+              :key="settingsComponent.type + '-component'"
```
absidue, line 40: Now that we no longer have the `<hr>` element, we can just used the type as the key directly.
```suggestion
              :key="settingsComponent.type"
```

absidue (commented):
I'm slightly concerned that it might be easy to accidentally click the wrong link in the menu on mobile, as they are close together and fingers are usually much bigger than a mouse cursor.

src/renderer/components/ft-settings-menu/ft-settings-menu.css
```diff
@@ -0,0 +1,150 @@
...
+  font-size: 26px;
+  inline-size: fit-content;
+  margin-block: 0 10px;
+}
+
+.title {
+  text-decoration: none;
+  color: var(--tertiary-text-color);
+  inline-size: 220px;
+  padding-block: 3px;
```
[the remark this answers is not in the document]
jasonhenriquez: Thanks for this feedback. One big issue I had was keeping all of the menu items in view for all device types. I may have to switch to a dynamic padding solution that takes up the full page height (up to a certain height, at least).

jasonhenriquez:
> When tabbing to the whole settings menu i would expect the go back to the main settings page to be the last tabable object. This isnt the case as the user has to tab to the settings menu and all the objects outside the menu to come back to the back button

Done!

> I think its safe to also remove the word settings in all the headers just like you did on the side. I dont see a reason for keeping it and from looking at other applications on my computer, none of them use that word when inside the settings category

I have no issue with that. Although could we get consensus on this first? I don't want to invest in that effort until we're certain that that's the game plan we want to go with.

PikachuEXE:
Are they all "settings/options"? If yes I am fine

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (commented):
Pending mobile side feedback before approving
Looks fine on desktop

src/renderer/components/ft-settings-menu/ft-settings-menu.css
```diff
@@ -0,0 +1,125 @@
...
+  .header {
+    font-size: 32px;
+  }
+}
+
+/* small height or width mobile breakpoint; intermediary text */
+@media only screen and (width <= 1015px) and (height <= 830px),
+only screen and (width <= 500px) {
+  .settingsMenu {
+    font-size: 18px;
```
MarmadileManteater, line 113:
```suggestion
    font-size: 25px;
```
Could this possibly be a little larger?

src/renderer/components/ft-settings-menu/ft-settings-menu.css
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  .titleIcon {
+    inline-size: 18px;
+    block-size: 18px;
+    margin-inline-end: 5px;
+  }
+
+  .header {
+    font-size: 22px;
```
MarmadileManteater, line 123:
```suggestion
    font-size: 25px;
```
(i'm suggesting this because it should be the same, or larger, than the other font-size i suggested to change)

jasonhenriquez:
Updated now: 
Screenshot_20240607_071925

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> Pending mobile side feedback before approving
> Looks fine on desktop

@MarmadileManteater do you still see any issues with this?

kommunarr:
If possible, I'd like this to be reviewed soon so I could work on other improvements to the settings.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
i tabbed to the page i wanted and selected it. Tabbing further through the list will make me eventually go up to the first element in the General Settings which is undesired as I wanted to be in the Player settings

https://github.com/user-attachments/assets/0c64506e-3cb4-457f-b895-5282429e5368

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

PR #5044, opened 2024-04-30
Make adding duplicate disabled by default

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Disable adding duplicate videos to playlists by default (cannot select playlists with all to be added videos already present)
Can be enabled via new toggle only visible on prompt when any duplicate detected

## Screenshots 

Nothing changed for video when absent in all playlists, only screenshots of adding video(s) with some/all already present in playlists

Adding single video
image

Adding multiple videos (duplicate disallowed)
image

Adding multiple videos (duplicate allowed)
image

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Style/UI/Text not finalized


efb4f5ff-1298-471a-8973-3d47447115dc:
Yeah not sure about the placement of that. The Add to Playlist prompt is already very overcrowded

PikachuEXE:
The idea is that user need to be able to enable the option before potentially tabbing into the playlists so they can select the potentially disabled playlist(s) instead of finding out playlists disabled and go back to enable it

No idea where to place it best

jasonhenriquez:
Could you justify the three controls at the top all next to each other on the right? It's not a perfect solution for now, but we'll look back on it in the future.

PikachuEXE:
Here is my imagined flow:
1. Enter query text
2. Optionally search for playlists with matching video title instead of just matching playlist names
3. Optionally realize that it's already added but want to add duplicate anyway
4. Sort it when playlist cannot be found

Ideally it should be (4) then (3) but it might look strange if there is an extra row just for one toggle on desktop
image

jasonhenriquez:
Sorry, to clarify, I mean right-justify the controls so they're grouped

PikachuEXE:
Like this?
image
image

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.css
```diff
@@ -24,17 +24,20 @@
 
 .optionsRow {
   display: grid;
-  grid-template-columns: repeat(2, 1fr);
+  grid-template-columns: repeat(3, auto);
+  column-gap: 32px;
   grid-template-rows: 1fr;
   align-items: center;
+  justify-content: end;
```
jasonhenriquez, line 31:
```suggestion
  justify-content: flex-end;
```

Not sure why the linter is not enforcing this one

PikachuEXE: What's the difference in grid layout?
In [CSS Box Alignment Module Level 3](https://www.w3.org/TR/css-align-3/#positional-values) it seems to say that unless in flex layout `flex-end` = `end`

For flex layout with `flex-direction: row-reverse` there is a difference between `justify-content: flex-end` and `justify-content: end` but in grid layout there is no property similar to `flex-direction` AFAIK
https://csslayout.news/whats-the-difference-between-the-alignment-values-of-start-flex-start-and-self-start/

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
```diff
@@ -48,14 +57,19 @@
       <ft-flex-box>
         <div
           v-for="(playlist, index) in activePlaylists"
-          :key="playlist._id"
+          :key="`${playlist._id}-${playlistDisabled(playlist._id)}`"
           class="playlist-selector-container"
+          :class="{
+            disabled: playlistDisabled(playlist._id),
+          }"
```
jasonhenriquez, line 64:
```suggestion
          }"
          :aria-disabled="playlistDisabled(playlist._id)"
```

There are still ways to focus non-focusable elements with screenreader tech, so best to just cover our bases semantically

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
```diff
@@ -28,11 +28,20 @@
...
         @change="doSearchPlaylistsWithMatchingVideos = !doSearchPlaylistsWithMatchingVideos"
       />
+      <ft-toggle-switch
+        v-if="anyPlaylistContainsVideosToBeAdded"
+        class="allowDuplicateToggle"
+        :label="$t('User Playlists.AddVideoPrompt.Allow Adding Duplicate Video(s)')"
+        :compact="true"
+        :default-value="addingDuplicateVideosEnabled"
+        @change="addingDuplicateVideosEnabled = !addingDuplicateVideosEnabled"
+      />
```
jasonhenriquez, line 44: **suggestion (blocking):** Thanks for moving these to the right. Could you also stack these two toggles vertically (to increase alignment and further visually group the controls)?

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
```diff
@@ -48,14 +57,19 @@
...
-          :key="playlist._id"
+          :key="`${playlist._id}-${playlistDisabled(playlist._id)}`"
           class="playlist-selector-container"
+          :class="{
+            disabled: playlistDisabled(playlist._id),
+          }"
         >
           <ft-playlist-selector
-            tabindex="0"
+            :tabindex="playlistDisabled(playlist._id) ? -1 : 0"
```
jasonhenriquez, line 67: **thought (future):** I would also like to refactor this to use `input type="checkbox"` someday, and similar for other form elements, because us imitating accessible functionality with custom controls ends up being error-prone and more work (this is pre-existing to this PR, just seems especially apparent with what we have to do to imitate the powerful `disabled` attribute)

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
```diff
@@ -238,10 +269,16 @@ export default defineComponent({
         const playlist = this.allPlaylists.find((list) => list._id === selectedPlaylistId)
         if (playlist == null) { return }
 
+        // Use [].concat to avoid `do not mutate vuex store state outside mutation handlers`
+        let videosToBeAdded = [].concat(this.toBeAddedToPlaylistVideoList)
+        if (!this.addingDuplicateVideosEnabled) {
+          const playlistVideoIds = playlist.videos.map((v) => v.videoId)
+          videosToBeAdded = videosToBeAdded.filter((v) => !playlistVideoIds.includes(v.videoId))
```
jasonhenriquez, line 276: **suggestion (non-blocking):** Would using a `Set` here be more performant?

jasonhenriquez (changes requested):

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
PikachuEXE: I wish there is a linter to provide warnings...

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
PikachuEXE: I don't think we have any existing design for having vertical grouping inside a row unless width is passed a breakpoint

image

This is not like settings which we have so many so that they are aligned horizontally with limited slots per row (well a table)
image

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
jasonhenriquez: What would that linting rule be? Put a warning on all types of array searches? Data structure choice is always going to be up to developer discernment, I'd imagine

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
jasonhenriquez: I'm having trouble understanding, are you expressing that it's difficult to implement or shouldn't be implemented for X reason?

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.js
PikachuEXE: True, Set can't have duplicate values, ignore~

src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue
PikachuEXE: > Could you also stack these two toggles vertically

I posted screenshot above (not pushed) to see (1) if this is what you want (2) if it is what you want then it shouldn't be like that since it's not a table and there is enough width

jasonhenriquez: I'm still not sure I understand. You can put the two toggles in a flex container that is `inline-end`-aligned, and then have that flex container's content be `inline-start`-aligned.

PikachuEXE: Short version: it looks weird
Long version: It's strange to see 2 elements/components aligned vertically in a row which is not a table and IMO is not an existing design in FT (feel free to name other projects with this design or existing places in FT with this design in case I miss). At the same time these 2 components are only common in type (toggle) but not their purpose. One is for searching/filtering and another one is for affecting the **outcome** of the action. (1st reason is more important than 2nd one)

jasonhenriquez: I personally disagree, but this is not a blocking concern, so will leave it to others' discretion and not push the matter

efb4f5ff-1298-471a-8973-3d47447115dc:
Only thing im missing tbh is a remove duplicates button within a playlist. Make it only appear when we know there are duplicates in the playlists other wise do not show it

PikachuEXE:
Preventing duplicates to be added and removing duplicates are different features IMO
I will open another PR for that

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
I think the left alignment looked better than grouping the settings together like this. 

Stacked isn't needed imo

PikachuEXE:
I agree left align look more natural / consistent with existing design in FT & in other places
Especially the `Playlist with Matching Videos` is meant to be grouped with search input above

Question is do you want it to look like OP screenshots or
image

<details> 

<summary> Hiding my code here since not committed yet </summary>

```css

.optionsRow {
  display: grid;
  grid-template-columns: repeat(auto-fit, minmax(100px, 1fr));
  align-items: center;
}

.tightOptions {
  display: grid;
  grid-template-columns: repeat(auto-fit, minmax(100px, max-content));
  column-gap: 16px;
  align-items: center;
}

@media only screen and (width <= 800px) {
  .optionsRow {
    /* Switch to rows from columns */
    grid-template-columns: auto;
    grid-template-rows: repeat(auto-fit, auto);
    align-items: stretch;
  }

  .tightOptions {
    /* Switch to rows from columns */
    grid-template-columns: auto;
    grid-template-rows: repeat(auto-fit, auto);
    align-items: stretch;
  }
}
```
</details>

jasonhenriquez:
I'd prefer if we move the sort box over to the left as well if we do that, because I don't like the cognitive aspect of the UX of requiring the user's eyes to dart to parse the meaning, but again, not a blocking issue, and does not matter much if we're going to pursue more expansive changes to the modal in the future

PikachuEXE:
Putting sort box in the right is existing design in FT and I consider that as a separate issue which should not be the concern of this PR

efb4f5ff-1298-471a-8973-3d47447115dc:
> Putting sort box in the right is existing design in FT and I consider that as a separate issue which should not be the concern of this PR

I agree.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

jasonhenriquez (approved):

absidue (approved):

PR #5050, opened 2024-05-02
Implement hot reloading for the locale files during development

absidue:
# Implement hot reloading for the locale files during development

## Pull Request Type

- [x] Other - Development enhancement

## Description
This pull request implements hot reloading for the locale files. I recommend reviewing this pull request by commit, so that it's easier to understand what changes are related to what.

Currently we have it setup to reload the entire app when a locale file is edited, which is not ideal for multiple reasons, a major one being that we have to load all the data from the databases again like we do at startup and all the caches get cleared.

What is hot reloading?
You know when you edit one of the Vue or JavaScript files and it only reloads the parts of the app that changed, that's hot reloading.

Please keep in mind that not all parts of the app react to changes to the locales, the same bits that don't automatically change when you change the language won't change here either, but you can just use the backwards and forwards arrows in the navigation bar, then which still avoids the full reload.

## Testing 
The easiest way to test that it works is to edit the `Search / Go to URL: Search / Go to URL` string in the language that you currently have selected, as that is the text that is shown in the search bar at the top of the window that is visible on all pages.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/_scripts/_hotReloadLocalesScript.js
+++ b/_scripts/_hotReloadLocalesScript.js
@@ -1,0 +1,18 @@
+const websocket = new WebSocket('ws://localhost:9080/ws')
+
+websocket.onmessage = (event) => {
+  const message = JSON.parse(event.data)
+
+  if (message.type === 'freetube-locale-update') {
+    const i18n = document.getElementById('app').__vue__.$i18n
+
+    for (const [locale, data] of message.data) {
+      // Only update locale data if it was already loaded
+      if (i18n.availableLocales.includes(locale)) {
+        const localeData = JSON.parse(data)
+
+        i18n.setLocaleMessage(locale, localeData)
+      }
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


_scripts/_hotReloadLocalesScript.js
```diff
@@ -0,0 +1,18 @@
+const websocket = new WebSocket('ws://localhost:9080/ws')
+
+websocket.onmessage = (event) => {
+  const message = JSON.parse(event.data)
+
+  if (message.type === 'freetube-locale-update') {
+    const i18n = document.getElementById('app').__vue__.$i18n
```
jasonhenriquez, line 7: There's so much I don't know. This is so cool. Side tip I found from looking this up for @ChunkyProgrammer: the Vue 3 equivalent is apparently [`__vueParentComponent.ctx`](https://stackoverflow.com/a/64429013).

jasonhenriquez (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

Issue #5000, opened 2024-04-22
[Feature Request]: Update label casing to sentence case


jasonhenriquez:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Our labels throughout FreeTube vary between [sentence case and title case](https://www.grammarly.com/blog/title-case-sentence-case/). There is no consistency, and this can at times be mildly jarring visually.

### Proposed Solution

Standardize all labels in FreeTube under sentence case, the one preferred by [many style guides](https://uxdesign.cc/title-case-vs-sentence-case-in-ux-writing-212087192261) ([see discussion here as well](https://www.caseconverter.com/sentence-case-vs-title-case/)). We can do this either [programmatically](https://stackoverflow.com/questions/4322314/convert-uppercase-letter-to-lowercase-and-first-uppercase-in-sentence-using-css) or manually (which sounds terrible). I don't think we can write exceptions for labels starting with a word that should be lowercase given our large amount of supported languages, but that's fine. 

Test modals and sections throughout the app to confirm that there are no issues with visually portraying the hierarchy with this change. If there are any, use additional styling like bolding and minor size differentials to resolve them.

### Alternatives Considered

None

### Issue Labels

visual improvement

efb4f5ff-1298-471a-8973-3d47447115dc:
I think Distraction free setting: Display Titles Without Excessive Capitalisation And Punctuation should be retested in PR

jasonhenriquez:
Removing "good first issue" as I feel like it might take some thought on how to actually do this. If we do it programmatically, for example, how would that affect languages like German, where capitalization is often necessary? We could probably have a set of languages not included, but then that's going beyond new contributor territory.

jasonhenriquez:
To do it programmatically, we'd also need a set of proper nouns saved to know not to make lowercase. It seems almost like a manual solution for `EN-US` would be better, then just hope that translations in other languages where sentence casing is preferred follow suit. I don't know of any better options at the moment.

ghost:
[...]

PR #5058, opened 2024-05-03
Always have quick bookmark set

### LINKED ISSUE #5051: [Feature Request]: Always have a quick bookmark playlist set if possible
### Guidelines

### Problem Description

I've seen multiple reports of users not having any playlist set as their Quick Bookmark target, which leads to what most users would consider a substantially lackluster UX for adding videos to playlists.

Assuming it's not due to a bug in our logic that sets the default playlist to `Favorites` upon first startup of post- custom User Playlists FreeTube, here are the main reasons for why I think users are not seeing any quick bookmark target:
- User clicks "Disable Quick Bookmark" icon without understanding what it means (it kind of looks like a link, maybe? We're leaning heavily on the icon alone to express a pretty complicated and specific meaning)
- User deletes their Favorites playlist at some point and uses a different one, not realizing that the Quick Bookmark will disappear until a new playlist is set as that one

### Proposed Solution

- Remove ability to "Remove quick bookmark"; only allow to switch it from another playlist where it's not set (icon shows as disabled on playlist where it's set).
- Have "Quick Bookmark" switch to another one automatically (if one exists) upon deletion (or insertion, if the # of playlists was zero), which every FreeTube developer compared to the primary mechanism in [the 1994 Christmas film "The Santa Clause"](https://en.wikipedia.org/wiki/The_Santa_Clause) when this topic was first discussed (please don't fact-check this claim).
  - Which one? The one at the top of the selected sort order sounds fine to me.

### Alternatives Considered

Lots of choices are possible here. I don't think there are any other better ones. Removing the ability to have no bookmarked playlist while you do have >0 playlists is one of those things that an extreme minority of users would somewhat mind, but it's greatly outweighed by the amount of users inconvenienced by us allowing that possibility.

### Issue Labels

### Additional Information

_No response_

jasonhenriquez:
# Always have quick bookmark set

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #5051 

## Description
- Changes "Add/Remove" quick bookmark icon to "regular/solid" bookmark icon instead of link/link-slash. See https://github.com/FreeTubeApp/FreeTube/commit/e8972540076c221ae19326d6b020091def7d7556 for discussion.
- Updates `ft-icon-button` `disabled` styling to show title on hover, and use `aria-disabled` (these is a needed usability improvement to better communicate what the disabled icon actually means, as our only disabled `ft-icon-button` case is the disabled `ft-refresh-widget` icon)
- Makes an enabled Quick Bookmark icon disabled to prevent unselection (see original story for rationale), with position in all cases moved to the be the leftmost icon now (to ensure non-disabled icons are grouped)
- Makes quick bookmark default to the 1st playlist in DB when no playlist is detected during startup / playlist add / playlist removal

## Screenshots 
Screenshot_20240503_125848
Screenshot_20240503_134255

## Testing 
- Have 1 user playlist -> delete playlist -> see no quick bookmark button on a given video in subscriptions (REG) -> add playlist -> see quick bookmark on playlist -> (REG, optional) see it persists on Ctrl+R
- Have >0 user playlists with no quick bookmark set (e.g., by removing it before switching to this PR, or deleting the `quickBookmarkTargetPlaylistId` entry in `settings.db`) -> run FreeTube -> see one of the existing playlists is set as the new quick bookmark playlist -> (REG, optional) see it persists on Ctrl+R
- (REG) Have 2 user playlists -> switch quick bookmark target to the other one -> see preference reflected and retained on Ctrl+R
- Have 0 user playlists -> Import user playlists in Data Settings -> see one is selected as new quick bookmark target

## Desktop

- **OS:** TW
- **OS Version:** OpenSUSE

## Future work
- Choosing new programmatic quick bookmark target to be the top of the sorting preference, not just the first one in the array

### DIFF
--- a/src/renderer/components/playlist-info/playlist-info.js
+++ b/src/renderer/components/playlist-info/playlist-info.js
@@ -230,9 +230,6 @@ export default defineComponent({
     quickBookmarkPlaylist() {
       return this.$store.getters.getPlaylist(this.quickBookmarkPlaylistId)
     },
-    quickBookmarkEnabled() {
-      return this.quickBookmarkPlaylist != null
-    },
     markedAsQuickBookmarkTarget() {
       // Only user playlists can be target
       if (this.selectedUserPlaylist == null) { return false }
@@ -366,7 +363,7 @@ export default defineComponent({
       if (this.selectedUserPlaylist.protected) {
         showToast(this.$t('User Playlists.SinglePlaylistView.Toast["This playlist is protected and cannot be removed."]'))
       } else {
-        this.removePlaylist(this.id)
+        this.removePlaylist(this.id, true)
         this.$router.push(
           {
             path: '/userPlaylists'
--- a/src/renderer/components/playlist-info/playlist-info.js
+++ b/src/renderer/components/playlist-info/playlist-info.js
@@ -402,10 +402,6 @@ export default defineComponent({
         showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist is now used for quick bookmark'))
       }
     },
-    disableQuickBookmark() {
-      this.updateQuickBookmarkTargetPlaylistId(null)
-      showToast(this.$t('User Playlists.SinglePlaylistView.Toast.Quick bookmark disabled'))
-    },
 
     updateQuery(query) {
       this.query = query
--- a/src/renderer/store/modules/playlists.js
+++ b/src/renderer/store/modules/playlists.js
@@ -1,4 +1,6 @@
 import { DBPlaylistHandlers } from '../../../datastores/handlers/index'
+import i18n from '../../i18n/index'
+import { showToast } from '../../helpers/utils'
 
 function generateRandomPlaylistId() {
   return `ft-playlist--${generateRandomUniqueId()}`
@@ -46,7 +48,7 @@ const getters = {
 }
 
 const actions = {
-  async addPlaylist({ commit }, payload) {
+  async addPlaylist({ state, commit, rootState, dispatch }, payload) {
     // In case internal id is forgotten, generate one (instead of relying on caller and have a chance to cause data corruption)
     if (payload._id == null) {
       // {Time now in unix time}-{0-9999}
@@ -79,15 +81,30 @@ const actions = {
 
     try {
       await DBPlaylistHandlers.create([payload])
+
+      const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !state.playlists.find((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
+      if (noQuickBookmarkSet) {
+        dispatch('updateQuickBookmarkTargetPlaylistId', payload._id, { root: true })
+      }
+
       commit('addPlaylist', payload)
     } catch (errMessage) {
       console.error(errMessage)
     }
   },
 
-  async addPlaylists({ commit }, payload) {
+  async addPlaylists({ state, commit, rootState, dispatch }, payload) {
     try {
       await DBPlaylistHandlers.create(payload)
+
+      const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !state.playlists.find((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
+      if (noQuickBookmarkSet) {
+        // TODO: use user's top sorting preference, or most recently used playlist.
+        // The sorting logic is currently not in an accessible utility class.
+        const chosenPlaylist = payload[0]
+        dispatch('updateQuickBookmarkTargetPlaylistId', chosenPlaylist._id, { root: true })
+      }
+
       commit('addPlaylists', payload)
     } catch (errMessage) {
       console.error(errMessage)
@@ -185,7 +202,7 @@ const actions = {
     }
   },
 
-  async grabAllPlaylists({ commit, dispatch, state }) {
+  async grabAllPlaylists({ rootState, commit, dispatch, state }) {
     try {
       const payload = (await DBPlaylistHandlers.find()).filter((e) => e != null)
       if (payload.length === 0) {
@@ -308,6 +325,15 @@ const actions = {
           }
         }
 
+        // if no quick bookmark is set, try to find another playlist
+        const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !payload.find((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
+        if (noQuickBookmarkSet && payload.length > 0) {
+          // TODO: use user's top sorting preference, or most recently used playlist.
+          // The sorting logic is currently not in an accessible utility class.
+          const chosenPlaylist = payload[0]
+          dispatch('updateQuickBookmarkTargetPlaylistId', chosenPlaylist._id, { root: true })
+        }
+
         commit('setAllPlaylists', payload)
       }
       commit('setPlaylistsReady', true)
@@ -334,18 +360,46 @@ const actions = {
     }
   },
 
-  async removePlaylist({ commit }, playlistId) {
+  async removePlaylist({ state, commit, rootState, dispatch }, playlistId) {
     try {
       await DBPlaylistHandlers.delete(playlistId)
+
+      const quickBookmarkPlaylistWasDeleted = playlistId === rootState.settings.quickBookmarkTargetPlaylistId
+      if (quickBookmarkPlaylistWasDeleted) {
+        // TODO: use user's top sorting preference, or most recently used playlist.
+        // The sorting logic is currently not in an accessible utility class.
+        const chosenPlaylist = state.playlists.find((playlist) => playlist._id !== playlistId)
+        if (chosenPlaylist) {
+          dispatch('updateQuickBookmarkTargetPlaylistId', chosenPlaylist._id, { root: true })
+          showToast(i18n.t('User Playlists.SinglePlaylistView.Toast["Playlist {playlistName} is the new quick bookmark playlist."]', {
+            playlistName: chosenPlaylist.playlistName
+          }))
+        }
+      }
+
       commit('removePlaylist', playlistId)
     } catch (errMessage) {
       console.error(errMessage)
     }
   },
 
-  async removePlaylists({ commit }, playlistIds) {
+  async removePlaylists({ state, commit, rootState, dispatch }, playlistIds) {
     try {
       await DBPlaylistHandlers.deleteMultiple(playlistIds)
+
+      const quickBookmarkPlaylistWasDeleted = playlistIds.indexOf(rootState.settings.quickBookmarkTargetPlaylistId) !== -1
+      if (quickBookmarkPlaylistWasDeleted) {
+        // TODO: use user's top sorting preference, or most recently used playlist.
+        // The sorting logic is currently not in an accessible utility class.
+        const chosenPlaylist = state.playlists.find((playlist) => playlistIds.indexOf(playlist._id) === -1)
+        if (chosenPlaylist) {
+          dispatch('updateQuickBookmarkTargetPlaylistId', chosenPlaylist._id, { root: true })
+          showToast(i18n.t('User Playlists.SinglePlaylistView.Toast["Playlist {playlistName} is the new quick bookmark playlist."]', {
+            playlistName: chosenPlaylist.playlistName
+          }))
+        }
+      }
+
       commit('removePlaylists', playlistIds)
     } catch (errMessage) {
       console.error(errMessage)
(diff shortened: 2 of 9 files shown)


PikachuEXE:
How about auto picking next quick bookmark target on delete but still allow removing quick bookmark target
It's not clear what deleting playlist would do now, but the same happens in this PR
Again I am fine with changing the default to lean for specific behaviour pattern, but not with removing existing abilities

jasonhenriquez:
> How about auto picking next quick bookmark target on delete but still allow removing quick bookmark target [...] Again I am fine with changing the default to lean for specific behaviour pattern, but not with removing existing abilities

I don't think this is a tenable standard if we're counting the ability to not have a quick bookmark playlist as a feature we are obligated to preserve by merit of it being included in a prior release. I don't see any cost to the user experience by removing this capability. The cost of keeping this capability is having people accidentally click it at some point and wonder why they can't add videos to playlists like they could before. There's no reasonable cost calculation I can come up with that justifies keeping this, but I'm open to hearing any. But I'm always going to value the choice that improves the UX if the only case against it is an arbitrary standard.

PikachuEXE:
Let me list out some issues from https://github.com/FreeTubeApp/FreeTube/issues/5051 and what I think how this PR solves it (or not)

### A. User clicks "Disable Quick Bookmark" icon (accidentally)

1. Too easy to click
2. not enough info about what it does

### B. User deletes their Favorites playlist at some point and uses a different one, not realizing that the Quick Bookmark will disappear until a new playlist is set as that one

- Not enough info about what "delete playlist" means now (which now also includes disabling quick bookmark button)

This PR tries to solve A1 & A2 by making it impossible to disable the feature. I agree that it's effective but I don't like the side effect. 
For B1 this PR only changes the action of disabling the quick bookmark target to auto setting **a random target without user intent**
So it's from `delete playlist -> cannot quick bookmark -> report bug` to `delete playlist -> potentially quick bookmark videos into an unintended playlist -> report bug`

There are other ways to solve these
- Show a prompt/page/whatever to allow user to quick bookmark when they try to quick bookmark but quick bookmark target unset (do nothing until user input received)
- Show more detailed messages before/after using disabling quick bookmark / deleting playlists
- Disable playlist deletion button with message until quick bookmark target changed
- Make it easier to see & set a quick bookmark target on user playlists view (maybe put quick bookmark on top)

jasonhenriquez:
I'm fine with also requiring modal confirmation for changing the quick bookmark; I can add that in if you'd prefer. But I'm still not seeing a justification for why we should allow the quick bookmark to be null if the user has a user playlist. And as to improving the auto-selection of the new playlist or making it manual, I'm open to it, but would prefer that to be in a separate PR, as I think this is a sufficient solution on its own.

jasonhenriquez:
> this PR's current state is just changing the unexpected result (disable quick bookmark -> change target to a random playlist if any).

That's incorrect; disabling quick bookmark is no longer an option for this PR. That's something that some users have already done, and we would at the very least have to do programmatic selection for these people because we don't want to prompt them on startup. Only other case that unexpected programmatic selection is done is when removing the current Quick Bookmark playlist, which is an edge case, and a toast informs them of the new playlist that is set as the new target (+ it shows up on the title of the actual control). 

I thin that it's unfair to say this is equivalent to the current state of users not being aware that the control is present, and it makes it clearer that they can go to the page for that playlist to configure it when seeing any instance of the Quick Bookmark icon. I don't want to build out a whole new selection prompt just for the "delete currently quick bookmarked playlist" case due to the new mitigations of the new toast (`Playlist {playlistName} is the new quick bookmark playlist`) and the icon with title of the playlist showing up on each video. But I do want to see what others think on this one. @MarmadileManteater @absidue @efb4f5ff-1298-471a-8973-3d47447115dc @ChunkyProgrammer

static/locales/en-US.yaml
```diff
@@ -209,6 +208,7 @@ User Playlists:
       There were no videos to remove.: There were no videos to remove.
       This playlist is protected and cannot be removed.: This playlist is protected and cannot be removed.
       Playlist {playlistName} has been deleted.: Playlist {playlistName} has been deleted.
+      Playlist {playlistName} is the new quick bookmark playlist.: Playlist {playlistName} is the new quick bookmark playlist.
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 211:
```suggestion
      Playlist {playlistName} is the new quick bookmark playlist.: Playlist {playlistName} is the new Quick Bookmark playlist.
```

efb4f5ff-1298-471a-8973-3d47447115dc:
IMO I think everything about this implementation is nicely done. 

I also dont see a reason why it should be `null`

We could indeed let the user select what Playlist they want to use for Quick Bookmark when the Quick Bookmark playlist is deleted as it could come over as random and would require "another input" from the user to go to the actual playlist they want to have as the Quick Bookmark one.

PikachuEXE:
> I also dont see a reason why it should be `null`

They can want to stop using a playlist as quick bookmark target (for converting it into playlist(s) with its content) (permanently or temporarily)
I don't see a reason NOT to allow user to defer that choice to a later point when we can

src/renderer/store/modules/playlists.js
```diff
@@ -79,15 +81,30 @@ const actions = {
 
     try {
       await DBPlaylistHandlers.create([payload])
+
+      const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !state.playlists.find((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
```
absidue, line 85: As you don't need the value and only need to know if it exists, you could use `Array#some`.

```suggestion
      const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !state.playlists.some((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
```

src/renderer/store/modules/playlists.js
```diff
@@ -334,18 +360,46 @@ const actions = {
...
       console.error(errMessage)
     }
   },
 
-  async removePlaylists({ commit }, playlistIds) {
+  async removePlaylists({ state, commit, rootState, dispatch }, playlistIds) {
     try {
       await DBPlaylistHandlers.deleteMultiple(playlistIds)
+
+      const quickBookmarkPlaylistWasDeleted = playlistIds.indexOf(rootState.settings.quickBookmarkTargetPlaylistId) !== -1
```
absidue, line 390:
```suggestion
      const quickBookmarkPlaylistWasDeleted = playlistIds.includes(rootState.settings.quickBookmarkTargetPlaylistId)
```

src/renderer/store/modules/playlists.js
```diff
@@ -79,15 +81,30 @@ const actions = {
...
       console.error(errMessage)
     }
   },
 
-  async addPlaylists({ commit }, payload) {
+  async addPlaylists({ state, commit, rootState, dispatch }, payload) {
     try {
       await DBPlaylistHandlers.create(payload)
+
+      const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !state.playlists.find((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
```
absidue, line 100:
```suggestion
      const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !state.playlists.some((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
```

src/renderer/store/modules/playlists.js
```diff
@@ -334,18 +360,46 @@ const actions = {
...
-  async removePlaylists({ commit }, playlistIds) {
+  async removePlaylists({ state, commit, rootState, dispatch }, playlistIds) {
     try {
       await DBPlaylistHandlers.deleteMultiple(playlistIds)
+
+      const quickBookmarkPlaylistWasDeleted = playlistIds.indexOf(rootState.settings.quickBookmarkTargetPlaylistId) !== -1
+      if (quickBookmarkPlaylistWasDeleted) {
+        // TODO: use user's top sorting preference, or most recently used playlist.
+        // The sorting logic is currently not in an accessible utility class.
+        const chosenPlaylist = state.playlists.find((playlist) => playlistIds.indexOf(playlist._id) === -1)
```
absidue, line 394:
```suggestion
        const chosenPlaylist = state.playlists.find((playlist) => !playlistIds.includes(playlist._id))
```

src/renderer/store/modules/playlists.js
```diff
@@ -308,6 +325,15 @@ const actions = {
           }
         }
 
+        // if no quick bookmark is set, try to find another playlist
+        const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !payload.find((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
```
absidue, line 329:
```suggestion
        const noQuickBookmarkSet = !rootState.settings.quickBookmarkTargetPlaylistId || !payload.some((playlist) => playlist._id === rootState.settings.quickBookmarkTargetPlaylistId)
```

src/renderer/store/modules/playlists.js
jasonhenriquez: Good call-out, not sure what I was thinking. I might've copy-pasted it from another context without thinking about it.

src/renderer/store/modules/playlists.js
jasonhenriquez: As to why I chose `indexOf` instead of `includes`, I found some research indicating that it's more performant on almost every browser, but I agree that it's right to change these because it does ultimately make for cleaner and more readable code.

efb4f5ff-1298-471a-8973-3d47447115dc:
Everything LGTM.

I have a few Q's before approval though. 

1. I do not quite understand why we have the bookmark button on 2 pages now. Its probably added for convenience on the main Playlist page but why would we keep it inside a Playlist or vice versa ?

2. Before this PR all the settings regarding to a playlist were stored inside a playlist but now one of them is also seen on the main playlist page. Couldnt that lead to confusion?

jasonhenriquez:
> I do not quite understand why we have the bookmark button on 2 pages now. Its probably added for convenience on the main Playlist page but why would we keep it inside a Playlist or vice versa ?

I think it makes sense to be able to set it inside of the playlist, because that's information & a control about the playlist grouped with other information about it. I think it's nice to be able to have the icon visible on the list item because it's useful information about a playlist. I don't have a strong opinion about whether it should still be a button on every playlist item, or instead just a disabled icon on the list item of the current Quick Bookmark target. @PikachuEXE If you have any justification for one or the other, please feel welcome to share.

PikachuEXE:
My use case for having buttons in user playlists view is to quickly changing a quick bookmark target, add a new videos maybe (from search result, channel video list, single remote playlist, whatever), and change it back to usual "watch later" playlist (in my case I don't have one I actually unset it). It's for saving clicks especially when there are a lot of playlists (search, click instead of search, click inside, click again, then go back to user playlists page) plus what's the point not making it a button when it's already an indicator. 

(Review comment about latest change to be given later)

PikachuEXE:
The target indicator style seems different from what I have in #5082 
Where the target would have an indicator which is different in icon AND color
image

> disable deleting a bookmarked playlist

I see nothing in regards about what should be done and/or why is it disabled to be able to delete that playlist
As a desktop user I at least expect to see title, but that doesn't work for mobile well so what I have done in #5082 is to show a toast while still allowing clicking on it (style is something else)
[`29a3ebd` (#5082)](https://github.com/FreeTubeApp/FreeTube/pull/5082/commits/29a3ebd05fe379a07190690de15af623962daae0)

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
This wording of the toast message throws me off a little. This playlist is **now** used for quick bookmark. 

The playlist was already being used as the quick bookmark one so including the **now** in there would mean that it wasnt before?


https://github.com/FreeTubeApp/FreeTube/assets/73130443/bb429641-ff91-46a5-8af6-189babe20bd0

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
I see Quick Bookmark and quick bookmark being used allot. Which one should be replaced to be consistent within the whole app?

efb4f5ff-1298-471a-8973-3d47447115dc:
Im not sure, either way we should still make it consistent.

jasonhenriquez:
Turns out it was actually just one label (pre-existing). The others are in caps because they're in title case (see: #5000).

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -366,7 +374,7 @@ export default defineComponent({
       if (this.selectedUserPlaylist.protected) {
         showToast(this.$t('User Playlists.SinglePlaylistView.Toast["This playlist is protected and cannot be removed."]'))
       } else {
-        this.removePlaylist(this.id)
+        this.removePlaylist(this.id, true)
```
PikachuEXE, line 377: What's the use of `true` here?

PikachuEXE (commented):
Everything seems fine except one line/question

src/renderer/components/playlist-info/playlist-info.js
jasonhenriquez: Good catch, this was vestigial

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #5068, opened 2024-05-05
Move local API player cache internals into the main process

absidue:
# Move local API player cache internals into the main process

## Pull Request Type

- [x] Security improvement

## Description
Currently the player cache implementation that we use in Electron, writes to the file system in the renderer process (the stuff that actually runs inside of the window), this pull request moves it to the main process with IPC calls. That allows us to add checks to ensure that the player cache can only be used to read and write files in the `player_cache` directory. It also avoids exposing where the `player_cache` directory is located on the file system.

As an added bonus this pull request shaves off 1591 bytes from the `renderer.js` file but only adds 240 bytes to the `main.js` file.

## Testing 
1. Open the dev data location (same as normal one but in the `Electron` folder instead of the `FreeTube` one)
2. Delete the `player_cache` folder
3. `yarn run dev`
4. Open a video with the local API
5. Check that the `player_cache` folder gets created and an item appears inside of it
6. Open another video
7. Check in the Network tab of the dev tools that only one request was made to `https://www.youtube.com/s/player/{identifier}/player_ias.vflset/en_US/base.js`. If there was only one, that means the second video used the cached player data.

After step 5 you've tested that adding an item to the cache works and after step 7 that reading from the cache works.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -4,8 +4,6 @@ const IpcChannels = {
   DISABLE_PROXY: 'disable-proxy',
   OPEN_EXTERNAL_LINK: 'open-external-link',
   GET_SYSTEM_LOCALE: 'get-system-locale',
-  GET_USER_DATA_PATH: 'get-user-data-path',
-  GET_USER_DATA_PATH_SYNC: 'get-user-data-path-sync',
   GET_PICTURES_PATH: 'get-pictures-path',
   SHOW_OPEN_DIALOG: 'show-open-dialog',
   SHOW_SAVE_DIALOG: 'show-save-dialog',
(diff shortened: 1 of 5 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
I cant seem to find the request in the Network tab, also tried filtering for it without any luck. Any suggestions how I can easily find the request?

absidue:
Don't copy the exact text the `{identifier}` part is the placeholder for whatever id the current player version has.

Filtering for `base.js` should work though.

efb4f5ff-1298-471a-8973-3d47447115dc:
Filtering by `base.js` gives me 0 results so its not working?

PikachuEXE:
Test result pass checking code later

1st window after playing 2 videos
image

2nd window after playing 3 videos
image

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Step 5: Opened first vid, confirmed that item appeared inside it. Also saw a base.js request that matched the item. 
Step 6: Opened second vid, saw another request to base.js. So that means that it didnt use the cached one

After that i opened a few other videos and no other base.js requests were made. 2nd video somehow not using the already cached request

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Approving as the id was different and thats why i saw a new base.js file

jasonhenriquez (approved):

Issue #5055, opened 2024-05-03
[Bug]: Have Wayland work without command line arguments


jasonhenriquez:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Run Wayland with FreeTube downloaded through `7z` / `linux_portable`
2. Encounter app crashing on load

### Expected Behavior

FreeTube works without external command line argument. This can be done by running the executable from the command line as such:
 ```./freetube --ozone-platform-hint=auto```

Expected behavior is that the app can run without that argument, or programmatically includes it.

### Issue Labels

causes crash

### FreeTube Version

v0.20

### Operating System Version

OpenSUSE TW

### Installation Method

Portable

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

v0.20.0

### Additional Information

Functionally, dependent on https://github.com/electron/electron/pull/35630 or a temporary reversion in the Electron version. The latter is highly undesirable. That most likely makes this so out of our control that it doesn't warrant an issue. But I'm opening up this issue to try to minimize duplicates of this being posted in the future and making it easier to find this information through a GitHub search.

So reminder, the fix for your personal use is either switching to a non-`7z`/`linux_portable` (e.g., `.deb`, `.rpm`, `.apk`) download or running FreeTube like this:
 ```./freetube --ozone-platform-hint=auto```

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

jasonhenriquez:
Update: this was fixed in [Electron 30.0.6.](https://github.com/FreeTubeApp/FreeTube/pull/5137#issuecomment-2119981995)

winstonma:
[...]

PR #5054, opened 2024-05-02
Fix startup problems for some Linux platforms for some installs

jasonhenriquez:
# Fix startup problems for some Linux platforms for some installs

## Pull Request Type

- [x] Bugfix

## Related issue
None; `Segmentation fault (core dumped)` occurring for some Linux users post-Electron 30 

## Description
Co-created with @absidue 
- Fixes downloads for Wayland (not sure if any others) users using  non-`7z`/`portable` Linux builds

## Testing 
- [Download a  non-`7z`/`portable` Linux build from here](https://github.com/jasonhenriquez/FreeTube/actions/runs/8930351772) and confirm the app opens & runs correctly

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
This does not work for `7z` / `portable` downloads, which is going to be a major problem for those users going forward. The commit in this PR can be reverted once https://github.com/electron/electron/pull/35630 is merged.

PR #5098, opened 2024-05-11
Revert Electron to 29.3.3

jasonhenriquez:
# Revert Electron to 29.3.3

## Pull Request Type

- [x] Other - Maintenance

## Related issue
#5054, #5055

## Description
Reverts Electron version to prevent issues for what seems to be a good deal of Linux users. Many distros need the argument `--ozone-platform-hint=auto` to be able to run correctly (e.g., most Linux systems running Wayland as the display system), some distros are blowing up when that argument is passed in (e.g., Ubuntu), and some installs can't have this argument applied programmatically when the default executable is run (e.g., Linux portable, Linux `.7z`). 

The immediate course of action that the FreeTube core contributor team discussed in our Matrix channel is to revert the Electron 30 upgrade, test the next major Electron versions to see if the issue is resolved, and if no fix is made, we will upgrade to the latest Electron version and figure out a solution.

Edit 2024-05-13: changed to 29.3.3 from prior 29.3.0 and revalidated.

## Testing
- (optional, if have an available means to do so) Test that Linux system runs [any given Linux build from here](https://github.com/jasonhenriquez/FreeTube/actions/runs/9068271952) properly.

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW


efb4f5ff-1298-471a-8973-3d47447115dc:
This should probably be the last PR to be merged before release

jasonhenriquez:
Closed in favor of #5137.

PR #5119, opened 2024-05-17
Display Subtitles/CC and 4k tag - #917

### LINKED ISSUE #917: Display Subtitles/CC tag
I think it would be nice if you add a tag Subtitles/CC if a video has them like YouTube website does.

Display Subtitles/CC tag if a video has them.

YouTube website displays Subtitles tag.

**Screenshots**
YouTube Subtitles

dkshxd:
# Title

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Closes #917

## Description
Implementation of the "4K" and "Subtitles" labels on video search results. See screenshots. 
Note: I've only seen this on the search screen so they haven't been implemented anywhere else.

## Screenshots 
4k-and-subtitles
No-labels

## Testing 
See screenshots for testing
1. Search of 4k and subtitles. Results show 4K and subtitles together
2. Search results including just 4K or subtitles separately
3. Search results including no 4K and subtitle labels
4. Rudimentary check of subscriptions, trending, most popular, history screens. Labels don't appear there. I don't see them on the Youtube website either.

## Desktop

- Windows
- 10
- v0.20.0 Beta

## Additional context
The implementation is flexible enough to support the addition of more labels such as "Live". I'll leave that for another ticket. Live shouldn't be that difficult, however someone will probably need to implement the icon. The alternative is a red label saying "Live" perhaps with a big text dot in front of it.

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.scss
+++ b/src/renderer/components/ft-list-video/ft-list-video.scss
@@ -8,3 +8,17 @@
   // Makes img element sized correctly before image loading starts
   aspect-ratio: 16/9 auto;
 }
+
+.videoTagLine {
+  margin-block-start: 3px;
+}
+
+.videoTag {
+  padding: 3px;
+  background-color: rgb(255 255 255 / 10%);
+  font-size: 10px;
+  border-radius: 2px;
+  display: inline-block;
+  font-weight: 500;
+  white-space-collapse: collapse;
+}
--- a/src/renderer/components/ft-list-video/ft-list-video.vue
+++ b/src/renderer/components/ft-list-video/ft-list-video.vue
@@ -145,6 +145,28 @@
           class="viewCount"
         > • {{ $tc('Global.Counts.Watching Count', viewCount, {count: parsedViewCount}) }}</span>
       </div>
+      <div
+        v-if="is4k || hasCaptions"
+        class="videoTagLine"
+      >
+        <div
+          v-if="is4k"
+          class="videoTag"
+          :aria-label="$t('Search Listing.Label.4K')"
+          role="img"
+        >
+          {{ $tc('Search Listing.Label.4K') }}
+        </div>
+        <div
+          v-if="hasCaptions"
+          class="videoTag"
+          :aria-label="$t('Search Listing.Label.Closed Captions')"
+          role="img"
+          :style="{marginLeft: is4k ? '4px' : '0px' }"
+        >
+          {{ $tc('Search Listing.Label.Subtitles') }}
+        </div>
+      </div>
       <ft-icon-button
         class="optionsButton"
         :icon="['fas', 'ellipsis-v']"
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -796,6 +796,8 @@ export function parseLocalListVideo(item) {
       lengthSeconds: isNaN(movie.duration.seconds) ? '' : movie.duration.seconds,
       liveNow: false,
       isUpcoming: false,
+      is4k: movie.is_4k,
+      hasCaptions: movie.has_captions
     }
   } else {
     /** @type {import('youtubei.js').YTNodes.Video} */
@@ -826,7 +828,9 @@ export function parseLocalListVideo(item) {
       lengthSeconds: isNaN(video.duration.seconds) ? '' : video.duration.seconds,
       liveNow: video.is_live,
       isUpcoming: video.is_upcoming || video.is_premiere,
-      premiereDate: video.upcoming
+      premiereDate: video.upcoming,
+      is4k: video.is_4k,
+      hasCaptions: video.has_captions
     }
   }
 }
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -101,6 +101,8 @@ export default defineComponent({
       watchProgress: 0,
       published: undefined,
       isLive: false,
+      is4k: false,
+      hasCaptions: false,
       isUpcoming: false,
       isPremium: false,
       hideViews: false,
@@ -648,6 +650,8 @@ export default defineComponent({
       this.description = this.data.description
       this.isLive = this.data.liveNow || this.data.lengthSeconds === 'undefined'
       this.isUpcoming = this.data.isUpcoming || this.data.premiere
+      this.is4k = this.data.is4k
+      this.hasCaptions = this.data.hasCaptions
       this.isPremium = this.data.premium || false
       this.viewCount = this.data.viewCount
 
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -61,6 +61,12 @@ Search / Go to URL: Search / Go to URL
 Search Bar:
   Clear Input: Clear Input
 Search character limit: Search query is over the {searchCharacterLimit} character limit
+Search Listing:
+  Label:
+    4K: 4K
+    Subtitles: Subtitles
+    # Aria labels
+    Closed Captions: Closed Captions
   # In Filter Button
 Search Filters:
   Search Filters: Search Filters


src/renderer/themes.css
```diff
@@ -95,6 +95,7 @@
   --side-nav-hover-color: #e0e0e0;
   --side-nav-active-color: #757575;
   --search-bar-color: #f5f5f5;
+  --search-list-tag-bg-color: rgb(242 242 242 / 100%);
```
jasonhenriquez, line 98: **suggestion:** Would `--secondary-card-bg-color` work for this use case? Would prefer not to add new theme variables if possible.

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -145,6 +145,28 @@
...
+          {{ $tc('Search Listing.Label.4K') }}
+        </div>
+        <div
+          v-if="hasCaptions"
+          class="videoTag"
+          :aria-label="$t('Search Listing.Label.Closed Captions')"
+          role="img"
+          :style="{marginLeft: is4k ? '4px' : '0px' }"
+        >
+          {{ $tc('Search Listing.Label.Subtitles') }}
```
jasonhenriquez, line 167:
```suggestion
          {{ $t('Search Listing.Label.Subtitles') }}
```

**suggestion (blocking):** No need for pluralization, so we can just use `$t`.

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -145,6 +145,28 @@
...
+        v-if="is4k || hasCaptions"
+        class="videoTagLine"
+      >
+        <div
+          v-if="is4k"
+          class="videoTag"
+          :aria-label="$t('Search Listing.Label.4K')"
+          role="img"
+        >
+          {{ $tc('Search Listing.Label.4K') }}
```
jasonhenriquez, line 158:
```suggestion
          {{ $t('Search Listing.Label.4K') }}
```

**suggestion (blocking):** No need for pluralization, so we can just use `$t`.

jasonhenriquez:
Thanks for putting in the initiative on this @dkshxd! Left some suggested changes and questions.

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -145,6 +145,28 @@
...
+        </div>
+        <div
+          v-if="hasCaptions"
+          class="videoTag"
+          :aria-label="$t('Search Listing.Label.Closed Captions')"
+          role="img"
+          :style="{marginLeft: is4k ? '4px' : '0px' }"
```
jasonhenriquez, line 165: **suggestion (blocking):** Please use dynamic classes to apply styling adjustments (`:class={ /* */ }`). This allows better expansion of the affected styles, better semantic clarification of the intention/purpose, and also better linting due to some limitations of our current plugins. Case in point, `margin-left` instead of `margin-inline-start` not being enforced in the template, and thus leading to unexpected behavior for users using FT in an RTL language (e.g., Arabic).

efb4f5ff-1298-471a-8973-3d47447115dc:
There is also a `8k` and a `new` label. Could you implement those also.

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -145,6 +145,28 @@
...
         > • {{ $tc('Global.Counts.Watching Count', viewCount, {count: parsedViewCount}) }}</span>
       </div>
+      <div
+        v-if="is4k || hasCaptions"
+        class="videoTagLine"
+      >
+        <div
+          v-if="is4k"
+          class="videoTag"
+          :aria-label="$t('Search Listing.Label.4K')"
```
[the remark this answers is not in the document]
dkshxd: I actually based these values off the Youtube website. I believe they used role "img" because it behaves like a image label. It's probably because they want it behave like this - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Img_role#using_roleimg_to_confer_meaning_that_is_obscured_or_implied

My guess about usage of the aria-label on a screen reader is that it will prefer to read the aria-label over the actual text inside. The label on the captions is more descriptive than the text so it made sense to me. 4k label I guess didn't need more elaboration.

src/renderer/components/ft-list-video/ft-list-video.vue
dkshxd: Done

src/renderer/themes.css
dkshxd: Side-by-side-3

I just had a go at --secondary-card-bg-color. It works fine with every theme apart from black and default. I've attached a screenshot. Middle is with custom background - Left and right are with --secondary-card-bg-color. I'll change it if you think it works for you.

src/renderer/components/ft-list-video/ft-list-video.vue
dkshxd: Done

src/renderer/components/ft-list-video/ft-list-video.vue
dkshxd: I've removed the conditional style scriptlet. I forgot about the CSS :first-child selector. The same thing is now implemented via CSS. 

The code exists to remove a weird indent on the video tags for the first element in the list.

src/renderer/themes.css
jasonhenriquez: That works for me!

src/renderer/components/ft-list-video/ft-list-video.vue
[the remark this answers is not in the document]
jasonhenriquez: Thanks, that makes sense.

absidue:
> The implementation is flexible enough to support the addition of more labels such as "Live". I'll leave that for another ticket. Live shouldn't be that difficult, however someone will probably need to implement the icon. The alternative is a red label saying "Live" perhaps with a big text dot in front of it.

FreeTube already labels live videos, so unless I am missing something, I doubt we'll need a second label to convey the same information. That being said, a quick look at the linked issue tells me that that paragraph was added due to a missunderstanding of an unrelated comment on that issue about support for live captions (they likely mean either Windows 11's live captions feature where the operating system displays auto-generated captions in a separate window on top of the programs window, so nothing we would have any control over or Google Chrome's closed source proprietary live captions feature, which we would have no way of implementing considering it's closed source and proprietary nature).

live-stream

src/renderer/themes.css
dkshxd: Ok it's on --secondary-card-bg-color now. The changes to themes.css have been backed out.

ChunkyProgrammer (approved):

dkshxd:
Agree with the "Live" label comment, it's already on the video thumbnail, so a duplication won't need to be implemented.

"8k" and "new", I don't see them being supported on the Youtube API we're using. I can see 8k on the JSON response but I suspect someone will have to implement it on the API, unless someone can point it out to me.

absidue:
> I don't see them being supported on the Youtube API we're using.

Do you mean the local API (so YouTube.js) or the Invidious API, because FreeTube supports both?

dkshxd:
> > I don't see them being supported on the Youtube API we're using.
> 
> Do you mean the local API (so YouTube.js) or the Invidious API, because FreeTube supports both?



> Do you mean the local API (so YouTube.js) or the Invidious API, because FreeTube supports both?

I'm looking at the local API using Youtubei.js. The movie and video objects don't seem have methods/attributes for 8k and new. It's there for live and 4k.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
I hope next time the branch name can be non `development`, can't checkout branch with `gh pr checkout 5119`

efb4f5ff-1298-471a-8973-3d47447115dc:
What error are you getting?

PikachuEXE:
Can only be checkout if latest dev merged into target branch, but no point making this happen since making a new branch to make changes (especially for opening PR) is a common practice already.
image

src/renderer/helpers/api/local.js
```diff
@@ -826,7 +828,9 @@ export function parseLocalListVideo(item) {
       lengthSeconds: isNaN(video.duration.seconds) ? '' : video.duration.seconds,
       liveNow: video.is_live,
       isUpcoming: video.is_upcoming || video.is_premiere,
-      premiereDate: video.upcoming
+      premiereDate: video.upcoming,
+      is4k: video.is_4k,
+      hasCaptions: video.has_captions
```
PikachuEXE, line 833: (Not line specific or even PR specific)
This is why I dislike missing trailing commas in general - the extra diff and more difficult to duplicate lines
Just a rant not request to change
image

src/renderer/components/ft-list-video/ft-list-video.scss
```diff
@@ -8,3 +8,22 @@
...
+
+.videoTag {
+  padding: 3px;
+  margin-inline: 2px;
+  background-color: var(--secondary-card-bg-color);
+  font-size: 10px;
+  border-radius: 2px;
+  display: inline-block;
+  font-weight: 500;
+  white-space-collapse: collapse;
```
PikachuEXE, line 24: Again not request to change but suggestion for future changes
IMO the rules are easier to read when grouped (and sorted is even better) like (comment is optional)

```css
.videoTag {
  /* Spacing/outward stuff */
  margin-inline: 2px;
  padding: 3px;

  /* Self */
  display: inline-block;
  background-color: var(--secondary-card-bg-color); /* Normally I group this with color but it's absent here */

  /* Content like text */
  font-size: 10px;
  font-weight: 500;
  white-space-collapse: collapse;
```

PikachuEXE (approved):

absidue:
> I hope next time the branch name can be non development, can't checkout branch with gh pr checkout 5119

@PikachuEXE You can add `--branch a-different-name` to the command to make it name the local branch differently.

https://cli.github.com/manual/gh_pr_checkout

PR #5120, opened 2024-05-17
Fix gap next to banner when Hide Side Bar Labels is enabled

absidue:
# Fix gap next to banner when Hide Side Bar Labels is enabled

## Pull Request Type
- [x] Bugfix

## Related issue
https://github.com/FreeTubeApp/FreeTube/projects/10#card-92520872

## Description
This pull request makes the `ft-refresh-widget` take up the correct amount of space when the `Hide Side Bar Labels` setting is enabled, as it has a gap on the left side at the moment.

## Screenshots 
Before:
before

After:
after

## Testing 
1. Turn on `Hide Side Bar Labels` in the `Theme Settings`
2. Open the subscription page
3. Check that the refresh banner is flush with the side bar

### Regression tests
- Check that it is still flush with the side bar when:
  - ... the side bar is expanded
  - ... when the `Hide Side Bar Labels` setting is turned off
- Check that it is flush with the side of the app when the screen is less or equal to 680px wide.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** b9d5778a4d9f5626a25796c5dc46f63dde03e983

## Additional context
Not sure why the CSS rules in this component are all namespaced with the `.floatingRefreshSection` class, when Vue's scoped CSS feature already takes care of only applying it to the right component for us (see `scoped` attribute on the `<style>` tag in the template file (`ft-refresh-widget.vue`).


jasonhenriquez:
I will note that this issue is affecting #4949 as well. If that PR is merged first, you just have to apply these changes to the mixin instead of the `ft-refresh-widget` to fix it in both places. If this PR is merged first, I'll have to remember to incorporate this fix into the mixin.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez (approved):

Issue #3142, opened 2023-01-31
[Feature Request]: Add 'Features' Search Filter


ChunkyProgrammer:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Invidious and Youtube.js both support searching on the features (Invidious is missing the documentation so I'll probably create a PR to add it)

The filters for videos/all types are:
![image](https://user-images.githubusercontent.com/78101139/215899605-cd42856d-48d6-4b8a-aa7d-a632be76dd5b.png)

The filters for movies are:
![image](https://user-images.githubusercontent.com/78101139/215900337-ec1d6b05-7ece-40e5-88db-0cb54a8a4e18.png)

And the filter should not work for channel/playlist types 

### Proposed Solution

Add the 'features' filter as  checkboxes in the search filter component

### Alternatives Considered

Don't add it

### Issue Labels

ease of use improvement, improvement to existing feature

PR #5125, opened 2024-05-18
Allow filtering searches by features

ChunkyProgrammer:
# Allow filtering searches by features

Opening as draft as I would like to do a bit more testing before this gets reviewed

## Pull Request Type
- [x] Feature Implementation

## Description
Youtube and Invidious both support filtering videos by features. This PR adds support for it to FreeTube

## Screenshots
image

## Testing 
- use search filters for searches with both invidious and local api

## Desktop
- **OS:** Linux Mint
- **OS Version:** 21.3


src/renderer/components/ft-checkbox-list/ft-checkbox-list.js
```diff
@@ -0,0 +1,61 @@
...
+  computed: {
+    inputName: function () {
+      const name = this.title.replace(' ', '')
+      return name.toLowerCase() + this.id
+    }
+  },
+  mounted: function () {
+    this.id = this._uid
+    this.selectedValues = this.initialValues
+  },
```
absidue, line 43:
```suggestion
  created: function () {
    this.id = this._uid
    this.selectedValues = this.initialValues
  },
```

src/renderer/components/ft-checkbox-list/ft-checkbox-list.vue
```diff
@@ -0,0 +1,33 @@
+<template>
+  <div class="pure-radiobutton filter">
```
absidue, line 2: Looks like the CSS file contains stuff for both radio buttons and checkboxes but you are using some radio button classes here, is that intentional? If yes, could you please remove the unused CSS from the CSS file.

ChunkyProgrammer: Oh yeah, I forgot to leave a comment about the css. This is the same css we use for the radio buttons css file. I wasn't sure if it'd be better to use the same css file for both or remove checkbox stuff from the radio button file and radio button stuff from the checkbox file. Do you think it'd be better to use the same file for both?

efb4f5ff-1298-471a-8973-3d47447115dc:
Closes your own issue right? 

https://github.com/FreeTubeApp/FreeTube/issues/3142

src/renderer/components/ft-checkbox-list/ft-checkbox-list.vue
absidue: I think it might be best to keep the CSS separated to avoid future confusion about why there is a lot of CSS in the file of a component that doesn't use a lot of it.

That being said the `ft-checkbox-list` Vue component seems to be using just `radiobutton` classes which is a bit confusing.

PikachuEXE (approved):

src/renderer/components/ft-search-filters/ft-search-filters.vue
```diff
@@ -48,6 +48,16 @@
...
           @change="updateDuration"
         />
+        <ft-checkbox-list
+          ref="featuresCheck"
+          :title="$t('Search Filters.Features.Features')"
+          :labels="featureLabels"
+          :values="featureValues"
+          :initial-values="searchDefaultFeatures"
+          class="searchRadio"
+          :is-checkbox="true"
```
absidue, line 58: This prop doesn't exist on the `ft-checkbox-list` component, so it can be removed.
```suggestion
```

absidue (approved):

jasonhenriquez (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer thanks for implementing this and making us on par with YT again in terms of search filters. I know this is already merged and should have reviewed this earlier but this PR is missing one thing IMO. 

Before this PR:
Search filters that could not be used with each other would reset to default state

With this PR:
Search filters do not reset if an incompatible Features search filter is applied

Issue #4982, opened 2024-04-19
[Feature Request]: show if video is saved (and maybe in what playlist)


MrHappy:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Now when I watch a video I cannot see if the video is already saved in a playlist (and in what playlist).

### Proposed Solution

In the previous version there was a yellow asterisk to indicate it was saved.
Something like that and (preferably the name of the playlist).

### Alternatives Considered

Saving it again and after that checking other playlist if I did not save it there already and remove duplicates.

### Issue Labels

display more information to user

jasonhenriquez:
Can you clarify what change is wanted and where? Like the "Add to Playlist" `+` icon having a number attached to it? Or some type of change on the prompt where you add the video to different playlists?

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @MrHappy,

> In the previous version there was a yellow asterisk to indicate it was saved.

This isnt gone but it could be that you havent enabled this yet.
Go into one of your playlists and click on the following button

![Capture](https://github.com/FreeTubeApp/FreeTube/assets/73130443/3e6cdfa4-a948-4b43-b060-60344862e0ea)

Now you will see a Clock icon instead of a Star icon on the places you are used to, when hovering on a thumbnail and under the video player.

When clicked on the icon it will save that video to the playlist you have enabled this feature for.


https://github.com/FreeTubeApp/FreeTube/assets/73130443/29917e89-94a8-482f-91bc-e49d42bc5ea4



We also indicate if a video is already added to a playlist


https://github.com/FreeTubeApp/FreeTube/assets/73130443/c4d8f383-5a5c-41d8-b5a8-278ed79411ca

Closing this issue as i think this has addressed your points.

toby63:
@efb4f5ff-1298-471a-8973-3d47447115dc I don't understand why you are against re-implementing a simple indicator to the video page?
I guess that many people see this as a regression.
[...]

PR #5141, opened 2024-05-21
Show no of playlist a video already added to on add playlist button

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/4982

## Description

Update add to playlist button to show no. of playlists a video already added to 

## Screenshots 

image
image
image
image

## Testing 

- Add some videos to playlist(s)
- Ensure counter shown & correct on add playlist button (in different places)
- Ensure other icon buttons style unchanged

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Just using default colors & position for preview
Feel free to suggest other styles

When reviewing `src/renderer/components/ft-icon-button/ft-icon-button.vue` maybe hide whitespace changes for easier review


src/renderer/components/ft-icon-button/ft-icon-button.vue
```diff
@@ -1,28 +1,48 @@
...
+        v-if="counterValue > 0"
+        class="iconCounter"
+        :class="{
+          [theme]: true,
+        }"
+        :style="{
+          fontSize: `${size + padding * 2}px`,
+        }"
+        counter
+        :value="counterValue"
```
jasonhenriquez, line 42: **issue (blocking):** There's currently no non-visual or semantic indication of the meaning, which makes it inaccessible to screen reader users, and somewhat confusing for people who don't intuitively grasp the meaning.

jasonhenriquez:
**suggestion (blocking):** I get why you would want it to be small, but I'm having trouble seeing it in your regular zoom images. If possible, I'd recommend that it's at least 14-16px.

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.scss
absidue, line 1: The Vue template for this component just uses a different component without changing it, so none of the CSS in this file is used, please remove the file.

PikachuEXE:
15px looks like this:
image

Seems way too big for me

12px?
image
image

src/renderer/components/ft-icon-button/ft-icon-button.vue
PikachuEXE: Would updating the title make it better? (good enough or not is something else)
image

jasonhenriquez: I think that's a safe way to do it. That label seems pretty good too, albeit maybe a bit too long. Maybe `Add to Playlist (Aleady Added to 2)`?

jasonhenriquez:
Feedback:

> 12px?

I think you're right. It commands too much attention at a larger size.

1.  It's still on the right side for RTL languages:

Screenshot_20240522_110831

2. At a bigger font-size, and its position on the right side, it pops out too much now, in that you can't tell if it's popping out of the icon to the right of it.

Screenshot_20240522_110728

Interestingly, this problem is not apparent with the current RTL issue. Therefore, maybe a good fix for 2 is just moving it to the `inline-start` side.

3. This might sound minor, but there's some redundancy to it in certain places that feels subtly irksome:
  i. You're on the Playlist page for the one playlist that has it, but you see the 1 on the icon. This is not conveying any new information.
  ii. You have it quick bookmarked on any route, and that's the only playlist containing that video, but you see the 1 on the icon. This is not conveying any new information.
  iii. Intersection of cases i and ii and the Playlist page.

Again, this might sound minor, but it's presenting the same information twice, that adds unnecessary visual scanning and processing time to understand what is being logically conveyed. It also makes this feature unnecessarily prominent for the more common use case of someone who uses their Quick Bookmark playlist for the vast majority of their videos. I'd strongly recommend implementing carveouts for the above three cases.

PikachuEXE:
RTL now looks like:
image

Reduced `translate` value and list view look like this: (If I guess wrong on what `At a bigger font-size` means please let me know how to reproduce)
image

For (3)
- The original issue saying there is no indication of a video is saved in any playlist. I agree that there is no need for videos in user playlists to solve that issue. (To be done later)
- I can make the counter disappeared when it's **only** saved in quick bookmark playlist. But I think there might be people in the issue wants to see the count if the video is saved in more than 1 playlists (quick bookmark playlist included or not). It's a proxy for "what playlists is the video saved in" request (whether it should be addressed is another thing).
- Not considering (iii) yet until we agree what changes to be made for (i) * (ii) and implemented

jasonhenriquez:
Thanks so much @PikachuEXE! Could you switch the default side to be `inline-start` (or ya know what I mean, the logical equivalent) for the icons?

> I can make the counter disappeared when it's only saved in quick bookmark playlist.

Yes, to clarify, that's what I intended with ii. If you're removing it from the User Playlist view altogether, that makes i and iii no longer applicable.

PikachuEXE:
I am not entirely sure if you mean moving from right to left, but I made the change anyway
Fix me if I guess wrong
image
image

Removed counter in user playlist only (still present in remote playlist)
image

src/renderer/components/watch-video-info/watch-video-info.js
```diff
@@ -4,15 +4,22 @@ import FtCard from '../ft-card/ft-card.vue'
 import FtIconButton from '../ft-icon-button/ft-icon-button.vue'
 import FtShareButton from '../ft-share-button/ft-share-button.vue'
 import FtSubscribeButton from '../ft-subscribe-button/ft-subscribe-button.vue'
-import { formatNumber, openExternalLink, showToast } from '../../helpers/utils'
+import FtPlaylistAddVideoButton from '../ft-playlist-add-video-button/ft-playlist-add-video-button.vue'
+import {
+  formatNumber,
+  openExternalLink,
+  openInternalPath,
```
jasonhenriquez, line 11: **chore:** linting issue

jasonhenriquez (commented):
This looks great! I was initially skeptical if this requested feature could be implemented well, but the way you've implemented it solves the problem while also not overwhelming the user with information, or affecting other use cases negatively.

Fix the minor linting issue below and I'll LGTM.

PikachuEXE:
@absidue 
I am aware of the performance issue when there are many playlists with many videos
I doubt a solution can be developed before next release
So maybe we can disable it in places that can take a performance hit
And/or disable this only if no. of playlist/saved videos is above some number

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,125 @@
...
+        playlistCount: this.videoAddedToPlaylistCount,
+      })
+    },
+
+    quickBookmarkPlaylistId() {
+      return this.$store.getters.getQuickBookmarkTargetPlaylistId
+    },
+    quickBookmarkPlaylist() {
+      return this.$store.getters.getPlaylist(this.quickBookmarkPlaylistId)
+    },
```
absidue, line 95: Once this is rebased or the development branch is merged in you can apply this suggestion (https://github.com/FreeTubeApp/FreeTube/pull/5169).

```suggestion
    quickBookmarkPlaylist() {
      return this.$store.getters.getQuickBookmarkPlaylist
    },
```

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,125 @@
...
+      if (!this.showVideoAddedToPlaylistCount) { return 0 }
+
+      let count = 0
+
+      this.allPlaylists.forEach((playlist) => {
+        const videoAlreadyAdded = playlist.videos.some((v) => {
+          return v.videoId === this.videoId
+        })
+        if (videoAlreadyAdded) { count += 1 }
+      })
```
absidue, line 73: Same optimisation as in https://github.com/FreeTubeApp/FreeTube/pull/5170
```suggestion
      const videoId = this.videoId
      let count = 0

      this.allPlaylists.forEach((playlist) => {
        const videoAlreadyAdded = playlist.videos.some((v) => {
          return v.videoId === videoId
        })
        if (videoAlreadyAdded) { count += 1 }
      })
```

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,125 @@
...
+    },
+    isQuickBookmarkEnabled() {
+      return this.quickBookmarkPlaylist != null
+    },
+    isInQuickBookmarkPlaylist: function () {
+      if (!this.isQuickBookmarkEnabled) { return false }
+
+      return this.quickBookmarkPlaylist.videos.some((video) => {
+        return video.videoId === this.videoId
+      })
```
absidue, line 104: Same optimisation as in https://github.com/FreeTubeApp/FreeTube/pull/5170
```suggestion
      const videoId = this.videoId

      return this.quickBookmarkPlaylist.videos.some((video) => {
        return video.videoId === videoId
      })
```

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,125 @@
...
+    allPlaylists: function () {
+      return this.$store.getters.getAllPlaylists
+    },
+
+    title() {
+      const initialText = this.$t('User Playlists.Add to Playlist')
+      if (this.videoAddedToPlaylistCountText == null) { return initialText }
+
+      return `${this.$t('User Playlists.Add to Playlist')} (${this.videoAddedToPlaylistCountText})`
+    },
```
absidue, line 61: We don't need to translate that string unless we use it.
```suggestion
    title() {
      if (this.videoAddedToPlaylistCountText == null) {
        return this.$t('User Playlists.Add to Playlist')
      }

      return `${this.$t('User Playlists.Add to Playlist')} (${this.videoAddedToPlaylistCountText})`
    },
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -729,30 +736,6 @@ export default defineComponent({
...
-        authorId: this.channelId,
-        description: this.description,
-        viewCount: this.viewCount,
-        lengthSeconds: this.data.lengthSeconds,
-      }
-
-      this.showAddToPlaylistPromptForManyVideos({ videos: [videoData] })
-
-      // Focus when prompt closed
-      this.addToPlaylistPromptCloseCallback = () => {
```
absidue, line 746: Looks like there are still some watchers and an entry in `data` for this property, can those be removed too?

absidue (commented):
Here are a few suggestions to improve the performance, you will want to merge in development or rebase this branch on development, first though.

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
PikachuEXE: Oh I actually meant to use `initialText` on later return result too
Please see updated diff

src/renderer/components/ft-icon-button/ft-icon-button.vue
PikachuEXE: Updated

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,135 @@
...
+    size: {
+      type: Number,
+      default: 20
+    },
+
+    showVideoAddedToPlaylistCount: {
+      type: Boolean,
+      default: true,
+    },
```
absidue, line 49:
```suggestion
    },
    isInQuickBookmarkPlaylist: {
      type: Boolean
      required: true
    }
```

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,135 @@
...
+      // however as we know that some users have playlists that have more than 10k items in them
+      // it adds up quickly, especially as there are usually lots of ft-list-video instances active at the same time.
+      // So create a temporary variable outside of the array, so we only have to do it once.
+      // Also the search is re-triggered every time any playlist is modified.
+      const videoId = this.videoId
+
+      return this.quickBookmarkPlaylist.videos.some((video) => {
+        return video.videoId === videoId
+      })
+    },
```
absidue, line 115:
```suggestion
```

src/renderer/components/watch-video-info/watch-video-info.vue
```diff
@@ -79,13 +79,14 @@
...
-          :icon="['fas', 'plus']"
-          class="option"
-          theme="base"
-          @click="togglePlaylistPrompt"
+          :video-id="id"
+          :video-title="title"
+          :channel-id="channelId"
+          :channel-name="channelName"
+          :view-count="viewCount"
+          :length-seconds="lengthSeconds"
```
absidue, line 89:
```suggestion
          :length-seconds="lengthSeconds"
          :is-in-quick-bookmark-playlist="isInQuickBookmarkPlaylist"
```

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -44,16 +44,19 @@
...
+          :video-id="id"
+          :video-title="title"
+          :channel-id="channelId"
+          :channel-name="channelName"
+          :view-count="viewCount ?? null"
+          :length-seconds="lengthSeconds"
           :padding="appearance === `watchPlaylistItem` ? 5 : 6"
           :size="appearance === `watchPlaylistItem` ? 14 : 18"
-          @click="togglePlaylistPrompt"
+          :show-video-added-to-playlist-count="showVideoAddedToPlaylistCount"
```
absidue, line 59:
```suggestion
          :show-video-added-to-playlist-count="showVideoAddedToPlaylistCount"
          :is-in-quick-bookmark-playlist="isInQuickBookmarkPlaylist"
```

absidue (commented):
As we already check if the video is in the quick bookmark playlist inside the parent component, we can just pass down the result, instead of needing to perform the check again inside this component

src/renderer/components/ft-playlist-add-video-button/ft-playlist-add-video-button.js
```diff
@@ -0,0 +1,135 @@
...
+        const videoAlreadyAdded = playlist.videos.some((v) => {
+          return v.videoId === videoId
+        })
+        if (videoAlreadyAdded) { count += 1 }
+      })
+
+      // If only saved in quick bookmark target, don't show the count which is confusing
+      if (count === 1 && this.isInQuickBookmarkPlaylist) {
+        return 0
+      }
```
absidue, line 84: We don't need to check the quick bookmark playlist again, as we already know if it contains the video thanks to the `isInQuickBookmarkPlaylist` prop.
```suggestion
      const isQuickBookmarkEnabled = this.isQuickBookmarkEnabled
      let quickBookmarkPlaylistId

      if (isQuickBookmarkEnabled) {
        quickBookmarkPlaylistId = this.quickBookmarkPlaylist._id
      }

      this.allPlaylists.forEach((playlist) => {
        // We don't need to check the quick bookmark playlist
        // as we already know it from isInQuickBookmarkPlaylist
        if (isQuickBookmarkEnabled && quickBookmarkPlaylistId === playlist._id) {
          return
        }

        const videoAlreadyAdded = playlist.videos.some((v) => {
          return v.videoId === videoId
        })
        if (videoAlreadyAdded) { count++ }
      })

      // If only saved in quick bookmark target, don't show the count which is confusing
      if (count > 0 && this.isInQuickBookmarkPlaylist) {
        count++
      }
```

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
The only place im missing this number on a thumbnail is inside of a playlist

PikachuEXE:
It done that way intentionally
See original issue and comment in this PR https://github.com/FreeTubeApp/FreeTube/pull/5141#issuecomment-2125245573 (and a few comments afterward)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez:
I still have concerns about the performance impact / energy usage. I'd prefer to wait on this until after the Vue 3 migration where we can use Maps / Sets, as it currently doesn't pass my personal cost versus benefit evaluation. I won't block it if the rest of the team feels differently, though.

PikachuEXE:
We can disable this in any video collection view (i.e. only on watch page?
And only update implementation & enable it elsewhere after vue 3 migration (which takes no idea how long

jasonhenriquez:
I think the tough thing is that it's not very useful of a feature if people don't know when/where it will appear, and probably more confusing than not in terms of what we're training the users to expect. I know it sucks to suggest putting a PR that you put a good deal of work into stasis for so long, so I'll just leave it to the rest of the team on how we should handle it.

PikachuEXE:
It's less urgent for me as https://github.com/FreeTubeApp/FreeTube/pull/5044 is merged (I simply open it, found out it's already added, close it)
The other way around is to add a setting to only show it when enabled with notes about performance implication (but not sure how this affects performance in vue 3)

jasonhenriquez:
I'd be open to having it in experimental settings, but I'd wait to see what others think before going forward with that

PikachuEXE:
Will revisit after Vue 3 migration done

PR #5150, opened 2024-05-22
Use playlist name for playlist page title

jasonhenriquez:
# Use playlist name for playlist page title

## Pull Request Type

- [x] Feature Implementation

## Related issue
https://github.com/FreeTubeApp/FreeTube/pull/4949#issuecomment-2123750685

## Description
Sets document title to playlist title. This is consistent with how we do it for other routes with custom named content, YT, and user expectations.

## Testing 
- Test document title is updated to the playlist title for a given user playlist or remote playlist (either backend)

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -82,7 +82,7 @@ export default defineComponent({
     },
     windowTitle: function () {
       const routePath = this.$route.path
-      if (!routePath.startsWith('/channel/') && !routePath.startsWith('/watch/') && !routePath.startsWith('/hashtag/')) {
+      if (!routePath.startsWith('/channel/') && !routePath.startsWith('/watch/') && !routePath.startsWith('/hashtag/') && !routePath.startsWith('/playlist/')) {
         let title = translateWindowTitle(this.$route.meta.title, this.$i18n)
         if (!title) {
           title = packageDetails.productName
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -21,6 +21,7 @@ import {
   showToast,
 } from '../../helpers/utils'
 import { invidiousGetPlaylistInfo, youtubeImageUrlToInvidious } from '../../helpers/api/invidious'
+import packageDetails from '../../../../package.json'
 
 const SORT_BY_VALUES = {
   DateAddedNewest: 'date_added_descending',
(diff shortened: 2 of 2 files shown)


ChunkyProgrammer (approved):

PikachuEXE:
Can we update text for `User Playlists.Your Playlists` in this PR too? Got inconsistent casing
image

Also we should prevent long titles like existing logic
https://github.com/FreeTubeApp/FreeTube/blob/v0.20.0-beta/src/renderer/components/ft-list-playlist/ft-list-playlist.js#L55-L60

```js
    titleForDisplay: function () {
      if (typeof this.title !== 'string') { return '' }
      if (this.title.length <= 255) { return this.title }

      return `${this.title.substring(0, 255)}...`
    },
```

jasonhenriquez:
> Can we update text for User Playlists.Your Playlists in this PR too? Got inconsistent casing

This one isn't as straightforward. "Your playlists" conveys the playlists you own. "Your Playlists" conveys the section containing the playlists you own.

> Also we should prevent long titles like existing logic

How about constraining the user playlist name length (video title length is already restricted by YT)? Can do this in a separate PR if desired.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE:
> This one isn't as straightforward. "Your playlists" conveys the playlists you own. "Your Playlists" conveys the section containing the playlists you own.

No I only mean the one in `en_GB` is `Your playlists` but in `en_US` is `Your Playlists`
I forgot to ensure the screenshot to include the files names

> How about constraining the user playlist name length (video title length is already restricted by YT)? Can do this in a separate PR if desired.

Please do

jasonhenriquez:
Oh, I didn't catch that. Yeah, `English (UK)` has a lot of pretty arbitrary differences from `English (US)`. Might honestly be worth starting the UK file from the US one and manually scanning for place where the words need to be changed.

absidue:
@PikachuEXE @jasonhenriquez That string is correct.

Using title case is an American English thing, British English uses sentence case a lot more.
The real problem is actually a different one, the string being discussed here is correct but other ones in `English (UK)` that use title case, should be corrected to sentence case (presumably they were just copied from `English (US)`).

PR #4992, opened 2024-04-22
Show error message when search is over the Google search limit of 100 characters

dkshxd:
Show error message when search is over the Google search limit of 100 characters

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Closes https://github.com/FreeTubeApp/FreeTube/issues/940

## Description
Resubmitted: As I messed up the review changes on previous pull request and Github didn't like it.

Show error message when search is over the Google search limit of 100 characters. Also logs the error to console. The event handler on the search box is suppressed as well when over 100 characters as typing in the box can also trigger the exception screen as well (the error message still appears when the search is executed via enter key or mouse click).

## Screenshots 
322418591-29af3a9b-e90b-43b3-9997-937b0d35a1de

## Testing 

* Has been tested
* Tested cases

1. Normal search - less 100 characters
2. Search over 100 characters
3. Trigger the event handler on the search box by typing inside search box when under and over 100 characters
4. Empty search

## Desktop

- **OS: Windows**
- **OS Version: 10**
- **FreeTube version: 0.20 Beta**

PR #5156, opened 2024-05-22
Set a max length of 100 for most inputs

jasonhenriquez:
Excluded are playlist descriptions, which presumably deserve to have some leeway, and the password input, to prevent the edge case of a user being locked out if they had a longer one set (and also why not for this one).

# Set a max length of 100 for most inputs

## Pull Request Type

- [x] Feature Implementation

## Related issue
https://github.com/FreeTubeApp/FreeTube/pull/5150#issuecomment-2124043618

## Description
There's just no need for that much information in a custom label, because beyond 100 (and before, in many cases), we're already limiting the visible text. 

Excluded are playlist descriptions, which presumably deserve to have some leeway, and the password input, to prevent the edge case of a user being locked out if they had a longer one set (and also why not for this one)

**Note:** Due to the value of 100 chosen, [this code](https://github.com/FreeTubeApp/FreeTube/pull/4992/files) is currently not in use. We can 1) keep this code here in case that number ever changes, or 2) remove it and just remember to revert a commit. I don't have a preference.

## Testing 
- Test that inputs are limited from continued text at 100
- (Optional) Test edge case behavior when a playlist title is already >100 characters and the title is edited. Expected outcome: you can't add new characters, but you can remove them, and save at any length.

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
The User Playlist desktop list view breaks for long enough titles and descriptions. We probably need code truncating those after x lines.

### DIFF
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -23,6 +23,10 @@ export default defineComponent({
       type: String,
       default: null
     },
+    maxlength: {
+      type: Number,
+      default: 100
+    },
     value: {
       type: String,
       default: ''
--- a/src/renderer/components/ft-input/ft-input.vue
+++ b/src/renderer/components/ft-input/ft-input.vue
@@ -45,6 +45,7 @@
         :value="inputDataDisplayed"
         :list="idDataList"
         class="ft-input"
+        :maxlength="maxlength"
         :type="inputType"
         :placeholder="placeholder"
         :disabled="disabled"
--- a/src/renderer/components/password-dialog/password-dialog.vue
+++ b/src/renderer/components/password-dialog/password-dialog.vue
@@ -8,6 +8,7 @@
       ref="password"
       :placeholder="$t('Settings.Password Dialog.Password')"
       :show-action-button="false"
+      :maxlength="null"
       input-type="password"
       class="passwordInput"
       @input="handlePasswordInput"
--- a/src/renderer/components/password-settings/password-settings.vue
+++ b/src/renderer/components/password-settings/password-settings.vue
@@ -20,6 +20,7 @@
         :show-action-button="false"
         :show-label="true"
         input-type="password"
+        :maxlength="null"
         :value="password"
         @input="e => password = e"
         @keydown.enter.native="handleSetPassword"
--- a/src/renderer/components/playlist-info/playlist-info.vue
+++ b/src/renderer/components/playlist-info/playlist-info.vue
@@ -71,6 +71,7 @@
       :show-action-button="false"
       :show-label="false"
       :value="newDescription"
+      :maxlength="null"
       @input="(input) => newDescription = input"
       @keydown.enter.native="savePlaylistInfo"
     />


absidue:
Please make the max length opt-in, that means the default value is null and then you pass in the limit in the few places it is needed, making it opt-out is going to cause more problems than it will solve. For example you currently set a 100 character limit on the search bar, with the reasoning that it makes the search query limit code obsolete, however that will break the second use of the search bar: handling URLs. The URLs can easily go over 100 characters, especially if they come from a tor Invidious instance, see this comment for more information: https://github.com/FreeTubeApp/FreeTube/pull/4288#issuecomment-1800490629

jasonhenriquez:
Thanks for this context @absidue, I wasn't aware of that. Will update.

jasonhenriquez:
Updated now to remove the constraints from places where it would restrict valid inputs. Kept the default value the same so we don't have to have a magic number, or a constant we'd have to import into each file.

PikachuEXE:
> Please make the max length opt-in, that means the default value is null and then you pass in the limit in the few places it is needed, making it opt-out is going to cause more problems than it will solve.

I don't think this is solved
And also now I have no idea what to test (coz the diff doesn't show the list of inputs

jasonhenriquez:
> I don't think this is solved

See answer:
> Kept the default value the same so we don't have to have a magic number, or a constant we'd have to import into each file.

>  And also now I have no idea what to test (coz the diff doesn't show the list of inputs

Thanks, just updated the OP with the list of non-affected inputs.

PikachuEXE (changes requested):
- Playlist Name should be 255 at least (or unlimited)
- Playlist Description should be 255 at least (or unlimited)
- Proxy Port Number should be 5

Also in https://github.com/FreeTubeApp/FreeTube/pull/5150#issuecomment-2124043618 I simply means limiting the displayed value not input value

src/renderer/components/ft-input/ft-input.js
```diff
@@ -23,6 +23,10 @@ export default defineComponent({
       type: String,
       default: null
     },
+    maxlength: {
+      type: Number,
+      default: 100
```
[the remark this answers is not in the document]
jasonhenriquez: 100 is just a number I chose that's arbitrarily high but minimizes UI breaking / misleading the user as to how much of their input will realistically be displayed when applicable. If anything, it's probably a bit too high, but I think it's fine. I don't think there's any reason to have different individualized values for each input.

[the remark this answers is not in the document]
PikachuEXE: Places with 100 char limit: (13)
- src/renderer/views/UserPlaylists/UserPlaylists.vue - searchBar
- src/renderer/views/SubscribedChannels/SubscribedChannels.vue - searchBar
- src/renderer/views/History/History.vue - searchBar
- src/renderer/views/Channel/Channel.vue - searchBar
- src/renderer/components/proxy-settings/proxy-settings.vue - proxyPort
- src/renderer/components/playlist-info/playlist-info.vue - Playlist Name
- src/renderer/components/playlist-info/playlist-info.vue - Playlist Description
- src/renderer/components/playlist-info/playlist-info.vue - Search for Videos
- src/renderer/components/ft-profile-edit/ft-profile-edit.vue - Profile Name
- src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue - searchBar
- src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.vue - Playlist Name
- src/renderer/components/distraction-settings/distraction-settings.vue - Hide Channels
- src/renderer/components/distraction-settings/distraction-settings.vue - Hide Videos and Playlists Containing Text

Places with unlimited length: 12 (`:maxlength="null"`)


Youtube playlist name now has 150 char limit, description has 5000 char limit:
image

image


Channel name now has 100 char limit:
https://www.reddit.com/r/JackSucksAtLife/comments/17e8h08/since_when_was_the_youtube_channel_name_limit/
image


This means these places should have 255 chars limit due to Playlist Name + Channel Name = 250 + 5 char for whatever purpose like delimiter:
- src/renderer/views/UserPlaylists/UserPlaylists.vue - searchBar
- src/renderer/components/playlist-info/playlist-info.vue - Playlist Name
- src/renderer/components/playlist-info/playlist-info.vue - Playlist Description (not saying it should match Youtube's 5000 char limit but the limit should be at least the same as playlist name or larger)
- src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue - searchBar
- src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.vue - Playlist Name

Other places:
- src/renderer/components/distraction-settings/distraction-settings.vue - Hide Videos and Playlists Containing Text (150 due to playlist anme)
- src/renderer/components/proxy-settings/proxy-settings.vue - proxyPort (max value 65535 = 5 chars)


And now there are 6 places left with 100 char limit while 12 places with no limit and I have no idea why there is a default value of 100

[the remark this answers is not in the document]
jasonhenriquez: I'm tired and having a bit of trouble following. What's the rationale for most of the suggested changes beyond conforming to YT? I did mean to remove the limit on the playlist description, but I must have accidentally removed it, so I can do that. My general idea is that we can use one arbitrary not-too-large value for most custom name-like fields and unbounded for ones where we don't have a reason to care.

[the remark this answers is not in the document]
jasonhenriquez: Re-reading this, I come off as unnecessarily callous. My main point is that I'm willing to spend at most ten minutes on further changes for this PR, because I do think we (and developers in the FOSS community as a whole) have a tendency to bikeshed about obscure minutia in a way that can be quite counterproductive. I worry that if I were to implement your suggested changes in full, I'm going to get a review from another person saying to do it all another way, and the process just gives me the heebie jeebies sometimes. So if you are confident that you have a set of changes that you think I can do pretty quickly that are also unlikely to be opposed by any other developer on our team, I'm game. Just give me the sign.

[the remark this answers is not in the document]
PikachuEXE: Just for changes to be made: change char limit to 255 in following places:
- src/renderer/views/UserPlaylists/UserPlaylists.vue - searchBar
- src/renderer/components/playlist-info/playlist-info.vue - Playlist Name
- src/renderer/components/ft-playlist-add-video-prompt/ft-playlist-add-video-prompt.vue - searchBar
- src/renderer/components/ft-create-playlist-prompt/ft-create-playlist-prompt.vue - Playlist Name

Reasons: 
- Remote playlist can have 150 chars
- Channel name can have 100 chars (and some people might want to add them)
- Existing display logic truncate string to 255 chars

PikachuEXE (approved):

absidue:
I still think it would be better to get rid of the implicit limit of 100 and instead set explicit limits where they are actually required. Your argument was that we would need a magic constant imported into multiple files, but I would argue that we don't need a constant at all, just write the number where it is needed. Secondly that the current situation is worse, because it's still a magic number, but you don't see it at the call/usage site, so you won't even know it exists unless you stumble upon it in testing or remember to look at the `ft-input` component. It also means we won't have to override it with `null` in so many places anymore.

src/renderer/components/download-settings/download-settings.vue
```diff
@@ -34,6 +34,7 @@
             :show-action-button="false"
             :show-label="false"
             :disabled="true"
+            :maxlength="100"
```
absidue, line 37: As this displays the users chosen downloads folder, the path could much longer than 100 characters.
```suggestion
```

src/renderer/components/player-settings/player-settings.vue
```diff
@@ -247,6 +247,7 @@
           :spellcheck="false"
           :show-action-button="false"
           :show-label="false"
+
```
absidue, line 250: Extra whitespace.
```suggestion
```

src/renderer/components/proxy-settings/proxy-settings.vue
```diff
@@ -29,6 +29,7 @@
           :show-action-button="false"
           :show-label="true"
           :value="proxyHostname"
+
```
absidue, line 32: Ditto
```suggestion
```

src/renderer/components/sponsor-block-settings/sponsor-block-settings.vue
```diff
@@ -40,6 +40,7 @@
           :show-action-button="false"
           :show-label="true"
           :value="sponsorBlockUrl"
+
```
absidue, line 43: Ditto
```suggestion
```

src/renderer/components/sponsor-block-settings/sponsor-block-settings.vue
```diff
@@ -52,6 +53,7 @@
           :show-action-button="false"
           :show-label="true"
           :value="deArrowThumbnailGeneratorUrl"
+
```
absidue, line 56: Ditto
```suggestion
```

src/renderer/components/top-nav/top-nav.vue
```diff
@@ -87,6 +87,7 @@
           :data-list="searchSuggestionsDataList"
           :spellcheck="false"
           :show-clear-text-button="true"
+
```
absidue, line 90: Ditto
```suggestion
```

absidue:
Other than that it looks good now, thank you for doing the changes.

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5158, opened 2024-05-22
Play next/prev video after removing current video from the playlist

### LINKED ISSUE #4918: [Feature Request]: Play next video after removing current video from the playlist
### Guidelines

### Problem Description

1. Enable autoplay
2. Open the non-first video of a playlist
3. Remove that video from the playlist while watching it
4. Let video autoplay. Result: first video in the playlist plays

### Proposed Solution

1. Enable autoplay
2. Open the non-first video of a playlist
3. Remove that video from the playlist while watching it
4. Let video autoplay. Result: video that would have been next in the playlist if the current video was not removed from the playlist plays

### Alternatives Considered

N/A

### Issue Labels

### Additional Information

_No response_

jasonhenriquez:
# Play next/prev video after removing current video from the playlist

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4918

## Description
When you delete the video that is shown as being currently watched in `watch-video-playlist`, the video index will now visually appear to be the video prior to that one. if you delete that one, it carries on the video prior, and so forth. When the video autoplays or you click `Play Next Video` in this state, it plays the video that is shown as the one after the current video. When you click `Play Previous Video` in this state, it plays the video that is shown as the "current" video. These are all existing behaviors of YT.

Note that if you were to navigate back to the previous route for a deleted playlist entry through the `Back` button, the current video index will be set to zero, as this feature is reset when you navigate to a new route. This pre-existing behavior is fine, and making this more "durable" is not worth the effort. Believe me, I've tried.

## Screenshots 
Screenshot_20240522_150833

## Testing 
1. Test that removing the current video in the playlist will have the play icon in `watch-video-playlist` update to the prior video
  A. Test `Play Previous Video`  plays the video with the same index as the video with the play icon
  B. Test `Play Next Video` and autoplay plays the video with index one higher than the video with the play icon
2. Test that removing the current video in the playlist will have the play icon in `watch-video-playlist` update to the prior video, then delete that one. See that the play icon then goes the video prior to that.
  A. (Optional) Repeat validations 1A and 1B in this state.
4. (Optional) Test 1 with Shuffle and/or Reverse enabled

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

### DIFF
--- a/src/renderer/components/watch-video-playlist/watch-video-playlist.js
+++ b/src/renderer/components/watch-video-playlist/watch-video-playlist.js
@@ -160,12 +160,12 @@ export default defineComponent({
     },
     selectedUserPlaylistVideoCount () {
       // Re-fetch from local store when current user playlist updated
-      this.parseUserPlaylist(this.selectedUserPlaylist, { allowPlayingVideoRemoval: true })
+      this.parseUserPlaylist(this.selectedUserPlaylist)
       this.shufflePlaylistItems()
     },
     selectedUserPlaylistLastUpdatedAt () {
       // Re-fetch from local store when current user playlist updated
-      this.parseUserPlaylist(this.selectedUserPlaylist, { allowPlayingVideoRemoval: true })
+      this.parseUserPlaylist(this.selectedUserPlaylist)
     },
     videoId: function (newId, oldId) {
       // Check if next video is from the shuffled list or if the user clicked a different video
@@ -466,22 +466,12 @@ export default defineComponent({
       })
     },
 
-    parseUserPlaylist: function (playlist, { allowPlayingVideoRemoval = true } = {}) {
+    parseUserPlaylist: function (playlist) {
       this.playlistTitle = playlist.playlistName
       this.channelName = ''
       this.channelId = ''
 
-      if (this.playlistItems.length === 0 || allowPlayingVideoRemoval) {
-        this.playlistItems = playlist.videos
-      } else {
-        // `this.currentVideo` relies on `playlistItems`
-        const latestPlaylistContainsCurrentVideo = playlist.videos.some(v => v.playlistItemId === this.playlistItemId)
-        // Only update list of videos if latest video list still contains currently playing video
-        if (latestPlaylistContainsCurrentVideo) {
-          this.playlistItems = playlist.videos
-        }
-      }
-
+      this.playlistItems = playlist.videos
       if (this.reversePlaylist) {
         this.playlistItems = this.playlistItems.toReversed()
       }
--- a/src/renderer/components/watch-video-playlist/watch-video-playlist.js
+++ b/src/renderer/components/watch-video-playlist/watch-video-playlist.js
@@ -48,6 +48,7 @@ export default defineComponent({
       loopEnabled: false,
       reversePlaylist: false,
       pauseOnCurrentVideo: false,
+      prevVideoBeforeDeletion: null,
       channelName: '',
       channelId: '',
       playlistTitle: '',
@@ -85,15 +86,7 @@ export default defineComponent({
     },
 
     currentVideoIndexZeroBased: function () {
-      return this.playlistItems.findIndex((item) => {
-        if (item.playlistItemId != null && this.playlistItemId != null) {
-          return item.playlistItemId === this.playlistItemId
-        } else if (item.videoId != null) {
-          return item.videoId === this.videoId
-        } else {
-          return item.id === this.videoId
-        }
-      })
+      return this.findIndexOfCurrentVideoInPlaylist(this.playlistItems)
     },
     currentVideoIndexOneBased: function () {
       return this.currentVideoIndexZeroBased + 1
@@ -123,16 +116,7 @@ export default defineComponent({
 
     videoIndexInPlaylistItems: function () {
       const playlistItems = this.shuffleEnabled ? this.randomizedPlaylistItems : this.playlistItems
-
-      return playlistItems.findIndex((item) => {
-        if (item.playlistItemId != null && this.playlistItemId != null) {
-          return item.playlistItemId === this.playlistItemId
-        } else if (item.videoId != null) {
-          return item.videoId === this.videoId
-        } else {
-          return item.id === this.videoId
-        }
-      })
+      return this.findIndexOfCurrentVideoInPlaylist(playlistItems)
     },
     videoIsFirstPlaylistItem: function () {
       return this.videoIndexInPlaylistItems === 0
@@ -186,6 +170,7 @@ export default defineComponent({
     },
     watchViewLoading: function (newVal, oldVal) {
       // This component is loaded/rendered before watch view loaded
+      this.prevVideoBeforeDeletion = null
       if (oldVal && !newVal) {
         // Scroll after watch view loaded, otherwise doesn't work
         // Mainly for Local API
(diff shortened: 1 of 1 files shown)


PikachuEXE (dismissed):

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Something is going wrong here: When loop is enabled it will skip a video

https://github.com/FreeTubeApp/FreeTube/assets/73130443/090bc6ae-f8cb-48c5-9fd3-21a3fd657359

jasonhenriquez:
Excellent catch John, we are very lucky to have someone with such great testing skills & discipline!

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Found another one

https://github.com/FreeTubeApp/FreeTube/assets/73130443/b7f0b5f5-3751-4145-925d-0538fa672a0a

jasonhenriquez:
@efb4f5ff-1298-471a-8973-3d47447115dc Could you share what you did before this? From the logs, it seems like somehow a `null` entry got into `randomizedPlaylistItems`.

efb4f5ff-1298-471a-8973-3d47447115dc:
https://github.com/FreeTubeApp/FreeTube/assets/73130443/81f98994-5331-46d0-b5f2-2e4301150dc1

jasonhenriquez:
Thanks @efb4f5ff-1298-471a-8973-3d47447115dc, fixed now. These were not in the original implementation, and still are not, but here are things I tried to add right now and gave up on:
- Defaulting to the 1st playlist as the first in the list when current video is not present at all (currently takes another trigger of `parseUserPlaylist` to have this behavior successfully occur). This happens because we're resetting `prevVideoBeforeDeletion` after the video is fully watch view is fully loaded, and this causes a race condition where it becomes unset. I couldn't figure out how to resolve this properly without adding a watcher to `playlistItemId` to do the removal instead, which I didn't know if that was worth it necessarily.
- Having the video index "become" the current video if it's removed and then re-added. It's a somewhat marginal case, and I don't know if it's actually feasible/sensible in the case of duplicates existing.

efb4f5ff-1298-471a-8973-3d47447115dc:
1. Turn shuffle on
2. Remove current playing video
3. click on play next video button
4. see that random video is selected which is expected because shuffle is turned on
5. repeat 2
6. click on play previous video button 
7. see that the actual previous video is being played which is unexpected

jasonhenriquez:
Thank you and great catches, I admittedly would never have caught these last two. I think it's all good now *knock on wood*

efb4f5ff-1298-471a-8973-3d47447115dc:
Removed video spawns back in?

https://github.com/FreeTubeApp/FreeTube/assets/73130443/d851f07d-fc20-4400-a730-a237f2d99656

jasonhenriquez:
Still looking into this one. I think I'm going to be fixing bugs with this PR until I die

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/components/watch-video-playlist/watch-video-playlist.js
```diff
@@ -232,6 +216,18 @@ export default defineComponent({
...
+      return playlist.findIndex((item) => {
+        if (item.playlistItemId != null && (this.playlistItemId != null || this.prevVideoBeforeDeletion?.playlistItemId != null)) {
+          return item.playlistItemId === this.playlistItemId || item.playlistItemId === this.prevVideoBeforeDeletion?.playlistItemId
+        } else if (item.videoId != null) {
+          return item.videoId === this.videoId || item.videoId === this.prevVideoBeforeDeletion?.videoId
+        } else {
+          return item.id === this.videoId || item.id === this.prevVideoBeforeDeletion?.videoId
+        }
+      })
+    },
```
absidue, line 229: Caching these 3 values, that are accessed multiple times for every iteration, might help a bit performance wise.
```suggestion
    findIndexOfCurrentVideoInPlaylist: function (playlist) {
      const playlistItemId = this.playlistItemId
      const prevVideoBeforeDeletion = this.prevVideoBeforeDeletion
      const videoId = this.videoId

      return playlist.findIndex((item) => {
        if (item.playlistItemId != null && (playlistItemId != null || prevVideoBeforeDeletion?.playlistItemId != null)) {
          return item.playlistItemId === playlistItemId || item.playlistItemId === prevVideoBeforeDeletion?.playlistItemId
        } else if (item.videoId != null) {
          return item.videoId === videoId || item.videoId === prevVideoBeforeDeletion?.videoId
        } else {
          return item.id === videoId || item.id === prevVideoBeforeDeletion?.videoId
        }
      })
    },
```

jasonhenriquez: Thanks! Two notes on this while we're looking at it:
- Do you know if the non-null checks have any effect on the performance? I'm not sure why they're there honestly. 
- I also don't fully know what `item.id` is versus `item.videoId` and what scenario it applies to where the former would be present when the latter isn't.

absidue: > Do you know if the non-null checks have any effect on the performance? I'm not sure why they're there honestly.

AFAIK because it's both a null and undefined check instead of just being one of them (notice that it only has one equals sign after the exclamation point) it still has to do type coercion, so we could probably convert them to just being truthy checks, as they will always either be strings or missing.

> I also don't fully know what item.id is versus item.videoId and what scenario it applies to where the former would be present when the latter isn't.

I haven't looked into it, but I presume it has something to do with us using the same component for playlists from 3 different sources (local API, Invidious API and users playlists), so they are probably mutually exclusive.

absidue (approved):

src/renderer/components/watch-video-playlist/watch-video-playlist.js
```diff
@@ -232,6 +216,23 @@ export default defineComponent({
     }
   },
   methods: {
+    findIndexOfCurrentVideoInPlaylist: function (playlist) {
```
PikachuEXE, line 219: nitpick: this should be `videos` / `playlistItems`
`findIndexOfCurrentVideoInPlaylist` is called with various source of video lists (with different sorting orders applied)

src/renderer/components/watch-video-playlist/watch-video-playlist.js
```diff
@@ -466,24 +482,29 @@ export default defineComponent({
...
-        this.playlistItems = playlist.videos
-      } else {
-        // `this.currentVideo` relies on `playlistItems`
-        const latestPlaylistContainsCurrentVideo = playlist.videos.some(v => v.playlistItemId === this.playlistItemId)
-        // Only update list of videos if latest video list still contains currently playing video
-        if (latestPlaylistContainsCurrentVideo) {
-          this.playlistItems = playlist.videos
-        }
+      const isCurrentVideoInParsedPlaylist = this.findIndexOfCurrentVideoInPlaylist(playlist.videos) !== -1
+      if (!isCurrentVideoInParsedPlaylist) {
```
PikachuEXE, line 491: Why use double negative here... (and below)

PikachuEXE (approved):
Non-blocking comments only so I only wish I spot them earlier next time

PR #5170, opened 2024-05-24
Access reactive video id once outside of the quick bookmark check

absidue:
# Access reactive video id once outside of the quick bookmark check

## Pull Request Type

- [x] Performance improvement

## Description
As the `id` property is reactive (it's inside of `data`), there is a small overhead for each access, during normal use that's not a problem, but in this case where we are accessing it potentially thousands of times (depends on how many items are in the quick bookmark playlist of course) inside a loop, it starts adding up. To avoid that, we can save the video to a temporary variable outside of the loop and access that directly, which avoids the reactivity overhead inside the loop.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 6481123d48815e87374ac5467c4419db23d46348

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -442,8 +442,10 @@ export default defineComponent({
     isInQuickBookmarkPlaylist: function () {
       if (!this.isQuickBookmarkEnabled) { return false }
 
+      const id = this.id
+
       return this.quickBookmarkPlaylist.videos.some((video) => {
-        return video.videoId === this.id
+        return video.videoId === id
       })
     },
     quickBookmarkIconText: function () {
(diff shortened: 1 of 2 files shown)


jasonhenriquez (commented):
**suggestion (non-blocking):** Document the rationale in code comments for both instances

PikachuEXE:
Same
Best to have a linter (if there is one) but code comments works fine for now

Can't find one in https://eslint.vuejs.org/rules/

absidue:
What do you want me to write in the code comment?

> Best to have a linter (if there is one)

What?

PikachuEXE:
> Best to have a linter

I know there isn't one, ignore it.

> What do you want me to write in the code comment?
```js
// Avoid unnecessary reactive data access overhead
const id = this.id
```

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue:
Added comments

jasonhenriquez (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #5173, opened 2024-05-25
Convert unnecessary SCSS files to CSS

absidue:
# Convert unnecessary SCSS files to CSS

## Pull Request Type

- [x] Other - Cleanup

## Description
We have a bunch of SCSS files in the project, some are worth keeping as they take proper advantage of SCSS usng variables and mixins and imports for example, others however just use it for unnecessary nesting or are simple enough to just be plain CSS file. This pull request cleans up those unnecessary SCSS files.

## Testing 
Check that the following pages and components look the same as before:
- `ft-age-restricted` (Turn on `Show Family Friendly Only` and visit an age-restricted video)
- `ft-auto-grid` (Test the grid and list layouts on the subscriptions page for example)
- `player-settings`
- `About`

## Desktop

- **OS:** Windows
- **OS Version:** 10


jasonhenriquez:
Not sure if intentional, but the age restricted title looks different now.

| Before | After |
|--|--|
| Screenshot_20240525_172722 | Screenshot_20240525_172318 |

src/renderer/components/ft-age-restricted/ft-age-restricted.vue
```diff
@@ -1,8 +1,6 @@
 <template>
-  <div
-    class="ft-age-restricted"
-  >
-    <h2>
+  <div>
+    <h2 class="messsage">
```
absidue, line 3:
```suggestion
    <h2 class="message">
```

jasonhenriquez (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #4803, opened 2024-03-25
[Feature Request]: Keep manually fetched subscriptions from the previous session


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

FT does not retain the subscriptions that were fetched in the previous session. Users always need to fetch manually and this will result in seeing the latest uploads. Users might have to scroll down quite a bit to see where they left off.

User has `Automatic fetching` set to `disabled` -> FT shows the message that the user has to manually load their subscriptions. -> closes FT -> opens FT -> FT shows the message that the user has to manually load their subscriptions.

### Proposed Solution

User should see their see their manually fetched subscriptions from the previous session when `Automatic fetching` is `disabled`

IMO this would give the user better inside in where they left off. Users can decide better if they want to refresh or not.

### Alternatives Considered

none

### Issue Labels

ease of use improvement

### Additional Information

https://github.com/FreeTubeApp/FreeTube/pull/4380 should be merged first

PR #5185, opened 2024-05-28
Implement persistent subscription cache

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Make subscription cache sync across multiple windows and persistent over app restarts

## Screenshots 

N/A

## Testing 

Many cases, generally:
- Disable `Fetch Feed Automatically`
- Switch/mess around with profiles if wanted
- Load videos, live, shorts, community posts
- Open new window(s) to see effects
- Reboot app
- Switch to different profiles to see effects

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

- No idea how to handle `updateSubscriptionShortsCacheWithChannelPageShorts`
- No idea why community posts in other window have like count like this
  image

### DIFF
--- a/src/renderer/store/modules/profiles.js
+++ b/src/renderer/store/modules/profiles.js
@@ -29,7 +29,14 @@ const getters = {
   profileById: (state) => (id) => {
     const profile = state.profileList.find(p => p._id === id)
     return profile
-  }
+  },
+
+  getSubscribedChannelIdSet: (state) => {
+    const mainProfile = state.profileList.find((profile) => {
+      return profile._id === MAIN_PROFILE_ID
+    })
+    return mainProfile.subscriptions.reduce((set, channel) => set.add(channel.id), new Set())
+  },
 }
 
 function profileSort(a, b) {
--- a/src/constants.js
+++ b/src/constants.js
@@ -17,11 +17,13 @@ const IpcChannels = {
   DB_HISTORY: 'db-history',
   DB_PROFILES: 'db-profiles',
   DB_PLAYLISTS: 'db-playlists',
+  DB_SUBSCRIPTIONS: 'db-subscriptions',
 
   SYNC_SETTINGS: 'sync-settings',
   SYNC_HISTORY: 'sync-history',
   SYNC_PROFILES: 'sync-profiles',
   SYNC_PLAYLISTS: 'sync-playlists',
+  SYNC_SUBSCRIPTIONS: 'sync-subscriptions',
 
   GET_REPLACE_HTTP_CACHE: 'get-replace-http-cache',
   TOGGLE_REPLACE_HTTP_CACHE: 'toggle-replace-http-cache',
@@ -52,7 +54,14 @@ const DBActions = {
     DELETE_VIDEO_ID: 'db-action-playlists-delete-video-by-playlist-name',
     DELETE_VIDEO_IDS: 'db-action-playlists-delete-video-ids',
     DELETE_ALL_VIDEOS: 'db-action-playlists-delete-all-videos',
-  }
+  },
+
+  SUBSCRIPTIONS: {
+    UPDATE_VIDEOS_BY_CHANNEL: 'db-action-subscriptions-update-videos-by-channel',
+    UPDATE_LIVE_STREAMS_BY_CHANNEL: 'db-action-subscriptions-update-live-streams-by-channel',
+    UPDATE_SHORTS_BY_CHANNEL: 'db-action-subscriptions-update-shorts-by-channel',
+    UPDATE_COMMUNITY_POSTS_BY_CHANNEL: 'db-action-subscriptions-update-community-posts-by-channel',
+  },
 }
 
 const SyncEvents = {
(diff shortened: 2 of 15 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Isnt https://github.com/FreeTubeApp/FreeTube/issues/4803 related?

Maybe this PR closes it? Idk didnt test yet😬

PikachuEXE:
Yup added to OP

PikachuEXE:
Updated code & DB file name to `subscription cache` per dev chat comment

efb4f5ff-1298-471a-8973-3d47447115dc:
From the looks of it this also solves https://github.com/FreeTubeApp/FreeTube/issues/4152

efb4f5ff-1298-471a-8973-3d47447115dc:
I will revisit this after the shaka PR.

PikachuEXE:
after the shaka PR :)

src/renderer/store/modules/subscription-cache.js
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        let hasData = false
+
+        if (Array.isArray(dataEntry.videos)) {
+          videos[channelId] = { videos: dataEntry.videos, timestamp: dataEntry.videosTimestamp }
+          hasData = true
+        }
+        if (Array.isArray(dataEntry.liveStreams)) {
+          liveStreams[channelId] = dataEntry.liveStreams
```
absidue, line 79: Looks like this line is probably left over from something, as it doesn't exists for the other types.

src/renderer/store/modules/profiles.js
```diff
@@ -29,7 +29,14 @@ const getters = {
   profileById: (state) => (id) => {
     const profile = state.profileList.find(p => p._id === id)
     return profile
-  }
+  },
+
+  getSubscribedChannelIdSet: (state) => {
+    const mainProfile = state.profileList.find((profile) => {
+      return profile._id === MAIN_PROFILE_ID
+    })
```
absidue, line 37: The all channels profile is always the first profile in the array, so you don't have to search through the array for it :)
```suggestion
    const mainProfile = state.profileList[0]
```

absidue (commented):
I can't comment on unchanged files, but as you aren't using the `getLastVideoRefreshTimestampByProfile` and `updateLastVideoRefreshTimestampByProfile` code anymore, it probably makes sense to remove it from the `src/renderer/store/modules/utils.js` file.

absidue (approved):

absidue:
For the videos, live and shorts tabs we calculate the published timestamp at the time that we fetch the data, however for the community posts we currently save the relative text timstamp `10 minutes ago`. While this is a pre-existing problem, it will be much more noticable with this pull request because now we will be saving that data between app launches, instead of just in-memory for the same session.

PikachuEXE:
Made change to `Save community post publish time in absolute time not relative time text`
Tested w/ local&IV API (test yourself

absidue (approved):

ChunkyProgrammer (approved):

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc  you approved the doc PR https://github.com/FreeTubeApp/FreeTube-Docs/pull/143 but not this PR?
image

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Why cant it persist the cache when subscribing to a new channel? Lets say the last refresh timestamp was on 8hrs ago and I want to keep it that way. Now this completely disappears. Users must refresh otherwise they cant see videos. I would expect that my feed doesnt disappear and when i press refresh i will see the videos from the new channel i subscribed to

https://github.com/user-attachments/assets/fa576b78-16ad-449e-a9c1-7740c445e194

Timestamp doesnt appear after refresh

https://github.com/user-attachments/assets/52ad74d5-6ced-4950-8c3d-bd303846a898

PikachuEXE:
> Why cant it persist the cache when subscribing to a new channel?

That's A3 in https://github.com/FreeTubeApp/FreeTube/pull/3668

> Timestamp doesnt appear after refresh

Will fix

efb4f5ff-1298-471a-8973-3d47447115dc:
> > Why cant it persist the cache when subscribing to a new channel?
> 
> That's A3 in https://github.com/FreeTubeApp/FreeTube/pull/3668

Sorry i have to pushback on this because I think that this is a dedradation of the UX if it goes in like this

PikachuEXE:
Fixed `Timestamp doesnt appear after refresh`

> Why cant it persist the cache when subscribing to a new channel?

I think you are looking for something like https://github.com/kommunarr/FreeTube/pull/2
It's quite big and I don't think this PR should include that additional amount of code change = more time for review & test
This PR improves for those with no new channel subscribed after refresh, just no improvement after new channel subscribed (worst case just refresh like now)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
After seeing the report in https://github.com/FreeTubeApp/FreeTube/issues/5724 I think we need to change something. I think allot of users will see this as broken behavior just like the one that reported it. I suggest to bypass the cache when on application startup.

PikachuEXE:
https://github.com/FreeTubeApp/FreeTube/issues/5724 looks like a bug we didn't catch in this PR (forgot to add test case with `Fetch Feed Automatically` enabled, marginal case test)

Fixing today

PR #5191, opened 2024-05-30
$ Remove duplicate line in a method setting a property to the same value

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/pull/5044#issuecomment-2138397557

## Description

- Only on user playlists with duplicate videos (by ID)
- Add button for removing duplicate videos

## Screenshots 

image

## Testing 

- Ensure button invisible in remote playlist, edit mode, playlist with no video, no duplicate video
- Ensure button works like remove watched videos except it targets duplicate videos (including toast message)
- Ensure button disappear if duplicate videos removed another way

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

If you got a better icon let me know~

### DIFF
--- a/src/renderer/components/playlist-info/playlist-info.js
+++ b/src/renderer/components/playlist-info/playlist-info.js
@@ -346,7 +346,6 @@ export default defineComponent({
 
       if (removedVideosCount === 0) {
         showToast(this.$t('User Playlists.SinglePlaylistView.Toast["There were no videos to remove."]'))
-        this.showRemoveVideosOnWatchPrompt = false
         return
       }
 
--- a/src/renderer/components/playlist-info/playlist-info.js
+++ b/src/renderer/components/playlist-info/playlist-info.js
@@ -110,6 +110,7 @@ export default defineComponent({
       editMode: false,
       showDeletePlaylistPrompt: false,
       showRemoveVideosOnWatchPrompt: false,
+      showRemoveDuplicateVideosPrompt: false,
       newTitle: '',
       newDescription: '',
       deletePlaylistPromptValues: [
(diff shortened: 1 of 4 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Ugh finding an appropriate icon for this sucks

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Everything LGTM but im not so sure about the icon that is being used but also couldnt find a better one that we arent already using. 

> Ensure button disappear if duplicate videos removed another way

Ugh we might also need to do this for the remove watched videos. No point in having it there when there arent any videos in the playlists. That would also make the UX consistent which wont confuse users

PikachuEXE:
Updated to make remove watched videos button visible only when any video watched

efb4f5ff-1298-471a-8973-3d47447115dc:
Thanks Pika i really appreciate the effort. I'll try to find a better icon. If anyone found an icon that represents this better pls share :)

jasonhenriquez:
[`users-slash`](https://fontawesome.com/icons/users-slash?f=classic&s=solid)  almost works because it has many of one thing being slashed. This is why I want #4259 to be a thing, because I don't think anything we could choose here would fully suffice in conveying the meaning well enough on its own.

efb4f5ff-1298-471a-8973-3d47447115dc:
Lets change it to that one as that is indeed the only slash kind of icon that involves multiple things

PikachuEXE:
icon updated
image

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

jasonhenriquez (commented):
Looks good!

**suggestion (non-blocking):** This is not strictly within the original scope, but I would suggest that the 2 prompts actually showed the number of duplicate videos / watched videos that will be removed. They currently only do so in the toasts after the deletions are complete. As it stands, it's unclear what the magnitude of the action will be.

PikachuEXE:
> the 2 prompts actually showed the number of duplicate videos / watched videos that will be removed

Updated
image
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

jasonhenriquez (approved):

absidue (approved):

Issue #2446, opened 2022-08-03
[Feature Request]: Remember page positioning when using the back or forward button 


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a feature request that matches the one I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.

### Problem Description

Multiple users have reported that when using the back or forward button it doesn't direct the to the same position on the page the were before. 

Here is a list what does and doesn't work properly at the moment:

- [x] Going back or forward to the Subscription page will bring u where u was before on the page.
- [ ] Going back or forward to the Default Trending page will bring u where u was before on the page.¹
- [x] Going back or forward to the Most Popular page will bring u where u was before on the page.
- [ ] Going back or forward to the Playlist page will bring u where u was before on the page.²
- [ ] Going back or forward to the History page will bring u where u was before on the page.³
- [ ] Going back or forward to the Settings page will bring u where u was before on the page.⁴
- [x] Going back or forward to the About page will bring u where u was before on the page.
- [ ] Going back or forward to a Channels video page will bring u where u was before on the page.⁵
- [ ] Going back or forward to a Channels page when u made a search will bring u where u was before on the page.⁶
- [ ] Going back or forward to a Channels page when u used Sorted videos By will bring u where u was before on the page.⁷
- [ ] Going back or forward to a Channels playlist page will bring u where u was before on the page.⁸
- [x] Going back or forward in a Playlist will bring u where u was before on the page.
- [ ] Going back or forward to a Channels about page will bring u where u was before on the page.⁹
- [ ] Going back or forward to the channels tab in the sidebar will bring u where u was before on the page.¹⁰
- [x] Going back or forward to a search page (with or without filters being applied) will bring u where u was before on the page.
- [ ] Going back or forward to a video will bring u where u was before on the page.¹¹
- [ ] Going back or forward to a profile page u selected in the profile manager will bring u back where u was before on the page.¹²

1. Go to another Trending Category -> scroll a bit -> click on random video -> click on back button -> see that u are on the same position on the page but in the default trending category. This would be also the case if u click on the back button another time and then click forward to go back to the trending page. If u do this on the default trending tab everything is working fine.
[...]

PR #5192, opened 2024-05-30
Make history page remember last query string & search limit

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Make history page remember last query string & search limit to deal with
- Search in history
- Found a few videos
- View one of them, go back last page, oops not filtered

## Screenshots 

No video I am lazy

## Testing 

- Ensure last used search query and data limit restored when revisiting history page
- Ensure they are not restored after clearing query text / in new window
- Ensure no 2 time rendering on page initial render

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

There are 2 commits & first one is refactor
Might easier to review per commit


efb4f5ff-1298-471a-8973-3d47447115dc:
I understand where you're are coming from but this behavior is also not apparent in Playlists tab, Channel tab, and a Channel page.

Channel page one also applies to Nr 6 on the list of issue https://github.com/FreeTubeApp/FreeTube/issues/2446 which is a way bigger issue

jasonhenriquez:
I'd also like to know rationale, as I don't see any requests for this. I'm not necessarily opposed to it, but I don't have any reason to believe that the app remembering my last query would cause fewer irksome moments than the app not remembering it.

PikachuEXE:
Making it work like search query page
I realize this when searching in history (and several videos popup), click one of them, not the one I want, go back, oops
Though I can always open in new windows (but mobile can't do it?)
Up to you guys

jasonhenriquez:
If you want, you can share the build & I can try it out for a while in personal use. It sounds nice in theory, but it feels more horizontal of a change for the use cases where you want to go to the initial page state & now have to click to clear the input. The search bar is also omnipresent, so you wouldn't ever forget what was going on, but here I could see it leading to momentary confusion and re-familiarizing yourself with the context of what's going on & what you were searching beforehand. Seems like maybe "open in a new tab" covers the type of use case you're mentioning better (albeit I guess not if you didn't expect to misclick).

This is more of a tangent for user playlists, but this is something I find myself wanting after searching for a particular set of videos in a user playlist: a temporary custom ordering where the non-filtered results are cached at the top of the `watch-video-playlist`, then the regular ordering for the rest, and maybe a button to restore the regular ordering / clear the "filter" ordering.

PikachuEXE:
Used `router.replace` to ensure the query text restored only on clicking back
Clicking on history link won't restore it

About the `catch` after `replace`:
https://v3.router.vuejs.org/guide/advanced/navigation-failures.html

efb4f5ff-1298-471a-8973-3d47447115dc:
> Just to clarify im not against this but I think we should do it for all pages instead of one for the sake of a consistent UX

Just retested it but my opinion on this hasnt changed.

PikachuEXE:
Can you list all the page types so I won't miss any on next update?

kommunarr:
That sounds like it would be a pretty sizable change to the codebase for a small feature. I would personally advise putting this one on the backburner just because I'm not sure if the feature would provide enough utility versus the effort to maintain it.

PikachuEXE:
> FT Channel tab
User Playlist tab main view
Inside User Playlist
Creator Channel page

Also done but for the Creator Channel page one I am not really confident (even though I tested it a bit

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
1. Go to History
2. Search Something
3. Enable Case Sensitive
4. Click on Video
5. Go back
6. See that results not the same, Its showing results from Case Sensitive Disabled

Same with Playlists main view:


https://github.com/user-attachments/assets/a5f94272-44d1-4743-b16a-0ec375fa6901

PikachuEXE:
Updated

efb4f5ff-1298-471a-8973-3d47447115dc:
Notice that results are ordered in a different when navigating back to the channel page


https://github.com/user-attachments/assets/374d566d-037c-4913-9538-e120d69b1d15

PikachuEXE:
I can reproduce it once but cannot reproduce afterward, no idea how results are sorted (or just taking from API)

https://github.com/user-attachments/assets/c57b8fd2-bfee-4f79-a408-b3f32c7ce7b0

src/renderer/views/Channel/Channel.js
```diff
@@ -453,13 +460,19 @@ export default defineComponent({
...
-      this.getChannelLocal().finally(() => {
+      await this.getChannelLocal().finally(() => {
         this.autoRefreshOnSortByChangeEnabled = true
       })
     }
+
+    const oldQuery = this.$route.query.searchQueryText ?? ''
+    if (oldQuery !== null && oldQuery !== '') {
+      this.newSearch(oldQuery)
+    }
```
PikachuEXE, line 475: If you figure out what's messing with the query result sort order let me know
I read the rest of the code I don't think FT even tries to mess with the result sort order

efb4f5ff-1298-471a-8973-3d47447115dc:
> I can reproduce it once but cannot reproduce afterward, no idea how results are sorted (or just taking from API)

If you switch often enough you can make it happen again

Uhh why its focused like that? Dev branch doesnt seem to do that.

https://github.com/user-attachments/assets/a3428fbf-0321-4c0a-b1a4-7e260429b1dc

PikachuEXE:
> It is showing the loading animation so maybe its trying to do something in the background?

- `newSearch`
- `this.isElementListLoading = true`
- `searchChannelLocal` or `searchChannelInvidious`

Ya it's searching

> Uhh why its focused like that? Dev branch doesnt seem to do that.

Caused by `changeTab` -> `showOutlines`. it's a bug that is never visible until now
Fixed

PikachuEXE:
> If you switch often enough you can make it happen again.

I can't make it happen anymore

Maybe you can find the request and look at the result order to see if that matches the order of result displayed?

image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
I tried to reproduce it again but it took me a solid minute going back and forth to make it happen again. Maybe its not that big of a deal?

PikachuEXE:
If you can prove that the result order is incorrectly/ever manipulated with this PR then it should be fixed
Otherwise it could be just temp search result order alternating for who knows why

efb4f5ff-1298-471a-8973-3d47447115dc:
> If you can prove that the result order is incorrectly/ever manipulated with this PR then it should be fixed

I would love to but im not sure how. Could you give me some more instructions on top of the ones u already provided in https://github.com/FreeTubeApp/FreeTube/pull/5192#issuecomment-2390338307

PikachuEXE:
- Go into a channel, search stuff
- Inspect latest request (request payload with `query`
- Preview tab, `contents`...`tabs`, find element with `expandableTabRenderer`, `content`...`contents`, expand those items to inspect video IDs

efb4f5ff-1298-471a-8973-3d47447115dc:
I would like to console log the videoID's because expanding everything like this makes my head hurt. Where should I put the console.log?

PikachuEXE:
Change 
```js
        if (isNewSearch) {
          this.searchResults = results
        } else {
          this.searchResults = this.searchResults.concat(results)
        }
```

to 

```js
        if (isNewSearch) {
          this.searchResults = results
          results.forEach(e => {
            if (e.type === 'video') console.log(`${e.type}: ${e.videoId}`)
            else if (e.type === 'playlist') console.log(`${e.type}: ${e.playlistId}`)
            else console.log(`${e.type}`)
          })
        } else {
          this.searchResults = this.searchResults.concat(results)
        }
```
image

`searchChannelLocal` is not modified by this PR

efb4f5ff-1298-471a-8973-3d47447115dc:
Left one is just a normal search. Right one is when coming back to the channel page

Capture3

PikachuEXE:
I can simply find a channel and type something on search input and keep pressing enter and gets results sorted slightly differently sometimes
The one with thumbnail "Dumb but good" I have seen it on 4th place most of the time, but 2nd & 3rd place sometimes

https://github.com/user-attachments/assets/cdf70221-0089-41e2-a8fd-ddb9aa1c1198

absidue:
As the goal of this pull request seems to have changed a lot, could you please update the pull request description and title with what this pull request actually does.

src/renderer/views/Channel/Channel.js
```diff
@@ -2018,6 +2036,35 @@ export default defineComponent({
...
+
+          throw failure
+        })
+        return
+      }
+
+      await this.$router.replace({
+        path: `/channel/${this.id}`,
+        query: {
+          currentTab: 'search',
```
absidue, line 2055: Erm why is this setting the current tab in the query parameters, when it's a path parameter?

PikachuEXE:
@absidue 
Both fixed

efb4f5ff-1298-471a-8973-3d47447115dc:
It briefly shows that there are 0 results for the search i made before showing the actual search results

https://github.com/user-attachments/assets/07f74bef-a1e4-4131-9d78-50b7551c9933

PikachuEXE:
I just fixed this for search tab only
Since the data `isElementListLoading` for all tabs if any tab's data finished loading it shows the result too early
I introduced another state for search tab only (no need for other tabs yet coz PR is not for remembering other tabs)

absidue (commented):
As we are now adding the data limits to the URLs, there isn't really a need to duplicate them into session storage too, so we can probably remove that code.

Additionally the code for the channel page seems a bit messy, but in an unusual move for me I think I would be okay with it, assuming that it will be refactored soon anyway when we migrate the component to the composition API or start saving the selected tab in the URL, whichever happens sooner.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
To future after RC reviewer me: LGTM

PikachuEXE:
@absidue 
`dataLimit` and `searchDataLimit` and separated and I don't intend to introduce more changes to this already big PR causing potentially more stuff to test = more chances to break something

I can make a PR later for storing `dataLimit` into route instead of session storage (which might cause behaviour change, another reason to not do so in this PR

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/views/SubscribedChannels/SubscribedChannels.vue
```diff
@@ -230,10 +265,16 @@ watch(activeSubscriptionList, () => {
...
   document.addEventListener('keydown', keyboardShortcutHandler)
   getSubscription()
-})
+
+  const oldQuery = route.query.searchQueryText ?? ''
+  if (oldQuery !== null && oldQuery !== '') {
+    // `handleQueryChange` must be called after `filterHistoryDebounce` assigned
+    handleQueryChange(oldQuery, true)
+  }
+})()
```
absidue, line 277: Could you please explain why you changed this from an onMounted callback to an immediately invoked function?

src/renderer/views/Channel/Channel.js
```diff
@@ -2009,6 +2036,37 @@ export default defineComponent({
...
+    async saveStateInRouter(query) {
+      this.skipRouteChangeWatcherOnce = true
+      if (query === '') {
+        await this.$router.replace({ path: `/channel/${this.id}` }).catch(failure => {
+          if (isNavigationFailure(failure, NavigationFailureType.duplicated)) {
+            return
+          }
+
+          throw failure
+        })
```
absidue, line 2048: To make the code more readable please pick one of the Promise syntaxes either async-await or chaining, instead of combining both of them. When using async-await, you can just use a normal try-catch.

This suggestion applies to all files that have this construct introduced by this pull request.

absidue (commented):
I haven't got round to testing the pull request yet, so for the moment here is a code review with a bunch suggestions and one question.

src/renderer/views/SubscribedChannels/SubscribedChannels.vue
PikachuEXE: It was changed from `mounted` to `created` but in composition style either it's in setup block
However `oldQuery` is only used inside `created` so I wrap it with an immediately invoked function

PikachuEXE:
Other points are addressed except `created`

src/renderer/views/SubscribedChannels/SubscribedChannels.vue
absidue: The keyboard shortcut handler should be moved back into `onMounted` as we shouldn't be intercepting key presses before the user can actually see the component. As for the other stuff I don't think it's worth making it more complicated with an immediately invoked function because setup is compiled to a function, so the variable will get garbage collected as soon as it reaches the bottom of it anyway.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5213, opened 2024-06-02
Limit which URLs FreeTube will open externally

absidue:
# Limit which URLs FreeTube will open externally

## Pull Request Type

- [x] Security improvement

## Description
This pull request implements [Electron security guideline 15: "Do not use `shell.openExternal` with untrusted content"](https://www.electronjs.org/docs/latest/tutorial/security#15-do-not-use-shellopenexternal-with-untrusted-content). As FreeTube displays user generated content that can also contain links (video descriptions and comments), we can't lock this down properly but we can at least limit the URL schemes that can be opened from FreeTube.
Here are the allowed URL schemes:
* `https:` All URLs allowed, links using this scheme exist on the about page as well as in video descriptions and comments.
* `http:` All URLs allowed, links using this scheme exist in video descriptions and comments.
* `tel:` All URLs allowed, Autolinker turns phone numbers into links so if they appear in video descriptions they will get linked (we can turn off that behaviour if we don't want that, YouTube doesn't turn them into links).
* `mailto:` All URLs allowed, the email address on the about page and Autolinker turns email addresses into links so if they appear in video descriptions they will get linked (we can turn off that behaviour if we don't want that, YouTube doesn't turn them into links).
* `bitcoin:` Only allowed with the wallet address that appears on the about page
* `monero:` Only allowed with the wallet address that appears on the about page

### Check that desired links still open correctly
Check that the links on the about page and in video descriptions still open according to your `External link handling` preference.

### Check that unwanted URLs don't get opened
1. Open the dev tools
2. Run this with your desired URL or text `require('electron').ipcRenderer.send('open-external-link', 'https://example.com')`
    - On Windows you can try `ms-settings:` (URL scheme to launch the settings app)
    - On macOS you can try `notes://` (URL scheme to launch the notes app)
    - On Linux you can try `man:which` (URL to open the man page for the `which` command)
    - Text that isn't a URL e.g. `some random text`
3. For text that is not a URL or URL schemes that aren't allow listed, nothing should happen.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0b62ea78d83867a97217c500aa67f5343fe65066

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -6,7 +6,13 @@ import {
 import path from 'path'
 import cp from 'child_process'
 
-import { IpcChannels, DBActions, SyncEvents } from '../constants'
+import {
+  IpcChannels,
+  DBActions,
+  SyncEvents,
+  ABOUT_BITCOIN_ADDRESS,
+  ABOUT_MONERO_ADDRESS
+} from '../constants'
 import * as baseHandlers from '../datastores/handlers/base'
 import { extractExpiryTimestamp, ImageCache } from './ImageCache'
 import { existsSync } from 'fs'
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

jasonhenriquez:
**suggestion/question (non-blocking):** Should we have a toast informing the user when an unsupported domain of ${domain} is attempted to be navigated to?

absidue:
@jasonhenriquez Added a toast message now. If you want to test that you'll either want to edit the page HTML and click on the edited links or add `global.openExternalLink = openExternalLink` to `src/renderer/helpers/utils.js` that way you can use it from the devtools.

jasonhenriquez (approved):

src/main/index.js
```diff
@@ -855,8 +861,36 @@ function runApp() {
...
+        // Donation links on the about page
+        (parsedURL.protocol === 'bitcoin:' && parsedURL.pathname === ABOUT_BITCOIN_ADDRESS) ||
+        (parsedURL.protocol === 'monero:' && parsedURL.pathname === ABOUT_MONERO_ADDRESS)
+      ) {
+        shell.openExternal(url)
+        return true
+      }
+    }
+
+    return false
```
PikachuEXE, line 893: Nitpick on code style (intent)
The smaller code block should be nested instead of the bigger one
Change it this PR if you want but I notice this in other places too so it's more like a reminder/rant
```js
if (typeof url !== 'string') { return false }

// Rest of bigger code block
```

PikachuEXE (approved):
Tested but the testing instruction seems wrong
I used `await require('electron').ipcRenderer.invoke('open-external-link', 'https://nginx.org')`

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #4408, opened 2023-11-30
[Bug]: Page navigation arrow indicators not updated by mouse buttons


jasonhenriquez:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Navigate to any view, e.g., Settings
2. Using a mouse, use the "back" button to go back. See the navigation arrows not reflecting the proper position in history.
![Screenshot_20231130_064014](https://github.com/FreeTubeApp/FreeTube/assets/84899178/12ae63e0-ba5e-4168-8aa1-483617a220f9)

### Expected Behavior

The navigation arrows are updated properly for any kind of navigation method, including using mouse buttons.

### Issue Labels

inconsistent behavior, visual bug

### FreeTube Version

v0.19.1

### Operating System Version

OpenSUSE Tumbleweed

### Installation Method

Portable

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

none

### Additional Information

`vue-router` is waiting on the Navigation API to become more stable before sharing this information. This would allow for the best and most competent implementation: See: https://github.com/vuejs/vue-router/issues/3453

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

PR #5227, opened 2024-06-04
Navigation history dropdowns

### LINKED ISSUE #2740: [Feature Request]: Right click on navigation buttons show history
### Guidelines

- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for open and closed issues that are similar to the feature request I want to file, without success.

### Problem Description

When an user want to navigate back/forward to get back to a specific page it is possible that the user needs to click allot of times on the back/forward button to find the page they are looking for

### Proposed Solution

Right click on one of the buttons to show history path

https://user-images.githubusercontent.com/73130443/196279706-415bf6ac-fd37-4b12-bd19-7fe1c8e7b4d7.mp4

### Alternatives Considered

None

### Issue Labels

display more information to user, ease of use improvement, improvement to existing feature

### Additional Information

_No response_

jasonhenriquez:
# Navigation history dropdowns

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #2740

## Description
- On long press (0.5s) or right-click of a non-disabled navigation arrow, load <=15 surrounding history entries (their `route.meta.title` - see "Work for a future PR" below). The entry behavior is taken from FF:
  - If at the 7th entry from start/last, show results from the start/last. Otherwise, show 7 earlier entries & 7 later entries.
- `@contextmenu.prevent=''` is set on the navigation arrows to prevent the context menu appearing on right-click in dev mode.

Minor cleanup:
- d18b549509e6fffda356570d9dc1f02b194df032: Incidentally as a way of cleaning up an extra event handler to worry about in `ft-icon-button`, replaces the `mouseDownOnIcon` `focusout` logic in `ft-icon-button in favor of more straightforward logic (adopted from #4231)
- Incidentally, makes `topNav.historyForward` and `topNav.historyBackward` logic more straightforward

## Work for a future PR
- We're currently showing the `route.meta.title`, which is far less informative than the `<title>` attribute we set in various places. The main issue with doing that is that it's pretty onerous to do it at the moment because many of our routes asynchronously change the `<title>` attribute. I don't know the best solution, so let's see if Cunningham's Law works here: the future PR should use IPC or VueX to fire an event whenever a title is set, then we update the entry to switch from a placeholder `route.meta.title` to the newest `<title>` attribute value.

## Screenshots 
Screenshot_20240603_201818
Screenshot_20240603_201837

## Testing 
- Check that the result ordering behavior works correctly:
> If at the 7th entry from start/last, show results from the start/last. Otherwise, show 7 earlier entries & 7 later entries.
- Check that navigating to any arbitrary history entry works
- Check that both right-click and long-press work to trigger the navigation history dropdown
- Check long-press on mobile
- (REG, optional) w.r.t. d18b549509e6fffda356570d9dc1f02b194df032, check that `focusout` / <kbd>Esc</kbd> / keyboard navigation with `ft-icon-button` dropdowns still work as intended

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW


PikachuEXE:
It doesn't feel like web browser UX:
- Showing all entries when long press on either back / forward button, expect to see only previous / next entries
- Expect not to see entry for current page

jasonhenriquez:
> It doesn't feel like web browser UX:
- Showing all entries when long press on either back / forward button, expect to see only previous / next entries
- Expect not to see entry for current page

I'm not sure what browser you use, but the implementation is the same behavior as Firefox:

Screenshot_20240603_213730

I wouldn't say it's the only way to do it, but it's a major browser

PikachuEXE:
Even that's the case Firefox does show which one is the current entry which I don't see in this PR
image
image

jasonhenriquez:
> Even that's the case

'Tis!

> Firefox does show which one is the current entry which I don't see in this PR

Two visual indicators here, my friend! I can add a checkmark on top of that if desired, but please look at the PR & description in closer detail before formulating your criticisms.

Screenshot_20240603_215215

PikachuEXE:
I am sorry that I didn't provide screenshot when providing my opinion
image

A more accurate description would be the
Firefox does show which one is the current entry **in darker themes** which I don't see in this PR

PikachuEXE:
Additionally the colors for back / forward button are now inconsistent with other top-nav buttons in themes like `light`, `pastel pink`
image

jasonhenriquez:
>  Firefox does show which one is the current entry in darker themes which I don't see in this PR

This is more of a pre-existing problem with certain themes like Dark not having a visually distinctive `--side-nav-hover-color`, which results in the top & side navs conveying less information, but I will add a checkmark to make it clearer.

> Additionally the colors for back / forward button are now inconsistent with other top-nav buttons in themes like light, pastel pink

I think you're on an older version. I fixed this in an earlier commit.

PikachuEXE:
With checkmark it seems better
However I re-tested after checking out `Add checkmark icon to active history entry` and the inconsistent button color is still an issue

jasonhenriquez:
You might need to rebase then, as I cannot replicate:
Screenshot_20240603_221630

PikachuEXE:
Using theme `pastel pink` for testing:

What I see in this PR
image
image
image

`development`
image
image

jasonhenriquez:
Ah, it's not happening to me with hot pink, but I can see it with red as main color. Looking into it now, thank you

PikachuEXE:
All issues resolved, pending idea from others especially on mobile

jasonhenriquez:
I had no idea there was a [`NavigationHistory`](https://www.electronjs.org/docs/latest/api/navigation-history) class in Electron until @absidue pointed it out in the Matrix chat, so it depends if we want to put more eggs into this PR or save that refactor for a separate PR that adds title support, as this is currently functional as-is. That's more of an organizing question than a substantive one, so I'm fine with whichever choice. I will say the suggested change to using a context menu is probably a lot more lateral, so I would prefer if that one was kept to a follow-up PR.

With all that said, I would advise that any reviewers prioritize the higher-salience PRs first if possible. We're a lot closer than we think, and we're reaching a critical mass of feature requests for features we've already merged.

jasonhenriquez:
Here's where I'm at after looking into it some more. 

I can see what was being gotten at with using a context menu as an easy way to grab the current `browserWindow.webContents.navigationHistory`, but I don't like that we have less control over the styling of the popup. We can discuss that aspect some more, but I'm not sure if that's something we would want, so I'm going to leave that as a separate future effort.

The `NavigationHistory` API as we're using it is very new to Electron, so I would imagine it's subject to change some. My current idea for how we would integrate with it today:
- We use an `ipcMain.handle` call on each route navigation to get the `navigationHistory.getActiveIndex()` and `navigationHistory.length()` values, and use those to track the navigation arrow buttons' forward/backward state in lieu of our `navigateHistory()` function call and `isForwardOrBackward`  logic, thus fixing #4408.
- When long-clicking or right-clicking on a navigation arrow, we can replace our `sessionNavigationHistory` logic with an `ipcMain.handle` call that gets the `getEntryAtIndex(index)` for the up to 15 entries that we want, and then we can get the corresponding `title` property for those. When one is clicked, I believe we should still be able to use `window.history.go()` to navigate to them without issue.

Note that the above implementation would be easier & improved if [these properties were to be exposed to developers](https://github.com/electron/electron/pull/41577#issuecomment-1998664101):
> In a separate PR, duplicate and move other navigationHistory related endpoints like CanGoBack, GoBack, CanGoForward, GoForward, CanGoToOffset, GoToOffset, CanGoToIndex, GoToIndex to the new namespace after the first PR merges.

Because of that, it may be reasonable to wait on the `NavigationHistory` side of things until it's fleshed out some more with those utilities that would simplify the implementation.

jasonhenriquez:
@ArthurKun21 That sounds interesting, although that's a non-standard pattern compared to the way that most browsers do it, so I'd prefer if we could try to meet that same need in more conventional ways. Do the existing ways added in the PR address your usecase? The main thing we're currently missing is "pull down to open," which is similar to what you describe.

simulified:
There is a problem
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 1000; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

efb4f5ff-1298-471a-8973-3d47447115dc:
> The alternative way I suggested should help this functionality to be known more

Imo making the functionality stand out just because to get it more known isnt a stong argument to create a non-standard way of doing it. For things like that we should create a dedicated docs page.

absidue:
> We use an ipcMain.handle call on each route navigation to get the navigationHistory.getActiveIndex() and navigationHistory.length() values, and use those to track the navigation arrow buttons' forward/backward state in lieu of our navigateHistory() function call and isForwardOrBackward logic, thus fixing #4408.

I've opened #5242 which fixes that issue without having to switch existing functionality to an Electron only API, as the Navigation API already provides helpful `canGoBack` and `canGoForward` properties.

> Note that the above implementation would be easier & improved if these properties were to be exposed to developers:
>  > In a separate PR, duplicate and move other navigationHistory related endpoints like CanGoBack, GoBack, CanGoForward, GoForward, CanGoToOffset, GoToOffset, CanGoToIndex, GoToIndex to the new namespace after the first PR merges.

They are already exposed to developers on `webContents` objects, hence the `duplicate and move` part of the message that you quoted, I use the `goBack` and `goForward` methods in the pull request that I opened.

As for why I would suggest an Electron API for the dropdowns after advising against using an Electron API, earlier on in this message, firstly the dropdowns are new functionality, so you wouldn't be breaking any existing functionality in the Android version the unofficial FreeTubeAndroid fork by using an Electron API for them and secondly the Android WebView has a navigation history API, which is slightly more powerful, [`WebBackForwardList`](https://developer.android.com/reference/android/webkit/WebBackForwardList) which can be obtained by calling [`copyBackForwardList()`](https://developer.android.com/reference/android/webkit/WebView#copyBackForwardList()) on the WebView.

jasonhenriquez:
Not sure if I understand the suggestion in full - use the Electron `NavigationHistory` class for the dropdown logic, right? Or is it that + use an Electron context menu instead of the `ft-icon-button` dropdown?

absidue:
Although a context menu is probably a lot easier to implement, there is some merit to using the icon button drop down, as it matches the style of the rest of the app. The only thing that we need to look out for, is that we only display the dropdowns in Electron and that outside of Electron they just act like back and forward buttons with no extra functionality (just like before this pull request) to avoid the user getting confused by things breaking, because the Electron API isn't available.

kommunarr:
Updated to use the `NavigationHistory` component. Almost done, just have to see what is blowing up in the `yarn run pack:web`. I have no earthly idea.

absidue:
@kommunarr Your problem wasn't `!process.env.IS_ELECTRON` but that you weren't wrapping the Electron only code in the if, webpack can only ignore imports if they are inside of the if-else. Early returns are too far away from the actual import for webpack to analyze them (otherwise it would have to meticulously analyze every single piece of code in the file which would take significantly longer and would have to be a lot more advanced to cover all possible edge cases).

kommunarr:
@absidue Thanks for sharing that. While that explanation makes sense, I'm certainly not too proud to admit that I wouldn't have figured that explanation out on my own just by looking at the code.

efb4f5ff-1298-471a-8973-3d47447115dc:
Got these errors on app launch

image

kommunarr:
Oop, it looks like the optional chaining got removed before I committed it. Fixed and revalidated now

PikachuEXE:
Same as before
pending idea from others especially on mobile

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Also not an issue of this PR but this labeling is very confusing when going to an post


https://github.com/user-attachments/assets/965e2fdd-704d-4d3f-af03-4b60b3244cac


This isnt an issue of your PR but seeing "search results" multiple times for different query's is confusing 


image

kommunarr:
I can make the document title for search results include the query (i.e.,`${query} - FreeTube`, similar to search browsers and YT search). Not sure I understand the second concern.

efb4f5ff-1298-471a-8973-3d47447115dc:
>Not sure I understand the second concern. 

When going to a community posts its says `FreeTube`

It also says `FreeTube` when doing a search now

kommunarr:
TIL we had navigable posts! Fixed both `FreeTube` issues, incl. on initial search navigation

Screenshot_20241123_151010

kommunarr:
The same type of issue exists with many of the pages. My plan is merge this, then #5003, then open a separate PR for adding the route icons added in the second to the nav history dropdown entries. Don't want to make these two larger than they currently are, though.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Icons are definitely needed...

PR #5250, opened 2024-06-10
Local API: Extract playlists on the auto-generated "Music" channel

absidue:
# Local API: Extract playlists on the auto-generated "Music" channel

## Pull Request Type

- [x] Feature Implementation

## Description
This pull request adds support for extracting and displaying playlists on the auto-generated system "Music" channel. Unlike #5241 which adds support for a whole category of channels, this pull request is only for one specific one.

I also noticed that some of the playlists listed there were showing `NaN views` on the playlists page, so I fixed that so it correctly shows `0 views` when YouTube says `No views`.

## Screenshots 
music-channel-with-playlists

## Testing 
Open `https://www.youtube.com/music` in FreeTube and notice the playlists tab.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 00509090456527e4418804faba4532aabf3dde35

### DIFF
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -340,7 +340,7 @@ export default defineComponent({
         this.playlistDescription = result.info.description ?? ''
         this.firstVideoId = result.items[0].id
         this.playlistThumbnail = result.info.thumbnails[0].url
-        this.viewCount = extractNumberFromString(result.info.views)
+        this.viewCount = result.info.views.toLowerCase() === 'no views' ? 0 : extractNumberFromString(result.info.views)
         this.videoCount = extractNumberFromString(result.info.total_items)
         this.lastUpdated = result.info.last_updated ?? ''
         this.channelName = channelName ?? ''
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -675,6 +675,24 @@ export function parseLocalListPlaylist(playlist, channelId = undefined, channelN
 }
 
 /**
+ * @param {import('youtubei.js').YTNodes.CompactStation} compactStation
+ * @param {string} channelId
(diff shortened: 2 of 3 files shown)


PikachuEXE (approved):
Example playlist for `no views`
https://youtube.com/playlist?list=RDCLAK5uy_k5n4srrEB1wgvIjPNTXS9G1ufE9WQxhnA

Or keep finding one in https://www.youtube.com/music

src/renderer/views/Playlist/Playlist.js
```diff
@@ -340,7 +340,7 @@ export default defineComponent({
         this.playlistDescription = result.info.description ?? ''
         this.firstVideoId = result.items[0].id
         this.playlistThumbnail = result.info.thumbnails[0].url
-        this.viewCount = extractNumberFromString(result.info.views)
+        this.viewCount = result.info.views.toLowerCase() === 'no views' ? 0 : extractNumberFromString(result.info.views)
```
jasonhenriquez, line 343: **suggestion (non-blocking):** We have a few places in the code where we use `extractNumberFromString` and it ends up getting used as `NaN`. I get from an "API" level why `NaN` is an appropriate return value, but it is this behavior that ultimately results in us displaying `NaN` throughout the app. I'd recommend returning `0` instead of `NaN` from that function in both the not-string case and `NaN` `parseInt` case just to cover the swathe of edge cases more easily.

absidue: While you are correct that in many places we don't handle `NaN` and just display it to the user, in other places we actually rely on it, so we know when to hide UI elements for example.

So we either need to create a separate wrapper function the converts the `NaN` to zero in the places where that is desired or add an extra parameter to control when the fallback should be zero.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5263, opened 2024-06-13
Fixes opening an external player on the watch page passes user playlist

absidue:
# Fixes opening an external player on the watch page passes user playlist

## Pull Request Type

- [x] Bugfix

## Related issue
https://github.com/FreeTubeApp/FreeTube/discussions/5188 (it's a discussion so the close keyword doesn't work)

## Description
The code to open an external player in the watch-video-info component was checking `this.inUserPlaylist` to decide whether or not to pass the playlist information to the external player. Unfortunately though `this.inUserPlaylist` didn't exist, so it was passing playlist information to the external player. This pull request adds the missing `inUserPlaylist` property.

## Screenshots 
Before:
before

After:
after

## Testing 
1. Add this code snippet to the top of the `openInExternalPlayer` in `src/renderer/store/modules/utils.js`
    ```js
    console.log(payload)
    return
    ```
3. Open FreeTube
4. Set your external player to anything that isn't `None` (you don't actually have to have the player installed, because of the code snippet above)
5. Watch a user playlist
6. Click the open in external player button in the video info section
7. Check the console and make sure the logged object doesn't contain any playlist information

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 6b2917ad271c0b5a3ab7c4add5969a97b3b81bd9

### DIFF
--- a/src/renderer/components/watch-video-info/watch-video-info.js
+++ b/src/renderer/components/watch-video-info/watch-video-info.js
@@ -102,6 +102,10 @@ export default defineComponent({
     videoThumbnail: {
       type: String,
       required: true
+    },
+    inUserPlaylist: {
+      type: Boolean,
+      default: false
     }
   },
   emits: ['change-format', 'pause-player', 'set-info-area-sticky', 'scroll-to-info-area'],
--- a/src/renderer/views/Watch/Watch.vue
+++ b/src/renderer/views/Watch/Watch.vue
@@ -119,6 +119,7 @@
         :get-playlist-loop="getPlaylistLoop"
         :length-seconds="videoLengthSeconds"
         :video-thumbnail="thumbnail"
+        :in-user-playlist="!!selectedUserPlaylist"
         class="watchVideo"
         :class="{ theatreWatchVideo: useTheatreMode }"
         @change-format="handleFormatChange"


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

jasonhenriquez (approved):

src/renderer/components/watch-video-info/watch-video-info.js
```diff
@@ -102,6 +102,10 @@ export default defineComponent({
     videoThumbnail: {
       type: String,
       required: true
+    },
+    inUserPlaylist: {
+      type: Boolean,
+      default: false
```
PikachuEXE, line 108: Why default false instead of make it required
The only place using `watch-video-info` is `src/renderer/views/Watch/Watch.vue`

absidue:
```suggestion
      required: true
```

absidue:
I've made the parameter required now.

jasonhenriquez (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #5273, opened 2024-06-17
Update icon for channels button on side nav

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/pull/5003#issuecomment-2161930504

## Description

Update icon for channels button so an icon similar to the existing one can be used for remote playlist in #5003

## Screenshots 

image
image
image

## Testing 

- Check the new icon in different modes

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/side-nav/side-nav.vue
+++ b/src/renderer/components/side-nav/side-nav.vue
@@ -42,7 +42,7 @@
           class="thumbnailContainer"
         >
           <font-awesome-icon
-            :icon="['fas', 'list']"
+            :icon="['fas', 'user-check']"
             class="navIcon"
             :class="applyNavIconExpand"
             fixed-width
--- a/src/renderer/main.js
+++ b/src/renderer/main.js
@@ -89,6 +89,7 @@ import {
   faTimes,
   faTimesCircle,
   faTrash,
+  faUserCheck,
   faUsers,
   faUsersSlash,
 } from '@fortawesome/free-solid-svg-icons'
@@ -188,6 +189,7 @@ library.add(
   faTimes,
   faTimesCircle,
   faTrash,
+  faUserCheck,
   faUsers,
   faUsersSlash,
 


jasonhenriquez:
**suggestion (non-blocking):** The alignment of the icons is now bugging me given that this new icon is pretty wide. Could we left-align the icons?

PikachuEXE:
image
image

I understand the content is wider than others but as it's already fixed width I doubt that left-align (I don't even know what it means) fixes anything

PikachuEXE:
Or just use icon without large asymmetry 

e.g. `user`
image

jasonhenriquez:
> left-align (I don't even know what it means)

Screenshot_20240617_072734

Here's how it looks like with a one-line change (the one highlighted) for reference. Needs a bit more tweaking for the other two cases, but just to demonstrate what I mean

PikachuEXE:
In that case I might as well post all screenshots for using `user` for easier comparison
image
image
image

jasonhenriquez:
I'd prefer `user-check` because it's more clearly conveying "subscribing" with the checkmark. `list` and `user` seem roughly equivalent in terms of how effectively they describe the meaning. I will leave it at that and go along with whichever decision we make for this.

PikachuEXE:
I am not quite sure that I agree with the left align for icons (for text I can understand)
Due to varying gap especially for `Trending` & `Playlists`
image
image

Youtube got no such issue
image

Still finding better solutions

jasonhenriquez:
YouTube doesn't seem to have an issue because the icons are all very similar in size. We could try scaling this particular icon to a small width (with height `auto`)

PikachuEXE:
Width reduced from 1.25em to 1em:
image
image
image

jasonhenriquez:
Ooh! We're very close now. That looks good on everything except the Hide Labels view, where the torso of the icon is poking out. Maybe we need some individual & commented patch for this specific Hide Labels case & icon.

PikachuEXE:
height reduced from 1.3em > 1.15em
image

jasonhenriquez (approved):
LGTM! Please update the OP with the newest screenshots when you have a sec!

jasonhenriquez (commented):
Please update with the suggested change

jasonhenriquez (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
I do agree that this icon is way better than the one we had before but im after all those changes im still a bit bothered by the alignment. If this is the best that can be done than i just have to deal with it but i just wanted to point out that it still looks a bit off to me

jasonhenriquez:
I think you can make an argument for making the bounding box of the icon centered around the `user` torso, and having the checkmark part not included in the positioning check. But I don't know of a good way to do that at the moment.

PikachuEXE:
Manually adjusted the horizontal position
If it looks off please test locally to find the value that look right for you and share :)
image
image
image

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Icon in mobile view needs to be adjusted too

Capture

PikachuEXE:
Updated now, did not see your message (probably deleted accidentially)
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

ChunkyProgrammer (approved):

PR #5289, opened 2024-06-18
Avoid cloning all profiles when subscribing and unsubscribing

absidue:
# Avoid cloning all profiles when subscribing and unsubscribing

## Pull Request Type

- [x] Performance improvement

## Description
Here is a summary of the main improvements:
- Update just the subscriptions list in the affected profiles, instead of creating a copy/clone of all profiles, even if they weren't affected
- Subscribing in multiple profiles and unsubscribing in multiple profiles now only requires on database call and window sync per batch update
- Only send the necessary data between windows.

## Testing 
Subscribe to and unsubscribe from channels in invididual profiles as well as the "All Channels" one. Additionally make sure that the changes are correctly synced between windows.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 81cf8177ba816c6b913855a9d7df3f99edbe8731

### DIFF
--- a/src/renderer/components/ft-subscribe-button/ft-subscribe-button.js
+++ b/src/renderer/components/ft-subscribe-button/ft-subscribe-button.js
@@ -6,7 +6,7 @@ import FtButton from '../../components/ft-button/ft-button.vue'
 import FtPrompt from '../../components/ft-prompt/ft-prompt.vue'
 
 import { MAIN_PROFILE_ID } from '../../../constants'
-import { deepCopy, showToast } from '../../helpers/utils'
+import { showToast } from '../../helpers/utils'
 import { getFirstCharacter } from '../../helpers/strings'
 
 export default defineComponent({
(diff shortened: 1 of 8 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

src/renderer/components/ft-subscribe-button/ft-subscribe-button.js
```diff
@@ -169,30 +165,34 @@ export default defineComponent({
...
-        }
+      this.removeChannelFromProfiles({
+        channelId: this.channelId,
+        profileIds
+      })
+
+      showToast(this.$t('Channel.Channel has been removed from your subscriptions'))
+
+      if (profile._id === MAIN_PROFILE_ID && profileIds.length > 1) {
+        showToast(this.$t('Channel.Removed subscription from {count} other channel(s)', {
```
jasonhenriquez, line 193: **suggestion (non-blocking, minor):** Can do as a separate PR, as this is a pre-existing "issue", but we might as well pluralize the label instead of doing `(s)`

src/renderer/store/modules/profiles.js
```diff
@@ -214,6 +232,22 @@ const mutations = {
...
+      state.profileList.find(profile => profile._id === id).subscriptions.push(channel)
+    }
+  },
+
+  removeChannelFromProfiles(state, { channelId, profileIds }) {
+    for (const id of profileIds) {
+      const profile = state.profileList.find(profile => profile._id === id)
+
+      // use filter instead of splice in case the subscription appears multiple times
+      // https://github.com/FreeTubeApp/FreeTube/pull/3468#discussion_r1179290877
```
jasonhenriquez, line 246: **question (non-blocking, minor):** Doesn't have to be considered for this PR, but wouldn't a user know that they're subscribed to a channel multiple times? That + the amount of time since this bug has been fixed (unsure exactly when) means maybe this is safe to make a `splice` now.

jasonhenriquez (approved):

PR #5339, opened 2024-06-25
Set a few more headers for local API requests

absidue:
# Set a few more headers for local API requests

## Pull Request Type

- [x] Feature Implementation

## Description
This adds a few more headers to match YouTube, which should hopefully help make the local API requests look more legitimate.

## Testing 
Open a video with the local API and check in the dev tools that the /player request has the two headers added in this pull request.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.21.0


kommunarr (approved):
For those not in the know (like myself), we're currently sending a `sec-fetch-mode` of `cors`.

ChunkyProgrammer (approved):

PR #5348, opened 2024-06-29
Larger search/URL bar

### LINKED ISSUE #5341: [Feature Request]: Larger search/URL bar
### Guidelines

### Problem Description

Space waste

### Proposed Solution

Make search/URL bar (top of window) larger when there is enough space.

### Alternatives Considered

only extend search bar when "Hide FreeTube Header Logo" is enabled

### Issue Labels

visual improvement

### Additional Information

There will be a lot of discussion ;)

4yman-el:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue
closes #5341

## Description
modify `top-nav` SCSS so that the grid gives more space to search bar. This PR also increases its max width.

## Screenshots 
before:
before
after (manually using DevTools, I don't want to build):
after

## Desktop

- **OS:** Linux Mint (does not apply)
- **OS Version:** 21.3
- **FreeTube version:** v0.21.0 Beta

## Additional context

I also want to add `flex-shrink` for first item for "hide FreeTube logo" because SPACE WASTE.

### DIFF
--- a/src/renderer/components/top-nav/top-nav.scss
+++ b/src/renderer/components/top-nav/top-nav.scss
@@ -23,7 +23,7 @@
 
   @media only screen and (width >= 961px) {
     display: grid;
-    grid-template-columns: 1fr 440px 1fr;
+    grid-template-columns: 1fr 1.5fr .5fr;
   }
 
   @include top-nav-is-colored {
@@ -190,7 +190,7 @@
 
 .middle {
   flex: 1;
-  max-inline-size: 440px;
+  max-inline-size: 720px;
 
   .searchContainer {
     align-items: center;


src/renderer/components/top-nav/top-nav.scss
```diff
@@ -23,7 +23,7 @@
 
   @media only screen and (width >= 961px) {
```
kommunarr, line 24: Thanks @4yman-el! Only feedback: It does look a bit odd having the surrounding spacing change. How about this?
```
media only screen and (width >= 1122px) {
        display: grid;
        grid-template-columns: 1fr 720px 1fr;
    }
```

Screenshot_20240629_135756

4yman-el: It looks better. Thanks!

kommunarr: Also update the width breakpoint to 1122px to prevent intermediate breakage. Sorry, I couldn't originally offer it as a multiline `suggestion` because I was in unified view.

4yman-el: No problem!

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue:
Looks like this pull request was only designed with wide window sizes in mind which results in this bug: https://github.com/FreeTubeApp/FreeTube/issues/5540

@4yman-0 Please create a follow up pull request that only uses the changes from this pull request on wide window sizes, to fix that bug. Would be nice if we didn't have to revert your pull request to fix the bug.

deepspaceaxolotl:
I don't think that's quite it, because even on wide window sizes it appears off-center, because it's centered using grid across the whole width of the top bar (and thus relative to the viewport, rather than the content area as would feel more natural (#5540)).

Issue #4872, opened 2024-04-04
[Feature Request]: Drag and drop videos in edit mode


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

When your playlist becomes very big it becomes annoying to move a video to a certain spot u like.

### Proposed Solution

Implement something like this in the playlist edit mode 

https://github.com/PikachuEXE/FreeTube/assets/73130443/c3af06b5-4381-4fe6-a8a8-7bd3c81582d2

### Alternatives Considered

None.

### Issue Labels

ease of use improvement, improvement to existing feature

kommunarr:
Not sure if this should be included as a separate suggestion, but we could also add "transient" ordering changes like YT has today to the non-Custom sort orders.

jschuhmann47:
Hey there, I wanna ask how hard this would be to implement as a first contributor. It's a feature that I'd love to have. However I've never contributed in this repo so I don't know the difficulty this might have, but I'm willing to give it a shot.

PR #5423, opened 2024-07-16
Add: Sort Playlist Videos by Length / Duration fixed Issue#5268

creotove:
# Title

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/helpers/playlists.js
+++ b/src/renderer/helpers/playlists.js
@@ -5,6 +5,8 @@ export const SORT_BY_VALUES = {
   AuthorDescending: 'author_descending',
   VideoTitleAscending: 'video_title_ascending',
   VideoTitleDescending: 'video_title_descending',
+  VideoLengthAscending: 'video_length_ascending',
+  VideoLengthDescending: 'video_length_descending',
   Custom: 'custom'
 }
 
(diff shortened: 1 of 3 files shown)


absidue:
As I mentioned on that issue, you will have to handle that the duration can be completely missing.

creotove:
@absidue ohh sorry, there is no way that we can sort them properly, but still we can sort them by optionally checking the property and if the there is no property of lengthSeconds then that video will be shown to the top/bottom of the playlist, should I apply this? If any other proposal you need to add to implement in this feature you can add that.

absidue:
If you still want to proceed with this pull request, it definitely needs some handling for videos without durations.

Here are my two suggestions, feel free to pick one or suggest alternative ideas:
1. Hide the sort by duration option when at least one video in the playlist is missing it's duration.
2. Your suggestion of placing videos without duration at the top or bottom of the list depending on the sort direction, with the addition of a message warning the user that the sorting is inaccurate due to one or more videos in the playlist missing their duration.

creotove:
@absidue okk sure i can do that and as i read your alternative consideration in the main issue that user need to manually click the "up" button several times on each video to sort videos according to user need, we can implement a drag and drop feature there instead of manual buttons.

efb4f5ff-1298-471a-8973-3d47447115dc:
> we can implement a drag and drop feature there instead of manual buttons.

I think that is out of scope of this PR though but if you want to look at it, see https://github.com/FreeTubeApp/FreeTube/issues/4872

PR #5438, opened 2024-07-21
More flexible hardware acceleration for Linux

### LINKED ISSUE #5437: [Bug]: Hardware video acceleration via VA-API does not work
### Guidelines

### Describe the bug

Hardware video acceleration via VA-API on Chromium/Electron often requires additional patches but it has been working well recently.

With https://github.com/FreeTubeApp/FreeTube/commit/d2f14b072a1fadf81ce49a7516bda2bfd7ce3860 freetube started enforcing a parameter to enable it Linux. Currently this breaks video acceleration as a different set of parameters is required to enable VA-API and it is impossible to disable the harcoded `VaapiVideoDecodeLinuxGL` setting.

freetube should not hardcode these settings. They should be handled by Electron or left to the user as the parameters can vary depending on the video driver. 

Reproduction steps with the AMDGPU driver on X11:
- disable DASH AV1 formats in Player Settings
- open a sample video with `freetube --enable-features=Vulkan,VulkanFromANGLE,DefaultANGLEVulkan,VaapiIgnoreDriverChecks https://youtu.be/LXb3EKWsInQ`
- open Developer Tools with F12, add the Media tab
- check the Decoder name of the currently playing video
- a software decoder is used. The video decoder name shows `FFmpegVideoDecoder`

### Expected Behavior

A hardware accelerated video decoder is used. The video decoder name shows `VaapiVideoDecoder`.

### Issue Labels

feature stopped working

### FreeTube Version

v0.21.1 Beta

### Operating System Version

Arch Linux

### Installation Method

AUR (Unofficial)

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

xsmile:
# More flexible VA-API hardware video acceleration for Linux

## Pull Request Type
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #5437.

## Description

Video hardware acceleration is only enabled if no other Electron feature flags are found. The OpenGL mode is used by default.

Users can now override the feature flag list passed to Electron and e.g. use Vulkan instead of OpenGL.

## Testing

Testing was done with both the bundled Electron library in `freetube-0.21.1-linux-portable-x64.zip` and with the system package `electron31` from Arch Linux.

**Using the official freetube portable package**
- `./freetube` - no feature flags are found, `VaapiVideoDecodeLinuxGL` is enabled - VA-API does not work (in my case). This is the default behavior.
- `./freetube --enable-features=Vulkan,VulkanFromANGLE,DefaultANGLEVulkan,VaapiIgnoreDriverChecks` - custom feature flags are detected and left unmodified - VA-API works

**Using the system package electron31**
- `electron31 --enable-features=Vulkan,VulkanFromANGLE,DefaultANGLEVulkan,VaapiIgnoreDriverChecks resources/app.asar` - custom feature flags are detected and left unmodified - VA-API works

## Desktop
- **OS:** Arch Linux
- **OS Version:** rolling
- **FreeTube version:** v0.21.1 Beta
- **GPU driver:** amdgpu
- **Display server:** X

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -199,9 +199,11 @@ function runApp() {
   let startupUrl
 
   if (process.platform === 'linux') {
-    // Enable hardware acceleration via VA-API
+    // Enable hardware acceleration via VA-API with OpenGL if no other feature flags are found
     // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/vaapi.md
-    app.commandLine.appendSwitch('enable-features', 'VaapiVideoDecodeLinuxGL')
+    if (!app.commandLine.hasSwitch('enable-features')) {
+      app.commandLine.appendSwitch('enable-features', 'VaapiVideoDecodeLinuxGL')
+    }
   }
 
   const userDataPath = app.getPath('userData')


absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue:
@xsmile Do you know if there is a way that we could detect which flags would be the right ones to use? Ideally we would have it working correctly out of the box, without users having to pass their own flags to get it to work.

xsmile:
@absidue I'm not sure if there is an easy and reliable way to do this.

In general the following requirements must be met:
- VA-API library is installed
- Electron is built with VA-API support: this should be the default

Additionally you will need to differentiate between OpenGL and Vulkan modes:
- OpenGL doesn't require additional workarounds
- Vulkan requires an up to date [Mesa package](https://wiki.archlinux.org/title/Chromium#Hardware_video_acceleration) >= 24.1 and a [patch](https://gitlab.archlinux.org/archlinux/packaging/packages/electron31/-/blob/main/allow-ANGLEImplementation-kVulkan.patch?ref_type=heads) for Electron if using Wayland

The GPU vendor is relevant as well. Intel is known to be less problematic, AMD and NVIDIA might require more workarounds. As of now, acceleration on AMD GPUs seems to only work with Vulkan.

Alternatively, you could use pre-set flags for both OpenGL and Vulkan and let the user decide on the mode to be used via toggles on the Settings page. I would use the flags provided on the [official Chromium page](https://chromium.googlesource.com/chromium/src/+/master/docs/gpu/vaapi.md#vaapi-on-linux) as a reference.

You can also find some user feedback on this topic at the [Arch Linux forum](https://bbs.archlinux.org/viewtopic.php?id=244031)

xeleven11:
I use nvidia but none of the flags in previous post dont work. I have freetube v0.21.3 installed and do i have to install electron31 too?

absidue:
@xeleven11 To quote Chromium's own documentation on VA-API (https://chromium.googlesource.com/chromium/src/+/master/docs/gpu/vaapi.md#vaapi-on-linux):
> The NVIDIA VaAPI drivers are known to not support Chromium

xeleven11:
@absidue i have already read your linked document but there are some people here saying hardware decoding works from nvidia driver. im not sure whether their driver actually works with freetube or got confused with software decode.

PR #5455, opened 2024-07-24
Move the channel details into their own component

absidue:
# Move the channel details into their own component

## Pull Request Type

- [x] Refactor

## Related issue
https://github.com/orgs/FreeTubeApp/projects/4?pane=issue&itemId=68600889

## Description
This pull request moves the channel details section (header and tabs) into their own component, continuing the efforts to split the channel page up into multiple components.

As we'll hopefully be moving the Vue 3 soon, I decided to the use the `<script setup>` composition API style for this new component. That means inside that component we can already benefit from better auto-complete and once we upgrade to Vue 3 we'll also get better performance and bundle size improvements.

## Testing 
* Open various channels and make sure they are displayed correctly
* Open an invalid channel (e.g. copying the URL of a valid channel and removing a few characters from the ID) and make sure it appears correctly.
* Switch from one channel directly to another (e.g. by clicking another channel on the side bar while you are already viewing a different channel) and ensure that it updates correctly.
* Check that the `Hide Channel Subscribers` distraction free setting hides the subscriber count.
* Check that the `Hide Unsubscribe Button` parental control setting hides the subscribe/unsubscribe button.
* Check that the various channel tab hiding distraction free settings work.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0e72be464e76d5a0c64fbde98f133d2e7e8bb2fb

### DIFF
--- a/src/renderer/views/Channel/Channel.js
+++ b/src/renderer/views/Channel/Channel.js
@@ -1,24 +1,20 @@
 import { defineComponent } from 'vue'
 import { mapActions } from 'vuex'
 import FtCard from '../../components/ft-card/ft-card.vue'
-import FtInput from '../../components/ft-input/ft-input.vue'
 import FtSelect from '../../components/ft-select/ft-select.vue'
 import FtFlexBox from '../../components/ft-flex-box/ft-flex-box.vue'
 import FtLoader from '../../components/ft-loader/ft-loader.vue'
 import FtElementList from '../../components/ft-element-list/ft-element-list.vue'
 import FtAgeRestricted from '../../components/ft-age-restricted/ft-age-restricted.vue'
-import FtShareButton from '../../components/ft-share-button/ft-share-button.vue'
-import FtSubscribeButton from '../../components/ft-subscribe-button/ft-subscribe-button.vue'
 import ChannelAbout from '../../components/channel-about/channel-about.vue'
+import ChannelDetails from '../../components/ChannelDetails/ChannelDetails.vue'
 import FtAutoLoadNextPageWrapper from '../../components/ft-auto-load-next-page-wrapper/ft-auto-load-next-page-wrapper.vue'
 
 import autolinker from 'autolinker'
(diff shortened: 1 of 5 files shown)


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/views/Channel/Channel.js
```diff
@@ -50,20 +46,18 @@ export default defineComponent({
...
     'ft-select': FtSelect,
     'ft-flex-box': FtFlexBox,
     'ft-loader': FtLoader,
     'ft-element-list': FtElementList,
     'ft-age-restricted': FtAgeRestricted,
-    'ft-share-button': FtShareButton,
-    'ft-subscribe-button': FtSubscribeButton,
     'channel-about': ChannelAbout,
     'ft-auto-load-next-page-wrapper': FtAutoLoadNextPageWrapper,
+    ChannelDetails
```
ChunkyProgrammer, line 56: Is there a reason it's `ChannelDetails` and not `channel-details` like how we do every other component? Is there a plan to update other components to not use the dashes and just use the component name?

absidue: I was mostly inspired by this section https://vuejs.org/guide/components/registration.html#component-name-casing. As for updating the other components, that's a vague plan I have but I thought we could do it at the same time we convert the respective components to the SFC setup style composition API, that way we can already tell from the file name what has been migrated. That's a long term plan though, there is no rush to convert everything straight away, I just thought it would be good to use the SFC style for that component as I was creating a new component anyway and it means we won't have to do it later on.

ChunkyProgrammer: Okay, sounds good, as long as we're planning to make it consistent eventually 🙂

ChunkyProgrammer (approved):

Issue #5057, opened 2024-05-03
[Feature Request]: Import Playlists from YouTube


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Importing Playlists from YT is not possible. Now that Local Playlist support has released we can start to implement this.

### Proposed Solution

The ability to import the Playlist file provided by YouTube into FreeTube

Also what we need from the exported YT playlist file to test cases like:

- videos that are set to private in that playlist 
- videos that dont exist anymore
- shorts and previously streamed live videos
- Currently live videos (not sure if YT allows u to save that to a playlist)
- Upcoming videos or upcoming live videos (not sure if YT allows u to save that to a playlist)

### Alternatives Considered

None.

### Issue Labels

new feature

### Additional Information

Retrieved from 
https://github.com/PikachuEXE/FreeTube/issues/50#issuecomment-1734687322

> The YT playlist export is a list of CSV files
> 
> Example (header text depends on user language I guess, maybe it can be forced via `hl=en`) Headers: video ID, timestamp for video added to playlist time
> 
> ```
> 影片 ID,播放清單影片的建立時間戳記
> dKMTzrG92TE,2018-04-24T09:05:28+00:00
> ```
> 
> Update 1: Confirmed that https://takeout.google.com/?hl=en&pli=1 will export file with header text in `en`
> 
> ```
> Video ID,Playlist Video Creation Timestamp
> dKMTzrG92TE,2018-04-24T09:05:28+00:00
> ```

efb4f5ff-1298-471a-8973-3d47447115dc:
> Would it be possible to append to this feature. The ability to save the URL of the playlist in which the playlist comes from.
> 
> And the ability to re-sync again, like for example there is a new button to sync things up with the remote playlist so that once you have click the button it would fetch any new items from that youtube playlist

See #312

NoSpiner:
https://github.com/NoSpiner/Freetube-import

[...]

PR #5498, opened 2024-07-29
Support for youtube playlist imports in cvs

Sp3rick:
Main Change:
Support for youtube playlist imports in cvs

Others:
moved some code to parseCsv
added updatePlaylistVideo for update single playlist video on database added preference.js to get unique dataset results using the preferred api

# Title

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


absidue:
I have both reviewed the code or tested this pull request yet, however I have noticed that you haven't filled in the pull request template properly including the rather important testing section which for something like this should also include example files that reviewers can use to test the pull request.

efb4f5ff-1298-471a-8973-3d47447115dc:
This could potentially close #5057

absidue:
I have not tested this but looking at the code alone this pull request looks like it'll require an overhaul before it could be merged.

* The `playlists.csv` file contains all the metadata about the playlists, from what I can tell this pull request doesn't have any logic for that. It would probably be better to create a separate button for the YouTube playlists import, then prompt the user to selected the metadata file first and then afterwards prompt them to select the csv files for the playlists they want to import.
* You should avoid making requests to YouTube as much as possible, during an import like this you'll likely be sending hundreds if not thousands of requests in rapid succession, which will most certainly get you ratelimited or worse (the subscriptions import already has that problem and that only makes one request per channel, which is why in the long term we want to refactor the subscriptions import to not make any requests at all). Before doing any requests you should be checking every possible cache and store (e.g. watch history, existing playlists, subscriptions cache, search cache, trending cache, popular cache etc), making requests to YouTube or Invidious should be the last resort.
* While your preferences.js idea sounds great on paper, in practice it means that with the local API for example you'll be making about 6 requests per video (more in the future once we add support for YouTube's poToken stuff). For the import you only need one request per video because you only need a few metadata fields, whereas the watch page needs to create a real session on YouTube make extra requests to get playable streaming URLs, fetch the full description and recommend videos, as well as attempt to bypass age-restrictions. For the local API you should be creating a dedicated function that calls YouTube.js' getBasicInfo function).

PR #5546, opened 2024-08-08
Migrate FtElementList, FtAutoGrid and FtListLazyWrapper to the composition API

absidue:
# Migrate FtElementList, FtAutoGrid and FtListLazyWrapper to the composition API

## Pull Request Type

- [x] Feature Implementation
- [x] Refactoring

## Description
This pull request starts migrating some of our existing components to the composition API, specifically FtElementList, FtAutoGrid and FtListLazyWrapper. I picked those three components to start with as they are quite simple compared to some of the other components we have in the project and I could migrate them without causing any extra work for the Vue 3 migration.

What do I mean by not causing any extra work? When we upgrade to Vue 3, these three components should just work without requiring any changes. Other components that use translations inside the JavaScript or access the current locale, would require changes during the Vue 3 migration if they were to already get migrated to the composition API. To migrate to the composition API I would need to switch from `this.$t()` to importing the `i18n` object and using `i18n.t()`, but in Vue 3 that would then have to be changed to `i18n.global.t()`.
Of course we would eventually want to switch to the composition API support for vue-i18n (`const { t } = useI18n()`) but that has breaking changes, such as removing `tc()` and replacing it with `t()`, so we likely want to do that in a separate step after the initial Vue 3 migration.

## Testing 
1. Test that the grid and list layouts (`Video View Type` setting) still work correctly e.g. on the subscriptions page
2. Test that the various video filtering options such as the channel ID filter and the video title filter still work correctly in the search results.


PikachuEXE (approved):
I open 2 tabs to compare code during review

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Issue #5397, opened 2024-07-11
[Bug]: Massive leap in CPU usage between 19.2 and 20.0, and another jump to 21.1


libv:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

This is a playback issue, there seems to be no difference between local API and invidious. Perhaps something is wrong with hw accel?

I have compared the same video, selected from history, same window size, same resolution, same playback speed, with back to back installs from the listed versions, sometimes restarting to make sure channel info and such is cached.

19.1: struggles to start a stream these days.
real	 3m51.779s
user 1m27.945s
sys	 0m9.666s

19.2:
real	 3m41.153s
user 1m33.223s  --> comparison basis
sys	 0m10.070s

20.0:
real	 3m43.764s
user 3m28.316s --> 2.2x more than 19.2
sys	 0m11.600s

21.1: fails to play back with both local and invidious.

21.1beta:
real	 3m39.247s
user 4m14.308s --> 2.7x more than 19.2
sys	 0m9.562s

### Expected Behavior

Broadly similar CPU usage.

### Issue Labels

usability issue

### FreeTube Version

v0.20.0

### Operating System Version

Debian 11.6.0 (oldstable), amd64, 5.10-28 kernel. Intel 5th gen. i915.ko, xf86-video-modesetting, xfce.

### Installation Method

.deb

### Primary API used

Invidious API

### Last Known Working FreeTube Version (If Any)

v19.2

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
[...]

PR #5550, opened 2024-08-09
Add a GPU Internals entry to the View menu

absidue:
# Add a GPU Internals entry to the View menu

## Pull Request Type

- [x] Feature Implementation

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/5397#issuecomment-2266713946

## Description
As Linux users are having lots of graphical issues and hardware acceleration problems, people have been asking how to access the chrome://gpu page, my original suggestion was to paste a code snippet into the devtools (see link above), which is okay for one time use. However due to the many problems Linux users seem to be having, that debug page is likely going to see more usage in the future. This pull request makes it a lot easier to access that page by adding an entry to under the View menu in the app menu, which opens a new window to that page.

## Screenshots 
view-menu-with-GPU-internals-entry

## Testing 
Activate the View -> GPU Internals (chrome://gpu) menu item and make sure it opens a new window with the GPU internals page.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1513,6 +1513,16 @@ function runApp() {
               }
             }
           },
+          {
+            label: 'GPU Internals (chrome://gpu)',
+            click() {
+              const gpuWindow = new BrowserWindow({
+                show: true,
+                autoHideMenuBar: true
+              })
+              gpuWindow.loadURL('chrome://gpu')
+            }
+          },
           { type: 'separator' },
           { role: 'resetzoom' },
           { role: 'resetzoom', accelerator: 'CmdOrCtrl+num0', visible: false },


libv:
Exposing more info through the freetube UI is a good idea, but, chrome://gpu has an extremely high noise to signal ratio, and has only very limited usefulness.

The media info xsmile pointed me at just now is the actual key.

It makes more sense to provide the name of the video decoder, while decoding, and the messages log instead (for some reason my 19.2 is now telling me that vaapi initialization failed whereas 2x.x just goes straight to ffmpeg -- fun, fun). This info should be provided on a per video basis, as the next video might have a different outcome.

A theoretical example might be that the hw decoder does not support 360p as it might need 16pixel alignment (even though block size is a multiple of 16x16 usually), perhaps there is a hw quirk where decoded height only works where height is a multiple of 16. In such a case switching video resolutions to 360p will then cause a switch from hw decoding to software decoding.

With chrome://gpu you get a lot of noise, and nothing useful will be in there. With the media specific info, and the matching messages you get the info you need to start debugging.

absidue:
This just adds a way to access the chrome://gpu page, as you can see in the screenshot it's literally just on extra button, as the only way to currently do it pasting a code snippet into the dev tools, that likely won't work in the future. If you want to view the Media tab in the dev tools you can still do it easily yourself, this pull request doesn't change anything about that.

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5590, opened 2024-08-20
Display video tags for Invidious

ChunkyProgrammer:
# Title

## Pull Request Type
- [x] Feature Implementation

## Related issue
Implements changes from https://github.com/FreeTubeApp/FreeTube/pull/5119 when using the invidious API

## Screenshots
image
image

## Testing
- set api preference to Invidious
- search "8K"
- see some videos with 8k tag appear
- search "4k" and enable "3d" filter
- see videos with 3d tag, vr180 tag, 360vr tag

## Desktop
- **OS:** Linux Mint
- **OS Version:** 22


static/locales/en-US.yaml
```diff
@@ -111,6 +116,7 @@ Search Filters:
     Location: Location
     HDR: HDR
     VR180: VR180
+    New: New
```
absidue, line 119: Was this accidentally added? I don't see any changes to the search filters code to use this string.

ChunkyProgrammer: Oh yeah, I will remove that

absidue (approved):

PikachuEXE (approved):
Wonder why Local API has no such tags...

absidue:
The local API has a few, it's just that until now nobody has needed all of them (pretty sure FreeTube is the first program to actually use them) so nobody added support for all the other ones in YouTube.js. Generally people don't spend time adding functionality to YouTube.js unless they actually need it.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5592, opened 2024-08-20
Add a new setting to open links in a new window

### LINKED ISSUE #1501: Add option to always open a redirected video in a new window / instance
Currently, when I click on a YouTube link in Firefox while another video is currently playing in FreeTube, the current video stops and FreeTube opens the new link in the one FreeTube instance / video currently open. I'm using the Privacy Redirect extension.

Give an option to always open redirects in a second window / instance (and not autostart the second video). This would resemble the behavior of a browser, because clicking a YouTube link while another video is playing in your browser would typically open a second tab instead of stopping a currently running YouTube video. At least that's the behavior I'm used to.

I discovered that a second window can be opened manually by clicking on a video link within FreeTube (video suggestions on the right hand side, etc.). When I watch my videos on this "second" instance, redirects from the browser don't interrupt the video. But this is rather complicated for everyday use.

pakoito:
# Title

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #1501 

## Description
It adds a global setting that makes `freetube://` links be opened in a new window

## Testing 

I have not been able to pull the dependencies from yarn, as it errors in one of them. Thus, I have not been able to test the generation of the settings or whether the implementation is correct.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.21.3 beta

## Additional context
The PR is as-is for someone else to finalize and land.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -200,6 +200,10 @@ function runApp() {
   let mainWindow
   let startupUrl
 
+  function isOpenLinkInNewWindow() {
+    return sessionStorage.getItem('openLinkInNewWindow') === "true"
+  }
+
   if (process.platform === 'linux') {
     // Enable hardware acceleration via VA-API with OpenGL if no other feature flags are found
     // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/vaapi.md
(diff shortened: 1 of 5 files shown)


src/main/index.js
```diff
@@ -1369,10 +1377,16 @@ function runApp() {
   app.on('open-url', (event, url) => {
     event.preventDefault()
 
+    const startUrl = baseUrl(url)
+
     if (mainWindow && mainWindow.webContents) {
-      mainWindow.webContents.send(IpcChannels.OPEN_URL, baseUrl(url))
+      if (isOpenLinkInNewWindow()) {
+        createWindow({ replaceMainWindow: false, windowStartupUrl: startUrl, showWindowNow: true })
```
pakoito, line 1384: This is probably an IPC to open a new window.

src/main/index.js
```diff
@@ -820,7 +824,11 @@ function runApp() {
 
   ipcMain.once(IpcChannels.APP_READY, () => {
     if (startupUrl) {
-      mainWindow.webContents.send(IpcChannels.OPEN_URL, startupUrl)
+      if (isOpenLinkInNewWindow()) {
```
pakoito, line 827: This should probably check if there's already content on the window.

PR #5624, opened 2024-08-30
Specify locale for strings that may differ from the chosen display locale

absidue:
# Specify locale for strings that may differ from the chosen display locale

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description
The [MDN page for the HTML `lang` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#accessibility_concerns), mentions that the Web Content Accessibility Guidelines require the page language to be specified and also that parts of the page in other languages require their language to be specified too, so that screenreaders can correctly pronounce and translate the text. FreeTube already fullfills the first part, as we set the `lang` attribute on the HTML element, however we don't currently fullfill the latter part.

This pull request adds the `lang` attribute where we know that the language of the text may differ from the user's chosen locale preference and we know the language of the text e.g. locale preference selector (the locale names are in their own language), changelog (always in English). It does not add it in places where it may differ but we don't know the language of the text such as any text received from YouTube e.g. video titles, descriptions, channel names, comments etc, for those we just have to hope that the user is watching content in the same language as their locale preference.

I decided to test the before and after with the Narrator built into Windows 10 with both the narrator and FreeTube set to English and was surprised at the results. Before these changes, for languages with non-Latin characters in their names such as Arabic and Bulgarian, it just annouced their position in the select box (x out of total), so you have no idea what you are currently focused on, whereas afterwards it announced the position and the English translation of the language name.

## Screenshots 
locale preference selector html with lang attributes

### Quick and easy:
With the dev tools check that the `<option>` tags in the local preference select box have appropriate `lang` attributes.

### Getting your hands dirty
Use a screen reader of your choice (e.g. built-in Narrator on Windows or VoiceOver on macOS) and see how it handles the newly added `lang` attributes.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 492f2245d16d7d6a5d0da4a088095c3be39e5dd8

### DIFF
--- a/src/renderer/views/About/About.js
+++ b/src/renderer/views/About/About.js
@@ -21,7 +21,7 @@ export default defineComponent({
         {
           icon: ['fab', 'github'],
           title: this.$t('About.Source code'),
-          content: `<a href="https://github.com/FreeTubeApp/FreeTube">GitHub: FreeTubeApp/FreeTube</a><br>${this.$t('About.Licensed under the')} <a href="https://www.gnu.org/licenses/agpl-3.0.en.html">${this.$t('About.AGPLv3')}</a>`
+          content: `<a href="https://github.com/FreeTubeApp/FreeTube" lang="en">GitHub: FreeTubeApp/FreeTube</a><br>${this.$t('About.Licensed under the')} <a href="https://www.gnu.org/licenses/agpl-3.0.en.html">${this.$t('About.AGPLv3')}</a>`
         },
         {
           icon: ['fas', 'file-download'],
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -23,6 +23,7 @@
     >
       <span
         class="changeLogText"
+        lang="en"
         v-html="updateChangelog"
       />
       <ft-flex-box>
--- a/src/renderer/components/ft-select/ft-select.js
+++ b/src/renderer/components/ft-select/ft-select.js
@@ -47,6 +47,10 @@ export default defineComponent({
     iconColor: {
       type: String,
       default: null
+    },
+    isLocaleSelector: {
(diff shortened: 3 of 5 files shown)


src/renderer/views/About/About.js
```diff
@@ -21,7 +21,7 @@ export default defineComponent({
         {
           icon: ['fab', 'github'],
           title: this.$t('About.Source code'),
-          content: `<a href="https://github.com/FreeTubeApp/FreeTube">GitHub: FreeTubeApp/FreeTube</a><br>${this.$t('About.Licensed under the')} <a href="https://www.gnu.org/licenses/agpl-3.0.en.html">${this.$t('About.AGPLv3')}</a>`
+          content: `<a href="https://github.com/FreeTubeApp/FreeTube" lang="en">GitHub: FreeTubeApp/FreeTube</a><br>${this.$t('About.Licensed under the')} <a href="https://www.gnu.org/licenses/agpl-3.0.en.html">${this.$t('About.AGPLv3')}</a>`
```
PikachuEXE, line 24: Should `lang="en"` be added to other links too?

absidue: Any link in particular that you have in mind? I added it to this one because the display text is English, form what I can tell the rest of them on the about page either have the link itself as the display text or are translated, but I might have missed something.

PikachuEXE: Oh they are translated? Did not check :)

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Issue #5268, opened 2024-06-15
[Feature Request]: Sort Playlist Videos by Length


enrrari:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Not sure whether it's better to bring this up on #4921, which addresses the topic of sorting playlist videos, given it has already been merged, or to open a new request on this. I can move this over there if that's better.

I'd be interested in the option to sort playlists based on video length. I've gotten into the habit of saving videos to watch later, back when playlists weren't finished yet, and I find that sorting videos by video length allows me to quickly run through shorter videos (such as shorts, or just shorter videos in general) before committing to longer ones. Videos I usually add to playlists can range from a minute or less to hour long videos. Having the shorter ones back to back means I can just get through them quickly (especially now that they're actual playlists that play videos back to back).

### Proposed Solution

The feature to sort videos seems to be ready, but, as mentioned on #4921, video length was not an option added initially. However, such could be later added if someone displayed interest. The proposed solution is to add such an option. I am aware that doing so would add more clutter to the sort menu, and I unfortunately have no solution to that.

### Alternatives Considered

I have been manually sorting playlists sometimes, though I do get lazy when I queue stuff and then it's 5 shorter videos on the bottom, below 10 longer videos, and I need to manually click the "up" button several times on each video. I don't mind that much, to be honest. It's a silly request that is probably specific to me. 

Could also figure out some quicker way to move videos, other than clicking buttons numerous times. Though I guess that could be more complex to design than having buttons that move things around.

### Issue Labels

new feature

MarianMeido:
I would love this feature as well.  Sorting videos by length in a playlist is so helpful in starting one and actually getting through them.  I rearrange them by hand, which is also tricky since some video thumbnails display a time code and some don't.   Adding just one more method to the dropdown would be amazing!  Thanks for bringing this up!

absidue:
[...]

PR #5627, opened 2024-08-31
Add Playlist Sort By Video Duration

Hoverth:
# Add Playlist Sort By Video Duration

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes: #5268

## Description

Closes issue #5268 and PR #5423. It works off what @creotove put in their PR, but also includes a small notification if there is a video in the playlist that does not have a time attached to it.

### DIFF
--- a/src/renderer/helpers/playlists.js
+++ b/src/renderer/helpers/playlists.js
@@ -5,6 +5,8 @@ export const SORT_BY_VALUES = {
   AuthorDescending: 'author_descending',
   VideoTitleAscending: 'video_title_ascending',
   VideoTitleDescending: 'video_title_descending',
+  VideoLengthAscending: 'video_length_ascending',
+  VideoLengthDescending: 'video_length_descending',
   Custom: 'custom'
 }
 
@@ -19,7 +21,9 @@ export function getSortedPlaylistItems(playlistItems, sortOrder, locale, reverse
     sortOrder === SORT_BY_VALUES.VideoTitleAscending ||
     sortOrder === SORT_BY_VALUES.VideoTitleDescending ||
     sortOrder === SORT_BY_VALUES.AuthorAscending ||
-    sortOrder === SORT_BY_VALUES.AuthorDescending
+    sortOrder === SORT_BY_VALUES.AuthorDescending ||
+    sortOrder === SORT_BY_VALUES.VideoLengthAscending ||
+    sortOrder === SORT_BY_VALUES.VideoLengthDescending
   ) {
     collator = new Intl.Collator([locale, 'en'])
   }
@@ -45,6 +49,16 @@ function compareTwoPlaylistItems(a, b, sortOrder, collator) {
       return collator.compare(a.author, b.author)
     case SORT_BY_VALUES.AuthorDescending:
       return collator.compare(b.author, a.author)
+    case SORT_BY_VALUES.VideoLengthAscending: {
+      const aLengthForSort = (isNaN(a.lengthSeconds) || a.lengthSeconds === 0) ? 0 : a.lengthSeconds
+      const bLengthForSort = (isNaN(b.lengthSeconds) || b.lengthSeconds === 0) ? 0 : b.lengthSeconds
+      return aLengthForSort - bLengthForSort
+    }
+    case SORT_BY_VALUES.VideoLengthDescending: {
+      const aLengthForSort = (isNaN(a.lengthSeconds) || a.lengthSeconds === 0) ? 0 : a.lengthSeconds
+      const bLengthForSort = (isNaN(b.lengthSeconds) || b.lengthSeconds === 0) ? 0 : b.lengthSeconds
+      return bLengthForSort - aLengthForSort
+    }
     default:
       console.error(`Unknown sortOrder: ${sortOrder}`)
       return 0
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -180,6 +180,7 @@ export default defineComponent({
       return this.sortOrder === SORT_BY_VALUES.Custom
     },
     sortedPlaylistItems: function () {
+      this.showNoticesSometimes()
       return getSortedPlaylistItems(this.playlistItems, this.sortOrder, this.currentLocale)
     },
     visiblePlaylistItems: function () {
@@ -214,6 +215,10 @@ export default defineComponent({
             return this.$t('Playlist.Sort By.AuthorAscending')
           case SORT_BY_VALUES.AuthorDescending:
             return this.$t('Playlist.Sort By.AuthorDescending')
+          case SORT_BY_VALUES.VideoLengthAscending:
+            return this.$t('Playlist.Sort By.VideoLengthAscending')
+          case SORT_BY_VALUES.VideoLengthDescending:
+            return this.$t('Playlist.Sort By.VideoLengthDescending')
           default:
             console.error(`Unknown sort: ${k}`)
             return k
@@ -418,6 +423,20 @@ export default defineComponent({
     },
     showUserPlaylistNotFound() {
       showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist does not exist'))
+    },
+
+    showNoticesSometimes: function () {
+      if (this.alreadyShownNotice) return
+      if (
+        this.sortOrder === SORT_BY_VALUES.VideoLengthAscending ||
+        this.sortOrder === SORT_BY_VALUES.VideoLengthDescending
+      ) {
+        const anyVideoMissingLength = this.playlistItems.some(v => isNaN(v.lengthSeconds) || v.lengthSeconds === 0)
+        if (anyVideoMissingLength) {
+          showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist has a video with a length error'), 5000)
+          this.alreadyShownNotice = true
+        }
+      }
     },
 
     getNextPage: function () {
--- a/static/locales/en-GB.yaml
+++ b/static/locales/en-GB.yaml
@@ -174,6 +174,8 @@ User Playlists:
       This playlist is protected and cannot be removed.: This playlist is protected
         and cannot be removed.
       This playlist does not exist: This playlist does not exist
+
+      This playlist has a video with a length error: This playlist contains at least one video that doesn't have a length, it will be sorted as if their length is zero.
       Playlist {playlistName} has been deleted.: Playlist {playlistName} has been
         deleted.
       Video has been removed: Video has been removed
@@ -1006,6 +1008,8 @@ Playlist:
     AuthorAscending: Author (A-Z)
     AuthorDescending: Author (Z-A)
     VideoTitleAscending: Title (A-Z)
+    VideoLengthAscending: Length (Shortest first)
+    VideoLengthDescending: Length (Longest first)
 Toggle Theatre Mode: 'Toggle Theatre Mode'
 Change Format:
   Change Media Formats: 'Change Media Formats'
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -236,6 +236,8 @@ User Playlists:
       Playlist {playlistName} is the new quick bookmark playlist.: Playlist {playlistName} is the new quick bookmark playlist.
 
       This playlist does not exist: This playlist does not exist
+
+      This playlist has a video with a length error: This playlist contains at least one video that doesn't have a length, it will be sorted as if their length is zero.
   AddVideoPrompt:
     Select a playlist to add your N videos to: 'Select a playlist to add your video to | Select a playlist to add your {videoCount} videos to'
     N playlists selected: '{playlistCount} Selected'
@@ -928,6 +930,8 @@ Playlist:
     AuthorDescending: Author (Z-A)
     VideoTitleAscending: Title (A-Z)
     VideoTitleDescending: Title (Z-A)
+    VideoLengthAscending: Length (Shortest first)
+    VideoLengthDescending: Length (Longest first)
     Custom: Custom
 
 # On Video Watch Page


src/renderer/views/Playlist/Playlist.js
```diff
@@ -291,6 +296,20 @@ export default defineComponent({
...
+      ) {
+        if (!this.shownNotification) {
+          for (const v in this.playlistItems) {
+            if (!Object.hasOwn(v, 'lengthSeconds') || isNaN(v.lengthSeconds)) {
+              showToast("This playlist contains at least one video that doesn't have a length, it will be sorted as if their length is zero.", 10000, () => {})
+              this.shownNotification = true
+            }
+          }
+        }
+      }
```
PikachuEXE, line 312: Just call `this.checkPlaylist()` 
Also this part of code has no `break` = will shown N toasts for N videos without length (another reason not to just call existing method

src/renderer/views/Playlist/Playlist.js
```diff
@@ -280,6 +284,7 @@ export default defineComponent({
         } else {
           this.showUserPlaylistNotFound()
         }
+        this.checkPlaylist()
```
PikachuEXE, line 287: Should this be called if user playlist not found, I guess not

src/renderer/views/Playlist/Playlist.js
```diff
@@ -418,6 +437,23 @@ export default defineComponent({
...
 
+    checkPlaylist: function () {
+      if (
+        this.sortOrder === SORT_BY_VALUES.VideoLengthAscending ||
+        this.sortOrder === SORT_BY_VALUES.VideoLengthDescending
+      ) {
+        if (!this.shownNotification) {
+          for (const v in this.playlistItems) {
+            if (!Object.hasOwn(v, 'lengthSeconds') || isNaN(v.lengthSeconds)) {
+              showToast('This playlist contains at least one video that doesn\'t have a length, it will be sorted as if their length is zero.', 10000, () => {})
```
PikachuEXE, line 448: Need to be translated, also `() => {}` is not needed remove it

src/renderer/views/Playlist/Playlist.js
```diff
@@ -418,6 +437,23 @@ export default defineComponent({
       showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist does not exist'))
     },
 
+    checkPlaylist: function () {
+      if (
+        this.sortOrder === SORT_BY_VALUES.VideoLengthAscending ||
+        this.sortOrder === SORT_BY_VALUES.VideoLengthDescending
+      ) {
+        if (!this.shownNotification) {
```
PikachuEXE, line 445: Why need this

src/renderer/helpers/playlists.js
```diff
@@ -45,6 +49,10 @@ function compareTwoPlaylistItems(a, b, sortOrder, collator) {
       return collator.compare(a.author, b.author)
     case SORT_BY_VALUES.AuthorDescending:
       return collator.compare(b.author, a.author)
+    case SORT_BY_VALUES.VideoLengthAscending:
+      return a.lengthSeconds - b.lengthSeconds
+    case SORT_BY_VALUES.VideoLengthDescending:
+      return b.lengthSeconds - a.lengthSeconds
```
PikachuEXE, line 55: > it will be sorted as if their length is zero

Is this code doing what's claimed?

src/renderer/views/Playlist/Playlist.js
Hoverth: Gotcha, will remove & make translatable string

src/renderer/views/Playlist/Playlist.js
Hoverth: It doesn't strictly need the `if` statement, but it will speed it up if it knows that the `sortOrder` doesn't care about the length of the videos, as then it doesn't have to iterate through all the videos in the playlist

src/renderer/views/Playlist/Playlist.js
Hoverth: Yeah, probably not, I'll move up to only if the playlist is found.

src/renderer/views/Playlist/Playlist.js
Hoverth: Yep, removing that code.

package.json
```diff
@@ -131,5 +131,6 @@
     "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^5.0.4",
     "yaml-eslint-parser": "^1.2.3"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
PikachuEXE, line 135: I don't think this change should be part of the PR

src/renderer/helpers/playlists.js
PikachuEXE: Are you sure? 
I saw `if (!Object.hasOwn(v, 'lengthSeconds') || isNaN(v.lengthSeconds))` in `checkPlaylist` so either that's wrong or `lengthSeconds` can be `NaN`/`undefined`
I haven't tested this myself yet to be done tomorrow

src/renderer/views/Playlist/Playlist.js
```diff
@@ -418,6 +423,20 @@ export default defineComponent({
       showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist does not exist'))
     },
 
+    checkPlaylist: function () {
```
PikachuEXE, line 426: I can't think of a better one yet but the method is to show notice (not warning) and it shows sometimes only
Let me know if you can think of a better name

```suggestion
    showNoticesSometimes: function () {
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -418,6 +423,20 @@ export default defineComponent({
...
+      if (
+        this.sortOrder === SORT_BY_VALUES.VideoLengthAscending ||
+        this.sortOrder === SORT_BY_VALUES.VideoLengthDescending
+      ) {
+        for (const v in this.playlistItems) {
+          if (!Object.hasOwn(v, 'lengthSeconds') || isNaN(v.lengthSeconds)) {
+            showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist has a video with a length error'), 10000)
+            break
+          }
+        }
```
PikachuEXE, line 436: https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Array/some
I just prefer early return style if possible
```suggestion
        const anyVideoMissingLength = this.playlistItems.some(v => !Object.hasOwn(v, 'lengthSeconds') || isNaN(v.lengthSeconds))
        if (anyVideoMissingLength) {
          showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist has a video with a length error'), 10000)
        }
```

src/renderer/helpers/playlists.js
Hoverth: I think this is running into javascript truthy behaviour, I'll do some more tests

Hoverth: It can be left `undefined` if not evaluated correctly, but I think I should also check if it's zero

Hoverth:
Alright that should fix the issue.

src/renderer/helpers/playlists.js
Hoverth: I've come to the conclusion tthat the video object will always have `.lengthSeconds`, but depending on how it's evaluated, it could spit out an undefined, so I'll keep that in the code. I have also included the zero-check behaviour.

PikachuEXE: If I add shorts to a playlist, they have string values
image

I think it's best to use the same check as the one below and use `0` if check failed (See another line comment)

PikachuEXE, line 55:
```suggestion
    case SORT_BY_VALUES.VideoLengthAscending: {
      const aLengthForSort = (isNaN(a.lengthSeconds) || a.lengthSeconds === 0) ? 0 : a.lengthSeconds
      const bLengthForSort = (isNaN(b.lengthSeconds) || b.lengthSeconds === 0) ? 0 : b.lengthSeconds
      return aLengthForSort - bLengthForSort
    }
    case SORT_BY_VALUES.VideoLengthDescending: {
      const aLengthForSort = (isNaN(a.lengthSeconds) || a.lengthSeconds === 0) ? 0 : a.lengthSeconds
      const bLengthForSort = (isNaN(b.lengthSeconds) || b.lengthSeconds === 0) ? 0 : b.lengthSeconds
      return aLengthForSort - bLengthForSort
    }
```

image
image

PikachuEXE (changes requested):

src/renderer/helpers/playlists.js
Hoverth: Commit 57f251d adds this in, I did catch that both of the cases had the same difference calculation, so I fixed that as well.

Hoverth: I think that may be a bug in whatever system parses shorts, but we can put in a check here too.

absidue: > I think that may be a bug in whatever system parses shorts, but we can put in a check here too.

Nope, YouTube doesn't return durations for shorts on the shorts channel tab or in playlists that only contain shorts, so we set it to an empty string, the alternative would be `null` but the empty string fit best into the existing logic at the time.

Hoverth: That's fair enough, though I still feel a `null` would be better than an empty string for a integer value, but all in all, this code won't care.

PikachuEXE:
Works fine so far except I think the notice should be shown also when changing sort order in single playlist view (unless already shown
This is what I have in mind
(A) (already handled)
- Enter any single playlist
- Show notice if last used sorting based on length

(B)
- Enter any single playlist, last used sorting NOT based on length
- Switch sort order to based on length
- Show notice (just once within the same view)

Hoverth:
[7143e62](https://github.com/FreeTubeApp/FreeTube/pull/5627/commits/7143e627eb096109eec246c36bfd5a4c3866ab51) moves the call to when the playlist items are sorted, so if you load a playlist and then change the sorting, it'll still show. 

I'm not terribly sure how to make it only fire once though. Thoughts?

Hoverth:
Alright, I've given single-firing only a shot, and also tweaked the time to be only 5 seconds, as 10 seemed a little long to me

src/renderer/views/Playlist/Playlist.js
```diff
@@ -418,6 +423,20 @@ export default defineComponent({
       showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist does not exist'))
     },
 
+    showNoticesSometimes: function () {
+      if (this.alreadyShownNotice) return
```
PikachuEXE, line 427: It's not a reactive property so I am not sure if this should be declared somewhere
@absidue ?

Hoverth:
Strange how it automatically closed it when I attempted to rebase off of development, but it seems to be alright now.

PikachuEXE:
I am still waiting for answer for https://github.com/FreeTubeApp/FreeTube/pull/5627#pullrequestreview-2319661081

src/renderer/views/Playlist/Playlist.js
Hoverth: Should I just declare it anyway?

PikachuEXE: Just declare it, I doubt the performance matters too much here :)

Hoverth: Done!

absidue: Sorry missed the ping previously. Yes you should always declare variables somewhere, just assigning things to `this` is considered an anti-pattern with the Vue Options API.

absidue:
Could you please change all of the user facing messages and labels to use the word duration instead of length, as that's more understandable to people that don't look at the code.

Hoverth:
> Could you please change all of the user facing messages and labels to use the word duration instead of length, as that's more understandable to people that don't look at the code.

Done!

absidue:
> > Could you please change all of the user facing messages and labels to use the word duration instead of length, as that's more understandable to people that don't look at the code.
> 
> Done!

Looks like you forgot to change the actual text that the users sees:
```yaml
This playlist has a video with a length error: This playlist contains at least one video that doesn't have a length, it will be sorted as if their length is zero.

VideoDurationAscending: Length (Shortest first)
VideoDurationDescending: Length (Longest first)
```

Hoverth:
> > > Could you please change all of the user facing messages and labels to use the word duration instead of length, as that's more understandable to people that don't look at the code.
> > 
> > 
> > Done!
> 
> Looks like you forgot to change the actual text that the user sees:
> 
> ```yaml
> This playlist has a video with a length error: This playlist contains at least one video that doesn't have a length, it will be sorted as if their length is zero.
> 
> VideoDurationAscending: Length (Shortest first)
> VideoDurationDescending: Length (Longest first)
> ```

Whoops, sorry about that, should be fixed now.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
When the timestamp of an short is known it doesnt sort them properly

https://github.com/user-attachments/assets/48a077fe-2087-4ded-a1a2-e10615918d8d

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc That's because it's showing the duration from the watch history, in the playlist it is still unknown (updating watch history entries is really quick because we store them in an object so they can be quickly looked up by the video ID and we need to update it anyway, but with playlists we would have to go through every single video in every single playlist, we can't even stop at the first occurence of the video in a playlist because we allow duplicates).

Hoverth:
@absidue @efb4f5ff-1298-471a-8973-3d47447115dc 
Just to confirm, is this out of scope for this PR?

absidue:
@Hoverth Modifying the playlists definitely is, but you could still theoretically look up the video in the watch history during the sorting process if the duration is missing (you can look at the logic that ft-list-video uses to do that). You'll have to experiment to see how bad the performance is though, because sorting usually requires looking at each item in the list lots of times, so you'll want to find a way of avoiding looking it up in the history every single time during the same sorting run.

Hoverth:
> @Hoverth Modifying the playlists definitely is, but you could still theoretically look up the video in the watch history during the sorting process if the duration is missing (you can look at the logic that ft-list-video uses to do that). You'll have to experiment to see how bad the performance is though, because sorting usually requires looking at each item in the list lots of times, so you'll want to find a way of avoiding looking it up in the history every single time during the same sorting run.

I can't really come up with anything better than [after line 436 in Playlist.js]

```js
for (const v in this.playlistItems) {
  const lengthSeconds = this.playlistItems[v].lengthSeconds
  if (isNaN(lengthSeconds) || lengthSeconds == 0) {
    const videoHistory = this.$store.getters.getHistoryCacheById[this.playlistItems[v].videoId]
    if (typeof videoHistory !== 'undefined') this.playlistItems[v].lengthSeconds = videoHistory.lengthSeconds
  }
}
```

Plus, this spits out `Error: [vuex] do not mutate vuex store state outside mutation handlers.` because of the assignment done. I'm not sure if I'm familiar enough with this codebase to tackle this issue.

absidue:
I thought I said that modifying playlists was out of scope of this pull request...

Hoverth:
You did, however this would be the only way to run the code once and still have the playlist be sorted, as otherwise there would need to be at least two runs (one for displaying the toast message, and at least another for actually sorting the videos)

Hoverth:
Also should be noted that videos added via subscriptions fetched by RSS will also have unknown durations (`lengthSeconds = "0:00"`), so it seems like there should be a mechanism in which, when a video is played, updates the metadata for it, but that is definitely out of scope of this PR.

Hoverth:
@absidue it still has to iterate through all the playlist items, but this way it'll only do it if it's needed and seems to fix all the issues raised. I'm not terribly sure how performant it is as of yet, as I'm not sure of a way to test large playlists.

PikachuEXE:
I think `When the timestamp of an short is known it doesnt sort them properly` should be solved by updating the entry on playlist when no valid duration saved instead of reading from history every time

Out of scope for this PR anyway

Hoverth:
> I think `When the timestamp of an short is known it doesnt sort them properly` should be solved by updating the entry on playlist when no valid duration saved instead of reading from history every time

Also, it would probably be good to do the same for videos fetched via RSS, but yeah that's out of scope for this PR.

absidue:
Is there any functional reason for doing it with the subscriptions or just a nice to have?

We previously never had a need to modify the entries because the duration was a purely visual thing, so overlaying it from the watch history was fine. It's only become a problem now that you are building functionality on top of it, which as far as I can tell has nothing to do with the subscriptions...

Hoverth:
If you add a video from the subscriptions page, and it has been fetched via RSS, it would be sorted with the value `'0:00'`, which is incorrect. You probably wouldn't have to modify the subscriptions cache, most likely just the playlists one, as they get updated when new metadata is available, whereas I don't think the same would be needed for the subscriptions page.

absidue:
The main problem I have with having to update the durations in the playlist is that we'll have to do it every time a video is watched and there is no way to do it quickly, because it will require going through every single video in every single playlist every time and then also doing the database updates. And a large part of the reason it will be slow is because the playlists are currently deeply reactive and they will have to stay that way if we want to do those modifications to the individual playlist items.

src/renderer/helpers/playlists.js
```diff
@@ -31,6 +35,10 @@ export function getSortedPlaylistItems(playlistItems, sortOrder, locale, reverse
   })
 }
 
+export function checkDurationFromVideo(a) {
+  return (isNaN(a.lengthSeconds) || a.lengthSeconds === 0 || typeof a.lengthSeconds === 'string') ? 0 : a.lengthSeconds
```
PikachuEXE, line 39: I see this 3 times already just make a helper function?
```suggestion
  return (isNaN(a.lengthSeconds) || a.lengthSeconds === 0 || typeof a.lengthSeconds !== 'number') ? 0 : a.lengthSeconds
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -420,6 +434,27 @@ export default defineComponent({
       showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist does not exist'))
     },
 
+    getDurationPlaylistItems: function () {
+      const modifiedPlaylistItems = deepCopy(this.playlistItems)
+      for (const v in modifiedPlaylistItems) {
```
PikachuEXE, line 439: Maybe just use `forEach` instead of having to write `modifiedPlaylistItems[v]` in the loop

src/renderer/views/Playlist/Playlist.js
```diff
@@ -420,6 +434,27 @@ export default defineComponent({
...
+        const lengthSeconds = modifiedPlaylistItems[v].lengthSeconds
+        if (isNaN(lengthSeconds) || lengthSeconds === 0 || typeof lengthSeconds === 'string') {
```
PikachuEXE, line 441:
```suggestion
        if (isNaN(lengthSeconds) || lengthSeconds === 0 || typeof lengthSeconds !== 'number') {
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -420,6 +434,27 @@ export default defineComponent({
...
+          const videoHistory = this.$store.getters.getHistoryCacheById[modifiedPlaylistItems[v].videoId]
+          if (typeof videoHistory !== 'undefined') modifiedPlaylistItems[v].lengthSeconds = videoHistory.lengthSeconds
+        }
+      }
+      return modifiedPlaylistItems
+    },
+
+    showNoticesSometimes: function (playlistItems) {
+      if (this.alreadyShownNotice) return
+      const anyVideoMissingDuration = playlistItems.some(v => isNaN(v.lengthSeconds) || v.lengthSeconds === 0 || typeof v.lengthSeconds === 'string')
```
PikachuEXE, line 451:
```suggestion
      const anyVideoMissingDuration = playlistItems.some(v => isNaN(v.lengthSeconds) || v.lengthSeconds === 0 || typeof v.lengthSeconds !== 'number')
```

PikachuEXE (changes requested):

src/renderer/helpers/playlists.js
Hoverth: done!

src/renderer/views/Playlist/Playlist.js
Hoverth: done!

Hoverth:
> Although as long as we make sure that that doesn't impact the actual watch history updates, so doing it in a two step process, updating the watch history first and the metadata updates afterwards, it should be okay.

As that's what I've got working now, that's all good with me. One thing I would suggest, is maybe a button that reloads the metadata for all videos in a playlist? That way you don't care how long it takes, as it's an action invoked by the user themselves.

efb4f5ff-1298-471a-8973-3d47447115dc:
Weird idea and not sure if feasible but what about using the [toast message](https://github.com/FreeTubeApp/FreeTube/pull/5627/files#diff-1d702e6381a99185e240fe47a79563920174af6e327b61d1c171543ebfc44a81R452) for this purpose. Message would need to be slightly modified but if we add something like this to the toast `Timestamp information could be fetched from other places of the application. Click here to try to fetch the missing timestamps. This can take a while. `

Hoverth:
@absidue That's fair enough. Is there any other issues with this PR?

src/renderer/helpers/playlists.js
```diff
@@ -31,6 +35,10 @@ export function getSortedPlaylistItems(playlistItems, sortOrder, locale, reverse
   })
 }
 
+export function checkDurationFromVideo(a) {
+  return (isNaN(a.lengthSeconds) || a.lengthSeconds === 0 || typeof a.lengthSeconds !== 'number') ? 0 : a.lengthSeconds
+}
```
PikachuEXE, line 40: No export on `videoDurationWithFallback` unless it's used outside
```suggestion
export function videoDurationPresent(v) {
  if (typeof v.lengthSeconds !== 'number') { return false }
  
  return !(isNaN(v.lengthSeconds) || v.lengthSeconds === 0)
}

function videoDurationWithFallback(v) {
  if (videoDurationPresent(v)) { return v.lengthSeconds }

  // Fallback
  return 0
}
```

src/renderer/helpers/playlists.js
```diff
@@ -45,6 +53,12 @@ function compareTwoPlaylistItems(a, b, sortOrder, collator) {
       return collator.compare(a.author, b.author)
     case SORT_BY_VALUES.AuthorDescending:
       return collator.compare(b.author, a.author)
+    case SORT_BY_VALUES.VideoDurationAscending: {
+      return checkDurationFromVideo(a) - checkDurationFromVideo(b)
+    }
+    case SORT_BY_VALUES.VideoDurationDescending: {
+      return checkDurationFromVideo(b) - checkDurationFromVideo(a)
+    }
```
PikachuEXE, line 61:
```suggestion
    case SORT_BY_VALUES.VideoDurationAscending: {
      return videoDurationWithFallback(a) - videoDurationWithFallback(b)
    }
    case SORT_BY_VALUES.VideoDurationDescending: {
      return videoDurationWithFallback(b) - videoDurationWithFallback(a)
    }
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -20,9 +20,10 @@ import {
   getIconForSortPreference,
   setPublishedTimestampsInvidious,
   showToast,
+  deepCopy,
 } from '../../helpers/utils'
 import { invidiousGetPlaylistInfo, youtubeImageUrlToInvidious } from '../../helpers/api/invidious'
-import { getSortedPlaylistItems, SORT_BY_VALUES } from '../../helpers/playlists'
+import { getSortedPlaylistItems, checkDurationFromVideo, SORT_BY_VALUES } from '../../helpers/playlists'
```
PikachuEXE, line 26:
```suggestion
import { getSortedPlaylistItems, videoDurationPresent, SORT_BY_VALUES } from '../../helpers/playlists'
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -420,6 +434,26 @@ export default defineComponent({
       showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist does not exist'))
     },
 
+    getDurationPlaylistItems: function () {
+      const modifiedPlaylistItems = deepCopy(this.playlistItems)
+      modifiedPlaylistItems.forEach(video => {
+        if (!checkDurationFromVideo(video)) {
```
PikachuEXE, line 440:
```suggestion
        if (!videoDurationPresent(video)) {
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -420,6 +434,26 @@ export default defineComponent({
...
+          const videoHistory = this.$store.getters.getHistoryCacheById[video.videoId]
+          if (typeof videoHistory !== 'undefined') video.lengthSeconds = videoHistory.lengthSeconds
+        }
+      })
+      return modifiedPlaylistItems
+    },
+
+    showNoticesSometimes: function (playlistItems) {
+      if (this.alreadyShownNotice) return
+      const anyVideoMissingDuration = playlistItems.some(v => !checkDurationFromVideo(v))
```
PikachuEXE, line 450:
```suggestion
      const anyVideoMissingDuration = playlistItems.some(v => !videoDurationPresent(v))
```

PikachuEXE (changes requested):
Only method with side effects should be named after the action

src/renderer/views/Playlist/Playlist.js
```diff
@@ -420,6 +434,26 @@ export default defineComponent({
...
+    },
+
+    showNoticesSometimes: function (playlistItems) {
+      if (this.alreadyShownNotice) return
+      const anyVideoMissingDuration = playlistItems.some(v => !videoDurationPresent(v))
+      if (anyVideoMissingDuration) {
+        showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist has a video with a duration error'), 5000)
+        this.alreadyShownNotice = true
+      }
+    },
```
PikachuEXE, line 455: Loop once only for performance?
Method name would be too long if I add "sometimes with notice" :P
```suggestion
    getPlaylistItemsWithDuration() {
      const modifiedPlaylistItems = deepCopy(this.playlistItems)
      let anyVideoMissingDuration = false
      modifiedPlaylistItems.forEach(video => {
        if (videoDurationPresent(video)) { return }

        const videoHistory = this.$store.getters.getHistoryCacheById[video.videoId]
        if (typeof videoHistory !== 'undefined') {
          video.lengthSeconds = videoHistory.lengthSeconds
        } else {
          // Mark at least one video have no duration, show notice later
          // Also assign fallback duration here
          anyVideoMissingDuration = true
          video.lengthSeconds = 0
        }
      })
      
      // Show notice if not already shown before returning playlist items
      if (anyVideoMissingDuration && !this.alreadyShownNotice) {
        showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist has a video with a duration error'), 5000)
        this.alreadyShownNotice = true
      }
      
      return modifiedPlaylistItems
    },
```

src/renderer/views/Playlist/Playlist.js
```diff
@@ -180,6 +182,14 @@ export default defineComponent({
       return this.sortOrder === SORT_BY_VALUES.Custom
     },
     sortedPlaylistItems: function () {
+      if (
+        this.sortOrder === SORT_BY_VALUES.VideoDurationAscending ||
+        this.sortOrder === SORT_BY_VALUES.VideoDurationDescending
+      ) {
+        const playlistItems = this.getDurationPlaylistItems()
+        this.showNoticesSometimes(playlistItems)
```
PikachuEXE, line 190:
```suggestion
        const playlistItems = this.getPlaylistItemsWithDuration()
```

PikachuEXE (commented):
Maybe just loop it once?
Test locally before committing that change~

Hoverth:
> Maybe just loop it once? Test locally before committing that change~

It does feel a little bit faster, but that could just be placebo. Either way, it should be more efficient now.

PikachuEXE (approved):

src/renderer/views/Playlist/Playlist.js
```diff
@@ -416,10 +429,37 @@ export default defineComponent({
...
 
+    getPlaylistItemsWithDuration() {
+      const modifiedPlaylistItems = deepCopy(this.playlistItems)
+      let anyVideoMissingDuration = false
+      modifiedPlaylistItems.forEach(video => {
+        if (videoDurationPresent(video)) { return }
+
+        const videoHistory = this.$store.getters.getHistoryCacheById[video.videoId]
+        if (typeof videoHistory !== 'undefined') {
+          video.lengthSeconds = videoHistory.lengthSeconds
```
absidue, line 445: You need to check if the duration is valid here too, as just like with the playlist, just because a video exists doesn't mean all the information is there. Example: Manually marking a video as watched through the 3 dots menu on the shorts channel tab or in a playlist that only contains shorts.

src/renderer/helpers/playlists.js
```diff
@@ -31,6 +35,19 @@ export function getSortedPlaylistItems(playlistItems, sortOrder, locale, reverse
...
+
+  return !(isNaN(v.lengthSeconds) || v.lengthSeconds === 0)
+}
+
+function videoDurationWithFallback(v) {
+  if (videoDurationPresent(v)) { return v.lengthSeconds }
+
+  // Fallback
+  return 0
+}
```
absidue, line 49: Please use proper variable names here.

src/renderer/views/Playlist/Playlist.js
Hoverth: done!

PikachuEXE (approved):

ChunkyProgrammer (approved):

absidue (approved):
Not a fan of cloning the entire playlist, but as I can't come up with a better solution at the moment and everything else about this pull request is good, I'll approve.

Issue #677, opened 2020-10-16
[Feature request] "Random videos from subscriptions" page


ThatIsAPseudo:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
On youtube we have the "Home" tab that shows videos that may please you, but since it's based on Google tracking the user's preferences I guess it can't easily be done in Freetube and/or is simply not desirable. 
Maybe there could be a page with a showcase of channels similar to those you're already subscribed to, but it requires a database with (relevant) keywords about every channels of youtube.

**Describe the solution you'd like to see implemented**
Instead, there could be a kind of "random" tab, in which you'll find random videos from your own subscribed channels, so that when no new video is available, you'll still be able to find "unknown" videos linked to your preferences. There could also be an option to hide particular channels, and restrict the maximum age of videos.

**Describe alternatives you've considered**
For now I juste use Youtube home tab then when I find an interesting video I open it with Freetube.

**Additional context**
I just realized I may not be such a simple idea, since I don't know if there's a way to scrap random videos from a channel.

GilgusMaximus:
For this you could use the popular tab, which basically is an aggregation of popular channels on Invidious.
Just showing you random old videos from subscriptions is not something I would consider useful. Additionally this will cost you a lot of requests, leading to a rate limit much faster

PrestonN:
There's only one real way this could be done while still being private.

You would start each session with an empty recommendations list.  As you watch videos, FreeTube could take the recommendations that are shown and combine them as you watch more videos.  We would then take this list and shuffle the videos every time you view your "Recommendations" feed.  We wouldn't save this list between sessions so every time you load up FreeTube you'd start over on the "Recommendations" you see.

This is the only way I can see this happening without interfering with the privacy aspects of FreeTube.  This wouldn't be too difficult to do but it would definitely be a lower priority to put together.

ThatIsAPseudo:
@GilgusMaximus indeed I do use the Popular tab, however videos do not refresh often enough (imo). I understand the concern of rate limit though.

@PrestonN That sounds great, if you think that's a good first issue I could possibly try to give it a shot.

PrestonN:
This might not be the best first issue if you haven't contributed to the project before, but if you have an intermediate level of knowledge with Vue then it could be good task to work on.  I'd imagine something like this would cover a few different parts of the app in order to work properly.

[...]

PR #5631, opened 2024-09-01
Recommendations

111111111111111111111111111111111111110:
# Recommendations

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/677#issuecomment-2323350721

## Description
this adds recommended tab to the left side menu. recommendations is based on search history. search query history is stored in local storage.

## Screenshots 
Screenshot_2024-09-01_21-14-53

## Desktop

- **OS:** Debian
- **OS Version:** 12
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/recommended-tab-ui/recommended-tab-ui.css
+++ b/src/renderer/components/recommended-tab-ui/recommended-tab-ui.css
@@ -1,0 +1,15 @@
+.card {
+  inline-size: 85%;
+  margin-block: 0 60px;
+  margin-inline: auto;
+}
+
+.message {
+  color: var(--tertiary-text-color);
+}
+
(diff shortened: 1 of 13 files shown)


111111111111111111111111111111111111110:
this is simple implementation of recommended videos feature. algorithm should be quite easy to adjust later if needed.

absidue:
Seems interesting although before this can be properly reviewed it looks like there is a lot of unneeded code (seems like a lot was copied from the subscriptions page but without actually adapting it properly to this use case). Could you please clean that up?

111111111111111111111111111111111111110:
ok, did some code cleanup

PR #5649, opened 2024-09-04
Proxy live streams when "Proxy Videos Through Invidious" is enabled

absidue:
# Proxy live streams when "Proxy Videos Through Invidious" is enabled

## Pull Request Type

- [x] Feature Implementation

## Description
Currently FreeTube doesn't proxy live streams through Invidious, this pull request adds support for that. As you can tell from the code it requires some hackiness to get it to work:
1. Invidious swaps out the path parameters for query parameters, which means the URL no longer looks like a file path with a file extension, so shaka-player isn't able to guess the mime type based on the file extension, it does then try to do a HEAD request to check the `Content-Type` header but that's `application/octet-stream`, so also not useful.
2. Then I came up with a fix that worked with Invidious instances that use the built-in video proxy in Invidious but broke with ones using the Piped video proxy
3. The fix to get it working on the Piped video proxy broke with Invidious' own video proxy again
4. Finally I found something that worked with both video proxies and allowed shaka-player to actually play it

The code has various comments that explain what each bit does, so hopefully we'll all know what it does and why it is there in the future.

## Testing 
1. Turn on the "Proxy Videos Through Invidious" setting
6. Watch live videos e.g. https://youtu.be/jfKfPfyJRdk with both the local and the Invidious API
7. Check the `Network` tab of the devtools that the `index.m3u8` and `seg.ts` requests are going to your currently selected Invidious instance.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 45c8cafa5b23b9d4564e4a4872c77a934b5eb50b

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -38,6 +38,7 @@ const HTTP_IN_HEX = 0x68747470
 const USE_OVERFLOW_MENU_WIDTH_THRESHOLD = 600
 
 const RequestType = shaka.net.NetworkingEngine.RequestType
+const AdvancedRequestType = shaka.net.NetworkingEngine.AdvancedRequestType
 const TrackLabelFormat = shaka.ui.Overlay.TrackLabelFormat
 
 /** @type {Map<string, string>} */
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Can't test yet, got issue playing any video in local/IV API in dev ("Proxy Videos Through Invidious" on

image

Or you got a working instance? (I tried a few from https://api.invidious.io already

absidue:
> Can't test yet, got issue playing any video in local/IV API in dev ("Proxy Videos Through Invidious" on
> Or you got a working instance? (I tried a few from https://api.invidious.io already

As that bug only affects non-live videos and this pull request is about live videos, any of the instances on the list should work :).

PikachuEXE:
Still can't with instance https://invidious.perennialte.ch with [invidious.jing.rocks](https://invidious.jing.rocks/)
(They use diff versions)

image

image
image

Just instances outdated?

absidue:
Invidious returning a 403 usually means it got a 403 from YouTube, so probably that instance is IP blocked for live streams or something.

During my testing `https://iv.ggtyler.dev` and `https://invidious.reallyaweso.me` worked well.

ChunkyProgrammer (approved):

PR #5665, opened 2024-09-08
Fix: Video description collapsible and no longer hijacks page scroll

### LINKED ISSUE #4821: [Feature Request]: Make the description box collapsible
### Guidelines

### Problem Description

The description box hijacks the scroll focus when the mouse pointer is over it. So if I wanted to go down to the comments section and my mouse pointer is over the description, it will scroll the contents of the description instead of continuing down.

### Proposed Solution

Would be nice if the description box is collapsed by default or can be toggled in settings so it won't get in the way when scrolling down the page. YouTube defaults to collapsed for example:

Screenshot 2024-03-28 at 14-42-40 Life at Extremes Biology of Brine Pools and Methane Seeps - 2014 (Remastered!) Nautilus Live - YouTube

### Alternatives Considered

Disable the description box in the settings, though not ideal since it contains additional information and external links related to the video you're watching.

### Issue Labels

### Additional Information

_No response_

sabs21:
# Fix: Video description collapsible and no longer hijacks page scroll

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #4821 

## Description

Long video descriptions will now have their overflow hidden to prevent hijacking the scroll of the page. To see the full description and reintroduce the scrollbar, the user must click 'Click to View Description' shown at the bottom of the description. The 'Click to View Description' button is only shown for descriptions with a computed CSS height greater than or equal to 300 pixels.

## Screenshots 

CURRENT:
image

CHANGE (Description not expanded):
image

CHANGE (Description expanded):
image

## Testing 

I tested this change on a video with a long description (description exceeds CSS height of 300px) and a video with a short description. I verified: 
- Clicking 'Click to View Description' makes the description scrollable and that the button removes itself.
- Short descriptions do not show the 'Click to View Description' button.

## Desktop

- **OS:** Windows 10
- **OS Version:** 22H2, Build 19045.4842
- **FreeTube version:** v0.21.3 Beta

### DIFF
--- a/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
+++ b/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
@@ -4,6 +4,7 @@
     class="videoDescription"
   >
     <FtTimestampCatcher
+      ref="descriptionContainer"
       :class="{ description: true, short: !showFullDescription }"
       :input-html="shownDescription"
       @timestamp-event="onTimestamp"
@@ -25,7 +26,7 @@
 <script setup>
 import autolinker from 'autolinker'
 
-import { ref } from 'vue'
+import { onMounted, ref } from 'vue'
 import FtCard from '../ft-card/ft-card.vue'
 import FtTimestampCatcher from '../ft-timestamp-catcher/ft-timestamp-catcher.vue'
 
@@ -43,7 +44,8 @@ const props = defineProps({
 const emit = defineEmits(['timestamp-event'])
 
 let shownDescription = ''
-const showFullDescription = ref(false)
+const descriptionContainer = ref()
+const showFullDescription = ref(true)
 
 if (props.descriptionHtml !== '') {
   const parsed = parseDescriptionHtml(props.descriptionHtml)
@@ -77,6 +79,23 @@ function onTimestamp(timestamp) {
 function expandDescription() {
   showFullDescription.value = true
 }
+
+/**
+ * Returns true when description content does not overflow description container
+ * Useful for hiding the 'Click to View Description' button for short descriptions
+ */
+function isShortDescription() {
+  const videoDescriptionMaxHeight = 300
+  const computedDescriptionHeight = descriptionContainer.value.$el.getBoundingClientRect().height
+  return computedDescriptionHeight < videoDescriptionMaxHeight
+}
+
+onMounted(() => {
+  // To verify whether or not the description is too short for the
+  // 'Click to View Description' button, we need to check the description's
+  // computed CSS height. The only way to make this work is to check on mount.
+  showFullDescription.value = isShortDescription()
+})
 
 /**
  * @param {string} descriptionText
--- a/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
+++ b/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
@@ -4,16 +4,28 @@
     class="videoDescription"
   >
     <FtTimestampCatcher
-      class="description"
+      :class="{ description: true, short: !showFullDescription }"
       :input-html="shownDescription"
       @timestamp-event="onTimestamp"
     />
+    <h4
+      v-if="!showFullDescription"
+      class="getDescriptionTitle"
+      role="button"
+      tabindex="0"
+      @click="expandDescription"
+      @keydown.space.prevent="expandDescription"
+      @keydown.enter.prevent="expandDescription"
+    >
+      {{ $t("Description.Click to View Description") }}
+    </h4>
   </FtCard>
 </template>
 
 <script setup>
 import autolinker from 'autolinker'
 
+import { ref } from 'vue'
 import FtCard from '../ft-card/ft-card.vue'
 import FtTimestampCatcher from '../ft-timestamp-catcher/ft-timestamp-catcher.vue'
 
@@ -31,6 +43,7 @@ const props = defineProps({
 const emit = defineEmits(['timestamp-event'])
 
 let shownDescription = ''
+const showFullDescription = ref(false)
 
 if (props.descriptionHtml !== '') {
   const parsed = parseDescriptionHtml(props.descriptionHtml)
@@ -56,6 +69,13 @@ if (props.descriptionHtml !== '') {
  */
 function onTimestamp(timestamp) {
   emit('timestamp-event', timestamp)
+}
+
+/**
+ * Enables user to scroll entire contents of description
+ */
+function expandDescription() {
+  showFullDescription.value = true
 }
 
 /**
--- a/src/renderer/components/WatchVideoDescription/WatchVideoDescription.css
+++ b/src/renderer/components/WatchVideoDescription/WatchVideoDescription.css
@@ -9,3 +9,18 @@
   white-space: pre-wrap;
   overflow-wrap: anywhere;
 }
+
+.getDescriptionTitle {
+  margin: 0;
+  padding-block: 8px;
+  text-align: center;
+  text-decoration: underline;
+  cursor: pointer;
+  color: var(--title-color);
+}
+
+.short {
+  block-size: 230px;
+  overflow: hidden;
+  mask-image: linear-gradient(180deg, rgb(2 0 36 / 100%) 80%, rgb(255 255 255 / 0%) 100%);
+}
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -995,6 +995,8 @@ Comments:
   Subscribed: Subscribed
   Hearted: Hearted
 Up Next: Up Next
+Description:
+  Click to View Description: Click to View Description
 
 #Tooltips
 Tooltips:
--- a/static/locales/en_GB.yaml
+++ b/static/locales/en_GB.yaml
@@ -1059,6 +1059,8 @@ Comments:
   View {replyCount} replies: View {replyCount} replies
   Subscribed: Subscribed
 Up Next: 'Up Next'
+Description:
+  Click to View Description: Click to View Description
 
 # Toast Messages
 Local API Error (Click to copy): 'Local API Error (Click to copy)'


efb4f5ff-1298-471a-8973-3d47447115dc:
Could you explain why you chose for 300px? Is the behavior the same as YT? If not why is this better?

sabs21:
> Could you explain why you chose for 300px? Is the behavior the same as YT? If not why is this better?

I chose 300px because that's the height/max-block-size the description was already set to (look at .videoDescription in WatchVideoDescription.css).

kommunarr:
My (purely visual, untested) feedback: I like YT's UI for this better, particularly that the click area is the full description box so it's easier to click, the "more" button is casual, placed inside the flow of the description, and not too strongly competing for visual attention with an underline (and at least for En-US users, in lower case), and the corresponding "less". Also like that it does so by fully unraveling it rather than just making a scrollable div.

sabs21:
> My (purely visual, untested) feedback: I like YT's UI for this better, particularly that the click area is the full description box so it's easier to click, the "more" button is casual, placed inside the flow of the description, and not too strongly competing for visual attention with an underline (and at least for En-US users, in lower case), and the corresponding "less". Also like that it does so by fully unraveling it rather than just making the div scrollable.

Agreed. I went ahead and implemented your feedback in another branch. Here are screenshots of what I did. If this looks like its on a better track, I'll merge my new branch into this one. 

The only piece here that I didn't do is make the description fully unravel. The reason is that for very long descriptions, the description becomes longer than the recommended videos component on the right. This causes the page to auto load recommended videos which makes the page longer, causing the description to partially scroll back up and makes scrolling to the bottom of the description difficult.

On load:
image

Hovering over the description (entire description is clickable while collapsed):
image

Clicked description:
image

Show less:
image

efb4f5ff-1298-471a-8973-3d47447115dc:
@kommunarr any additional thought on the way its implemented now?

kommunarr:
Sorry for the delay, that looks great @sabs21! Please merge that in.

kommunarr (changes requested):
This is awesome!

**suggestion (blocking):** The main issue I can see is that the "...more" button doesn't work as a button when accessed through tabbing, and the "full box button" does appear in the tab order. My suggestion is to have the "...more" button work when tabbing, and have the "full box button" not appear in the tab order.

kommunarr:
Hi @sabs21, is there any way I can help or provide clarification on the issue & recommended fix?

sabs21:
> Hi @sabs21, is there any way I can help or provide clarification on the issue & recommended fix?

Apologies for the delay, admittedly since review has began on this I started a software dev job and programming more afterwards has been the last thing I wanted to do lol. I'm sure the accessibility piece is simply down to adding a "tabindex=0" to the description. I'd have to test it, but I'd assume hitting enter on focus should trigger its click event.

kommunarr:
No worries, I've been there before! To clarify, my suggestion is to remove the `tabindex=0` on the full box button (to remove it from the tab order) and move the `keydown` event listeners from the full box button to specifically the "Show more" button.
https://github.com/FreeTubeApp/FreeTube/blob/0a6c6e5c1534e91bb99e3154f1c82f5f8920a23e/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue#L25-L30

This way, when a user navigates using keyboard navigation, the full box button will not appear in the tab order button, and the "Show more" button will work when tabbing to it and pressing <kbd>Enter</kbd>.

sabs21:
> No worries, I've been there before! To clarify, my suggestion is to remove the `tabindex=0` on the full box button (to remove it from the tab order) and move the `keydown` event listeners from the full box button to specifically the "Show more" button.
> 
> https://github.com/FreeTubeApp/FreeTube/blob/0a6c6e5c1534e91bb99e3154f1c82f5f8920a23e/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue#L25-L30
> 
> This way, when a user navigates using keyboard navigation, the full box button will not appear in the tab order, and the "Show more" button will work when tabbing to it and pressing Enter.

I'll give this a shot now and see where it goes. Make sense to me on the surface!

sabs21:
Unfortunately I'm battling the eslinter here. I'd like to have the description overlay expand the description on click similar to how YT's behavior is. The problem being that eslint wants me to make the overlay focusable which of course goes against what we want.

I could just push a version without the overlay, but that greatly reduces the clickable area and would likely make the 'more...' button hard to find since it exists on the right side of the description.

kommunarr:
Gotcha, if you want to follow YT's implementation for advice, here's how they do it: make the "box button" not an actual button (i.e., remove the role=button), make the "Show more" one have role=button and have no click/keydown handling (which instead handled by bubbling up to its box button parent).

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -1,19 +1,49 @@
...
+      </span>
+      <span
+        v-else
+        class="descriptionStatus"
+        tabindex="0"
+        @keydown.space.prevent="expandDescription"
+        @keydown.enter.prevent="expandDescription"
+      >
+        {{ $t("Description.Expand Description") }}
+      </span>
```
kommunarr, line 31: This is working great through tab navigation! Last comment; could you add role="button" to these two?

sabs21: My bad! I accidentally removed those, just pushed that now.

kommunarr (commented):
LGTM; will approve after the release.

PikachuEXE:
I need example video for testing `Short descriptions do not show the 'Click to View Description' button`

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Pre-approval

kommunarr (commented):
Discovered one more issue; the "Show more" button is at the bottom of the tab order, so if a video description has links, you will have to tab through all of them to actually click the "Show more" button. It should instead be at the top of the tab order for the container. One possible solution is to [use the CSS order property](https://developer.mozilla.org/en-US/docs/Web/CSS/order). [Here's a video with links in the description as an example.](https://youtu.be/N43mwATceXc)

sabs21:
> Discovered one more issue; the "Show more" button is at the bottom of the tab order, so if a video description has links, you will have to tab through all of them to actually click the "Show more" button. It should instead be at the top of the tab order for the container. One possible solution is to [use the CSS order property](https://developer.mozilla.org/en-US/docs/Web/CSS/order). [Here's a video with links in the description as an example.](https://youtu.be/N43mwATceXc)

Great idea, I wound up flipping the ordering such that the buttons came first in the HTML, then used flexbox + order to adjust physical placement while keeping the desired tab ordering.

kommunarr:
Pre-existing: seeing template errors for videos with no description ([ex](https://youtu.be/PNo3yQ-h_ZQ)) due to no use of the null access operator.

For reference, [here's a video](https://youtu.be/floHBfe5-YM) with a short description, where it properly does not show the "Show more" button.

**issue:** Sorry for all the different rounds of feedback, but I think this should be the last! I don't mind the "Show less" button being at the top, but it should have the same top padding as the `p.description` that it replaces.

sabs21:
I went ahead and changed the visual ordering of the show less button so that the padding is normal once again. I didn't notice that the show less button was ordered differently due to me messing around with CSS locally in the dev tools. The spacing now appears normal once again too.

image

kommunarr:
As a byproduct of this change, clicking the Show more button now opens at the end of the container (i.e., scrolled all the way down) instead of at the top.

sabs21:
> As a byproduct of this change, clicking the Show more button now opens at the end of the container (i.e., scrolled all the way down) instead of at the top.

I reverted the previous change and just added the margin to the top of the show less button

kommunarr (approved):

PikachuEXE:
Anyone got a video with short description for testing?

PikachuEXE:
Just got error when trying to find a video with short description
https://youtu.be/ASeJV9KGAOc

image

efb4f5ff-1298-471a-8973-3d47447115dc:
Show Less is now seen in the top of the description box but i expected it to be shown. Was this an intentional change?

sabs21:
> Show Less is now seen in the top of the description box but i expected it to be shown. Was this an intentional change?

No I didn't expect that to happen, I'll work on fixing that now

sabs21:
Just fixed that, my guess is that when I merged the latest development, there was some css that caused the show less button to appear at the top.

efb4f5ff-1298-471a-8973-3d47447115dc:
I cant remember `...more` to be this far to the right. Did this also break?

image

sabs21:
> I cant remember `...more` to be this far to the right. Did this also break?
> 
> image

No, that hasn't changed. You can see this same padding in a screenshot within my September 16th post https://github.com/FreeTubeApp/FreeTube/pull/5665#issuecomment-2354046921

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
When expanding the description it scrolls all the way to the bottom instead of starting at top


https://github.com/user-attachments/assets/f83e2faf-dcba-4e03-ab68-cc01c5884046



Noticed a difference on FT vs YT

YT you can highlight text and open the link in the preview

https://github.com/user-attachments/assets/2dd6f0ea-4374-4ceb-8e43-7d7901417271

This isnt possible within FT. I can only highlight all the text within the collapsed container


https://github.com/user-attachments/assets/9755f161-9057-48f4-8a02-27abaf0be34b

Not able to highlight text is minor but maybe the url is?

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc https://github.com/FreeTubeApp/FreeTube/pull/6461

PR #5693, opened 2024-09-12
Use constants for API names

ChunkyProgrammer:
# Use constants for API names

## Pull Request Type
- [x] Other

## Related issue
Since we reuse 'local' and 'invidious' a lot in our code, I thought it'd be a good idea to use constants for those instead

## Description
- Update code so we use constants
- Fixes a bug for Hashtag page where it wouldn't use invidious if you have backend fallback enabled (this is leftover from when we only supported hashtags for local api)

## Desktop
- **OS:** Linux Mint
- **OS Version:** 22

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,5 +1,5 @@
 // IPC Channels
-const IpcChannels = {
+export const IpcChannels = {
   ENABLE_PROXY: 'enable-proxy',
   DISABLE_PROXY: 'disable-proxy',
   OPEN_EXTERNAL_LINK: 'open-external-link',
@@ -42,7 +42,7 @@ const IpcChannels = {
   SET_INVIDIOUS_AUTHORIZATION: 'set-invidious-authorization'
 }
 
-const DBActions = {
+export const DBActions = {
   GENERAL: {
     CREATE: 'db-action-create',
     FIND: 'db-action-find',
@@ -80,7 +80,7 @@ const DBActions = {
   },
 }
 
-const SyncEvents = {
+export const SyncEvents = {
   GENERAL: {
     CREATE: 'sync-create',
     UPSERT: 'sync-upsert',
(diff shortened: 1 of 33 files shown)


src/renderer/views/Hashtag/Hashtag.js
```diff
@@ -58,13 +59,13 @@ export default defineComponent({
...
       this.hashtagContinuationData = null
       this.videos = []
-      this.apiUsed = 'local'
+      this.apiUsed = API_DATA_SOURCES.LOCAL_API
       this.pageNumber = 1
     },
 
     getHashtag: async function() {
       const hashtag = decodeURIComponent(this.$route.params.hashtag)
-      if (this.backendFallback || this.backendPreference === 'local') {
```
PikachuEXE, line 67: Sneaky fix?

src/constants.js
```diff
@@ -94,27 +94,21 @@ const SyncEvents = {
...
-  MAIN_PROFILE_ID,
-  MOBILE_WIDTH_THRESHOLD,
-  PLAYLIST_HEIGHT_FORCE_LIST_THRESHOLD,
-  SEARCH_CHAR_LIMIT,
-  ABOUT_BITCOIN_ADDRESS,
+export const ABOUT_BITCOIN_ADDRESS = '1Lih7Ho5gnxb1CwPD4o59ss78pwo2T91eS'
+
+export const API_DATA_SOURCES = {
+  LOCAL_API: 'local',
+  INVIDIOUS_API: 'invidious'
```
PikachuEXE, line 113: Since you named it as `API_DATA_SOURCES`, why include `API` in constants names too?

src/renderer/helpers/api/invidious.js
```diff
@@ -384,7 +385,7 @@ export function filterInvidiousFormats(formats) {
   })
 }
 
-export async function getHashtagInvidious(hashtag, page) {
+export async function getHashtagInvidious(hashtag, page = 1) {
```
PikachuEXE, line 388: Sneaky but fine

src/constants.js
PikachuEXE: This is not addressed yet

absidue (dismissed):

PikachuEXE (dismissed):

PikachuEXE:
3 places left with `'invidious'`
image

ChunkyProgrammer:
@PikachuEXE  I don't think the constant can be used in the .vue file

absidue:
Not easily no, with the options API you would have to add it in `data` and then access it in the template (everything referenced inside the template implicitly starts with `this.`).

With the `<script setup>` Composition API in Vue 2 you could do it a bit cleaner, although the compiler would just add it to `this.` for you, but in the Vue 3 `<script setup>` Composition API, it would actually directly reference the variables.

PikachuEXE:
How about making `backendFallback || backendPreference === 'invidious'` a computed property? (constants can be used there?)

kommunarr:
+1 on the suggestion to make each of the `this.backendPreference === API_DATA_SOURCES.LOCAL && this.backendFallback`, `process.env.SUPPORTS_LOCAL_API && this.backendPreference === API_DATA_SOURCES.INVIDIOUS && this.backendFallback`, etc. shared boolean variables somehow, even if in the worst case the computed property declarations have to be repeated for each using file. I tend to have trouble mentally parsing these and quickly grasping what they logically represent.

absidue:
@kommunarr I feel like I've explained this before but just a reminder that anything starting with `process.env` is a build flag and are replaced by values at build time and should be as close to where it is actually used as possible so that terser actually realises the code is used or unused and can remove the unused code during the build process (e.g. anything inside of an `if (false)` can get removed).

And yes I am fully aware that is using the options API means that a lot of the local API code doesn't get treeshaken properly yet, but lets try to avoid making it intentionally worse.

@PikachuEXE Was only suggesting moving the conditions out of the template, not refactoring everything.

kommunarr:
Function that takes a process argument then?

PR #5698, opened 2024-09-13
Fix a few memory leaks while tearing down the player

absidue:
# Fix a few memory leaks while tearing down the player

## Pull Request Type

- [x] Bugfix

## Description
This pull request addresses two current memory leaks that occur while tearing down the player or more specifically two things that don't get torn down at the moment.

The first issue is that shaka-player's component registrations are global, however as we create our custom components with references to variables that are only meant to be used for the current player session, we need to clean those registrations up otherwise it will prevent the variables in question from being garbage collected until you play another video. Unfortunately shaka-player doesn't have a way to unregister a component, so I decided to overwrite the registration with `null`.

The second issue is that shaka-player's `destroy()` method is asynchronous, however as Vue's lifecycle hooks are synchronous, it means that the `destroy()` method isn't finished by the time FreeTube and Vue consider the player to be destroyed. To solve that I decided to introduce a method to the `ft-shaka-video-player` component which calls the destroy method on shaka-player. That way we can destroy the player first, wait for it to be finished and then prodeed with unmounting the Vue component.

This doesn't fix all of the problems, as shaka-player doesn't fully clean up after itself, hopefully some of their recent pull requests will help with that and I've also opened an issue about some of the UI elements not getting removed properly. Additionally there are some other things that seem to hang around (according to the Memory tab in the devtools), but I'm not sure why, my guess is that it is a Vue 2 bug, if that is the case then the only solution is updating to Vue 3.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** c90674d677f354bed60ff3f9340dde181494da4f

## Additional context

There may be code conflicts between this pull request and https://github.com/FreeTubeApp/FreeTube/pull/5677, in which case please make sure that pull request is merged first.


PikachuEXE (dismissed):
Last guy remember to merge after shaka player PR merged~

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #5724, opened 2024-09-22
[Bug]: v0.21.3-nightly-4778 Beta No longer refreshes subscriptions upon launch.


DoYouKnowMyNamePlz:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

Open freetube

### Expected Behavior

It used to refresh upon launching of freetube but in nightly 4776 and 4778 it does not.

### Issue Labels

usability issue

### FreeTube Version

v0.21.3-nightly-4778 Beta

### Operating System Version

WIndows 11 23H2 build 22631.4169

### Installation Method

.exe

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

Release v0.21.3 Beta

### Additional Information

This did not happen in normal beta builds, but when switching to nightly build 4776 it no longer refreshed after launching. This is even after having my computer shut down for 12hrs+. Right now it says "Videos feed last updated: 17 hours ago". Either I am missing something or something broke in an update.

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
Go to Settings -> Subscription Settings -> Make sure Fetch Feed Automatically is enabled

efb4f5ff-1298-471a-8973-3d47447115dc:
Wait..I can reproduce this behavior

PR #5743, opened 2024-09-24
Fix persistent cache incorrectly having higher priority then auto fetch feed option

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/5724

## Description

There is a bug introduced in https://github.com/FreeTubeApp/FreeTube/pull/5185
The cache when available is showing cached entries instead of consulting auto fetch feed option first
This PR fixes it

## Testing 

(A) (from https://github.com/FreeTubeApp/FreeTube/issues/5724)
- Disable `Fetch Feed Automatically`
- Switch/mess around with profiles if wanted
- Load videos, live, shorts, community posts
- Open new window(s) to see effects (should NOT auto fetch but shows cached entries)
- Reboot app
- Switch to different profiles to see effects (should NOT auto fetch but shows cached entries)

(B) (missing test case in previous PR)
- Enable `Fetch Feed Automatically`
- Switch/mess around with profiles if wanted
- Open new window(s) to see effects (should auto fetch)
- Reboot app
- Switch to different profiles to see effects (should auto fetch)

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


absidue:
I haven't tested it but looking at the code this pull request is wrong too, as it always bypasses the cache, but it is only supposed refresh automatically the first time you visit the subscription page in each window.

E.g.
1. Open window to subscription page -> Should refresh
2. Open a video
3. Go back to subscription page -> Should use cache

absidue:
I think C1 should behave the same as it did before the persistent cache, as that's also what the description of the setting says. One way to implement it would be to add a boolean to the store e.g. `hasAutoRefreshed`, which isn't synced across windows, that way it would refresh every time a new window is opened but not when you visit the subscriptions page afterwards.

PikachuEXE:
What about profile switch?
Coz that's included in option description as well

absidue:
Before the persistent cache, it only refreshed when the profile changed if there were missing cache entries (it calls `loadVideosFromCacheSometimes`). So if you launched with the all channels profile, switching to any other profile wouldn't refresh, as all cache entries already existed.

PikachuEXE:
So what's expected is:
(A) New window - auto fetch once per window
(B) Switch profile - auto fetch only on cache miss case

Let me know if anything incorrect

PikachuEXE:
Updated

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PikachuEXE:
@ChunkyProgrammer :wave:

ChunkyProgrammer (approved):

Issue #5641, opened 2024-09-02
[Bug]: Linux AppImage Built with FUSE v2: Only FUSE v3 Included with Newer Distros


the-burrito-triangle:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1)  Download FreeTube AppImage on Fedora 41 or CachyOS (Gnome DE) from github.
2)  Right click AppImage and set Executable as Program to enabled.
3)  Open a terminal in the directory containing FreeTube AppImage (or cd to directory).
4)  Run FreeTube: 
$ ./freetube_0.21.3_amd64.AppImage
FreeTube returns an error about FUSE not being present:
"Cannot mount AppImage, please check your FUSE setup."

On Fedora 41:
$ rpm -qa | grep fuse
fuse3
fuse3-libs
fuse-common

Old FUSE v2 is not installed on Fedora 41 by default (or CachyOS, an Arch Linux based distro).

After installing appropriate FUSE v2 packages FreeTube opens and runs normally:
$ sudo dnf install fuse fuse-libs

NOTE: FUSE v3 is included with at least Fedora 39. I'm not sure what distros only ship FUSE v2 packages, but switching to FUSE v3 should not be an issue from what I can tell as Flatpak and distro maintained FreeTube packages should not be affected by this change. The AppImage should be okay to update unless there was a technical reason for using FUSE v2 over v3.

### Expected Behavior

AppImage should run on newer distros.

### Issue Labels

usability issue

### FreeTube Version

freetube_0.21.3_amd64.AppImage

### Operating System Version

Fedora 41, CachyOS

### Installation Method

.AppImage

### Primary API used

Local API

### Nightly Build

[...]

PR #5747, opened 2024-09-25
Convert X64 AppImage to static runtime

### LINKED ISSUE #5714: [Feature Request]: libfuse2 dependency is EOL for the appimage version.
### Guidelines

### Problem Description

libfuse2 dependency is EOL for appimage version.

### Proposed Solution

migration to libfuse3.

### Alternatives Considered

n/a

### Issue Labels

### Additional Information

_No response_

Samueru-sama:
# Convert X64 AppImage to static runtime

## Pull Request Type
- [x] Feature Implementation

## Related issue
Closes #5714 #5641 but not fully since I was only able to do this for the X64 build.

## Description
This change drops libfuse2 dependency from the AppImage, which prevents them from working on newer distros that no longer ship libfuse2.

## Testing
[Artifact](https://github.com/Samueru-sama/FreeTube/actions/runs/11028756249/artifacts/1975640207)

image

## Desktop

- **OS:** Artix Linux
- **OS Version:** Rolling release (Kernel: Linux 6.10.10-artix1-1)
- **FreeTube version:** 0.21.3-nightly-13

## Additional context
I was not able to fix this on the arm builds because even though I tried using the same runners that the upload actions for those use, it complains that it can't execute the appimage to extract it, which is needed in order to convert it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,6 +103,18 @@ jobs:
       if: contains(matrix.runtime, 'arm64')
       run: yarn run build:arm64
 
+    - name: Convert X64 AppImage to static runtime
+      if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-x64')
+      run: |
+        sudo apt install desktop-file-utils
+        cd build
+        wget "https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage" -O ./appimagetool.AppImage
+        chmod +x *.AppImage && ./FreeTube-${{ steps.versionNumber.outputs.result }}.AppImage --appimage-extract
+        rm -f ./FreeTube*.AppImage
+        ./appimagetool.AppImage --comp zstd --mksquashfs-opt -Xcompression-level --mksquashfs-opt 20 \
+          -n ./squashfs-root ./FreeTube-${{ steps.versionNumber.outputs.result }}.AppImage
+        rm -rf ./squashfs-root ./appimagetool.AppImage
+
     - name: Upload Linux .zip x64 Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-x64')


absidue (commented):
Could you please use the filename everywhere, as you currently have a mix of using the filename in some places and wild cards in others. Additionally it probably makes sense to assign the filename to a variable after the `cd build` command and reference that variable.

I haven't tested the generated AppImage yet (need to boot up my Ubuntu VM and figure out how to run and then afterwards remove AppImages).

I'm okay with doing this after build hack for the moment but in the long run it would be much better if electron-builder handled that for us, that way people would be able to reproduce the builds easier on their own machines without having to do these steps manually after the build. As for the why it does't work for the arm AppImage, that's because the GitHub runners are x64, so it would probably be a lot better if electron-builder produced the AppImages with the static runtimes in the first place instead of having to do this patching after the fact.

Samueru-sama:
> Could you please use the filename everywhere, as you currently have a mix of using the filename in some places and wild cards in others. Additionally it probably makes sense to assign the filename to a variable after the `cd build` command and reference that variable.
> 
> I haven't tested the generated AppImage yet (need to boot up my Ubuntu VM and figure out how to run and then afterwards remove AppImages).
> 
> I'm okay with doing this after build hack for the moment but in the long run it would be much better if electron-builder handled that for us, that way people would be able to reproduce the builds easier on their own machines without having to do these steps manually after the build. As for the why it does't work for the arm AppImage, that's because the GitHub runners are x64, so it would probably be a lot better if electron-builder produced the AppImages with the static runtimes in the first place instead of having to do this patching after the fact.

Done, worth mentioning that using appimagetool gives the ability of creating the .zsync files that allow for delta updates in the AppImage with AppImageUpdate, which I will try to add after this PR.

absidue:
Just noticed you only did the changes for the workflow that creates the nightly builds but not for the release workflow, could you please make the changes there too?

It would also be great if you could create a feature request on the electron-builder GitHub repo to have the static runtime functionality built-in/out of the box and comment the link to the feature request here, before we consider merging these changes, so that there is a bit of hope that this after build modification won't have to stick around permanently.

Samueru-sama:
> Just noticed you only did the changes for the workflow that creates the nightly builds but not for the release workflow, could you please make the changes there too?
> 
> It would also be great if you could create a feature request on the electron-builder GitHub repo to have the static runtime functionality built-in/out of the box and comment the link to the feature request here, before we consider merging these changes, so that there is a bit of hope that this after build modification won't have to stick around permanently.

Ok will do, sorry I've been busy with other stuff. 

I also recently found out that the x86_64 appimagetool can be used to make appimages for other arch, the only thing is that I need to figure out how to extract them on the x86_64 runner.

efb4f5ff-1298-471a-8973-3d47447115dc:
> > Just noticed you only did the changes for the workflow that creates the nightly builds but not for the release workflow, could you please make the changes there too?
> > It would also be great if you could create a feature request on the electron-builder GitHub repo to have the static runtime functionality built-in/out of the box and comment the link to the feature request here, before we consider merging these changes, so that there is a bit of hope that this after build modification won't have to stick around permanently.
> 
> Ok will do, sorry I've been busy with other stuff.
> 
> I also recently found out that the x86_64 appimagetool can be used to make appimages for other arch, the only thing is that I need to figure out how to extract them on the x86_64 runner.

@Samueru-sama what are things that still need to be addressed in this PR?

Samueru-sama:
Well the PR only fixes the issue for the `x64` builds. 

Also while it works for the nightly builds, I'm not 100% sure if what I did for the release workflow is what needs to be done.

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

Issue #5749, opened 2024-09-25
[Bug]: RSS Feed Stops Working If The Profile is Overused


CTRLexe:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

![image](https://github.com/user-attachments/assets/d09201ee-abd9-41e7-905b-5cd27c14a7a1)
I click refresh feed after it didn't show me all the channels. It instead shows me this.

### Expected Behavior

Shows new videos from creators regardless of profile.

### Issue Labels

content not loading

### FreeTube Version

v0.21.3

### Operating System Version

Windows 10

### Installation Method

.exe

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

v0.21.2

### Additional Information

It doesn't matter which API I use or whether or not I use a VPN. If I had to guess, I'm still being rate limited.
It also seems to affect smaller profiles if I have RSS enabled.

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

rauldipeas:
Same situation here.

karolyi:
Same.

riverpiers:
You cannot see subscriptions when there are more than 124 in the default category of all channels. If you don’t have separate categories, you won’t see any videos currently. “Temporary solution”:
1. Go to the subscriptions tab and create several thematic categories.
2. Add subscriptions according to, for example, the channel’s topic, up to 124 subscriptions in one category.
3. Now you can see subscriptions outside the main aggregate category. (Channels not included in the subscription list will not be fetched).

karolyi:
@riverpiers all I can say is that I didn't add any new subs, and it started failing today. I don't think it's related to what you claim.

[...]

PR #5756, opened 2024-09-28
Do not overwrite cache entries when hitting RSS ratelimits

absidue:
# Do not overwrite cache entries when hitting RSS ratelimits

## Pull Request Type

- [x] Bugfix

## Related issue
Related to #5749

## Description
When hitting the RSS ratelimits we currently treat it the same as if the channel had no uploads and overwrite the existing cache entries with empty arrays, this pull request fixes that so that we don't overwrite them if YouTube returns a 403.

## Testing
I haven't directly tested this, as I am not personally blocked but it should work.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 30f95e5700ac57bc18700fa24e64af91fd3388f4

### DIFF
--- a/src/renderer/components/subscriptions-live/subscriptions-live.js
+++ b/src/renderer/components/subscriptions-live/subscriptions-live.js
@@ -206,10 +206,13 @@ export default defineComponent({
         channelCount++
         const percentageComplete = (channelCount / channelsToLoadFromRemote.length) * 100
         this.setProgressBarPercentage(percentageComplete)
-        this.updateSubscriptionLiveCacheByChannel({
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
As our team cant didnt ran into the blocks maybe create test build and provide it to the users of that issue?

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5758, opened 2024-09-28
Change the Proxy Videos Through Invidious setting to only apply when using the Invidious API

absidue:
# Change the Proxy Videos Through Invidious setting to only apply when using the Invidious API

## Pull Request Type

- [x] Feature Implementation

## Description
Users have often been confused that the `Proxy Videos Through Invidious` setting applies to the local API too. Additionally if you have that setting turned on and the Invidious instance you are using is having problems you just end up in an endless loop of Invidious failing (deceptively the error message gets shown as a `Local API error` but it's still Invidious that is causing it).

So as discussed in the Matrix chat this pull request changes the `Proxy Videos Through Invidious` setting to only apply to the Invidious API. I haven't changed anything about the settings page other than the tooltip text, as it will still proxy the video if you fallback to the Invidious API when the local API errors.

I have updated the tooltip text for most languages, but skipped the right-to-left languages as editing the strings in those files is messy if you don't understand the language, so I'll leave those to the translators on weblate.

## Testing
1. Turn on the `Proxy Videos Through Invidious` setting
2. Test the local API and check in the network tab that the video and audio requests are going to `*.googlevideo.com` and not an Invidious instance.
3. Regression test: Test the Invidious API and check in the network tab that the video and audio requests are still going to your chosen Invidious instance.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 30f95e5700ac57bc18700fa24e64af91fd3388f4


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Everything looks good but i think it would make sense to add the following to prevent confusion:

- Disable the toggle when Local API is set as Primary API and Fallback is disabled.

absidue:
> Everything looks good but i think it would make sense to add the following to prevent confusion:
> 
>     * Disable the toggle when Local API is set as Primary API and Fallback is disabled.

Implemented

ChunkyProgrammer (approved):
lgtm

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #5777, opened 2024-10-01
Upgrade eslint 8 > 9

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

- Can't use `--ext` anymore, and using blob like `./**/*.js` is very slow compared to specific folders, so only 2 folders are specified for js/vue files
- Separate config files for json/yaml files due to I can't avoid recommended config from being applied on them
- `beforeDestroy` -> `beforeUnmount` via lint-fix
- Draft until https://github.com/import-js/eslint-plugin-import/pull/2996 released

## Screenshots 

image

## Testing 

- `yarn run lint`
- `yarn run lint-json`
- `yarn run lint-yml`

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


PikachuEXE:
Since we can switch from `standard` to `neostandard` later maybe we can just upgrade to ESLint 9 first?

eslint.yaml.config.mjs
```diff
@@ -0,0 +1,35 @@
...
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all
+})
+
+export default [
+  ...compat.extends(
```
ChunkyProgrammer, line 16: I dont think the compat is needed here. https://github.com/ota-meshi/eslint-plugin-yml?tab=readme-ov-file#new-config-eslintconfigjs

eslint.json.config.mjs
```diff
@@ -0,0 +1,35 @@
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+import { FlatCompat } from '@eslint/eslintrc'
+import js from '@eslint/js'
+import jsoncEslintParser from 'jsonc-eslint-parser'
+import eslintPluginJsonc from 'eslint-plugin-jsonc'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const compat = new FlatCompat({
```
ChunkyProgrammer, line 10: This doesnt seem to be used

ChunkyProgrammer (commented):
We should definitely try to fix: Separate config files for json/yaml files due to I can't avoid recommended config from being applied on them 

Since IDEs/code editors wont be able to highlight the issues

eslint.config.mjs
```diff
@@ -0,0 +1,128 @@
...
+      },
+
+      parser: vueEslintParser,
+      ecmaVersion: 2022,
+      sourceType: 'module',
+    },
+
+    settings: {
+      'vue-i18n': {
+        localeDir: './static/locales/{en-US,en-GB,ar,bg,ca,cs,da,de-DE,el,es,es-AR,es-MX,et,eu,fa,fi,fr-FR,gl,he,hu,hr,id,is,it,ja,ko,lt,nb-NO,nl,nn,pl,pt,pt-BR,pt-PT,ro,ru,sk,sl,sr,sv,tr,uk,vi,zh-CN,zh-TW}.yaml',
```
absidue, line 59: Please import the list instead of hardcoding it, so that we don't have to maintain it in multiple places. Using `import activeLocales from './static/locales/activeLocales.json' with { type 'json' }` and then calling `.join(',')` on the imported array should work.

According to the history drop down on https://nodejs.org/api/esm.html#import-attributes the `with` keyword is supported in Node 21+ and in the LTS versions 18.20+ and 20.10+, so we should be fine using it.

PikachuEXE: With dynamic import, got warning (probably non issue assuming it eventually no longer Experimental 
```js
const { default: activeLocales } =
  await import('./static/locales/activeLocales.json', { with: { type: 'json' } })
```
image

Also static import would only cause syntax error with node 20.17.0
image
image

eslint.json.config.mjs
PikachuEXE: Updated~

eslint.yaml.config.mjs
PikachuEXE: Yup I forgot to check it updated now

PikachuEXE:
@ChunkyProgrammer config files merged
Turns out `localeDir` needed in all config blocks (even for json/yaml

eslint.config.mjs
```diff
@@ -0,0 +1,207 @@
...
+    },
+
+    rules: {
+      'no-tabs': 'off',
+      'comma-spacing': 'off',
+      'no-irregular-whitespace': 'off',
+    },
+
+    settings: {
+      'vue-i18n': {
```
ChunkyProgrammer, line 156: I dont think this would be needed here since this is for json files and our i18n files are yaml

eslint.config.mjs
absidue: The reason it didn't work with the static import is because of the missing `:` between `type` and `'json'`.

package.json
```diff
@@ -37,12 +37,12 @@
     "lint-all": "run-p lint lint-json",
     "lint": "run-p eslint-lint lint-style",
     "lint-fix": "run-p eslint-lint-fix lint-style-fix",
-    "eslint-lint": "eslint --ext .js,.vue ./",
-    "eslint-lint-fix": "eslint --fix --ext .js,.vue ./",
-    "lint-json": "eslint --ext .json ./",
+    "eslint-lint": "eslint --config eslint.config.mjs './src/**/*.js' './src/**/*.vue' './static/**/*.js'",
+    "eslint-lint-fix": "eslint --config eslint.config.mjs --fix './src/**/*.js' './src/**/*.vue' './static/**/*.js'",
+    "lint-json": "eslint --config eslint.config.mjs './static/**/*.json'",
```
absidue, line 42:
```suggestion
    "eslint-lint": "eslint --config eslint.config.mjs \"./src/**/*.js\" \"./src/**/*.vue\" \"./static/**/*.js\"",
    "eslint-lint-fix": "eslint --config eslint.config.mjs --fix \"./src/**/*.js\" \"./src/**/*.vue\" \"./static/**/*.js\"",
    "lint-json": "eslint --config eslint.config.mjs \"./static/**/*.json\"",
```

package.json
```diff
@@ -37,12 +37,12 @@
...
     "lint-style": "stylelint \"**/*.{css,scss}\"",
     "lint-style-fix": "stylelint --fix \"**/*.{css,scss}\"",
-    "lint-yml": "eslint --ext .yml,.yaml ./",
+    "lint-yml": "eslint --config eslint.config.mjs './**/*.yml' './**/*.yaml'",
```
absidue, line 45:
```suggestion
    "lint-yml": "eslint --config eslint.config.mjs \"./**/*.yml\" \"./**/*.yaml\"",
```

absidue (commented):
Windows CMD doesn't interpret single quotes, so it passes them through to eslint and then eslint fails to find files that start and end with single quotes. I've suggested changes to switch to double quotes instead.

eslint.config.mjs
```diff
@@ -0,0 +1,207 @@
...
+      'vue/no-template-key': 'warn',
+      'vue/no-useless-template-attributes': 'off',
+      'vue/multi-word-component-names': 'off',
+      'vuejs-accessibility/no-onchange': 'off',
+
+      // rule for Vue 3.x , FT is on 2.x
+      'vue/no-v-for-template-key-on-child': 'off',
+      'vue/no-v-for-template-key': 'error',
+      // To be fixed on Vue 3 migration
+      'vue/no-deprecated-v-on-native-modifier': 'warn',
```
absidue, line 92: Please remove these rules, the `vue/no-deprecated-v-on-native-modifier` line is turning on a rule that is off by default for Vue 2 and resulting in a warning getting logged.

```suggestion
```

eslint.config.mjs
```diff
@@ -0,0 +1,207 @@
...
+import eslintPluginYml from 'eslint-plugin-yml'
+import yamlEslintParser from 'yaml-eslint-parser'
+
+const { default: activeLocales } =
+  await import('./static/locales/activeLocales.json', { with: { type: 'json' } })
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
```
absidue, line 24: We can use `import.meta.dirname` here, as it is supported in Node 21.2.0+ and LTS 20.11.0+ and we already use it in the `_scripts/patchShaka.mjs` file.

```suggestion
const compat = new FlatCompat({
  baseDirectory: import.meta.dirname,
```

PikachuEXE:
Updated~

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #5779, opened 2024-10-01
Added functionality to export a single playlist

### LINKED ISSUE #5366: [Feature Request]: Importing/Exporting only 1 playlist.
### Guidelines

### Problem Description

Currently, you export the entire database of your playlists, and importing replaces the database of your playlists. However, if I want to "share" a playlist to another FreeTube user (giving an exported file), I would be sharing my entire playlist collection (which isn't what I want to do) and it would replace their playlists (which they probably don't want either)

### Proposed Solution

A solution could be having buttons similar to Import/Export playlist, but that only exports 1 playlist of choosing, and another button to import playlist without replacing the full database of playlist (it could replace playlist with the exact same same).

### Alternatives Considered

A similar solution would to put a button here in the playlist that would export the playlist (though, it seems to have enough buttons already)
image

Then there could be a button in the "Your Playlist" section that would import a playlist file.

### Issue Labels

improvement to existing feature, new feature

### Additional Information

_No response_

Zeabyte:
# Single Playlist Export

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #5366

## Description

This PR adds a button "Export Playlist" to the playlist-info component, that when clicked exports a single playlist with the file name "freetube-playlist-{playlistName}-{date}.db".

## Screenshots 

After:
image

## Testing 

I have tested this and found no problems. However, the UX is not perfect because this is only exporting. Importing currently requires going to data settings and there is no "Do you wish to overwrite the playlist" or "... add the videos to the local playlist" with the same name.

## Desktop

- **OS:** Fedora Workstation
- **OS Version:** 40
- **FreeTube version:** latest from development

## Additional context

One could further abstract the way files are saved to disk. I have copied it from data-settings.js.
I have reused the locale token `Settings.Data Settings.Unable to write file`.

### DIFF
--- a/src/renderer/components/playlist-info/playlist-info.js
+++ b/src/renderer/components/playlist-info/playlist-info.js
@@ -265,6 +265,13 @@ export default defineComponent({
       return this.selectedUserPlaylist.videos.length - this.userPlaylistUniqueVideoIds.size
     },
 
+    exportPlaylistButtonVisible: function() {
+      // Only online playlists can be shared
+      if (!this.isUserPlaylist) { return false }
+
+      return this.videoCount > 0
+    },
+
     deletePlaylistButtonVisible: function() {
       if (!this.isUserPlaylist) { return false }
       // Cannot delete during edit
@@ -278,7 +285,6 @@ export default defineComponent({
       // Only online playlists can be shared
       if (this.isUserPlaylist) { return false }
 
-      // Cannot delete protected playlist
       return !this.hideSharingActions
     },
 
--- a/src/renderer/components/playlist-info/playlist-info.vue
+++ b/src/renderer/components/playlist-info/playlist-info.vue
@@ -170,18 +170,18 @@
             @click="toggleCopyVideosPrompt"
           />
           <ft-icon-button
+            v-if="exportPlaylistButtonVisible"
+            :title="$t('User Playlists.Export Playlist')"
+            :icon="['fas', 'file-arrow-down']"
+            theme="secondary"
+            @click="handlePlaylistExport"
+          />
+          <ft-icon-button
             v-if="!editMode && userPlaylistDuplicateItemCount > 0"
             :title="$t('User Playlists.Remove Duplicate Videos')"
             :icon="['fas', 'users-slash']"
             theme="destructive"
             @click="showRemoveDuplicateVideosPrompt = true"
-          />
-          <ft-icon-button
-            v-if="!editMode"
-            :title="$t('User Playlists.Export Playlist')"
-            :icon="['fas', 'file-arrow-down']"
-            theme="secondary"
-            @click="handlePlaylistExport"
           />
           <ft-icon-button
             v-if="!editMode && userPlaylistAnyVideoWatched"


src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -412,6 +415,41 @@ export default defineComponent({
       showToast(this.playlistDeletionDisabledLabel)
     },
 
+    handlePlaylistExport: async function () {
+      const dateStr = getTodayDateStrLocalTimezone()
+      const title = this.selectedUserPlaylist.playlistName.replaceAll(' ', '-')
```
PikachuEXE, line 420: IMO spaces are usually replaced with `_` more often then `-`
I think only for SEO (URL) people prefer using `-`
Let me know if that's wrong :)

Also we might need to replace all invalid chars, e.g. `/` on MacOS, different for other OSes
(handling user input is nasty)

image

```suggestion
      const title = this.selectedUserPlaylist.playlistName.replaceAll(' ', '_')
```

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -412,6 +415,41 @@ export default defineComponent({
       showToast(this.playlistDeletionDisabledLabel)
     },
 
+    handlePlaylistExport: async function () {
+      const dateStr = getTodayDateStrLocalTimezone()
+      const title = this.selectedUserPlaylist.playlistName.replaceAll(' ', '_').replaceAll(/["%*/:<>?\\|]/g, '_')
```
PikachuEXE, line 420: Might need to replace more chars
See
https://www.mtu.edu/umc/services/websites/writing/characters-avoid/

I use `Special !@#$%^&*()-_=+[]{};:'".>/?\| List` as test name and I got `freetube-playlist-Special_!@#$_^&_()-_=+[]{};_'_.______List-2024-10-07` on MacOS
Need a windows guy to test the result

Zeabyte: @PikachuEXE I thought that the characters I listed in the Regexp would be those that are absolutely illegal for use in certain OS’s. The guide you linked calls some absolutely common, usual and legal characters "illegal" and "forbidden". So, just to confirm, this is more about conventions, right?

Thanks for having a look at it.

PikachuEXE: I didn't have access to Windows machine so I can't test it
Here is a screenshot of forbidden char message
2024-10-07 18_55_11-重新命名

I will create a custom build to test on Windows machine later

PikachuEXE: On windows I have `freetube-playlist-Special_!@#$_^&_()-_=+[]{};_'_.______List-2024-10-07.db`, same as MacOS

PikachuEXE (dismissed):
I assume Linux is fine

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc @absidue 👋

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
- Dont show the export button when playlist has no videos in it.
- Move the export button to right of the green copy playlists button so that all the green and red buttons are grouped correctly
image
- Discussion topic: Maybe this closes the issue. I know we dont have a "Import one Playlist button" but now we provide a way to export one playlist. So doesnt that automatically mean that the user can also import one playlist?

PikachuEXE:
> I know we dont have a "Import one Playlist button" but now we provide a way to export one playlist. So doesnt that automatically mean that the user can also import one playlist?

I think so

PikachuEXE:
I also see export button on online playlist (which does nothing)
Let me recheck later...

absidue (approved):

PikachuEXE:
@ChunkyProgrammer Though I can approve it but if you have time I rather have you take a look :)
Or I will just (partially self-)approve tomorrow

src/renderer/components/playlist-info/playlist-info.js
```diff
@@ -412,6 +421,41 @@ export default defineComponent({
...
```
ChunkyProgrammer, line 426: Non-blocking: Maybe we should merge the logic here with the logic for replacing the names of files created for screenshots? https://github.com/FreeTubeApp/FreeTube/blob/94fe62aff4c5b721033090a896738bd8006d4c3d/src/renderer/helpers/utils.js#L565

ChunkyProgrammer (approved):

Issue #5038, opened 2024-04-29
[Bug]: Videos duplicate themselves when importing playlists


Xindaris:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Add videos to a playlist. (Using the quick bookmark feature; unsure if that's relevant)
2. Export the playlist.
3. Import the playlist on a different computer.
4. Add videos to the playlist on that other computer, and export again.
5. Import on the first computer.
6. Some videos randomly get two "copies" in the playlist now.

### Expected Behavior

Playlists should merge properly when importing, and not add another copy of a video that's already in the playlist.

### Issue Labels

inconsistent behavior, usability issue

### FreeTube Version

v0.20.0 Beta

### Operating System Version

Windows 11

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

lbpth:
It is also duplicated when I try to add video that has existed in the playlist before.

For example, I have playlist named Piano
1 - I add the video named Turkish March to Piano playlist then the number video is 1
2 - I click the button + again (still above Turkish March video), the number video count 2 in the Piano playlist
So, how to avoid duplicates if I have more than 100 videos in the playlist, manual removal is not good solution.

efb4f5ff-1298-471a-8973-3d47447115dc:
Is this still present in v0.21.0?

PikachuEXE:
Is the duplicates always created for the same set of videos or complete random?
If always the same set of videos would you mind sharing them? (Video links, ids)

No idea how to reproduce it happens randomly (I tried once myself

Xindaris:
[...]

PR #5783, opened 2024-10-03
Update playlist import to only add duplicate playlist items sometimes

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/5038

## Description

Updates the behaviour of importing playlists **when** existing playlist found

Before: 
- Add playlist item unique by `playlistItemId`

After: 
- If both existing & incoming playlists have unique videos only (i.e. unique `videoId` values)
  Only add playlist items from incoming playlist with `videoId` absent in existing playlist
- If either existing or incoming playlist has duplicate videos, add videos unique by `playlistItemId` (old behaviour)

## Testing 

- Download files from https://github.com/FreeTubeApp/FreeTube/issues/5038#issuecomment-2380302656
- Clear playlist `Favorites` coz that's the name of playlist in those files
- Test following cases (skip some if you want, but A probably should not be skipped)

(A) Both existing & incoming playlists have unique videos only
- import PC1 twice
- Or manually add some videos (which is the case of https://github.com/FreeTubeApp/FreeTube/issues/5038)

(B) Only incoming playlist has "duplicate" videos
- import PC1, import PC2

(C) Only existing playlist has "duplicate" videos
- import PC2, import PC1

(D) Only existing playlist has "duplicate" videos
- import PC2 twice (should still have 4 items coz unique by `playlistItemId`)

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -901,48 +901,50 @@ export default defineComponent({
         const playlistObjectKeys = Object.keys(playlistObject)
         const playlistObjectHasAllRequiredKeys = requiredKeys.every((k) => playlistObjectKeys.includes(k))
 
-        if (playlistObjectHasAllRequiredKeys) {
-          const existingPlaylist = this.allPlaylists.find((playlist) => {
-            return playlist.playlistName === playlistObject.playlistName
-          })
-
-          if (existingPlaylist !== undefined) {
-            playlistObject.videos.forEach((video) => {
-              let videoExists = false
-              if (video.playlistItemId != null) {
-                // Find by `playlistItemId` if present
-                videoExists = existingPlaylist.videos.some((x) => {
-                  // Allow duplicate (by videoId) videos to be added
-                  return x.videoId === video.videoId && x.playlistItemId === video.playlistItemId
-                })
-              } else {
-                // Older playlist exports have no `playlistItemId` but have `timeAdded`
-                // Which might be duplicate for copied playlists with duplicate `videoId`
-                videoExists = existingPlaylist.videos.some((x) => {
-                  // Allow duplicate (by videoId) videos to be added
-                  return x.videoId === video.videoId && x.timeAdded === video.timeAdded
-                })
-              }
-
-              if (!videoExists) {
-                // Keep original `timeAdded` value
-                const payload = {
-                  _id: existingPlaylist._id,
-                  videoData: video,
-                }
-
-                this.addVideo(payload)
-              }
-            })
-            // Update playlist's `lastUpdatedAt`
-            this.updatePlaylist({ _id: existingPlaylist._id })
-          } else {
-            this.addPlaylist(playlistObject)
-          }
-        } else {
+        if (!playlistObjectHasAllRequiredKeys) {
           const message = this.$t('Settings.Data Settings.Playlist insufficient data', { playlist: playlistData.playlistName })
           showToast(message)
+          return
         }
+
+        const existingPlaylist = this.allPlaylists.find((playlist) => {
+          return playlist.playlistName === playlistObject.playlistName
+        })
+
+        if (existingPlaylist === undefined) {
+          this.addPlaylist(playlistObject)
+          return
+        }
+
+        playlistObject.videos.forEach((video) => {
+          let videoExists = false
+          if (video.playlistItemId != null) {
+            // Find by `playlistItemId` if present
+            videoExists = existingPlaylist.videos.some((x) => {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
> (A) Both existing & incoming playlists have unique videos only

Expected behavior: Not seeing any duplicates (This is the testcase from the issue)

> (B) Only incoming playlist has "duplicate" videos

Expected behavior: You will see the duplicate because PC2 got it in its db

> (C) Only existing playlist has "duplicate" videos

Expected behavior: You will see the duplicate because PC2 got it in its db

> (D) Only existing playlist has "duplicate" videos

Expected behavior: You will see the duplicate because PC2 got it in its db

LGTM

src/renderer/components/data-settings/data-settings.js
```diff
@@ -901,48 +903,63 @@ export default defineComponent({
...
-              } else {
-                // Older playlist exports have no `playlistItemId` but have `timeAdded`
-                // Which might be duplicate for copied playlists with duplicate `videoId`
-                videoExists = existingPlaylist.videos.some((x) => {
-                  // Allow duplicate (by videoId) videos to be added
-                  return x.videoId === video.videoId && x.timeAdded === video.timeAdded
-                })
-              }
+        const existingPlaylist = this.allPlaylists.find((playlist) => {
+          return playlist.playlistName === playlistObject.playlistName
```
absidue, line 913: It probably makes sense to try to match by `_id` first, because if you rename the playlist on one computer you still want it to be the same when you import it on the other computer instead of being treated as a completely separate playlist.

That being said I've just realised that the `_id` is considered an ignored key, so there is no way that a renamed playlist could ever match, because the import will make sure that the `_id` won't match, but that problem already existed before this pull request. Also weird that we don't save the lastPlayedAt field when it's a new playlist.

src/renderer/components/data-settings/data-settings.js
```diff
@@ -901,48 +903,63 @@ export default defineComponent({
...
+                return x.videoId === video.videoId && x.timeAdded === video.timeAdded
+              })
+            }
           } else {
-            this.addPlaylist(playlistObject)
+            // Find by `playlistItemId` if present
+            videoExists = existingPlaylist.videos.some((x) => {
+              // Allow duplicate (by videoId) videos to be added
+              return x.videoId === video.videoId
+            })
```
absidue, line 948: Could you please update this comment to match what the code does.

src/renderer/components/data-settings/data-settings.js
PikachuEXE: Agree but no way this belongs to this PR (plus need test cases)
I can make another PR to make it find by `_id` first later

`lastPlayedAt` not saved for new/existing playlist intentionally
Obvious for existing playlist but for new playlist it's never played locally and I consider it as a "local property" (e.g. importing other people's playlist)

absidue: I have to say I personally never use the playlists, but the way I use the other export and import options is for backup purposes, which means I only ever import my own stuff in my FreeTube install (dev mode is an exception because there I only use it to test stuff and that data is saved in a separate location), I would expect everything possible to be imported.

Importing someone else's files is probably a pretty rare use case, that is only really becoming possible because of the open PR to export single playlists.

absidue:
I think you may have missed what I was trying to say with please update the comment, as you updated a different one. The one I originally commented on says that it is searching by playlist item ID, but then the code right below it is searching by video ID...

src/renderer/components/data-settings/data-settings.js
```diff
@@ -901,48 +903,63 @@ export default defineComponent({
...
+              // Older playlist exports have no `playlistItemId` but have `timeAdded`
+              // Which might be duplicate for copied playlists with duplicate `videoId`
+              videoExists = existingPlaylist.videos.some((x) => {
+                // Allow duplicate (by videoId) videos to be added
+                return x.videoId === video.videoId && x.timeAdded === video.timeAdded
+              })
+            }
           } else {
-            this.addPlaylist(playlistObject)
+            // Find by `playlistItemId` if present
```
PikachuEXE, line 944:
```suggestion
```

PikachuEXE (commented):
For editing online

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #5793, opened 2024-10-04
Fix subscriptions CSV import shortening channel names

absidue:
# Fix subscriptions CSV import shortening channel names

## Pull Request Type

- [x] Bugfix

## Description
This pull request fixes the subscriptions CSV import removing the last character from channel names when they are quoted. We probably never noticed this problem before as we used to make API requests during the import, so the name was replaced with the one from the API response.

Noticed this while working on another pull request.

## Screenshots
youtub

## Testing
1. Download this file (remove the .txt file extension) and import it into FreeTube: [subscriptions.csv.txt](https://github.com/user-attachments/files/17261510/subscriptions.csv.txt)
2. Check in the side bar that the YouTube channel (the only one in the file) is actually called YouTube

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 43c6262eb5c46461b44ec40d0e5ee097b66dbd9a

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -220,7 +220,7 @@ export default defineComponent({
         return [...yt.matchAll(splitCSVRegex)].map(s => {
           let newVal = s[1]
           if (newVal.startsWith('"')) {
-            newVal = newVal.substring(1, newVal.length - 2).replaceAll('""', '"')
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer (approved):

PikachuEXE (approved):
Here have a bath tub guard
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5794, opened 2024-10-04
Fix hashtag page not using Invidious API when it should

ChunkyProgrammer:
# Fix hashtag page not using Invidious API when it should

## Pull Request Type
- [x] Bugfix

## Description
When Invidious API is preferred back end, it would use local api for hashtag page if fallback was enabled. This PR fixes it so that it will use IV API.

## Testing 
- with iv api + fallback
- go to a hashtag page
- make sure it doesn't use local api

## Desktop
- **OS:** Linux Mint
- **OS Version:** 22
- **FreeTube version:** latest

### DIFF
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -388,7 +388,7 @@ export function filterInvidiousFormats(formats) {
   })
 }
 
-export async function getHashtagInvidious(hashtag, page) {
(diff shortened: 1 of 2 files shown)


PikachuEXE (approved):
image

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5843, opened 2024-10-09
Add configurations for linting scripts and add global ignores

ChunkyProgrammer:
# Add configurations for linting scripts and add global ignores

## Pull Request Type
- [x] Other

## Related issue
It seems like the default recommended javascript config was being applied to our scripts files and this was causing issues with lefthook trying to lint the files

## Description
I added a specific config for our _scripts folder so we can still use eslint for those files. I also moved some of the ignores to the top of the script which serves as a global ignores (it will apply to all rule below)

## Screenshots 
Before:
- add console.log to a file in _scripts (ex: getShakaLocales.js)
image
- try to commit:
image
After:
- commit successful

## Desktop
- **OS:** Linux Mint
- **OS Version:** 22


absidue:
Could you please optimise the ignore patterns according to the docs: https://eslint.org/docs/latest/use/configure/ignore

* When ignoring directories it is best to add a `/` to the end of the pattern so that ESLint knows that it is a folder and can skip scanning the directory entirely.
* The `dist` and `node_modules` only exist in the same directory as the `eslint.config.mjs` file, so we don't need recusive patterns for those.

absidue:
Sorry for only commenting this now, but I wanted to test it on my computer before I commented. The eslint docs mention that `**/node_modules/` is on the default ignores list so we don't have to add it ourselves.
> [...]. This pattern is added after the default patterns, which are `["**/node_modules/", ".git/"]`.

absidue (approved):

eslint.config.mjs
```diff
@@ -37,9 +43,8 @@ export default [
       '**/*.{js,vue}',
     ],
     ignores: [
-      '**/node_modules',
-      '**/_scripts',
-      '**/dist',
+      '_scripts/',
+      'eslint.config.mjs'
```
PikachuEXE, line 47: Duplicate entry?

eslint.config.mjs
```diff
@@ -192,4 +193,43 @@ export default [
...
+    },
+
+    rules: {
+      'no-console': 'off',
+      'n/no-path-concat': 'off',
+      'unicorn/better-regex': 'error',
+    }
+  },
+  {
+    files: ['**/*.mjs'],
```
PikachuEXE, line 216: Unless there are other `.mjs` file folders?
```suggestion
    files: ['_scripts/*.mjs'],
```

eslint.config.mjs
```diff
@@ -22,6 +22,12 @@ const compat = new FlatCompat({
 })
 
 export default [
+  {
+    ignores: [
+      'dist/',
+      "eslint.config.mjs"
```
PikachuEXE, line 28: Might as well
```suggestion
      'eslint.config.mjs',
```

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #5841, opened 2024-10-07
Enable Belarusian locale

aliaksandr-trush:
# Enable Belarusian locale

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description
Application translated to Belarusian language, and it could be enabled in the application

## Screenshots 
выява

## Testing 
- Choose Беларуская locale in General settings

## Desktop
- **OS:** Fedora
- **OS Version:** 40
- **FreeTube version:** latest development branch

PR #5847, opened 2024-10-09
Update regions

absidue:
# Update regions

## Pull Request Type

- [x] Other

## Related issue
Follow up to #5841 

## Description
I ran `yarn get-regions` as Belarusian was recently added to the activeLocales list. A few other locales had label changes on YouTube's end which is why those files got updated too.

shaka-player doesn't support Belarusian, which means that only FreeTube's custom elements will be translated when the user has Belarusian selected, however I still think that 41 out of the 46 active FreeTube locales being supported by shaka-player is pretty impressive.

## Testing
1. Change the locale to `Беларуская`
2. Check that the regions list is displayed in Belarusian and not English

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 3b4d88dca60dec146f4558a73eb858dc5330b1b0


ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

MarkusGaugg:
https://github.com/shaka-project/shaka-player/commit/51d00b8f082e571deb76f584f7eb39de2614d491

PR #5853, opened 2024-10-11
Add Playlist Sort By Video Duration (#5627)

efb4f5ff-1298-471a-8973-3d47447115dc:
# Add Playlist Sort By Video Duration

## Pull Request Type

- [x] Feature Implementation

## Related issue

#5627 

## Description

Last minute nice to have add to RC. I will make sure that I will delete the build that this PR creates so it wont confuse users.

### DIFF
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -20,9 +20,10 @@ import {
   getIconForSortPreference,
   setPublishedTimestampsInvidious,
   showToast,
+  deepCopy,
 } from '../../helpers/utils'
 import { invidiousGetPlaylistInfo, youtubeImageUrlToInvidious } from '../../helpers/api/invidious'
-import { getSortedPlaylistItems, SORT_BY_VALUES } from '../../helpers/playlists'
+import { getSortedPlaylistItems, videoDurationPresent, videoDurationWithFallback, SORT_BY_VALUES } from '../../helpers/playlists'
 import packageDetails from '../../../../package.json'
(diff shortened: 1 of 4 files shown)


PikachuEXE (approved):

src/renderer/views/Playlist/Playlist.js
```diff
@@ -416,10 +429,40 @@ export default defineComponent({
...
+          // Mark at least one video have no duration, show notice later
+          // Also assign fallback duration here
+          anyVideoMissingDuration = true
+          video.lengthSeconds = 0
+        }
+      })
+
+      // Show notice if not already shown before returning playlist items
+      if (anyVideoMissingDuration && !this.alreadyShownNotice) {
+        showToast(this.$t('User Playlists.SinglePlaylistView.Toast.This playlist has a video with a duration error'), 5000)
```
kommunarr, line 459: **question:** This seems like more of an understandable implementation detail rather than a reportable error. Thoughts on removing this notice?

static/locales/en-US.yaml
```diff
@@ -241,6 +241,8 @@ User Playlists:
       Playlist {playlistName} is the new quick bookmark playlist.: Playlist {playlistName} is the new quick bookmark playlist.
 
       This playlist does not exist: This playlist does not exist
+
+      This playlist has a video with a duration error: This playlist contains at least one video that doesn't have a duration, it will be sorted as if their duration is zero.
```
kommunarr, line 245: **question:** Just for discussion, thoughts on moving videos with no duration to the bottom of the stack (i.e., `Number.MAX_SAFE_INTEGER` when sort ascending, `0` when sort descending)? Would result in fewer complaints, and would be preferable if for some reason a playlist had multiple of these consistently because they wouldn't have to be scrolled past each time

kommunarr (approved):

PR #5857, opened 2024-10-13
Add keyboard shortcuts to titles

### LINKED ISSUE #1587: Show on hover if icon has shortcut
I often go back to the docs because i need to find out what shortcut i need for a task. This wont be the case if i can look it up fast on just a hover.

See video. This also needs to be the case when the rewrite is done and u can choose ur own shortcuts.

None.

**Video**

https://user-images.githubusercontent.com/73130443/129462389-88c2a52b-eb9c-4517-933f-89f358e496c3.mp4

**Additional context**
None.

kommunarr:
# Add keyboard shortcuts to titles

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #1587

## Description
- Adds keyboard shortcut information to their button labels, e.g., `Mute (m)`
  - This includes the video player buttons, the feed refresh buttons, and the top nav HistoryForward/HistoryBackward/OpenNewWindow buttons, but NOT Electron context menu labels
- **Implementation note:** To do this properly for the built-in Shaka controls that we don't modify in our code, this PR updates the base Shaka label localization values to include the shortcut whenever the Shaka locale is set.
- **Implementation note:** Constants were added for the aforementioned keyboard shortcuts, including updating their existing code instances to use the constants.

## Screenshots 

Screenshot_20241013_181816

## Testing 
- Test that buttons with corresponding keyboard shortcuts have the shortcut referenced in the label, be they in "on" or "off" mode (e.g., `Full screen (f)` / `Exit full screen (f)`). See [the official list of shortcuts](https://docs.freetubeapp.io/usage/keyboard-shortcuts/) for reference.

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
The addition of the new constants should help make enhancements like #251 easier in the future.


src/constants.js
```diff
@@ -113,6 +113,40 @@ const SyncEvents = {
...
+    INCREASE_VIDEO_SPEED: 'p',
+    FULLWINDOW: 's',
+    THEATRE_MODE: 't',
+    TAKE_SCREENSHOT: 'u',
+    VOLUME_UP: 'Up Arrow',
+    VOLUME_DOWN: 'Down Arrow',
+    SMALL_REWIND: 'Left Arrow',
+    SMALL_FAST_FORWARD: 'Right Arrow',
+    LAST_CHAPTER: 'Cmd/Ctrl+Left Arrow',
+    NEXT_CHAPTER: 'Cmd/Ctrl+Right Arrow',
```
PikachuEXE, line 144: Are these used?

src/constants.js
```diff
@@ -113,6 +113,40 @@ const SyncEvents = {
   },
 }
 
+const KeyboardShortcuts = {
+  APP: {
+    HISTORY_BACKWARD: 'Alt+Left Arrow',
+    HISTORY_FORWARD: 'Alt+Right Arrow',
+    NEW_WINDOW: 'Cmd/Ctrl+N'
```
PikachuEXE, line 120: Maybe add a comment about these are used as labels and not in `case` only

src/constants.js
kommunarr: Nope, these are currently just here for future use. I'll mark that

PikachuEXE:
Label for captions
image
image

kommunarr:
Another good catch; I wasn't testing with a video that had a captions option. Just fixed the issue and added logic preventing label modification if the Shaka localization key does not exist. Also note that this specific Captions label will be the only one I've seen with some text to the right of the shortcut. This is fine IMO and more work than desired to individually "fix".

Screenshot_20241014_080659

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Pressing `c` doesnt enable/disable captions

kommunarr:
Added localization of shortcuts & multi-key shortcuts as well as a dangling instance of the wrong constant name for `CAPTIONS`

efb4f5ff-1298-471a-8973-3d47447115dc:
Question: History tab and Settings tab also got shortcuts. Is there a reason those weren't added?

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc I specifically added the shortcuts with corresponding button labels. The closest equivalent to those would be the corresponding side nav options, which I suppose would work. I can add those as well.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (dismissed):
Tested video player, sidebar, top nav button labels, refresh button

Only on MacOS though

kommunarr:
Still on the fence about whether we should use the Mac key symbols for `darwin` (⌘/⇧/⌥/etc.)

efb4f5ff-1298-471a-8973-3d47447115dc:
> Still on the fence about whether we should use the Mac key symbols for `darwin` (⌘/⇧/⌥/etc.)
> 
> Edit: did it. I used the version of the arrow icons that are used in Mac context menus, but it is worth noting that FF for Mac uses the more standard-looking ←→↑↓. I'm also open to using the arrow icons for non-Mac systems; I just went with the text version of those from looking at FF for Linux.

@PikachuEXE what do you think of the change in the latest commit

PikachuEXE:
I think there is no need to use short form on labels
The places I see short forms are app menu bar & keymap menus with limited spaces
Even short forms should be used the arrows chars should be ←→↑↓ at least
I haven't checked how this looks on Windows yet

Zen Browser
image
RubyMine
image
Vivaldi
image

kommunarr:
I can change the arrow symbols. I'm not sure if we have other active contributors who are Mac users to provide additional opinions, but maybe if any users have any thoughts

kommunarr:
Updated the arrow icons for Mac:

Screenshot 2024-10-23 at 8 24 31 AM

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue:
I preferred the stats button before this pull request a lot more as it matches the appearance of the other controls, now it looks out of place in the player.

Before:

before

After:
after

Other controls:
other-controls

kommunarr:
Ah, for the stats button, you're referring to the text content? I noticed that the labels we were setting for it weren't being used, so I thought that was an error. I can change that back if desired.

I can also look into removing the shortcuts from the overflow menu if that is desired as well. My only concern with that is that I'm not sure if we will be adding more shortcuts or moving more controls to a default overflow menu in the future.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Fullwinodw doesnt show shortcut key in overflow menu

kommunarr:
Sorry, I'm still a bit confused on what is being asked to be changed. Just fixing full window to have the shortcut show in the overflow menu? @absidue @efb4f5ff-1298-471a-8973-3d47447115dc

efb4f5ff-1298-471a-8973-3d47447115dc:
> Just fixing full window to have the shortcut show in the overflow menu?

correct

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -39,6 +40,19 @@ const RequestType = shaka.net.NetworkingEngine.RequestType
...
 const TrackLabelFormat = shaka.ui.Overlay.TrackLabelFormat
 
+const shakaControlKeysToShortcuts = {
+  MUTE: KeyboardShortcuts.VIDEO_PLAYER.GENERAL.MUTE,
+  UNMUTE: KeyboardShortcuts.VIDEO_PLAYER.GENERAL.MUTE,
+  PLAY: KeyboardShortcuts.VIDEO_PLAYER.PLAYBACK.PLAY,
+  PAUSE: KeyboardShortcuts.VIDEO_PLAYER.PLAYBACK.PLAY,
+  PICTURE_IN_PICTURE: KeyboardShortcuts.VIDEO_PLAYER.GENERAL.PICTURE_IN_PICTURE,
+  ENTER_PICTURE_IN_PICTURE: KeyboardShortcuts.VIDEO_PLAYER.GENERAL.PICTURE_IN_PICTURE,
+  EXIT_PICTURE_IN_PICTURE: KeyboardShortcuts.VIDEO_PLAYER.GENERAL.PICTURE_IN_PICTURE,
```
PikachuEXE, line 50: These 2 might be used but really difficult to understand
Might need a comment to explain their existence

kommunarr: Added a comment now

PikachuEXE (dismissed):

src/renderer/views/Popular/Popular.js
```diff
@@ -0,0 +1,108 @@
+import { defineComponent } from 'vue'
```
ChunkyProgrammer, line 1: This file no longer exists in dev and was migrated to Popular.vue (composition api)

kommunarr: Good catch, I missed this on resolving the merge conflict. Updated now

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #1220, opened 2021-04-21
Skip to next video


efb4f5ff-1298-471a-8973-3d47447115dc:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Youtube / Freetube often get used as a music player. If a song comes up and u already listened to or u don't like it u can just press the button and skip to the next video. 

**Describe the solution you'd like to see implemented**
Integrate this feature into the player and also into PIP.

This icon can be used for implementing it into the player. 
https://fontawesome.com/icons/step-forward?style=solid

Place the icon in the exact same position as seen in the screen.
Place the icon in PIP on the right of play button.

If #397 is implemented u also need to check if the video ur trying to skip to is already been watched or not. 
**While autoplay is enabled and the video ur skipping to is already been watched then skip the video.**

**Screenshots**
![FT #397](https://user-images.githubusercontent.com/73130443/115615788-eb5b4a00-a2de-11eb-9a63-6000ff8c91a7.PNG)

KSterner2:
Hello there! 
I've been trying to work on this feature but I have some questions. It's still a work in progress but as of right now this is what I have for the function. The toggleNextVideo() function works but I can't grab the next video within the playlist.  

<img width="908" alt="Screenshot 2023-11-14 at 1 17 38 PM" src="https://github.com/FreeTubeApp/FreeTube/assets/95242846/09e7eab7-2c92-4af8-97d6-38c9f31d2e26">

<img width="931" alt="Screenshot 2023-11-14 at 1 17 58 PM" src="https://github.com/FreeTubeApp/FreeTube/assets/95242846/ceee4edd-12a0-468a-b449-c69089704e2a">


Here's what shows in the console for issues as well.  
<img width="423" alt="Screenshot 2023-11-14 at 1 32 41 PM" src="https://github.com/FreeTubeApp/FreeTube/assets/95242846/f454e061-3a66-4224-b01d-ab8e6d9c1bea">

Heres what it looks like in FT:
<img width="636" alt="Screenshot 2023-11-14 at 1 48 46 PM" src="https://github.com/FreeTubeApp/FreeTube/assets/95242846/f5cfb037-be2b-42e8-b03d-ab4f7f3f75d9">

jasonhenriquez:
Hi @KSterner2, I think you should be able to play the next video in a playlist by invoking the `this.$refs.watchVideoPlaylist.playNextVideo()` function.

KSterner2:
[...]

PR #5866, opened 2024-10-14
Add autoplay toggle to the video player

### LINKED ISSUE #1181: Integrate Autoplay mode into mediaplayer
At this moment the Autoplay mode button is not part of the media player. Lets say for example u are in fullscreen but u want autoplay mode enabled then u need to do the following things:

Go out of fullscreen > enable autoplay > go into fullscreen. 

That doesn't make allot of sense to me. This is also the case when u are in Fullwindow mode. I know that this feature doesnt really relate to the mediaplayer unlike the other functions but in my opinion this would enhance user experience.

Integrate the autoplay mode into the media player. See added screenshot for YT implementation.

None

**Screenshots**
Capture

**Additional context**
None

kommunarr:
# Add autoplay toggle to the video player

## Pull Request Type

- [X] Feature Implementation

## Related issue
closes #1181

## Description
Adds autoplay toggle to the media player. 
- Behavior: 
  - When in a playlist, the `Autoplay Playlist Videos` setting is the one exposed/manipulated. Otherwise, the `Autoplay Recommended Videos` setting is the one exposed/manipulated.
  - The autoplay button is hidden when autoplay is not possible.
    - Note: "possible" being defined as the existing checks in `Watch.js` to call `this.$refs.watchVideoPlaylistplayNextVideo`.
  - When the autoplay countdown timer is going down, disabling autoplay will abort the countdown.
- (Adopted from #4342) Updated very confusing autoplay labels' `EN-US` values to more understandable meanings (see Screenshots section for details) 
- For icon considerations, only the deprecated list of Material icons are available, so we don't have access to `autoplay`/`autopause`. I decided on `play_circle`/`play_pause` instead. One other considered alternative was `toggle_on`/`toggle_off`, but it would have been less descriptive.
- **Minor bug fix:** Previously, the Shaka mobile overflow menu intersected with the top bar on many mobile devices (e.g., in the quality settings or playback rate settings), blocking the top entries from being visible. This has now been fixed.

## Screenshots 

Autoplay off:

Autoplay off

Autoplay on:

Autoplay on

Mobile overflow menu:

Mobile overflow menu

Autoplay Settings Label/Ordering Changes:
(Autoplay Videos -> Start Videos Automatically, moved to bottom)
(Play Next Video -> Autoplay Recommended Videos, moved to top)
(Autoplay Playlists -> Autoplay Playlist Videos)

|Before|After|
|-----------|--------|
|Screenshot_20241014_100718|Screenshot_20241014_100659|

## Testing 
- Test that disabling autoplay aborts the autoplay countdown
- Test that the Autoplay toggle does not appear for a non-playlist video when `Hide Recommended Videos` is enabled
- Test that the Autoplay toggle does not appear for a playlist video when it's the last video in the playlist and loop is disabled
  - Check that it does appear when loop is then enabled
- Test that using the autoplay toggle causes the video to autoplay / not autoplay accordingly and updates the corresponding setting (playlist versus non-playlist)

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
I considered firing `handleEnded` on enabling autoplay on an ended video in order to start the autoplay countdown, but I decided against it. Partially because it's a race condition risk to wait sufficiently long for `isAutoplayEnabled` to update, partially because it's not that important.

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -14,6 +14,7 @@ import { LegacyQualitySelection } from './player-components/LegacyQualitySelecti
 import { ScreenshotButton } from './player-components/ScreenshotButton'
 import { StatsButton } from './player-components/StatsButton'
 import { TheatreModeButton } from './player-components/TheatreModeButton'
+import { AutoplayToggle } from './player-components/AutoplayToggle'
 import {
   findMostSimilarAudioBandwidth,
   getSponsorBlockSegments,
@@ -103,6 +104,14 @@ export default defineComponent({
       type: Boolean,
       default: false
     },
+    autoplayPossible: {
+      type: Boolean,
+      default: false
+    },
+    autoplayEnabled: {
+      type: Boolean,
+      default: false
+    },
     vrProjection: {
       type: String,
       default: null
@@ -113,6 +122,7 @@ export default defineComponent({
     'loaded',
     'ended',
     'timeupdate',
+    'toggle-autoplay',
     'toggle-theatre-mode'
   ],
   setup: function (props, { emit, expose }) {
@@ -726,6 +736,7 @@ export default defineComponent({
       if (useOverFlowMenu.value) {
         uiConfig.overflowMenuButtons = [
           'ft_screenshot',
+          'ft_autoplay_toggle',
           'playback_rate',
           'loop',
           'ft_audio_tracks',
@@ -745,6 +756,7 @@ export default defineComponent({
           'recenter_vr',
           'toggle_stereoscopic',
           'ft_screenshot',
+          'ft_autoplay_toggle',
           'playback_rate',
           'loop',
           'ft_audio_tracks',
@@ -771,6 +783,11 @@ export default defineComponent({
         if (index !== -1) {
           elementList.splice(index, 1)
         }
+      }
+
+      if (!props.autoplayPossible) {
+        const index = elementList.indexOf('ft_autoplay_toggle')
+        elementList.splice(index, 1)
       }
 
       if (props.format === 'audio') {
@@ -1627,6 +1644,24 @@ export default defineComponent({
       shakaOverflowMenu.registerElement('ft_audio_tracks', new AudioTrackSelectionFactory())
     }
 
+    function registerAutoplayToggle() {
+      events.addEventListener('toggleAutoplay', () => {
+        emit('toggle-autoplay')
+      })
+
+      /**
+       * @implements {shaka.extern.IUIElement.Factory}
+       */
+      class AutoplayToggleFactory {
+        create(rootElement, controls) {
+          return new AutoplayToggle(props.autoplayEnabled, events, rootElement, controls)
+        }
+      }
+
+      shakaControls.registerElement('ft_autoplay_toggle', new AutoplayToggleFactory())
+      shakaOverflowMenu.registerElement('ft_autoplay_toggle', new AutoplayToggleFactory())
+    }
+
     function registerTheatreModeButton() {
       events.addEventListener('toggleTheatreMode', () => {
         emit('toggle-theatre-mode')
@@ -1755,6 +1790,9 @@ export default defineComponent({
     function cleanUpCustomPlayerControls() {
       shakaControls.registerElement('ft_audio_tracks', null)
       shakaOverflowMenu.registerElement('ft_audio_tracks', null)
+
+      shakaControls.registerElement('ft_autoplay_toggle', null)
+      shakaOverflowMenu.registerElement('ft_autoplay_toggle', null)
 
       shakaControls.registerElement('ft_theatre_mode', null)
       shakaOverflowMenu.registerElement('ft_theatre_mode', null)
@@ -2345,6 +2383,8 @@ export default defineComponent({
         registerScreenshotButton()
       }
       registerAudioTrackSelection()
+      registerAutoplayToggle()
+
       registerTheatreModeButton()
       registerFullWindowButton()
       registerLegacyQualitySelection()
--- a/src/renderer/components/ft-shaka-video-player/player-components/AutoplayToggle.js
+++ b/src/renderer/components/ft-shaka-video-player/player-components/AutoplayToggle.js
@@ -1,0 +1,77 @@
+import shaka from 'shaka-player'
+
+import i18n from '../../../i18n/index'
+
+export class AutoplayToggle extends shaka.ui.Element {
+  /**
+   * @param {EventTarget} events
+   * @param {HTMLElement} parent
+   * @param {shaka.ui.Controls} controls
+   */
+  constructor(autoplayEnabled, events, parent, controls) {
+    super(parent, controls)
+
+    /** @private */
+    this.button_ = document.createElement('button')
+    this.button_.classList.add('autoplay-toggle')
+    this.button_.classList.add('shaka-tooltip')
+
+    /** @private */
+    this.icon_ = document.createElement('i')
+    this.icon_.classList.add('material-icons-round')
+    this.icon_.textContent = 'pause_circle'
+
+    this.button_.appendChild(this.icon_)
+
+    const label = document.createElement('label')
+    label.classList.add('shaka-overflow-button-label')
+    label.classList.add('shaka-overflow-menu-only')
+
+    /** @private */
+    this.nameSpan_ = document.createElement('span')
+    label.appendChild(this.nameSpan_)
+
+    /** @private */
+    this.currentState_ = document.createElement('span')
+    this.currentState_.classList.add('shaka-current-selection-span')
+    label.appendChild(this.currentState_)
+
+    this.button_.appendChild(label)
+
+    this.parent.appendChild(this.button_)
+
+    /** @private */
+    this.autoplayEnabled_ = autoplayEnabled
+
+    // listeners
+
+    this.eventManager.listen(this.button_, 'click', () => {
+      events.dispatchEvent(new CustomEvent('toggleAutoplay', {
+        detail: !this.autoplayEnabled_
+      }))
+    })
+
+    this.eventManager.listen(events, 'toggleAutoplay', (/** @type {CustomEvent} */ event) => {
+      this.autoplayEnabled_ = event.detail
+
+      this.updateLocalisedStrings_()
+    })
+
+    this.eventManager.listen(events, 'localeChanged', () => {
+      this.updateLocalisedStrings_()
+    })
+
+    this.updateLocalisedStrings_()
+  }
+
+  /** @private */
+  updateLocalisedStrings_() {
+    this.nameSpan_.textContent = i18n.t('Video.Autoplay')
+
+    this.icon_.textContent = this.autoplayEnabled_ ? 'play_circle' : 'pause_circle'
+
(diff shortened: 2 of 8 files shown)


src/renderer/components/ft-shaka-video-player/player-components/AutoplayToggle.js
```diff
@@ -0,0 +1,77 @@
+import shaka from 'shaka-player'
+
+import i18n from '../../../i18n/index'
+
+export class AutoplayToggle extends shaka.ui.Element {
+  /**
+   * @param {EventTarget} events
```
absidue, line 7:
```suggestion
   * @param {boolean} autoplayEnabled
   * @param {EventTarget} events
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -1627,6 +1644,24 @@ export default defineComponent({
       shakaOverflowMenu.registerElement('ft_audio_tracks', new AudioTrackSelectionFactory())
     }
 
+    function registerAutoplayToggle() {
```
absidue, line 1647: As it's possible that the user could adjust the auto play settings in another window while this player is open, I would suggest adding a watcher to `props.autoplayEnabled` and notifying the button about the change with an event, so that it doesn't get out of sync.

src/renderer/components/ft-shaka-video-player/player-components/AutoplayToggle.js
```diff
@@ -0,0 +1,77 @@
...
+
+    /** @private */
+    this.autoplayEnabled_ = autoplayEnabled
+
+    // listeners
+
+    this.eventManager.listen(this.button_, 'click', () => {
+      events.dispatchEvent(new CustomEvent('toggleAutoplay', {
+        detail: !this.autoplayEnabled_
+      }))
```
absidue, line 51: As you don't seem to be using the payload in the event listener on the other side, we can send the event without the payload.
```suggestion
      events.dispatchEvent(new CustomEvent('toggleAutoplay'))
```

kommunarr: That's being used on Line 55, no?
https://github.com/FreeTubeApp/FreeTube/blob/4317f3e0dc0c0ade04335461246adad1595f0f6c/src/renderer/components/ft-shaka-video-player/player-components/AutoplayToggle.js#L55

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
kommunarr: Isn't this being done by us passing `autoplayEnabled` as a prop from the `Watch.js` `autoplayEnabled` computed property?

src/renderer/components/ft-shaka-video-player/player-components/AutoplayToggle.js
absidue: Whoops, yes you are correct, I only looked in the player file itself (maybe I shouldn't be reviewing pull requests on my phone :P ).

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
absidue: That just for passing the initial value to the button, it passes a raw boolean. shaka-player and the custom player elements are outside of Vue, so they don't have reactivity, which means we have to take care of reacting to changes ourselves.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -324,6 +334,9 @@ export default defineComponent({
       return store.getters.getSponsorBlockShowSkippedToast
     })
 
+    /** @type {import('vue').ComputedRef<boolean>} */
+    const autoplayEnabled = computed(() => props.autoplayEnabled)
+
```
absidue, line 339:
```suggestion
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -949,6 +969,14 @@ export default defineComponent({
       }
     })
 
+    watch(autoplayEnabled, (newValue, oldValue) => {
```
absidue, line 972:
```suggestion
    watch(() => props.autoplayEnabled, (newValue, oldValue) => {
```

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
I think the pause button can be removed. Seems to be redundant as clicking on the autoplay button does the same thing

image

kommunarr:
> I think the pause button can be removed.

To be honest, I've never even noticed this feature existed until you just pointed it out. Removed

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
I wonder why in additional to adding the button to player (great for mobile especially which I have no problem with)
There is no button on playlist panel (previously pause button)

As a desktop user I expect controls for playlist to be all on that panel

kommunarr:
The "Stop autoplay after this video" button is a very niche feature amongst other apps and seems to have been our temporary replacement for an easily accessible autoplay toggle. The cost of users having to switch to using the autoplay toggle for their use cases is less than that of having to maintain two redundant features.

PikachuEXE:
I think you misunderstand my comment

I am saying I expect to see that new button on playlist panel instead of just inside the player

If "the mental load cost of multiple controls on the same page with the same behavior is high", I expect to see it in playlist panel

Screenshot 2024-10-15 at 08 19 59

kommunarr:
> I am saying I expect to see that new button on playlist panel instead of just inside the player
If "the mental load cost of multiple controls on the same page with the same behavior is high", I expect to see it in playlist panel

Still not sure I understand your point. The autoplay toggle will be visible in the player bar in such a case. There is no point in moving the location of the autoplay toggle or duplicating it for different kinds of videos.

PikachuEXE:
There is a point, playlist related info should be visible on playlist panel including auto play enabled or not
https://github.com/FreeTubeApp/FreeTube/issues/1181 is about full screen/full window mode about not being able to toggle it without exiting the mode and not about default/theater mode
Which is why I support adding that button to player but the status should be on the playlist panel **first**

kommunarr:
Will leave it for other team members than us to decide, as we seem to be in another of our famous impasses. Will leave my main points up:

>  Autoplay, as of this PR, will be a property identified with the video player rather than a playlist control, the same as the audio level or theater mode. Even if we were to make the older pause button act exactly like the autoplay toggle, the mental load cost of multiple controls on the same page with the same behavior is high and should be avoided whenever possible. Given these facts, it is not worthwhile to move the location of the autoplay toggle or duplicate it specifically for the case of playlist videos, and we should instead keep the functionality in one consistent, non-duplicated location for all videos.

efb4f5ff-1298-471a-8973-3d47447115dc:
> There is a point, playlist related info should be visible on playlist panel including auto play enabled
 
This isnt playlist related info. All buttons in there affects the whole playlist (except for the standard skip to next/previous video buttons). This is a button that is related to the individual video and by the introduction of this button we made the association to the playlist.

> https://github.com/FreeTubeApp/FreeTube/issues/1181 is about full screen/full window mode about not being able to toggle it without exiting the mode and not about default/theater mode

That is just an example. The purpose of my issue is to make it the same as YT. 

If we look at Autoplay in YT it is always enabled in playlists. BUT on YT everything that regards Autoplay is built into the player. So when new users come from YT to FT I'd argue that new users are mentally way more familiar with the way its done now.

PikachuEXE:
I assume you are referring to autoplay control which I can find at
https://youtu.be/Wkr2n5Fs_NA?t=30
FreeTube-20241015-181113

I am **not** opposing adding a button in the player

If the objective is the make FT looks like YT (not limited to this PR), then prev/next playlist item buttons should be in player as well
(not suggesting they should be done in the same PR)

If this is the case I can agree with the placement but only if the prev/next buttons are moved too (in the same/another PR)

kommunarr:
I think we're on the same page now. I / whomever can work on that in a separate PR (whoa, has no one made a feature request for that one at all yet? That actually surprises me)

efb4f5ff-1298-471a-8973-3d47447115dc:
Relevant issue https://github.com/FreeTubeApp/FreeTube/issues/1220 :D

PikachuEXE:
Waiting for https://github.com/FreeTubeApp/FreeTube/pull/5857 to be merged first

kommunarr:
One change that needs to be added to this PR, as aptly noticed by a teammate, is changing the force overflow menu breakpoint. I would prefer to wait until [this chore of moving some of our controls into an overflow menu in desktop/tablet view](https://github.com/orgs/FreeTubeApp/projects/4?pane=issue&itemId=84054778) is merged before doing so, as it could change the intent. I can probably take the chore up this week unless someone else has a clearer vision for it that they would like to enact.

efb4f5ff-1298-471a-8973-3d47447115dc:
> One change that needs to be added to this PR, as aptly noticed by a teammate, is changing the force overflow menu breakpoint. I would prefer to wait until [this chore of moving some of our controls into an overflow menu in desktop/tablet view](https://github.com/orgs/FreeTubeApp/projects/4?pane=issue&itemId=84054778) is merged before doing so, as it could change the intent. I can probably take the chore up this week unless someone else has a clearer vision for it that they would like to enact.

If i may ask what is the progress on this? Is this chore holding up this PR?

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc That's correct. There are decision points regarding the context menu that I would prefer to not be directly involved in, and I would also prefer for #6400 to go ahead first in the queue.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
https://github.com/FreeTubeApp/FreeTube/pull/6400 expected to be merged after this?

kommunarr:
This is not ready for review until I either adjust the breakpoints or we start putting more icons into the force overflow menu. In retrospect, I think it would make sense for this to go first instead of #6400 so we can have one PR for implementing the logic for both autoplay settings.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -35,7 +35,7 @@ import { pathExists } from '../../helpers/filesystem'
 // The UTF-8 characters "h", "t", "t", and "p".
 const HTTP_IN_HEX = 0x68747470
 
-const USE_OVERFLOW_MENU_WIDTH_THRESHOLD = 600
+const USE_OVERFLOW_MENU_WIDTH_THRESHOLD = 634
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 38: im a bit confused what changed here. Im seeing the overflow spawn around 663

image

kommunarr: It's not referring to the full viewport width:
https://github.com/FreeTubeApp/FreeTube/blob/0e6ce22ed33e0dbcc6804a5f98fae0cf8d6d9e28/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L982-L984

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
Briefly tested (except mobile & code review
I will let others review those parts

PikachuEXE (approved):
I tested this for 2 weeks no issue

absidue (approved):

PR #5871, opened 2024-10-15
Autoplay time limit

### LINKED ISSUE #4270: [Feature Request]: optional time limited autoplay
### Guidelines

### Problem Description

I enjoy listening to youtube videos as I fall asleep through freetube. However, autoplay keeps my laptop on all night. For most people this probably isn't a big issue, but I live off-grid and rely on battery power so limiting power consumption is very important to me. I would like a way to have limited autoplay so I can fall asleep using freetube and after an hour or two have my laptop be able to properly go to sleep.

### Proposed Solution

In the settings, have an option for 'limited autoplay' and have the end user be able to set the countdown time after which autoplay will disable itself until the user manually clicks on another video.

### Alternatives Considered

Alternatively you can go the youtube/netflix route of having a toggleable 'are you still watching?' popup which I don't think would be as good as implementing a time based autoplay but would be a decent enough stopgap option in the meantime that probably wouldn't be too hard to code in.

### Issue Labels

### Additional Information

Picture of offgrid laptop & battery
Imagepipe_0

kommunarr:
# Autoplay time limit

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #4270

## Description
Implements time-limited autoplay feature which prevents the next video from playing automatically after X hours of 0 click or keydown events. This is set to a default of 3 hours, but it can be configured to be up to 12. This timer resets back to the start for every click or keydown event made at any time in the window. The closely related `Next Video Interval` EN-US label is also renamed to `Autoplay Countdown Timer` to minimize user confusion regarding its meaning.

This feature will be nice for folks like me who occasionally fall asleep to videos and have to reset watch history for a good chunk of them. It should also reduce net power draw and bandwidth expenditure for asleep or otherwise AFK users who neglected to hit the pause button.

To preemptively answer the question of "why not allow infinity," I don't think allowing indefinite automatic streaming regardless of user interaction is ever a desirable state for 99+% of our users. The default of 3 hours without any detected click or keyboard activity is a fine sweet spot for minimizing streaming to sleeping/AFK users, much more generous than [YouTube's](https://support.google.com/youtube/answer/12819304?hl=en) or [Netflix's](https://help.netflix.com/en/node/114059), to name a couple of similar examples. 12 hours of total inactivity before blocking autoplay should be more than enough for users with very particular edge cases.

## Testing 
**_Note: I have this a draft because there are still a couple of autoplay PRs out, so I'd prefer not to make anyone have to re-validate post-merge conflicts._**
- on line 1630 of `Watch.js`:
- https://github.com/FreeTubeApp/FreeTube/blob/27f59e064ba2ea4b2986a3a50da83e96603b73bc/src/renderer/views/Watch/Watch.js#L1630
  - Replace it with this line to make the timeout time much lower:
 
```
this.autoplayInterruptionTimeout = setTimeout(() => { this.blockVideoAutoplay = true }, this.defaultAutoplayInterruptionInterval * 10_000)
```
  - Then enable Recommended Videos Autoplay and/or Playlist Videos Autoplay in settings.
  - Then you can run such validations like:
    - Ensure the timer is reset by any click or key event (e.g., by seeking close to the end of a video by a time slightly greater than the countdown time + autoplay interruption timer combined, then clicking 15 seconds in)
    - Ensure that the next video is blocked from starting automatically if no user interaction is made for an amount of time that is >= the autoplay interruption timer 

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

### DIFF
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -172,6 +172,7 @@ const state = {
   baseTheme: 'system',
   mainColor: 'Red',
   secColor: 'Blue',
+  defaultAutoplayInterruptionInterval: 3,
   defaultCaptionSettings: '{}',
   defaultInterval: 5,
   defaultPlayback: 1,
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -117,6 +117,8 @@ export default defineComponent({
       playNextCountDownIntervalId: null,
       infoAreaSticky: true,
       commentsEnabled: true,
+      blockVideoAutoplay: false,
+      autoplayInterruptionTimeout: null,
 
       onMountedRun: false,
 
@@ -157,6 +159,9 @@ export default defineComponent({
     },
     proxyVideos: function () {
       return this.$store.getters.getProxyVideos
+    },
+    defaultAutoplayInterruptionInterval: function () {
+      return this.$store.getters.getDefaultAutoplayInterruptionInterval
     },
     defaultInterval: function () {
       return this.$store.getters.getDefaultInterval
@@ -320,6 +325,7 @@ export default defineComponent({
       }
 
       window.addEventListener('beforeunload', this.handleWatchProgress)
+      this.resetAutoplayInterruptionTimeout()
     },
 
     changeTimestamp: function (timestamp) {
@@ -1617,6 +1623,12 @@ export default defineComponent({
 
       const playlist = this.selectedUserPlaylist
       this.updatePlaylistLastPlayedAt({ _id: playlist._id })
+    },
+
+    resetAutoplayInterruptionTimeout() {
+      clearTimeout(this.autoplayInterruptionTimeout)
+      this.autoplayInterruptionTimeout = setTimeout(() => { this.blockVideoAutoplay = true }, this.defaultAutoplayInterruptionInterval * 3_600_000)
+      this.blockVideoAutoplay = false
     },
 
     ...mapActions([
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -106,12 +106,17 @@ export default defineComponent({
     vrProjection: {
       type: String,
       default: null
+    },
+    blockVideoAutoplay: {
+      type: Boolean,
+      default: false
     }
   },
   emits: [
     'error',
     'loaded',
     'ended',
+    'reset-autoplay-interruption-timeout',
     'timeupdate',
     'toggle-theatre-mode'
   ],
@@ -154,6 +159,15 @@ export default defineComponent({
      * }[]}
      */
     let sortedCaptions
+
+    const blockAutoplay = props.blockVideoAutoplay
+    if (blockAutoplay) {
+      resetAutoplayInterruptionTimeout()
+    }
+
+    function resetAutoplayInterruptionTimeout() {
+      emit('reset-autoplay-interruption-timeout')
+    }
 
     // we don't need to sort if we only have one caption or don't have any
     if (props.captions.length > 1) {
@@ -1959,6 +1973,8 @@ export default defineComponent({
      * @param {KeyboardEvent} event
      */
     function keyboardShortcutHandler(event) {
+      resetAutoplayInterruptionTimeout()
+
       if (!player || !hasLoaded.value) {
         return
       }
@@ -2365,6 +2381,9 @@ export default defineComponent({
       document.removeEventListener('keydown', keyboardShortcutHandler)
       document.addEventListener('keydown', keyboardShortcutHandler)
 
+      document.removeEventListener('click', resetAutoplayInterruptionTimeout)
+      document.addEventListener('click', resetAutoplayInterruptionTimeout)
+
       player.addEventListener('loading', () => {
         hasLoaded.value = false
       })
@@ -2673,6 +2692,7 @@ export default defineComponent({
       document.body.classList.remove('playerFullWindow')
 
       document.removeEventListener('keydown', keyboardShortcutHandler)
+      document.removeEventListener('click', resetAutoplayInterruptionTimeout)
 
       if (resizeObserver) {
         resizeObserver.disconnect()
@@ -2768,6 +2788,7 @@ export default defineComponent({
       stats,
 
       autoplayVideos,
+      blockAutoplay,
       sponsorBlockShowSkippedToast,
 
       skippedSponsorBlockSegments,
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.vue
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.vue
@@ -14,7 +14,7 @@
       preload="auto"
       crossorigin="anonymous"
       playsinline
-      :autoplay="autoplayVideos ? true : null"
+      :autoplay="autoplayVideos && !blockAutoplay ? true : null"
       :poster="thumbnail"
       @play="handlePlay"
       @pause="handlePause"
--- a/src/renderer/components/player-settings/player-settings.js
+++ b/src/renderer/components/player-settings/player-settings.js
@@ -95,6 +95,10 @@ export default defineComponent({
       return this.backendPreference !== 'invidious' && !this.backendFallback
     },
 
+    defaultAutoplayInterruptionInterval: function () {
+      return parseInt(this.$store.getters.getDefaultAutoplayInterruptionInterval)
+    },
+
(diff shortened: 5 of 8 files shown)


absidue:
I know this pull request is still in draft state, but I think I have a different understanding of the feature request, so I wanted to mention it now, so we can already start a discussion around it.

In FreeTube we have 3 different types of auto play:
- start playing the video automatically
- go to next playlist video after the current one ends
- go to next recommended video after the current one ends

From reading the feature request and pull request description I would have expected this pull request to affect the later two settings, so that when the timer ends it finishes playing the current video and then just stays there. However the code in this pull request seems to target the first one, so when the timer ends it will finish playing the current video, go to the next one and then stop there. That means when the timer expires, it still does a bunch of work to get the new data and process it, updates the UI, downloads new images. Loading the next video also means that it is likely to get marked as watched.

kommunarr:
That is indeed a fair interpretation and set of implications. My main consideration with modifying the autoplay setting (qua the second / third definitions) was that I didn't want to disable the persistent setting on this condition, and I was thinking that such a feature would necessitate us implementing the ability for non-permanent settings modifications, although I realize now that we could instead achieve this logic by a `Watch.js` `handleVideoEnded` check (similar to the Playlist Pause button). I'm leaning toward that new solution now unless there are any other alternatives I have not considered.

absidue:
Handling it in Watch.js sounds like the best solution, you don't even need to change the logic in `handleVideoEnded` (only thing needed there would be to cancel the timeout), as you can change the logic in the `autoplayPlaylists` and `playNextVideo` computed props to `return value of setting and not auto play timer expired` (currently they just return the value of the setting).

efb4f5ff-1298-471a-8973-3d47447115dc:
One issue i see with this is the following case:

Slider is set to 3hrs

- User is actively watching a stream that exceeds the 3hr mark without interaction
- User is listening to a podcast that is longer than 3hrs without interaction

Yeah i know you can move the slider but that isnt the issue here. Some people might not want this feature be on by default with no way for it to turn off.

kommunarr:
If the problem is the "people will complain about the default before checking the settings" issue like we had with the play button, I can change the default to 4 or even 5 hours. There's no one-size-fits-all default necessarily, because you run into a balancing issue of the % of people who are bothered by their videos going on too long while they're asleep / AFK versus the % of people who want to watch back-to-back 3+ hour videos without any pausing, adjusting fullscreen, changing the audio level, etc. I still think 3 is the closest magic number from my guesstimate, but open to hearing your opinion. 

To your latter concern, if we get a singular user for whom the max of 12 hours without user interaction is too much, we can talk once they make an issue for it; but our concerns should be proportional to # of affected users and degree of affectedness, and both of those metrics are quite small in this case.

src/renderer/store/modules/settings.js
```diff
@@ -172,6 +172,7 @@ const state = {
   baseTheme: 'system',
   mainColor: 'Red',
   secColor: 'Blue',
+  defaultAutoplayInterruptionInterval: 3,
```
PikachuEXE, line 175: Other interval settings are seconds by default
Let's name it so it's clear that the value is hours not seconds
```suggestion
  defaultAutoplayInterruptionIntervalHours: 3,
```

src/renderer/views/Watch/Watch.js
```diff
@@ -321,7 +327,11 @@ export default defineComponent({
         this.getVideoInformationLocal()
       }
 
+      document.removeEventListener('click', this.resetAutoplayInterruptionTimeout)
+      document.addEventListener('click', this.resetAutoplayInterruptionTimeout)
```
PikachuEXE, line 331: I am wondering what happen to people interacting with keyboards only...

kommunarr: Great callout, addressed now

src/renderer/store/modules/settings.js
kommunarr: Done!

src/renderer/views/Watch/Watch.js
```diff
@@ -1233,6 +1246,12 @@ export default defineComponent({
         return
       }
 
+      if (this.blockVideoAutoplay) {
+        showToast(this.$t('Canceled next video autoplay due to inactivity'))
```
PikachuEXE, line 1250: This should stay visible for longer (1 hour?)
New users (new to this setting) will take time to realize the playlist is stopped and switch back and wondering why it's stopped
Also the text should mention the setting name if possible, but at least should include the value (e.g. 3h)

kommunarr: How about this?

Screenshot_20241204_010228

PikachuEXE: Looks better

src/renderer/views/Watch/Watch.js
```diff
@@ -1233,6 +1246,18 @@ export default defineComponent({
...
       }
 
+      if (this.blockVideoAutoplay) {
+        showToast(this.$t('Autoplay Interruption Timer',
+          this.defaultAutoplayInterruptionIntervalHours,
+          {
+            autoplayInterruptionIntervalHours: this.defaultAutoplayInterruptionIntervalHours
+          }),
+        3_600_000
+        )
```
PikachuEXE, line 1256: I am surprised that linter is ok with this o_0

PikachuEXE (approved):

src/renderer/views/Watch/Watch.js
kommunarr: It not only tolerates it: it enforces it to be like that. ¯\\_(ツ)_/¯

efb4f5ff-1298-471a-8973-3d47447115dc:
> To your latter concern, if we get a singular user for whom the max of 12 hours without user interaction is too small, we can talk once they make an issue for it; but our concerns should be proportional to # of affected users and degree of affectedness, and both of those metrics are quite small in this case.

Agreed.

Also

1. So i set the value to 1hr
2. Went to watch a long video
3. No activity for 1hr and 6mins
4. Didnt see a toast message and video didnt pause

Did i do something wrong?

kommunarr:
Interesting, did you increase volume, touch any unrelated keys (e.g., F10), something like that at any point? I'd recommend replacing line 1630 with a shorter time period as per the PR description to make your testing experience a bit easier

efb4f5ff-1298-471a-8973-3d47447115dc:
No i really didnt touch my machine for more than an hour

kommunarr:
Cannot replicate for recommended videos autoplay, user playlist autoplay, or remote playlist autoplay. I added logging statements on lines 1670 and 1671 of Watch.js as well, and did not see any unexpected behavior or condition triggering. If you could, replace this function in Watch.js as such to include these logging statements and see if you can replicate. This is using a 1s = 1hr time adjustment as well, but I tested at a smaller conversion rate and still could not replicate:

```
    resetAutoplayInterruptionTimeout() {
      console.log('clearing autoplay timeout')
      clearTimeout(this.autoplayInterruptionTimeout)
      this.autoplayInterruptionTimeout = setTimeout(() => { this.blockVideoAutoplay = true; console.log('blocking video autoplay') }, this.defaultAutoplayInterruptionIntervalHours * 1_000)
      this.blockVideoAutoplay = false
    },
```

Screenshot_20241204_201040

kommunarr:
Silly question, but are you testing with autoplay enabled?

efb4f5ff-1298-471a-8973-3d47447115dc:
> Silly question, but are you testing with autoplay enabled? 

Yup autoplay is enabled.

> The toast message is not intended to appear until the video ends.

ooooh my bad i interpreted this wrong. I thought the video would pause or something and show the toast message. I didnt know the video has to end first

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Sorry about that, LGTM

PikachuEXE:
image

src/renderer/views/Watch/Watch.js
absidue: The reason that is happening is because `showToast(` and `this.$t(` are both on the same line, so the linter sees `this.$t` as having one level of indentation and `showToast` as having none. That is why it forcing `3_600_000` and `)` to have no indentation.

Moving `this.$t` to its own line fixes the issue, as then the linter allows them both to have indentation:
screenshot-of-fixed-formatting

It would also improve the general readability of that code snippet if the placeholder name in the translation string weren't so long e.g. `{hours}` would already suffice, as there is only one placeholder in that string and there is enough context in the string and the code where it is used, to understand what the purpose of the placeholder is.

However as both things are stylistic issues and don't influence the functionality of the code, we can leave the cleanup for some future pull request.

absidue (approved):

absidue:
Thanks GitHub for duplicating the comment lol.

PR #5890, opened 2024-10-17
Switch from deprecated standard to neostandard in ESLint config

absidue:
# Switch from deprecated standard to neostandard in ESLint config

## Pull Request Type

- [x] Other - Development tooling

## Description
This pull request switches from the deprecated `standard` ESLint config (basically abandoned and requires `@eslint/compat` to get it running on ESLint 9) to `neostandard`. I also decided to use the code style rules that noestandard suggests with a few minor tweaks, as that lets us retain our current code style, instead of trying to fix our currently broken prettier setup (easiest way to no it is broken was the fact that when I fixed it, it wanted to reformat almost every single file in the repo).

The dependency changes are the following:
- Removed `eslint-config-standard`
- Added `neostandard`
- Removed `@eslint/compat` and `@eslint/eslintrc` as they were only needed for `standard`
- Removed `eslint-plugin-n` and `eslint-plugin-promise` from the direct dependencies, as `neostandard` lists both of them as dependencies
- Removed `prettier`, `eslint-config-prettier` and `eslint-plugin-prettier`
- Also removed `@babel/eslint-parser` as I noticed we weren't using it anymore.

Unfortunately despite us now having less dependencies listed in the package.json file, we still have more in total as neostandard has a big dependency tree.

## Testing
`yarn eslint`

You can also try things like incorrect indentation or trailing whitespace and check that eslint detects it.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 976727aadd04c2aed1a840ba73da4db199f7293f


.github/workflows/no-response.yml
PikachuEXE, line 1: Why the diff for this file is shown as all lines replaced...

.github/workflows/remove-outdated-labels.yml
PikachuEXE, line 1: Ditto

.github/workflows/no-response.yml
absidue: Line endings changed from `\r\n` (CRLF) to `\n` (LF)

PikachuEXE (approved):

kommunarr (approved):

ChunkyProgrammer (approved):

Issue #5265, opened 2024-06-13
[Feature Request]: Only Show Latest x Videos for Each Channel


c-ciobanu:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

I would like to have my `Subscriptions` panel as some sort of overview on all my channels where I can see the latest videos from all the channels I'm subscribed too.
If you are subscribed to many channels you probably have some that upload much more than others cluttering your `Subscriptions` panel just with their videos.
There is the option `Only Show Latest Video for Each Channel` that will kinda do what I'm asking for but only showing 1 video per channel is not enough for me since I can't really understand if that video is the latest or not. In my case I would like to only be able to see the latest 3 or 5 videos from each channel so I can have a nice understanding of what is going on.

### Proposed Solution

Update the existing option `Only Show Latest Video for Each Channel` so that is possible to manually define the maximum amount amount of videos to show per channel, or create a completely new option that allows for that and call it `Only Show Latest x Videos for Each Channel`

### Alternatives Considered

None.

### Issue Labels

improvement to existing feature, new feature, new optional setting

c-ciobanu:
Hi, I started working on this and I was wandering if it's a good idea to get rid of the existing `Only Show Latest Video for Each Channel` settings in favour of this one since it basically does the same thing.

kommunarr:
Yes, the new logic should also be being invoked in the same file & location as the original setting (`subscriptions.js`).

Thunderschnozzle:
Not sure if it's too late to add an additional request to the original one, but I'll do it anyways. Would it be possible to replace "Latest x Videos for Each Channel" with "Latest x Videos for Each Channel from the Last y Period"?
The user would be given a drop-down menu for x that will say: 1,2,3,4,5..., and a drop-down menu for y that will say:1 day, 2 days, 3 days, 1 week, 2 weeks, 1 month...
What do you guys think?

PR #5901, opened 2024-10-20
Update 'Only Show Latest Video for Each Channel' setting to handle a custom number of videos per channel

c-ciobanu:
# Update 'Only Show Latest Video for Each Channel' setting to handle a custom number of channels

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/5265

## Description

Updated the `Only Show Latest Video for Each Channel` setting to now handle a custom number of videos per channel to be shown.
Added a new variable to keep track of the number so the user can change it if the option is enabled.
I didn't change the existing setting so this way this feature is retro compatible.

## Screenshots 

image
image

## Testing 

Enabled the feature and changed the number different times while checking that the Videos, Shorts, and Live tabs showed the correct amount of videos for each channel.

## Desktop

- **OS:** Linux Mint
- **OS Version:** 21
- **FreeTube version:** 0.21.3

## Additional context

I didn't remove the old localisation string, is it removed automatically or do I need to do it.

### DIFF
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -18,6 +18,12 @@
           :compact="true"
           @change="updateUseRssFeeds"
         />
+        <ft-toggle-switch
+          :label="$t('Settings.Subscription Settings.Confirm Before Unsubscribing')"
+          :default-value="unsubscriptionPopupStatus"
+          :compact="true"
+          @change="updateUnsubscriptionPopupStatus"
+        />
       </div>
       <div class="switchColumn">
         <ft-toggle-switch
@@ -39,14 +45,6 @@
           input-type="number"
           :value="onlyShowLatestFromChannelNumber"
           @input="updateOnlyShowLatestFromChannelNumber"
-        />
-      </div>
-      <div class="switchColumn">
-        <ft-toggle-switch
-          :label="$t('Settings.Subscription Settings.Confirm Before Unsubscribing')"
-          :default-value="unsubscriptionPopupStatus"
-          :compact="true"
-          @change="updateUnsubscriptionPopupStatus"
         />
       </div>
     </div>
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -27,10 +27,18 @@
           @change="updateHideWatchedSubs"
         />
         <ft-toggle-switch
-          :label="$t('Settings.Subscription Settings.Only Show Latest Video for Each Channel')"
+          :label="$t('Settings.Subscription Settings.Limit the number of videos displayed for each channel')"
           :default-value="onlyShowLatestFromChannel"
           :compact="true"
           @change="updateOnlyShowLatestFromChannel"
+        />
+        <ft-input
+          v-if="onlyShowLatestFromChannel"
+          :placeholder="'1'"
+          :show-action-button="false"
+          input-type="number"
+          :value="onlyShowLatestFromChannelNumber"
+          @input="updateOnlyShowLatestFromChannelNumber"
         />
       </div>
       <div class="switchColumn">
--- a/src/renderer/components/subscription-settings/subscription-settings.js
+++ b/src/renderer/components/subscription-settings/subscription-settings.js
@@ -2,12 +2,14 @@ import { defineComponent } from 'vue'
 import { mapActions } from 'vuex'
 import FtSettingsSection from '../ft-settings-section/ft-settings-section.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
+import FtInput from '../ft-input/ft-input.vue'
 
 export default defineComponent({
   name: 'SubscriptionSettings',
   components: {
     'ft-settings-section': FtSettingsSection,
-    'ft-toggle-switch': FtToggleSwitch
+    'ft-toggle-switch': FtToggleSwitch,
+    'ft-input': FtInput,
   },
   computed: {
     hideWatchedSubs: function () {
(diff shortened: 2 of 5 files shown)


kommunarr (commented):
I haven't tested this yet, but good work! One potential suggestion: thoughts on having the label change to `Limit the number of videos displayed for each channel to` when the toggle is enabled and making the input an inline 2-digit-wide one? That way it could read out like a sentence.

efb4f5ff-1298-471a-8973-3d47447115dc:
> I didn't remove the old localisation string, is it removed automatically or do I need to do it.

Its not automatically, you need to remove it.

c-ciobanu:
> I haven't tested this yet, but good work! One potential suggestion: thoughts on having the label change to `Limit the number of videos displayed for each channel to` when the toggle is enabled and making the input an inline 2-digit-wide one? That way it could read out like a sentence.

I personally prefer something like this:
image
to this:
image

Let me know what you think.

c-ciobanu:
> > I didn't remove the old localisation string, is it removed automatically or do I need to do it.
> 
> Its not automatically, you need to remove it.

I imagine I have to remove it from all the files not just `static/locales/en-US.yaml`, right?

kommunarr:
I personally prefer the latter, although either way is fine from a general styling standpoint as long as other contributors feel the same. My main concern is that if we were to add a new toggle to that column, it would be visually unclear if it correlated to the top toggle or the bottom one. That concern could possibly be assuaged by reducing the margin between the two controls, but I'm not sure, as it's already so small between the toggles already.

On that note, you'll also need an `aria-labelledby` [(example)](https://www.w3.org/TR/WCAG20-TECHS/ARIA9.html#ARIA9-ex2) on the base `input` pointing to the toggle's title to ensure that this is accessible for users with screen readers.

src/renderer/components/subscription-settings/subscription-settings.vue
```diff
@@ -18,31 +18,42 @@
...
-          :default-value="unsubscriptionPopupStatus"
+          :label="$t('Settings.Subscription Settings.Hide Videos on Watch')"
+          :default-value="hideWatchedSubs"
           :compact="true"
-          @change="updateUnsubscriptionPopupStatus"
+          @change="updateHideWatchedSubs"
         />
+        <div class="onlyShowLatestFromChannel">
+          <ft-toggle-switch
+            :id="'onlyShowLatestFromChannel'"
```
absidue, line 37:
```suggestion
            id="onlyShowLatestFromChannel"
```

src/renderer/components/subscription-settings/subscription-settings.vue
```diff
@@ -18,31 +18,42 @@
...
+            :label="onlyShowLatestFromChannel ? $t('Settings.Subscription Settings.Limit the number of videos displayed for each channel to') : $t('Settings.Subscription Settings.Limit the number of videos displayed for each channel')"
+            :default-value="onlyShowLatestFromChannel"
+            :compact="true"
+            @change="updateOnlyShowLatestFromChannel"
+          />
+          <ft-input
+            v-if="onlyShowLatestFromChannel"
+            :placeholder="'1'"
```
absidue, line 45:
```suggestion
            placeholder="1"
```

src/renderer/components/subscription-settings/subscription-settings.vue
```diff
@@ -18,31 +18,42 @@
...
+            :show-action-button="false"
+            input-type="number"
+            :value="onlyShowLatestFromChannelNumber"
+            :aria-labelledby="'onlyShowLatestFromChannel'"
```
absidue, line 49:
```suggestion
            aria-labelledby="onlyShowLatestFromChannel"
```

src/renderer/components/subscription-settings/subscription-settings.scss
absidue, line 1: Please don't use nested CSS unless there is absolutely no other way, as it breaks Vue's scoped CSS functionality, in this case a class would probably be a lot more suitable than doing some nth-child stuff. That would also allow you to use a plain CSS file.

efb4f5ff-1298-471a-8973-3d47447115dc:
> > > I didn't remove the old localisation string, is it removed automatically or do I need to do it.
> > 
> > 
> > Its not automatically, you need to remove it.
> 
> I imagine I have to remove it from all the files not just `static/locales/en-US.yaml`, right?

Thats correct

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Lets only allow positive numbers >= 1

image

c-ciobanu:
> * Lets only allow positive numbers >= 1
> 
> image
> 
> * Also we should set the limit to 30 as we dont display more than that
> * I dont like how all the settings move when you enable this
> 
>  VirtualBoxVM_08l4jeu4gq.mp4

Seeing how the input and the settings work right now I don't think it's possible to have a limit enforced nicely right now since there is no validation at all.

For the alignment something like this would be better:
https://github.com/user-attachments/assets/6ad663e1-1f45-4622-985f-e9865cc22957
Layout not moving and it looks the same on lower resolutions.

Maybe instead of having an input a slider would be better?

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm i think thats a good suggestion to go the slider route. Also makes it more uniform with the other settings

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (commented):
Works well but I will let you guys approve

c-ciobanu:
Hey, what's the status on this, what's blocking it from being merged?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):
Looks great! Apologies for the delay; it can be tough for us to keep track of things as more tasks flow in.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

kommunarr (approved):

Issue #1214, opened 2021-04-20
Integrate countdown inside the player instead of static message ouside of the player when autoplay is enabled


efb4f5ff-1298-471a-8973-3d47447115dc:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
I noticed that there is a static message ("Playing next video in 5 seconds. Click to cancel.") when waiting for the next video to appear. I think its better to let the user know how much time they have left instead of just guessing when the 5sec will be over. 

This will enhance user experience in 2 ways:
1. Users will exactly know how much time is left when the new video starts. 
2. Users will exactly know how much time is left to prevent the next video from playing. 

**Describe the solution you'd like to see implemented**
Integrate the countdown into the player.

This is because there are 3 flaws in the current implementation:
1. When fullscreen is enabled u cant see the message appear.
2. When Fullwindow is enabled u cant see the message appear.
3. When PIP is enabled u cant see the message appear.

PikachuEXE:
I tried to take a stab but the result is not good...

https://user-images.githubusercontent.com/1018543/115650319-93aef400-a35b-11eb-9892-5cebb4a7e79b.mp4

Code change:
https://github.com/FreeTubeApp/FreeTube/compare/development...PikachuEXE:feature/playlist-autoplay-countdown

GilgusMaximus:
The idea is good, I am not so sure however that it will work well because these toasts, as you might have noticed get pushed to the top. Maybe you could make the toast less than 1 second, so that it disappears before the next one occurs. Then the previous one does not get pushed upwards

efb4f5ff-1298-471a-8973-3d47447115dc:
I think i worded wrong what i mean with integrate countdown into the player, because this implementation still occurs on the outside of the player. To clear things up i screen recored YT implementation.

https://user-images.githubusercontent.com/73130443/115767310-83693a00-a398-11eb-8ba0-0dd6fe5d361b.mp4

There has to be a countdown and cancel button. The other things like the preview of the next video and the play now button are optional.

GilgusMaximus:
Yeah, I mean @PikachuEXE's approach is an easy approach which solves the problem with existing functionality. For the moment that would be fine. But in the long run, something like on YouTube will obviously be more pleasant to look at

efb4f5ff-1298-471a-8973-3d47447115dc:
[...]

PR #5903, opened 2024-10-21
Stay in fullscreen/fullwindow/PiP + default viewing mode setting

### LINKED ISSUE #5427: [Feature Request]: Enable Preferred Viewing Mode by Default
### Guidelines

### Problem Description

Only option users have is to `Enable Theatre Mode by Default` but not for the other viewing modes.

### Proposed Solution

Create a dropdown in the player settings called `Default Viewing Mode` with the following values:

- Default
- Theatre
- Picture-In-Picture
- Full Window
- Fullscreen
- External player

### Alternatives Considered

None.

### Issue Labels

ease of use improvement

### Additional Information

_No response_

kommunarr:
# Stay in fullscreen/fullwindow/PiP + default viewing mode setting

## Pull Request Type

- [X] Feature Implementation

## Related issue
closes #5427

## Description
- Stay in fullscreen, fullwindow, or PiP for the next video in an autoplay session if one or more of these viewing modes is active upon the video ending
- Add `Default Viewing Mode` dropdown setting for whether to enable fullscreen / fullwindow / PiP / external player by default
  - **_Area for future improvement._**: the `External Player` option corresponds specifically to when an `ft-list-video` is interacted with, not any YT links found in a description or entered into the search bar. This was not implemented because we would have to implement logic for creating the external player video config from a URL, which is somewhat marginal and not sufficiently germane to this PR.
  - Note: the `External Player` option does not appear as an option when no external player is set
    - If the setting was changed to `External Player` prior to this, it is de facto set to `Default` until a new external player is set
- Minor settings changes:
  - `en-us`: 
    - `Enable Theatre Mode by Default` -> `Enable Theatre Mode by Default`
    - `Turn on Captions by Default` -> `Enable Captions by Default` (for visual consistency / grouping with its compatriot)
  - `View Playback Rate Interval` input moved from the same `ft-flex-box` with the sliders -> colocated with the other `ft-input`s (it appeared odd visually and was not a sufficiently logical grouping)
  - Video Player Settings flex box of `ft-input`s moved to above the flex box of `ft-slider`s (better logical/related grouping)

Note: I'm aware of two main cases where there are arguable discrepancies between the expected and actual behavior of "staying in {viewingMode}", but they are both what I would consider to be minor and not worth addressing in this PR:
1. When using the `watch-video-playlist` Play Prev/Next button, PiP specifically does not persist to the newly loaded video. This is personally acceptable to me for the reasons that a) it is a very minor case and b) we will be moving this functionality into the player in the near future, where this will be fixed by having playlist autoplay being one consistent feature (i.e., we will emit `ended` from `ft-shaka-video-player` to hit all of the same hooks regardless of how playlist play-next-video was triggered).
2. When using a mouse button or keyboard shortcut to go forward or back in history to a `/watch` page entry while you are currently watching a video in fullscreen/fullwindow/PiP, the fullscreen/fullwindow/PiP will not be maintained in the freshly loaded history route. This is due to us specifically clearing the `startIn{modeName}` variables on video load and repopulating them on autoplay. This is again a fairly marginal case, and it is arguable whether staying in FS/FW/PiP should even be the expected behavior.

## Screenshots 

| Before | After |
|---|---|
|Before Settings|Settings After|

## Testing 
1. With A) playlist / B) recommended videos autoplay enabled, set the Default Viewing Mode to a) fullscreen / b) fullwindow / c) PiP. I also recommend lowering the autoplay countdown to 0s or 1s.
   1. Play a video and ensure that the viewing mode is properly set to enabled by default.
   2. While still in the viewing mode, skip to near the end of the video and let it autoplay. Upon the loading of the next video, the previously active viewing mode should still be active.
   3. Leave the viewing mode and X) let the video autoplay or Y) click on another video. The `Default` viewing mode should persist, **_NOT_** FS/FW/PiP.
   4. (Optional) Test 1iii with autoplay disabled.
3. Ensure that with Default Viewing Mode of `External Player`, clicking on any `ft-list-video` will attempt to play it in the currently active external player. Any `ft-list-video` should also not have a redundant External Player icon.
  1. Then set the chosen `External Player` to `None`. The viewing mode should now function as `default`, and clicking on any `ft-list-video` will play it in the Watch page normally.
4. (REG, optional) Test that theater mode still works as intended.

## Desktop

- **OS:** MacOS
- **OS Version:** Ventura

## Additional context
I combined "Stay in fullscreen/fullwindow/PiP" with "default viewing mode setting" under one PR because the latter one is actually very little code, and testing them both together is more time efficient for you all. I can separate this out if desired, although I'd imagine it would not be.


efb4f5ff-1298-471a-8973-3d47447115dc:
- I still would like to see Theater Mode in there as it is a viewing mode
- Scroll on Fullwindow shouldnt happen

https://github.com/user-attachments/assets/023800cb-9ab3-42c9-a107-5c6c38906b3f

kommunarr:
> I still would like to see Theater Mode in there as it is a viewing mode

See my thoughts on that [here](https://github.com/FreeTubeApp/FreeTube/issues/5427#issuecomment-2425250771):
> I still think we should keep the Enable Theater Mode By Default setting, as it is not mutually exclusive with the other viewing modes.

Good catch on that bug! Will look into that.

> Q: Not sure but wouldnt this PR make https://github.com/FreeTubeApp/FreeTube/issues/1214 more urgent to implement? If so should we do it here or in a followup?

Not directly, since this PR only affects the behavior immediately following the countdown, but it would be good to have that UX issue resolved

absidue:
@kommunarr That bug is caused by not entering full window correctly, you just set the property to true but don't scroll to the top or add the CSS class that disables scrolling. Please make sure the features in this pull request behave exactly the same as the buttons and keyboard shortcuts do (that's what I meant by please don't reinvent the wheel in the private messages): https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L1649-L1661

efb4f5ff-1298-471a-8973-3d47447115dc:
> I still would like to see Theater Mode in there as it is a viewing mode
> > See my thoughts on that https://github.com/FreeTubeApp/FreeTube/issues/5427#issuecomment-2425250771:

Hmm presenting my arguments for why i think it should be included in the dropdown.

1. **Consistency in Settings**: Including Theater mode in the dropdown ensures consistency in how the user selects their preferred viewing mode. This uniformity makes the settings more intuitive and easier to use, as all modes are presented in the same way.

2. **User Preference**: While Theater mode might not be mutually exclusive with Fullscreen or Picture-in-Picture, some users may prefer to set it explicitly as their default mode. Including it in the dropdown respects user preferences and provides a straightforward way for them to select their desired viewing experience.

3. **Clarity in Configuration**: Having Theater mode in a separate toggle and the other modes in a dropdown can be confusing for users. Merging all viewing mode options into a single dropdown simplifies the settings interface and makes it clear that the user can choose between all available modes, including Theater mode.


To me as an user it isnt clear to me at all what will be my default viewing mode, is it fullscreen or theater?

image

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -10,19 +10,22 @@
...
-    >
-      <router-link
+    > 
+      <component
+        :is="watchPageLinkType"
         class="thumbnailLink"
         tabindex="-1"
         :to="watchPageLinkTo"
+        href="javascript:void(0)"
+        @click="handleWatchPageLinkClick"
```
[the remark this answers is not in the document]
kommunarr: Got it, is there a way I can do this without duplicating the inner content for the v-else case?

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc I don't have much to say other than that your points are valid and it's a tradeoff either way. [937935f](https://github.com/FreeTubeApp/FreeTube/pull/5903/commits/937935f06a79691402c8ddd7bc171562b165c948) is the commit to drop if we want to put Theater Mode in there. I'm still leaning towards keeping them separate, but I am open to hearing more perspectives as well.

PikachuEXE:
Just about `Enable Theater Mode by Default`
I think the issue is that in this PR the new option makes this option / label sound weird
It's actually default when `Default Viewing Mode` is `Default`, it's "secondary viewing mode" when others are selected
No idea how to represent this well

"Default Viewing Mode" now has
- "default mode" (not theater mode)
- theater mode
- Full screen + "default mode" fallback
- Full screen + theater mode fallback
- Full window + "default mode" fallback
- Full window + theater mode fallback
- PiP + "default mode" fallback
- PiP + theater mode fallback

kommunarr:
I was thinking about the same thing you were Pika, but I ultimately think we're overthinking it because we are assuming that people will inherently be asking themselves "but how does theater mode play into it" and get confused, when I don't think there is any inherent reason to do that or imply it if our settings structure doesn't. If we have it as separate settings, that's an intuitive indication that viewing mode is orthogonal to whether theater mode is enabled, which is in fact the case in reality as well. For Fullwindow or Fullscreen, you have to actually exit that viewing mode to interact with things like comments and seeing the playlist / recommended videos, and you'll have to see a player that is in theater mode or isn't. If you're in PiP, the size of the grayed out video player is still in play if you want to toggle it off, and that video player is in theater mode or isn't. I think we might be assuming that it is going to be a comprehension problem when we theorize it out in terms of number of total distinct states, but I believe that the UX of them being separate settings altogether makes the relationship much less complex and far more intuitive than we're making it out to be.

PikachuEXE:
I simply find having 2 "default" a bit confusing
A short tooltip on theater mode option like "This also decides the fallback mode for some viewing modes" or a tooltip on the new option (no idea about the text) would be clearer

Also can you add some new lines for long paragraphs :S

kommunarr:
I'm open to any potential smaller suggestions on what can be done in terms of updating the labels or icons. The big one may be that we have the `faDisplay` icon for the `Default Viewing Mode` dropdown, which is pretty similar to the `tv` icon for Theater Mode and creating an unnecessary connection. I'm trying to find something in the `fontawesome` library that's more like YT's `miniplayer` icon. I will say that even making it seem like they're connected (which again, practically speaking, they are not, other than in the narrow sense that "You can't see how big the player is while you are currently in FS/FW") through a clarifying tooltip could more confusing than just not mentioning it at all.

src/renderer/components/ft-list-video/ft-list-video.vue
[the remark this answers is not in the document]
kommunarr: Update: I cleaned this up in [bb3be09](https://github.com/FreeTubeApp/FreeTube/pull/5903/commits/bb3be09122deecd8141651288fa49ed959cc0618)

kommunarr:
One case I haven't implemented: staying in PiP when you click another video without letting it autoplay. It's a bit specific, but I am thinking about it.

absidue:
As PiP and full screen by default are only supported in Electron, please make sure that those options are hidden in the UI outside of Electron (the screenshot settings are like that so you can take inspiration for how it is handled there) and that if the user does have those values set (e.g. copying the settings.db from FreeTube desktop to FreeTubeAndroid) that it acts as if the settings are not set, so that it doesn't break anything.

kommunarr:
Updated logic to have the reincarnate-with player state variables saved on player destroy instead of video ended, thus allowing staying in PiP when clicking on an arbitrary video, staying in fullscreen/fullwindow when using Ctrl+Left Arrow / Ctrl+Right Arrow, and staying in PiP when using the watch-video-playlist Play Prev / Play Next buttons.

src/main/index.js
```diff
@@ -927,6 +927,14 @@ function runApp() {
     return app.getPath('pictures')
   })
 
+  ipcMain.on(IpcChannels.REQUEST_FULLSCREEN, ({ sender }) => {
+    sender.executeJavaScript('document.getElementById("videoContainer").requestFullscreen({navigationUI: "hide"})', true)
+  })
+
+  ipcMain.on(IpcChannels.REQUEST_PIP, ({ sender }) => {
+    sender.executeJavaScript('document.getElementById("video").requestPictureInPicture()', true)
+  })
```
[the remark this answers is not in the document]
kommunarr: Thanks for the suggestion on this - two notes:
- The `requestFullscreen`  callee apparently needs to be the parent of the video element, otherwise it doesn't bind properly with Shaka's fullscreen configurations
- There is [a minute performance benefit to calling `getElementById`](https://stackoverflow.com/a/71878632), albeit very minor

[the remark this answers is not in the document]
absidue: > The requestFullscreen callee apparently needs to be the parent of the video element, otherwise it doesn't bind properly with Shaka's fullscreen configurations

We are calling shaka-player's function which does the right thing, it's the same function we call for our `f` hotkey.

src/main/index.js
```diff
@@ -928,6 +928,14 @@ function runApp() {
     return app.getPath('pictures')
   })
 
+  ipcMain.on(IpcChannels.REQUEST_FULLSCREEN, ({ sender }) => {
+    sender.executeJavaScript('document.getElementById("videoContainer").requestFullscreen({navigationUI: "hide"})', true)
```
PikachuEXE, line 932: Question: why is this required? (`ipcRenderer.send(IpcChannels.REQUEST_FULLSCREEN)` then this

kommunarr: [Transient activation](https://developer.mozilla.org/en-US/docs/Web/Security/User_activation#transient_activation) is required to trigger fullscreen as per the Fullscreen API, i.e., programmatic interaction will not work. Electron's [`executeJavascript` function](https://www.electronjs.org/docs/latest/api/web-contents#contentsexecutejavascriptcode-usergesture), specifically with the `userGesture` boolean set to true, is a clean way for us to bypass this restriction.

PikachuEXE: Do you mind adding code comments for those 2 (I guess FT won't have too many of these) to indicate they are workaround (not usual practice)

kommunarr: Added comments now

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
@kommunarr thought long and hard about the separate toggle for Theater mode but i still do strongly feel that it should be included

kommunarr:
Thoughts on going ahead with it as is and modifying it in a separate PR if we see reports of comprehension problems?

PikachuEXE:
There might be too many options which might look worse in settings UI :P

kommunarr:
We could put the "Default X" (Theater Mode / Stable Volume / Captions / etc) in some kind of shared popup of checkboxes or some other compact form in the future, FWIW. Screenshot from #4231 for reference.

Screenshot_20241203_172828

efb4f5ff-1298-471a-8973-3d47447115dc:
> @efb4f5ff-1298-471a-8973-3d47447115dc Thoughts on going ahead with it as is and modifying it in a separate PR if we see reports of comprehension problems?

Sorry but on this one im really pushing it from my user perspective. It just doesnt sit right with me. The functionality is basically the same.

kommunarr:
I don't agree for reasons I have stated earlier but will concede if other teammates feel the same. @absidue @ChunkyProgrammer @MarmadileManteater 

If they do, I may also have to implement temporary logic for grabbing users' theater mode setting from the store and updating this new setting for 1-2 releases to minimize disruption to users.

Will leave up my main argument from earlier to other contributors who have not been tracking the discourse:

> I was thinking about the same thing you were Pika, but I ultimately think we're overthinking it because we are assuming that people will inherently be asking themselves "but how does theater mode play into it" and get confused, when I don't think there is any inherent reason to do that or imply it if our settings structure doesn't.

> If we have it as separate settings, that's an intuitive indication that viewing mode is orthogonal to whether theater mode is enabled, which is in fact the case in reality as well. For Fullwindow or Fullscreen, you have to actually exit that viewing mode to interact with things like comments and seeing the playlist / recommended videos, and you'll have to see a player that is in theater mode or isn't. If you're in PiP, the size of the grayed out video player is still in play if you want to toggle it off, and that video player is in theater mode or isn't.

> I think we might be assuming that it is going to be a comprehension problem when we theorize it out in terms of number of total distinct states, but I believe that the UX of them being separate settings altogether makes the relationship much less complex and far more intuitive than we're making it out to be.

> Because in reality, default theater mode and this new setting are as non-mutually exclusive as default captions and default viewing mode. We can remove both of those toggles and put it in the new dropdown if we want, but I don't think it's worth it compared to the inconvenience of having to toggle on theater mode every single session you exit FS/FW/PiP when you have FS/FW/PiP as your default viewing mode.

src/main/index.js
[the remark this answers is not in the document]
absidue: Sorry to revive an already closed thread. I've changed my mind on this please use the shaka-player methods, that way we know that it will always behave the same. e.g. shaka-player's PiP methods support both normal PiP and document PiP (customisable PiP) but your code only supports the former, so when Electron adds support for document PIP, we'll either have to duplicate all the code for it from shaka-player or call shaka-player's method, which is what I am already asking you to do.

absidue:
Copying my comment from above as I had left it on an already closed thread so it might have been missed:
> Sorry to revive an already closed thread. I've changed my mind on this please use the shaka-player methods, that way we know that it will always behave the same. e.g. shaka-player's PiP methods support both normal PiP and document PiP (customisable PiP) but your code only supports the former, so when Electron adds support for document PIP, we'll either have to duplicate all the code for it from shaka-player or call shaka-player's method, which is what I am already asking you to do.

absidue:
I agree with @efb4f5ff-1298-471a-8973-3d47447115dc while yes technically full screen by default and theatre mode by default are not mutually exclusive, it seems very unlikey that someone would want to combine them, as you would have to explicitly leave full screen to see theatre mode. So I'm on the side of moving theatre mode by default into the dropdown as a normal option, so making it mutually exclusive (definitely not one of those drop down inside a drop down things, that I thought we had agreed on a while back had bad UX) and for adding a migration into the datastore for migrating over existing users of the theatre mode by default setting.

kommunarr:
> Sorry to revive an already closed thread. I've changed my mind on this please use the shaka-player methods, that way we know that it will always behave the same. e.g. shaka-player's PiP methods support both normal PiP and document PiP (customisable PiP) but your code only supports the former, so when Electron adds support for document PIP, we'll either have to duplicate all the code for it from shaka-player or call shaka-player's method, which is what I am already asking you to do.

Will make these changes now.

>  Sorry to revive an already closed thread. I've changed my mind on this please use the shaka-player methods, that way we know that it will always behave the same. e.g. shaka-player's PiP methods support both normal PiP and document PiP (customisable PiP) but your code only supports the former, so when Electron adds support for document PIP, we'll either have to duplicate all the code for it from shaka-player or call shaka-player's method, which is what I am already asking you to do.

I don't understand where I'm supposed to do this. Is there a way I can call the Shaka built-in functions from `executeJavascript`, or is there another section of the code this is being referred to?

static/locales/en-US.yaml
```diff
@@ -406,11 +406,15 @@ Settings:
...
     Player Settings: Player
     Play Next Video: Play Next Video
-    Turn on Subtitles by Default: Turn on Subtitles by Default
+    Turn on Subtitles by Default: Enable Subtitles by Default
     Autoplay Videos: Autoplay Videos
     Proxy Videos Through Invidious: Proxy Videos Through Invidious
     Autoplay Playlists: Autoplay Playlists
-    Enable Theatre Mode by Default: Enable Theatre Mode by Default
+    Default Viewing Mode:
+      Default Viewing Mode: Default Viewing Mode
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 414: Should Theater mode and Default be added here? 

Also maybe a weird question to ask but should it be called Theater mode or Theater? I mean the header of the dropdown is called Default Viewing Mode so the user does already know we are talking about Modes here. 

image

efb4f5ff-1298-471a-8973-3d47447115dc:
> > Sorry to revive an already closed thread. I've changed my mind on this please use the shaka-player methods, that way we know that it will always behave the same. e.g. shaka-player's PiP methods support both normal PiP and document PiP (customisable PiP) but your code only supports the former, so when Electron adds support for document PIP, we'll either have to duplicate all the code for it from shaka-player or call shaka-player's method, which is what I am already asking you to do.
> 
> @absidue I don't understand where I'm supposed to do this. Is there a way I can call the Shaka built-in functions from `executeJavascript`, or is there another section of the code that is being referred to?

Will approve after this change.

absidue:
> @absidue I don't understand where I'm supposed to do this. Is there a way I can call the Shaka built-in functions from executeJavascript, or is there another section of the code that is being referred to?

As mentioned in my original thread, you can access the `ui` property on the video element and then use it the same way as the `ui` variable in the `ft-shaka-video-player` component.

```js
ipcMain.on(IpcChannels.REQUEST_FULLSCREEN, ({ sender }) => {
  sender.executeJavaScript('document.querySelector("video.player").ui.getControls().toggleFullScreen()', true)
})

ipcMain.on(IpcChannels.REQUEST_PIP, ({ sender }) => {
  sender.executeJavaScript('document.querySelector("video.player").ui.getControls().togglePiP()', true)
})
```

kommunarr:
I sincerely apologize, I completely missed that part when reading the diff and just saw the use of `video.player`! Just added and revalidated now.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
I am waiting for @absidue to review IPC related stuff first

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -1092,6 +1109,15 @@ export default defineComponent({
       emit('ended')
     }
 
+    function handleCanPlay() {
+      // PiP can only be activated once the video's readState and video track are populated
+      if (startInPip && ui.getControls().isPiPAllowed() && process.env.IS_ELECTRON) {
```
absidue, line 1114: PiP is not possible in audio-only mode, I'm not sure if that is already caught by the isPipAllowed() method but if it isn't we should check for it.
```suggestion
      if (startInPip && props.format !== 'audio' && ui.getControls().isPiPAllowed() && process.env.IS_ELECTRON) {
```

src/renderer/components/player-settings/player-settings.js
```diff
@@ -62,6 +62,14 @@ export default defineComponent({
...
       screenshotFilenameExample: '',
       screenshotDefaultPattern: '%Y%M%D-%H%N%S',
+      viewingModeValues: [
+        'default',
+        'theatre',
+        ...(process.env.IS_ELECTRON ? ['fullscreen'] : []),
+        'fullwindow',
+        ...(process.env.IS_ELECTRON ? ['pip'] : []),
+        'external_player'
+      ]
```
absidue, line 72: External players are also an Electron only thing.
As half of the list is Electron only, it would probably be simpler to just do `process.env.IS_ELECTRON ? [electron stuff] : [outside of electron stuff]`

src/renderer/views/Watch/Watch.js
```diff
@@ -339,6 +345,22 @@ export default defineComponent({
...
+          this.useTheatreMode = this.theatrePossible
+          return
+        case 'fullscreen':
+          this.startNextVideoInFullscreen = true
+          return
+        case 'fullwindow':
+          this.startNextVideoInFullwindow = true
+          return
+        case 'pip':
+          this.startNextVideoInPip = true
```
absidue, line 360: Could you please use `break` instead of `return` here, as you aren't using it to return specific values.

src/renderer/components/ft-shaka-video-player/player-components/FullWindowButton.js
```diff
@@ -64,6 +65,12 @@ export class FullWindowButton extends shaka.ui.Element {
       this.updateLocalisedStrings_()
     })
 
+    if (startInFullWindow) {
+      events.dispatchEvent(new CustomEvent('setFullWindow', {
+        detail: true
+      }))
+    }
```
absidue, line 72: I might be missing something but what was the reason for doing this inside of the UI component instead of just doing it in the player component?

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2782,6 +2814,12 @@ export default defineComponent({
       ignoreErrors = true
 
       if (ui) {
+        if (ui.getControls()) {
+          // save the state of player settings to reinitialize them upon next creation
+          const controls = ui.getControls()
+          emit('player-destroyed', controls.isFullScreenEnabled(), fullWindowEnabled.value, controls.isPiPEnabled())
+        }
```
absidue, line 2821: As this function is called by the Watch page, we could probably just return these three values in an object as the return value of the function and avoid the extra emit.

src/datastores/handlers/base.js
```diff
@@ -25,6 +25,16 @@ class Settings {
...
     }
 
+    // In FreeTube 0.23.0, the "Enable Theatre Mode by Default" setting was incoporated as an option
+    // of the "Default Viewing Mode" setting. This is a one time migration to preserve users'
+    // Theater Mode preference through this change.
+    const defaultTheatreMode = await db.settings.findOneAsync({ _id: 'defaultTheatreMode' })
+
+    if (defaultTheatreMode?.value) {
+      await this.upsert('defaultViewingMode', 'theatre')
+      await this.upsert('defaultTheatreMode', false)
```
absidue, line 35: Instead of setting it to false and keeping it around it might be easier to just delete it.

```suggestion
      await db.settings.removeAsync({ _id: 'defaultTheatreMode'})
```

static/locales/en-US.yaml
```diff
@@ -406,11 +406,16 @@ Settings:
   Player Settings:
     Player Settings: Player
     Play Next Video: Play Next Video
-    Turn on Subtitles by Default: Turn on Subtitles by Default
+    Turn on Subtitles by Default: Enable Subtitles by Default
     Autoplay Videos: Autoplay Videos
     Proxy Videos Through Invidious: Proxy Videos Through Invidious
     Autoplay Playlists: Autoplay Playlists
-    Enable Theatre Mode by Default: Enable Theatre Mode by Default
```
absidue, line 413: This should be deleted for all locales, you might want to wait until everyone is ready to merge it before you do so to avoid code conflicts.

src/renderer/views/Watch/Watch.js
kommunarr: This was my way of making sure that this event listener would fire properly. I can change this if there's a more elegant way to do it:

https://github.com/FreeTubeApp/FreeTube/blob/8c3311380bca5a9fda628130a723afbd3145a7a7/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L1665-L1677

src/renderer/components/ft-shaka-video-player/player-components/FullWindowButton.js
kommunarr: This was my way of making sure that this event listener would fire properly. I can change this if there's a more elegant way to do it:

https://github.com/FreeTubeApp/FreeTube/blob/8c3311380bca5a9fda628130a723afbd3145a7a7/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L1665-L1677

kommunarr:
Renamed existing `Remember History` setting to `Remember Watch History` and added `Remember Search History` setting as per suggestions in the Matrix chat.

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -380,6 +380,10 @@ export default defineComponent({
       return this.$store.getters.getExternalPlayer
     },
 
+    externalPlayerIsDefaultViewingMode: function () {
+      return this.externalPlayer !== '' && this.$store.getters.getDefaultViewingMode === 'external_player'
```
absidue, line 384: Even if the setting is not possible/hidden on the settings page, it is still possible that the setting could have an incompatible value for the current install as people could copy their databases between installs.
```suggestion
      return process.env.IS_ELECTRON && this.externalPlayer !== '' && this.$store.getters.getDefaultViewingMode === 'external_player'
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -482,13 +486,18 @@ export default defineComponent({
       return this.isInQuickBookmarkPlaylist ? 'base favorite' : 'base'
     },
 
-    watchPageLinkTo() {
-      // For `router-link` attribute `to`
+    watchVideoRoute() {
```
absidue, line 489: Lets combine these two computed properties into a single one, as they are quite simple.
```suggestion
    watchVideoRouterLink() {
```

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -482,13 +486,18 @@ export default defineComponent({
...
       return {
         path: `/watch/${this.id}`,
         query: this.watchPageLinkQuery,
       }
     },
+
+    // For `router-link` attribute `to`
+    watchVideoRouterLink() {
+      return !this.externalPlayerIsDefaultViewingMode ? this.watchVideoRoute : {}
+    },
```
absidue, line 499:
```suggestion
    // For `router-link` attribute `to`
      if (!this.externalPlayerIsDefaultViewingMode) {
        return {
          path: `/watch/${this.id}`,
          query: this.watchPageLinkQuery,
        }
      } else {
        return {}
      }
    },
```

src/renderer/components/player-settings/player-settings.js
```diff
@@ -125,8 +131,18 @@ export default defineComponent({
       return this.$store.getters.getDefaultQuality
     },
 
-    defaultTheatreMode: function () {
-      return this.$store.getters.getDefaultTheatreMode
+    defaultViewingMode: function () {
+      const defaultViewingMode = this.$store.getters.getDefaultViewingMode
+      if ((defaultViewingMode === 'external_player' && this.externalPlayer === '') ||
```
absidue, line 136:
```suggestion
      if ((defaultViewingMode === 'external_player' && (!process.env.IS_ELECTRON || this.externalPlayer === '')) ||
```

src/renderer/components/player-settings/player-settings.js
```diff
@@ -62,6 +62,12 @@ export default defineComponent({
       screenshotFolderPlaceholder: '',
       screenshotFilenameExample: '',
       screenshotDefaultPattern: '%Y%M%D-%H%N%S',
+      viewingModeValues: [
+        'default',
+        'theatre',
+        'fullwindow',
+        ...(process.env.IS_ELECTRON ? ['fullscreen', 'pip', 'external_player'] : []),
+      ]
```
absidue, line 70: Probably best to make this a computed property, because it needs to contain the matching entries to the names but this currently doesn't check the external player setting, so they'll be out-of-sync if the external player is disabled.

src/datastores/handlers/base.js
```diff
@@ -25,6 +25,16 @@ class Settings {
...
 
+    // In FreeTube 0.23.0, the "Enable Theatre Mode by Default" setting was incoporated as an option
+    // of the "Default Viewing Mode" setting. This is a one time migration to preserve users'
+    // Theater Mode preference through this change.
+    const defaultTheatreMode = await db.settings.findOneAsync({ _id: 'defaultTheatreMode' })
+
+    if (defaultTheatreMode?.value) {
+      await this.upsert('defaultViewingMode', 'theatre')
+      await db.settings.removeAsync({ _id: 'defaultTheatreMode' })
+    }
```
absidue, line 36: We should still delete the setting even if it was false.
```suggestion
    if (defaultTheatreMode) {
      if (defaultTheatreMode.value) {
        await this.upsert('defaultViewingMode', 'theatre')
      }

      await db.settings.removeAsync({ _id: 'defaultTheatreMode' })
    }
```

src/renderer/views/Watch/Watch.js
```diff
@@ -171,8 +174,11 @@ export default defineComponent({
     defaultInterval: function () {
       return this.$store.getters.getDefaultInterval
     },
-    defaultTheatreMode: function () {
-      return this.$store.getters.getDefaultTheatreMode
+    defaultViewingMode: function () {
+      return this.$store.getters.getDefaultViewingMode
+    },
+    externalPlayer: function () {
+      return this.$store.getters.getExternalPlayer
```
absidue, line 181: Looks like this newly added computed property isn't used in this component.
```suggestion
```

absidue (commented):
Still haven't got round to actually testing this pull request as I keep finding things in the code but I should probably just accept the code at some point so that it doesn't take too much longer and clean it up myself after this has been merged (the components will need to be migrated to the compositon API anyway so it could also just be done at the same time).

kommunarr:
Haha, I love your sense of humor! Changes made!

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

static/locales/en-US.yaml
```diff
@@ -406,11 +406,16 @@ Settings:
...
+    Default Viewing Mode:
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 413: I dont see Full Window in here. How does that get translated?

kommunarr: That's the one. I thought it would be better to reuse the existing label than create a new duplicate one. We do this kind of thing in a handful of other places in the app as well, and I suppose a better practice would probably be to move all such keys to a `Global` section like was done in #6495, most likely in a separate effort.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
No idea what's changed/I should test since last time...

kommunarr:
Testing criteria should be up to date - [the functional changes since your last review](https://github.com/FreeTubeApp/FreeTube/pull/5903/files/bb3be09122deecd8141651288fa49ed959cc0618..b82daca7b8905130ef217e39e2bb907273c6f553) were
1. Theater Mode by Default being removed & subsumed into the new `Default Viewing Mode` setting and 
2. Setting migration for theater mode to the new setting

PikachuEXE (approved):
Retested
- Fullscreen
- PiP
- Theater Mode setting migration

PikachuEXE (dismissed):

kommunarr:
Fixed the language-related merge conflicts, so please review at your earliest convenience if possible. Languages updating to the new Autoplay labels are each causing merge conflicts 😭

PikachuEXE (dismissed):
Sleep time~

kommunarr:
Permission to leave the label removal work to our linter @absidue 😭😭

kommunarr:
Thanks, I'm not trying to get out of any due diligence, I just wish that we could do something with `@intlify/vue-i18n/no-unused-keys` and its `ignore`s as a once-per-release thing instead of the more manual upkeep and merge conflict pizzazz

absidue:
Conflicts aren't usually such a big problem, it's just that you had two open PRs that changed strings close to each other, that you want merged not too long after each other. If you were willing to wait a few weeks with this PR you could have solved the conflicts once in a few weeks time when there would have been less new translations, so it probably would have been able to survive for a week or more without getting any conflicts. Not trying to blame you for something, just pointing out the conflicts were inevitable and that waiting longer before solving them would have meant that it would have taken longer before they reappeared.

PikachuEXE (approved):
Merge conflicts:
FdQ_ECYWAAAmGPw

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

Palpabols:
Thank you for fixing this! The viewing mode changing without me wanting to make using FreeTube a hassle. This is one of the best features of the latest update!

PR #5928, opened 2024-10-23
Cleanup Invidious backwards compatibility

absidue:
# Cleanup Invidious backwards compatibility

## Pull Request Type

- [x] Code cleanup

## Related issue
* https://github.com/iv-org/invidious/pull/4586

## Description
This pull request removes the backwards compatibility that we had to handle Invidious returning incorrect height and width values for the legacy format streams and not returning the height and width at all for the adaptive formats. The reason to do so is that the pull request that fixed that was released in Invidious v2.20240825.0 which was released on the 25th of August 2024, so about 3 months ago.

Considering all the issues that Invidious has been having recently it seems highly unlikely that anyone would still be running an Invidious version older than that. However just to be on the safe side, I decided that it's definitely best to wait until after the upcoming release to merge this, just to give instance maintainers and users a bit more time.

## Testing
Open a video with the Invidious API and check that all 3 media format types work and that video height and width are shown in the stats for the legacy formats.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** b9adcb3d89bcb1878cdbaceb83342fd2a8f51477

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -18,7 +18,6 @@ import {
   findMostSimilarAudioBandwidth,
   getSponsorBlockSegments,
   logShakaError,
-  qualityLabelToDimension,
   repairInvidiousManifest,
   sortCaptions,
   translateSponsorBlockCategory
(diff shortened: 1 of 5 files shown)


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
pre-approval

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PR #5943, opened 2024-10-26
Invidious API: Add support for auto-dubbed audio tracks

absidue:
# Invidious API: Add support for auto-dubbed audio tracks

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
https://github.com/LuanRT/YouTube.js/pull/786

## Description
Depends on https://github.com/LuanRT/YouTube.js/pull/786

YouTube had added a feature for uploaders to tell YouTube to automatically generate dubbed audio tracks for their videos. Support for the local API is handled in the linked YouTube.js pull request, this pull request implements it for the Invidious backend, as we handle multiple audio tracks support for Invidious in the FreeTube code (because Invidious doesn't support them itself yet).

## Testing
Check that the player selects the original audio track by default and that you can use some of the other audio tracks.
Test video: https://youtu.be/snZ9w5xc_ic (thanks to the users in the Invidious Matrix chat for mentioning a channel that has those types of audio tracks)

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** eaef9a034135ce80622bf849e8e907c4bb89923a


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Isnt this labeling a bit weird? Also it doesnt look to be sorted a-z?

https://github.com/user-attachments/assets/73ca3407-3a77-4039-b4d6-1f83d7fd8cef

absidue:
Pretty sure we've never done any sorting on the audio tracks on either API, we just display them in the order that they came in the API response.

As for the labels I can force it to always use the `language (region)` format for the labels that we create for the Invidious API, so it matches the labels for the local API that are returned by YouTube.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Pretty sure we've never done any sorting on the audio tracks on either API, we just display them in the order that they came in the API response.

Ah okay good to know. Please disregard my comment.

> As for the labels I can force it to always use the language (region) format for the labels that we create for the Invidious API, so it matches the labels for the local API that are returned by YouTube.

Yes lets do that for the sake of uniformity

absidue:
Updated the display names to always be in the `language (region)` format.

ChunkyProgrammer (approved):

PikachuEXE (approved):

VaelophisNyx:
Is there a timeframe we could expect the update with this to be out by?

absidue:
@VaelophisNyx No eta

PR #6002, opened 2024-10-28
Make `Big Play Button` smaller

### LINKED ISSUE #5951: [Feature Request]: Make huge play button in 0.22 Beta smaller on desktop
### Guidelines

### Problem Description

The play/pause in the center of the video player on 0.22 beta is way too big for desktop screens. It should be smaller. 

{2012DCF4-0295-46CC-97C6-1CD9DFA833F5}

### Proposed Solution

Please make the play/pause button in the center of the player smaller on desktops. 

### Alternatives Considered

N/A

### Issue Labels

visual improvement

### Additional Information

_No response_

ChunkyProgrammer:
# Make `Big Play Button` smaller

## Pull Request Type
- [x] Other - visual improvement

## Related issue
closes #5951

## Description
This PR adds some transparency to the Big Play Button and adds a max size for it (110px x 110px).

## Screenshots (from 22 inch monitor)
Before:
Normal mode:
image
Theater Mode:
image

After:
Normal mode:
image
Theater mode (the button is the same size but it might look smaller on GitHub since the screenshot is larger):
image

## Testing 
- I was able to test this on a 22inch monitor, 24inch monitor and 64 inch TV but the TV was zoomed in to 200%. It looked good on these monitors. I'd recommend testing with large monitors (smaller monitors/phones should be fine/the same).

## Desktop
- **OS:** Linux Mint 22
- **OS Version:** 22


kommunarr:
For reference, our pre-Shaka play button opacity was 0.45:
https://github.com/FreeTubeApp/FreeTube/blob/256b27cfed9cb487735076cbb14108300cdbade7/src/renderer/videoJS.css#L2031C35-L2031C37

kommunarr:
It's also worth noting that YT does not have this feature on either web or their mobile app. Maybe we change the default setting? Fine for keeping that discussion & implementation to a separate one-line PR if desired, though

PikachuEXE:
Oh default value is on?
I guess this is from the player we use not from YT
If auto play video is on by default the big button seems not so useful?

ChunkyProgrammer:
By default in shaka-player, the big play button only appears for mobile users. I'm okay with changing the default to off. @MarmadileManteater you may want to keep it on for FreeTubeAndroid.

absidue:
@kommunarr I'm slightly confused what you mean by YouTube's mobile app doesn't have this, because from what I can tell it definitely does:
image

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
```diff
@@ -171,6 +171,14 @@
   opacity: 1;
 }
 
+:deep(.shaka-controls-container[shown='true'] > .shaka-play-button-container > .shaka-play-button) {
+  opacity: 0.75 !important;
```
absidue, line 175: This change should only apply when `prefers-reduced-transparency` is `no-preference` or you could add a second block to override it back to the default transparency when it is set to `reduce`.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-transparency

kommunarr:
@absidue On second check, you're correct, it does have it, but the overlay dismisses itself on clicking the player. My mistake

kommunarr:
**non-blocking:** I'd like to see what it would look like at the v0.21 0.45 opacity

ChunkyProgrammer:
> **non-blocking:** I'd like to see what it would look like at the v0.21 0.45 opacity

0.45:
image

0.75:
image

kommunarr:
I prefer the lower opacity just because it only really matters if it's visible at all versus not. Will go with whichever one the team prefers.

PikachuEXE:
Ask https://github.com/FreeTubeApp/FreeTube/issues/5951 guys?

absidue (approved):

Pentaphon:
I like the 0.45 opacity setting. The play button should be there but it should feel small and out of the way and lower opacity helps with that.

digisus:
Just to add that I too prefer the 0.45 opacity - and in general the smaller button for desktops. I think the suggestions made in #6216 are spot on. Thank you!

PR #6012, opened 2024-10-29
Add support for hms timestamp in query string

### LINKED ISSUE #6008: [Feature Request]: Support for alternative timestamp url
### Guidelines

### Problem Description

Usually timestamps are formatted like this in seconds https://youtu.be/0IxaD8QwKTs?t=98s

But turns out that this human readable format (Hours:Minutes:Seconds) also works in Youtube, but not yet in freetube

https://youtu.be/0IxaD8QwKTs?t=0h1m38s

### Proposed Solution

Be able to load based on the alternative timestamp format

https://youtu.be/0IxaD8QwKTs?t=0h1m38s

### Alternatives Considered

None

### Issue Labels

improvement to existing feature, support for external software

### Additional Information

Hopefully loading url from outside freetube like using LibRedirect will also be updated with this feature.

ChunkyProgrammer:
# Title

## Pull Request Type
- [x] Feature Implementation

## Related issue
closes #6008 

## Description
Adds support for alternate timestamp urls.

## Testing (i think it might subtracts 2 seconds from it)
- go to https://youtu.be/27iu423Yjtc?t=2h10m20s 
- make sure it goes to 2 hours 10 minutes and 20 seconds

- go to https://youtu.be/27iu423Yjtc?t=3h
-  make sure it goes to 3 hour mark

- go to https://youtu.be/27iu423Yjtc?t=15m
- make sure it goes to 15 minute mark

- go to https://youtu.be/27iu423Yjtc?t=40s
- make sure it goes to 40 second mark  

## Desktop

- **OS:** Linux Mint
- **OS Version:** 22

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -146,7 +146,28 @@ function runApp() {
               }
 
               if (params.has('timestamp')) {
-                newParams.set('t', params.get('timestamp'))
+                let timestamp = params.get('timestamp')
+                if (timestamp && (timestamp.includes('h') || timestamp.includes('m') || timestamp.includes('s'))) {
+                  if (timestamp && (timestamp.includes('h') || timestamp.includes('m') || timestamp.includes('s'))) {
+                    const { seconds, minutes, hours } = timestamp.match(/(?:(?<hours>(\d+))h)?(?:(?<minutes>(\d+))m)?(?:(?<seconds>(\d+))s)?/).groups
+                    let time = 0
+
+                    if (seconds) {
+                      time += Number(seconds)
+                    }
+
+                    if (minutes) {
+                      time += 60 * Number(minutes)
+                    }
+
+                    if (hours) {
+                      time += 3600 * Number(hours)
+                    }
+
+                    timestamp = time
+                  }
+                }
+                newParams.set('t', timestamp)
                 hasParams = true
               }
 
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -611,7 +611,26 @@ export function getVideoParamsFromUrl(url) {
 
   function extractParams(videoId) {
     paramsObject.videoId = videoId
-    paramsObject.timestamp = urlObject.searchParams.get('t')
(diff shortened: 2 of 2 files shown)


src/main/index.js
absidue, line 1: The changes in this file are not necessary as that code is for the context menu on in-app links and as we always convert the timestamp into seconds before adding it to the watch page URL there is no scenario where it would have this format, if it did the watch page would break.

src/renderer/helpers/utils.js
```diff
@@ -611,7 +611,26 @@ export function getVideoParamsFromUrl(url) {
 
   function extractParams(videoId) {
     paramsObject.videoId = videoId
-    paramsObject.timestamp = urlObject.searchParams.get('t')
+    let timestamp = urlObject.searchParams.get('t')
+    if (timestamp && (timestamp.includes('h') || timestamp.includes('m') || timestamp.includes('s'))) {
```
absidue, line 615: This breaks support for links that don't include a letter after the number of seconds which is something we previously supported. Which means you can't copy a video link with a timestamp from FreeTube's watch page and use it in FreeTube anymore without modifying the URL.

src/main/index.js
ChunkyProgrammer: Ahh okay, I should have looked at it a bit more. I thought this was part of app startup when passing a link through the command line

src/renderer/helpers/utils.js
absidue: I haven't actually tested the code yet, but that if checks for whether a timestamp exists AND that it contains one of those 3 letters, so theoretically at least based on reading the code number only timestamps won't work anymore.

ChunkyProgrammer: It should have the same behavior if it's number only since L614 and L633 are outside of the if statement

absidue: Ah okay I missed that, sorry.

PikachuEXE (approved):

kommunarr (approved):

absidue (approved):

PR #6042, opened 2024-10-30
Add a polyfill for vue-i18n's `useI18n()` composable

absidue:
# Add a polyfill for vue-i18n's useI18n() composable

## Pull Request Type

## Description
This pull request introduces a polyfill for `vue-i18n` 9+'s `useI18n` composable as well as a custom auto-fixable ESLint rule to enforce the use of the polyfill.

Why?
- `vue-i18n` 8, the version of `vue-i18n` that is compatible with Vue 2, doesn't have the `useI18n` composable, so this polyfill allows us to already migrate components that need to use vue-i18n in the JavaScript section (e.g. accessing the current locale to pass to `Intl` APIs or using the translate function in computeds like we do for the drop down labels).
- Even once we upgrade to Vue 3, `vue-i18n` 9+'s `useI18n` composable cannot be used while the `vue-i18n` instance is in legacy mode `legacy: true`, but we need legacy mode for components that are still using the Options API (unless we migrate every single component to the Composition API before we use Vue 3, which is unlikely to happen)
- `vue-i18n` 9 does have an `allowComposition` option, however that option comes with some serious limitations (https://vue-i18n.intlify.dev/guide/migration/vue3#limitations) such as requiring the call to `useI18n` to be inside the `nextTick()` callback which is only fired after the component has already been mounted/rendered and was removed in `vue-i18n` 10 (https://vue-i18n.intlify.dev/guide/migration/breaking10.html#drop-allowcomposition-option). The polyfill in this PR doesn't have those limitations.
- Compared to importing the vue-i18n instance and accessing the properties on it directly (what we currently do in the `ft-shaka-video-player.js` file), this provides a drop-in replacement for the `useI18n` function, so once we turn off `legacy` mode, all we need to do is remove the ESLint rule and update the imports to point to `vue-i18n` instead.
- I created an auto-fixable ESLint rule so you don't even have to think about which `useI18n` function you import or that your editor has imported the wrong one, as the ESLint rule will catch it and can fix the import for you.
- The only extra work this introduces for the initial Vue 3 migration is a few minor changes to the polyfill file, all files that import the polyfill should "just work".

## Screenshots
eslint-error

eslint-quick-fix

using-polyfill

### Testing the polyfill
1. Open two FreeTube windows
2. In the first window navigate to a video
3. In the second window open the general settings
4. Change the display locale and check that the video player updates it's language (e.g. hover over the playback rate selector)
5. Take a screenshot and make sure the toast message is translated

### Testing the ESLint rule
1. Try adding `import { useI18n } from 'vue-i18n'` to a file and check that ESLint detects it and that the fix works (it should change the text after the `from` part).
2. Try adding `import { useI18n, createI18n } from 'vue-i18n'` to a file and check that ESLint detects and and that the fix works (it should edit the existing import to remove the useI18n part and create a second import importing the `useI18n` polyfill)

## Desktop

- **OS:** Windows 10
- **OS Version:** 10
- **FreeTube version:** d6788636f145448be45f3ceb710f672aad3e08c2

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -10,6 +10,7 @@ import eslintPluginJsonc from 'eslint-plugin-jsonc'
 import eslintPluginYml from 'eslint-plugin-yml'
 import yamlEslintParser from 'yaml-eslint-parser'
 import neostandard from 'neostandard'
+import eslintPluginFreeTube from './_scripts/eslint-rules/plugin.mjs'
 
 import activeLocales from './static/locales/activeLocales.json' with { type: 'json' }
 
@@ -40,6 +41,7 @@ export default [
     ],
     plugins: {
       unicorn: eslintPluginUnicorn,
+      freetube: eslintPluginFreeTube
     },
 
     languageOptions: {
@@ -115,6 +117,8 @@ export default [
       '@intlify/vue-i18n/no-deprecated-tc': 'off',
       'vue/require-explicit-emits': 'error',
       'vue/no-unused-emit-declarations': 'error',
+
+      'freetube/use-i18n-polyfill': 'error'
     },
   },
 
@@ -209,7 +213,7 @@ export default [
     }
   },
   {
-    files: ['_scripts/*.mjs'],
+    files: ['_scripts/**/*.mjs'],
     languageOptions: {
       globals: {
         ...globals.node,
--- a/src/renderer/composables/use-i18n-polyfill.js
+++ b/src/renderer/composables/use-i18n-polyfill.js
@@ -1,0 +1,79 @@
+/* eslint-disable @intlify/vue-i18n/no-dynamic-keys */
+import { computed } from 'vue'
+
+import i18n from '../i18n/index'
+
+/**
+ * Polyfill for vue-i18n's useI18n composable, as it is not available in Vue 2
+ * and doesn't work when vue-i18n 9+ (used for Vue 3) is set to `legacy: true`,
+ * which is needed for Options API components.
+ *
+ * Yes, vue-i18n 9 has an `allowComposition` option,
+ * but it comes with limitations that this polyfill doesn't have and was removed in vue-i18n 10.
+ *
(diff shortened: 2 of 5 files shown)


eslint.config.mjs
```diff
@@ -209,7 +213,7 @@ export default [
     }
   },
   {
-    files: ['_scripts/*.mjs'],
+    files: ['_scripts/**/*.mjs'],
```
PikachuEXE, line 216: I think the one on `eslint-lint` & `eslint-lint-fix` should be updated too

PikachuEXE (commented):
image

Before fix (`yarn run eslint-lint-fix`)
image
After fix
image

src/renderer/composables/use-i18n-polyfill.js
```diff
@@ -0,0 +1,79 @@
...
+ */
+function t(key, arg1, arg2) {
+  // Remove these lines in the Vue 3 migration and pass all args to the `.t()` call
+  if (typeof arg1 === 'number') {
+    return i18n.tc(key, arg1)
+  } else if (typeof arg2 === 'number') {
+    return i18n.tc(key, arg2, arg1)
+  }
+
+  return i18n.t(key, arg1)
```
ChunkyProgrammer, line 78: I guess arg1 would be ignored if null or undefined here?

eslint.config.mjs
absidue: Good catch, thank you!

absidue: Fixed!

PikachuEXE (approved):

ChunkyProgrammer (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
rip didnt see that the conflicts label got applied

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #6048, opened 2024-10-31
Add some missing jsdoc comments, add jsdoc eslint plugin

ChunkyProgrammer:
# Add some missing jsdoc comments, add jsdoc eslint plugin

## Pull Request Type
- [x] Other - code quality improvement

## Description
This PR  does the following:
- adds some basic jsdoc rules for the jsdoc eslint plugin
- fixes some errors from the eslint plugin
- adds some missing jsdoc comments

## Additional Info
The typescript eslint plugin is able to convert jsdoc comment annotations to typescript so it should make it easier to switch to typescript whenever we start migrating some of our code to typescript.

## Desktop
- **OS:** Linux Mint
- **OS Version:** 22

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,6 +20,8 @@ updates:
           - "@eslint/*"
           - "yaml-eslint-parser"
           - "vue-eslint-parser"
+          - "typescript-eslint"
+          - "neostandard"
       stylelint:
         patterns:
           - "stylelint"
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -10,6 +10,7 @@ import eslintPluginJsonc from 'eslint-plugin-jsonc'
 import eslintPluginYml from 'eslint-plugin-yml'
 import yamlEslintParser from 'yaml-eslint-parser'
 import neostandard from 'neostandard'
+import jsdoc from 'eslint-plugin-jsdoc';
 
 import activeLocales from './static/locales/activeLocales.json' with { type: 'json' }
 
@@ -31,6 +32,16 @@ export default [
   ...eslintPluginVue.configs['flat/vue2-recommended'],
   ...vuejsAccessibility.configs["flat/recommended"],
   ...intlifyVueI18N.configs['flat/recommended'],
+  {
+    plugins: {
+      jsdoc
+    },
+    rules: {
+      'jsdoc/check-param-names': 'error',
+      'jsdoc/no-multi-asterisks': 'error',
+      'jsdoc/check-alignment': 'error'
+    }
+  },
   {
     files: [
       '**/*.{js,vue}',
--- a/src/renderer/helpers/player/utils.js
+++ b/src/renderer/helpers/player/utils.js
@@ -189,13 +189,13 @@ export function sortCaptions(captions) {
  * This function cleans it up, so that we can use it.
  *
  * Here is a list of things this function does:
- * * Removes bogus roles and labels
- * * Extracts the languages from the audio URLs if available and adds it to the adapation sets
- * * Adds roles and labels when possible to add support for multiple audio tracks
+ * Removes bogus roles and labels
+ * Extracts the languages from the audio URLs if available and adds it to the adapation sets
+ * Adds roles and labels when possible to add support for multiple audio tracks
  *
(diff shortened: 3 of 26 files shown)


.github/dependabot.yml
```diff
@@ -20,6 +20,8 @@ updates:
           - "@eslint/*"
           - "yaml-eslint-parser"
           - "vue-eslint-parser"
+          - "typescript-eslint"
```
absidue, line 23: I'm guessing this is left over from some testing, as we don't have it as a direct dependency and likely won't either as it is a dependency of neostandard so we just need to turn it on and if we need it separately we need to use their reexport to avoid problems (see discussion here https://github.com/neostandard/neostandard/issues/184).
```suggestion
```

eslint.config.mjs
```diff
@@ -31,6 +32,16 @@ export default [
...
+  {
+    plugins: {
+      jsdoc
+    },
+    rules: {
+      'jsdoc/check-param-names': 'error',
+      'jsdoc/no-multi-asterisks': 'error',
+      'jsdoc/check-alignment': 'error'
+    }
+  },
```
absidue, line 44: I haven't tested this yet but placing it in it's own group without any constraint on which file extensions to apply it to, from my understanding would mean that these eslint rules are run on the JSON and YAML files too. We should probably add it directly to the relevant existing config objects or limit this to the relevant files.

src/renderer/helpers/player/utils.js
```diff
@@ -189,13 +189,13 @@ export function sortCaptions(captions) {
...
- * * Adds roles and labels when possible to add support for multiple audio tracks
+ * Removes bogus roles and labels
+ * Extracts the languages from the audio URLs if available and adds it to the adapation sets
+ * Adds roles and labels when possible to add support for multiple audio tracks
  *
  * Things this function does not do:
- * * Separate DRC (Stable Volume) from their original counterparts
- * * Tag HDR video streams (Invidious puts all video streams in the same adaptation set,
+ * Separate DRC (Stable Volume) from their original counterparts
+ * Tag HDR video streams (Invidious puts all video streams in the same adaptation set,
```
absidue, line 198: Could we replace them with `-`, as they were meant to be bullet points or would that get flagged by the eslint plugin too? If that doesn't work either it's not a big deal :)

absidue (dismissed):

eslint.config.mjs
```diff
@@ -10,6 +10,7 @@ import eslintPluginJsonc from 'eslint-plugin-jsonc'
 import eslintPluginYml from 'eslint-plugin-yml'
 import yamlEslintParser from 'yaml-eslint-parser'
 import neostandard from 'neostandard'
+import jsdoc from 'eslint-plugin-jsdoc';
```
PikachuEXE, line 13:
```suggestion
import jsdoc from 'eslint-plugin-jsdoc'
```

eslint.config.mjs
```diff
@@ -40,6 +41,7 @@ export default [
     ],
     plugins: {
       unicorn: eslintPluginUnicorn,
+      jsdoc
```
PikachuEXE, line 44:
```suggestion
      jsdoc,
```

PikachuEXE (changes requested):
Mainly for remove unnecessary semicolon

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/views/Watch/Watch.js
```diff
@@ -722,7 +722,8 @@ export default defineComponent({
           }
 
           if (result.storyboards?.type === 'PlayerStoryboardSpec') {
-            let source = result.storyboards.boards
+            /** @type {import('youtubei.js/dist/src/parser/classes/PlayerStoryboardSpec').StoryboardData[]} */
+            let source = result.storyboardSpec.boards
```
PikachuEXE, line 726: @ChunkyProgrammer 
This blows up
image

ChunkyProgrammer: Whoops, didnt mean to change this from storyboards to storyboardSpec

PR #6064, opened 2024-11-01
Migrate FtTimestampCatcher to the composition API

absidue:
# Migrate FtTimestampCatcher to the composition API

## Pull Request Type

## Description
This pull request migrates the FtTimestampCatcher component to the compostion API. During the Vue 3 migration we'll have to change the `vue-router/composables` import to `vue-router`, as it will error at build time if we don't change it it will be hard to miss and can be fixed quickly with a project wide find and replace, I decided it wasn't a big deal.

## Testing

Check that clicking on a timestamp jumps to that place in the video and that middle or CTRL+clicking on a timestamp opens a new window which starts playing at that timestamp.
Example video: https://youtu.be/jNQXAC9IVRw

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 59eff14712ae3c65b1847ce69375f4ba66d2882a


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
> and that middle or CTRL+clicking on a timestamp opens a new window which starts playing at that timestamp.

This doesnt open up the video in a new tab for me. Also doenst work on latest nightly build.

I did made it work by pressing right click on the timestamp which brings up the context menu which gives me to option to open in a new tab. So i think i can approve?

PikachuEXE:
Middle click does open but not Cmd+Click but it's like that in dev

absidue:
As far as I know we don't do anything special for the open in new window stuff, so it's possible that Electron just has different behaviors oh different operating systems.

PikachuEXE (approved):
`onclick` unchanged, plus it's how we handle YT link

```js
        this.handleYoutubeLink(el.href, {
          doCreateNewWindow: event.type === 'auxclick'
        })
```

ChunkyProgrammer (approved):

kommunarr (approved):

PR #6099, opened 2024-11-04
Add warning message to proxy settings

ChunkyProgrammer:
# Add warning message to proxy settings

## Pull Request Type
- [x] Other - usability improvement

## Related issue
It isn't clear for non-power users that you actually have to set up a proxy/set up Tor to get it to work. This change should remove some of that confusion.

## Description
Adds a warning message under the proxy settings saying you need to setup a proxy/Tor and configure it or FreeTube won't work

## Screenshots 
image

## Testing 
- go to proxy settings
- see warning

## Desktop
- **OS:** Linux Mint
- **OS Version:** 22


src/renderer/components/proxy-settings/proxy-settings.css
```diff
@@ -1,3 +1,15 @@
 .protocol-dropdown {
   margin-block-end: 1rem;
 }
+
+.proxy-warning {
+  font-weight: bold;
+  padding-inline: 4%;
+  background-color: var(--accent-color);
+  color: var(--text-with-accent-color);
+  padding-block: 3px;
```
PikachuEXE, line 10:
```suggestion
  padding-block: 12px;
  padding-inline: 4%;
  background-color: var(--accent-color);
  color: var(--text-with-accent-color);
  font-weight: bold;
```
image

static/locales/en-US.yaml
```diff
@@ -573,6 +573,7 @@ Settings:
     Manage Subscriptions: Manage Subscriptions
   Proxy Settings:
     Proxy Settings: Proxy
+    Proxy Warning: FreeTube doesn't have a built-in proxy but can connect to an external proxy, such as one running on your machine like Tor or an external one provided by a VPN. If enabled, make sure your proxy/Tor is set up, or FreeTube won't be able to fetch any data.
```
absidue, line 576: I'm a bit concerned that the VPN part of the sentence will make people think they need to configure something in FreeTube when using a VPN.

efb4f5ff-1298-471a-8973-3d47447115dc:
Going to play devils advocate here, why go for this banner instead of our standard tooltip `?` Not against it but i would like to understand the reasoning. If we go this way than it should only be shown when the setting is enabled.

offtopic discussion: should we also use a banner for other settings?

absidue:
I think part of the problem here is the naming of the setting itself "enable proxy/tor" sounds like you can just turn it on like other settings and it will just work, something along the lines of "connect to external proxy" with a subtitle of "enter the connection details for your proxy below" might make it more obvious what the setting actually does.

PikachuEXE:
`Enable Tor / Proxy` => `Enable Tor / External Proxy`? (Something like that not sure if Tor can be called "external" I don't use it

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
> [...] If we go this way than it should only be shown when the setting is enabled.

Thoughts?

PikachuEXE:
Ya that one would be noticeable

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PR #6134, opened 2024-11-10
Fix wayland app icon for flatpak.

mariakeating:
# Fix wayland app icon for flatpak.

## Pull Request Type
- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Currently, the app icon (i.e. titlebar icon in KDE) is incorrect as the
app_id is set to FreeTube instead of matching the desktop file with
io.freetubeapp.FreeTube.

## Screenshots
Screenshot_20241111_001919

## Testing
Testing done on an arch linux install.
Untested for distribution packages nor windows and macos builds.

## Desktop
- **OS:** Arch Linux

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "freetube",
   "productName": "FreeTube",
+  "desktopName": "io.freetubeapp.FreeTube",
   "description": "A private YouTube client",
   "version": "0.22.0",
   "license": "AGPL-3.0-or-later",


PikachuEXE (approved):
App icon looks the same on MacOS (`yarn build:arm64`)

absidue (approved):
No idea how to test this, without doing a full on flatpak release (we unpack the Linux zip files and use their contents for the flatpak, so I don't understand how an extra package.json field would change anything about the flatpak, unless the flatpak runtime reads the package.json file?) but as this doesn't seem to break anything (IDK if this would break the deb install for example), I'm happy to give it a try.

absidue:
Dismissing my review because it's probably worth checking that this doesn't break the deb builds, which afaik use the `.desktop` file that electron-builder auto-generates instead of the static one in the flatpak repo.

mariakeating:
> I don't understand how an extra package.json field would change anything about the flatpak.

It's not changing the flatpak, it's changing the value of the app_id set for the xdg_toplevel to match the .desktop file the flatpak provides. If flatpak didn't require such a naming convention, I would have made a PR on the flatpak's repo to rename the file.

> Dismissing my review because it's probably worth checking that this doesn't break the deb builds, which afaik use the `.desktop` file that electron-builder auto-generates instead of the static one in the flatpak repo.

Definitely worth checking. My understanding from the docs is that this also changes the name of the generated .desktop file but I'm not familiar with this tooling.

mariakeating:
Unfortunately, with this change, the file generated for .deb files is still named freetube.desktop instead of matching the desktopName key, resulting in wayland icons broken *outside* of flatpak.

Seems like electron builder makes this very frustrating. https://github.com/electron-userland/electron-builder/issues/4071

Unless anybody knows how to work around this, I will close and instead make a PR for the flatpak's repo as there is another option that would achieve this (2nd paragraph).
https://docs.flatpak.org/en/latest/electron.html#using-correct-desktop-file-name

Issue #5618, opened 2024-08-27
[Feature Request]: In the 'Playlist View' I can sort by MANY different options. It would be AWESOME if we could sort by RANDOM order.


aaronsql2019:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

In the 'Playlists' feature. I can click on a playlist and sort by many different options.
#1. I clicked on 'Custom' in the sort by clause. I don't know what this is sorting by.
#2. I'd LIKE to sort by RANDOM (so that every time I listen to this playlist, it is in a new order).

### Proposed Solution

I'd like to be able to 'Sort By: Random' 
I have 103 songs in this playlist. I think that 10 songs are the same SONG but different ReMixes.
I want to keep all the 'Duplicates' but I want to listen to them. I just don't want to sort alphabetically, because then I have 12 copies of the song 'IDGAF' (by Dua Lipa). 

Again, I WANT to listen to all of these mixes. I just don't want to listen to them ONE RIGHT AFTER THE OTHER.

### Alternatives Considered

As an alternative, if you could just explain what the 'Sort Order: Custom' does. This MIGHT already do what I want (but I think that it merely allows me to do a CUSTOM re-ordering of the songs (doing drag and drop).

![Dua_Lipa_IDGAF_Sort_Order_20240827](https://github.com/user-attachments/assets/cd4f8242-fb2b-417e-8b73-94d4a4c9ada1)

### Issue Labels

ease of use improvement

### Additional Information

I started by taking a copy of THIS playlist. 
https://www.youtube.com/playlist?list=PLfqxLOyMxF2-c6P8d185EM4HcjgrYzAEl
I clicked each song in the list to add it to my 'Default Playlist'. That generally works. 

I just ALSO wish that there was an easier way to import a BUNCH of playlists to SQLITE.
My PRIMARY FEATURE is to be able to access these playlists via SQLITE. I assume that I can do that (once I copy the playlist into FreeTube)?

[...]

PR #6153, opened 2024-11-12
Add Play All and Shuffle Button

JL0000:
# Add Play All and Shuffle Button

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

#5618

## Description

Added `Play all` and `Shuffle` buttons to both user and youtube playlists.
Clicking on `Play all` always plays the first video in the playlist.
Clicking on `Shuffle` plays a random video in the playlist and toggles shuffle mode on.

### Before
**User Playlist**
playlist-no-button
playlist-no-button-list
**Youtube Playlist**
yt-playlist-no-button

### After
**User Playlist**
playlist-button
playlist-button-list
**Youtube Playlist**
yt-playlist-button

## Testing 

`Play all` plays the first video after video order is changed.
`Shuffle` works as intended after a video is deleted whilst playing through the playlist. 
`Shuffle` resets the playlist after adding a new video whilst it is playing its last item in the iteration. This behaviours is inherited from the development branch.

## Desktop

- **OS:** MacOs
- **OS Version:** 15.1


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Everything looks good but i did notice this error when entering an empty playlist.

https://github.com/user-attachments/assets/3e85c84d-d92d-4310-94a2-34eb9aec6e3c

JL0000:
I have fixed this error and I have hid the buttons when the playlist is empty.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
I don't think the buttons should be shown in edit mode
Screenshot 2024-11-17 at 08 10 16

I will fix other buttons/inputs myself later

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> I do also think that all of our icons should have visible labels, not just these two, for accessibility & ease of use reasons. That said, that's no reason to not have them here, just that in the future I recommend that all of these buttons become of the same type, with optionally the same positioning differences kept if we feel like it.

This is a signal boost for https://github.com/FreeTubeApp/FreeTube/pull/4950 and i like the member to chime in on that PR. I think it would be a good improvement.

JL0000:
> **suggestion (non-blocking):** I'd ideally like it if on a desktop view breakpoint for the grid view playlist page, the top row became a flex container with the icons on the left and the new buttons on the right. This way the fixed container wouldn't have to be as tall.
> 
So I tried this out and it turned out fine
Screenshot 2024-11-23 at 21 57 12
On a public playlist, because `playlistOptionsWrapper` was moved out of `channelShareWrapper` (the new buttons in the `playlistOptionsWrapper` will push `playlistChannel` to the left), the copy and share buttons will not be to the immediate right of the channel name. Is this fine?
Screenshot 2024-11-23 at 21 54 58

efb4f5ff-1298-471a-8973-3d47447115dc:
Thoughts:

Playlist Grid view: To it seems a bit weird that the Secondary color theme colors are divided by the red icons. Maybe place the buttons to the left of the bookmark icon so that its grouped by color?

image

Channel playlist view: Would maybe look nicer if the buttons are placed to the left of the copy and share link buttons

image

Playlist List view: The suggestion above cant be applied here as there isnt space for it. I do think that it kind of looks off when sitting between the icons above and the searchbar but im not sure if it can be made to look 'nicer'

image

JL0000:
I tried a few things and I think the best it looked was grid mode not changing but list mode having centered buttons.
Screenshot 2024-11-26 at 19 35 54 Screenshot 2024-11-26 at 19 36 09

efb4f5ff-1298-471a-8973-3d47447115dc:
that does look good but im still bothered by the grid view as it is beside the other colored options. Would like to hear from others on this.

One additional suggestion could be to go back to this style:

https://github.com/FreeTubeApp/FreeTube/pull/6153#pullrequestreview-2456416968

efb4f5ff-1298-471a-8973-3d47447115dc:
@

> that does look good but im still bothered by the grid view as it is beside the other colored options. Would like to hear from others on this.
> 
> One additional suggestion could be to go back to this style:
> 
> [#6153 (review)](https://github.com/FreeTubeApp/FreeTube/pull/6153#pullrequestreview-2456416968)

@kommunarr thoughts?

kommunarr:
Putting the new buttons to the left sounds okay. Make sure that the functional horizontal margin between them is the same as the one between the icon buttons, as it appears not to be glancing at it now

Issue #5652, opened 2024-09-05
[Feature Request]: "Hide Videos From Channels" List As Submenu


ne04:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The distraction free settings were recently expanded with "Hide Videos From Channels" and "Hide Videos And Playlists Containing Text" options. This is great and I have found these options very useful, but the problem is the list of channels and keywords hidden are very prominently displayed on the settings page, and channels and keywords I do not want to see or think about show up while I'm trying to change other freetube settings. I'm sure that I'm not the only one with this issue, and I imagine this feature would be very useful for people struggling with trauma or addiction, or victims of abuse; if not for this primary limitation. Attached is a screenshot of how the page looks currently with placeholder channels and keywords. 
![Screenshot from 2024-09-04 20-07-45](https://github.com/user-attachments/assets/4aec41c4-1616-4515-96e5-ed797f7d7a09)

### Proposed Solution

This feature would be more able to be utilized by everyone if the list of hidden channels / keywords were made into a hidden submenu that you use a click to open in order to make removals. The initially hidden state makes complete sense for a feature for hiding the data entered. Suggestion screenshots attached.
![viewhidden](https://github.com/user-attachments/assets/c5b6e4b8-eddb-453a-badf-c3bbfd67f0e8)
![viewhiddenmenu](https://github.com/user-attachments/assets/1348a82c-c866-416c-94db-f5701ea0c357)

### Alternatives Considered

Another alternative that could be lower-effort would be something similar to a spoiler tag, where the entries still appear but are initially grey-ed out and require a click to view. Or even a "Hide Hidden Channels List" toggle switch would suffice.

### Issue Labels

improvement to existing feature, visual improvement

### Additional Information

[...]

PR #6156, opened 2024-11-12
Added buttons to hide hidden channels/text

JL0000:
# Added buttons to hide hidden channels/text

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

#5652 

## Description

- Added toggle switches to hide/show the channels and text hidden by filters set by user
- Grouped the toggles together with the filtering into a new section

## Screenshots 

**Before**
before
**After**
after

## Testing 

Toggles are on by default.
Toggles hide/show the filtered channels and text.

## Desktop

- **OS:** MacOs
- **OS Version:** 15.1

### DIFF
--- a/src/renderer/components/ft-input-tags/ft-input-tags.vue
+++ b/src/renderer/components/ft-input-tags/ft-input-tags.vue
@@ -22,7 +22,7 @@
       @click="updateTags"
     />
     <div
-      v-if="showHiddenContent"
+      v-if="showTags"
       class="ft-tag-box"
     >
       <ul>
--- a/src/renderer/components/ft-input-tags/ft-input-tags.vue
+++ b/src/renderer/components/ft-input-tags/ft-input-tags.vue
@@ -21,7 +21,10 @@
       :force-action-button-icon-name="['fas', 'arrow-right']"
       @click="updateTags"
     />
-    <div class="ft-tag-box">
+    <div
+      v-if="showHiddenContent"
+      class="ft-tag-box"
+    >
       <ul>
         <li
           v-for="tag in tagList"
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -296,6 +296,8 @@ const state = {
   screenshotFolderPath: '',
   screenshotFilenamePattern: '%Y%M%D-%H%N%S',
   settingsSectionSortEnabled: false,
+  showHiddenChannels: true,
+  showHiddenText: true,
   fetchSubscriptionsAutomatically: true,
   settingsPassword: '',
   useDeArrowTitles: false,
--- a/src/renderer/components/distraction-settings/distraction-settings.js
+++ b/src/renderer/components/distraction-settings/distraction-settings.js
@@ -111,6 +111,12 @@ export default defineComponent({
     showDistractionFreeTitles: function () {
       return this.$store.getters.getShowDistractionFreeTitles
     },
+    showHiddenChannels: function () {
+      return this.$store.getters.getShowHiddenChannels
+    },
+    showHiddenText: function () {
+      return this.$store.getters.getShowHiddenText
+    },
     channelsHidden: function () {
       return JSON.parse(this.$store.getters.getChannelsHidden).map((ch) => {
         // Legacy support
@@ -224,6 +230,8 @@ export default defineComponent({
       'updateHideSubscriptionsShorts',
       'updateHideSubscriptionsLive',
       'updateHideSubscriptionsCommunity',
+      'updateShowHiddenChannels',
+      'updateShowHiddenText'
     ])
   }
 })
--- a/src/renderer/components/distraction-settings/distraction-settings.vue
+++ b/src/renderer/components/distraction-settings/distraction-settings.vue
@@ -255,11 +255,11 @@
         :tag-name-placeholder="$t('Settings.Distraction Free Settings.Hide Channels Placeholder')"
         :show-action-button="true"
         :tag-list="channelsHidden"
+        :show-tags="showHiddenChannels"
         :tooltip="$t('Tooltips.Distraction Free Settings.Hide Channels')"
         :validate-tag-name="validateChannelId"
         :find-tag-info="findChannelTagInfo"
         :are-channel-tags="true"
-        :show-hidden-content="showHiddenChannels"
         @invalid-name="handleInvalidChannel"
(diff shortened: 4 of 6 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Errors when entering the settings page and using the toggles

https://github.com/user-attachments/assets/6617e03d-b447-44b3-a208-aaf146eab197

JL0000:
Fixed it.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
I think the label reads a bit confusing to me
`Show Hidden Channels` at first I thought it was disabling hiding channels then I read `Show Hidden Text`.
I think it would be better to label them as something like "show all already entered list of channels/text" (Not sure what exact words to use)
Or move the controls below each text input field (though not the best position I guess) so users can relate that control to the list of videos/texts

JL0000:
What do you think about `List Hidden Channels` and `List Hidden Words`?

PikachuEXE:
`Show Stored/Saved Hidden Channels` & `Show Stored/Saved Hidden Keywords`?
Weird to see `List`...

JL0000:
Because I name this section `Content Filter` I think `Show Channels Added to Content Filter` and `Show Keywords Added to Content Filter` works well. If content filter is vague, then `Show Stored Hidden Channels` and `Show Stored Hidden Keywords` works well too.

PikachuEXE:
`Show Added Channels` + `Show Added Keywords`?
Assuming they are under heading `Content Filter`

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

kommunarr (commented):
**suggestion:** I hate to be the guy who comes in late with new things to say. This feature does work well, but I am still skeptical of the toggle setup. Couldn't we do a Show / Hide button where the content is supposed to appear / disappear (e.g., like our recent Hide/Show description feature) instead of toggles above the whole container? That way the positional and logical relationship of the control to the corresponding section is much more clear. I ask particularly because I thought I was in the wrong branch for a couple minutes when trying to test this.

JL0000:
> **suggestion:** I hate to be the guy who comes in late with new things to say. This feature does work well, but I am still skeptical of the toggle setup. Couldn't we do a Show / Hide button where the content is supposed to appear / disappear (e.g., like our recent Hide/Show description feature) instead of toggles above the whole container? That way the positional and logical relationship of the control to the corresponding section is much more clear. I ask particularly because I thought I was in the wrong branch for a couple minutes when trying to test this.

I think it does make it more coherent. Maybe to the right of each tooltip?
Also, can you link me the PR for the "recent Hide/Show description feature" please?

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
I would like the `showTags` to be true by default so that this doesnt visually change anything for our users in the next version. If someone decided to uncheck it than that should be remembered within the session but also on next startup. This is how it works with all the toggles in the settings

Settings dont get remembered when navigating away and back to the settings page

https://github.com/user-attachments/assets/627d4b66-9576-4c8e-844f-3ac67188fd2e

src/renderer/components/ft-input-tags/ft-input-tags.vue
```diff
@@ -21,7 +21,26 @@
...
+      v-if="tagList.length >= 1"
+      class="checkbox-container"
+    >
+      <input
+        :id="label"
+        type="checkbox"
+        :checked="showTags"
+        @change="toggleShowTags"
+        @keydown.enter.prevent="toggleShowTags"
+        @keydown.space.prevent="toggleShowTags"
```
kommunarr, line 34: **suggestion:** These can be removed since you're using the base HTML component.

src/renderer/components/ft-input-tags/ft-input-tags.css
```diff
@@ -38,6 +38,12 @@
   user-select: text;
 }
 
+.checkbox-container {
+  display: flex;
+  align-items: center;
+  font-size: 12px;
+}
```
kommunarr, line 45:
```suggestion
.checkbox-container {
  display: flex;
  align-items: center;
  font-size: 12px;
  accent-color: var(--accent-color);
}
```

**suggestion:** Use app's accent color (the secondary theme color) for checkbox checked color for improved visual consistency.

kommunarr (commented):
This looks great!

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM!

kommunarr (approved):

src/renderer/store/modules/settings.js
```diff
@@ -296,6 +296,8 @@ const state = {
   screenshotFolderPath: '',
   screenshotFilenamePattern: '%Y%M%D-%H%N%S',
   settingsSectionSortEnabled: false,
+  showAddedChannels: true,
+  showAddedKeywords: true,
```
PikachuEXE, line 300: 1. These should be moved to below `forbiddenTitles` (coz they are related)
2. I think maybe using `showAddedChannelsHidden` & `showAddedForbiddenTitles` would be better to clearly show that they are related.

PikachuEXE (changes requested):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/components/ft-input-tags/ft-input-tags.vue
```diff
@@ -21,7 +21,24 @@
...
+      v-if="tagList.length >= 1"
+      class="checkbox-container"
+    >
+      <input
+        :id="label"
+        type="checkbox"
+        :checked="showTags"
+        @change="toggleShowTags"
+      >
+      <label :for="label">
```
absidue, line 34: Please find a way to make this id unique, as this component is shown on the page 3 times at the same time, so using the same ID for all of them will result in the wrong label getting referenced.

ChunkyProgrammer: This should be unique since it's taking the label prop but it should probably be somewhat sanitized (ex: spaces removed)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

kommunarr (approved):

PikachuEXE (approved):

Issue #3900, opened 2023-08-21
[Feature Request]: Reveal original title when using DeArrow


redactedscribe:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The new feature to use DeArrow for more accurate video titles is great, but it would be nice to have a way to see whether the title is the original or not.

### Proposed Solution

For example, when a video has had its title changed, display a small icon, or style the title text differently in some way. Then by hovering over the title or perhaps clicking some icon prepended to the title, be able to see what the original title is. The goal would be to be able to easily recognise an unoriginal (DeArrowed) title just by looking at it.

### Alternatives Considered

We can just click a video to see its original title, but there is no way to know the title in the feed is original or not. Secondly, one has to remove the video from the watch history if you were simply curious about the original title (and not watching the video).

### Issue Labels

display more information to user, improvement to existing feature

### Additional Information

Currently, the DeArrow title is only shown in the feeds, not on the actual video page itself, which I find to be a bit confusing: You click one thing and a "different" video shows up. From my perspective, it'd be better if the same title was displayed everywhere, but instead indicate that when titles have been DeArrowed. The original title would still be used under-the-hood (if it gets used for anything, I don't know).

Thanks!

kommunarr:
I'd be interested to see if there's another app that integrates with DeArrow and has a similar feature. I can see the appeal, but many of the possible implementations of this could be a downgrade to the user experience if not done right.

efb4f5ff-1298-471a-8973-3d47447115dc:
This is how DeArrow does it in the browser. We could do the same to match the UX.


https://github.com/FreeTubeApp/FreeTube/assets/73130443/c11611c9-a240-45d9-84d7-fa56febd8e4e

PR #6164, opened 2024-11-14
Added Button to Show Original Title and Thumbnail for DeArrow

JL0000:
# Added Button to Show Original Title and Thumbnail for DeArrow

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

#3900

## Description

If DeArrow is enabled for title or/and thumbnail, when hovering over the video info area, a button is added next to the video title when the video is part of a list. This button switches the title or/and thumbnails back to its original version and vice versa.

I have reduced the padding for the ellipsis button as with both it and the new DeArrow button in the grid view, it takes away too much title space.

The smaller button suggests that the title content did not change (case insensitive).

The button cannot be pinned if DeArrow doesn't change anything.

## Video 

https://github.com/user-attachments/assets/61abcf2c-d1b3-42ae-b0be-fa1e667c2733

## Desktop

- **OS:** MacOS
- **OS Version:** 15.1


PikachuEXE:
Even though it feels kind of working for desktop
Reducing padding for buttons is bad for mobile users (touch screen users more precisely)
Functionality works but UI needs to be improved but I have no idea what would be good

absidue:
I agree with @PikachuEXE here, maybe it would be better suited inside the dropdown?

JL0000:
Didn't know it was available on mobile. What's the best way to test on mobile?

JL0000:
@absidue I think putting it in the dropdown menu would make it too cumbersome to use this feature.

absidue:
If you don't want to move it to the dropdown where the rest of the options are then you'll have to stick with your current solution but undo the change to shrink the buttons.

efb4f5ff-1298-471a-8973-3d47447115dc:
I think that we shouldn't put this in the dropdown and keep this as is.

Some differences im seeing compared to https://github.com/FreeTubeApp/FreeTube/issues/3900#issuecomment-2220157513

title moves around on hover

https://github.com/user-attachments/assets/682353a0-5f84-490f-8f85-68cd3dcbdb0e


button has different sizes 

https://github.com/user-attachments/assets/aa0f7f1e-01bd-4eab-8ca9-e3c1ca0ad02a

JL0000:
@efb4f5ff-1298-471a-8973-3d47447115dc both your concerns are actually present in the DeArrow extension but they're not shown in https://github.com/FreeTubeApp/FreeTube/issues/3900#issuecomment-2220157513

https://github.com/user-attachments/assets/f5e4dda9-e18d-40c4-96b4-36feb6f1ea28

JL0000:
I have changed back the button padding to its original size.

Thinking of adding a new setting in the DeArrow setting to enable this toggle to see original title/thumbnail.

Also I noticed a small issue at the moment:
- If the button cannot be fixed and you click it repeatedly very quickly, it can get fixed (working on this)

JL0000:
- The dot is now enabled if either the title or thumbnail has been modified. It will be in the primary colour.
- If not enabled, it is smaller and in black
- The title hovering has now been fixed.

https://github.com/user-attachments/assets/fbd97608-4769-476c-a39d-93c18a50eb06



- The toggle feature is in the settings. It is off by default.
Screenshot 2024-11-22 at 16 33 00

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -206,6 +214,21 @@
           {{ $t('Search Listing.Label.Subtitles') }}
         </div>
       </div>
+
+      <ft-icon-button
+        v-if="(showDeArrowToggle && isHovered) || deArrowTogglePinned"
+        class="deArrowToggleButton"
+        :class="{alwaysVisible: deArrowTogglePinned}"
+        :icon="['fas', 'dot-circle']"
+        :color="deArrowChangedContent ? 'var(--primary-color)' : '#808080'"
```
absidue, line 223: Please use the existing theme variables, don't hardcode colours, as it has to be usable on all themes. Additionally the ft-icon-button component already has a `theme` for changing the colours of it, please use that instead of introducing another way to do the same thing.

src/renderer/components/sponsor-block-settings/sponsor-block-settings.js
```diff
@@ -46,6 +46,12 @@ export default defineComponent({
     useDeArrowThumbnails: function () {
       return this.$store.getters.getUseDeArrowThumbnails
     },
+    deArrowShowOriginal: function () {
+      return this.$store.getters.getDeArrowShowOriginal
+    },
+    deArrowShowOriginalAlwaysOn: function () {
+      return this.$store.getters.getDeArrowShowOriginalAlwaysOn
+    },
```
absidue, line 54: It looks like this value doesn't exists in the store and is also not used in the component. This is presumably a leftover from previous iterations of this pull request?

src/renderer/components/ft-list-video/ft-list-video.vue
JL0000: I tried using the primary theme in `ft-icon-button` but the colours are reversed. I don't think I should create a new theme or change the current one. Not sure how to proceed.

**ft-icon-button.scss**
```
&.primary {
    background-color: var(--primary-color);
    color: var(--text-with-main-color);
```
**Current**
image
**If I reverse `color` with `background-color`**
image

src/renderer/components/sponsor-block-settings/sponsor-block-settings.js
JL0000: Yeah that is leftover, my bad

efb4f5ff-1298-471a-8973-3d47447115dc:
- I would like the `Show "Show Original" Button` to be greyed out when the DeArrow toggles are disabled

- Doesnt change to grey for me

https://github.com/user-attachments/assets/b959c0d5-b0a9-4798-8164-c973fa7d9cd8

src/renderer/components/ft-list-video/ft-list-video.vue
kommunarr: One fix would be to use the `--favorite-icon-color` as the "enabled" mode color, as that's one we already have set as a universal "other" state. You could also do `filter: grayscale(1);` for the disabled state and the primary/secondary icon color for the enabled state, which would match better with the existing DeArrow web UI.

kommunarr:
**question:** Why is there a "Show Original" toggle that's disabled by default? Is this degree of optionality important?

kommunarr:
**issue (blocking):** it looks like the "not applicable" icon is smaller than the "enabled" one.

**suggestion:** I'd recommend not even showing a "not applicable" icon if the video does not have a DeArrow version. It doesn't seem to provide any utility and is a bit confusing.

kommunarr:
**suggestion:** I'd prefer if the icon did not displace the title and instead showed up quietly like the "Show options" button.

JL0000:
> **question:** Why is there a "Show Original" toggle that's disabled by default? Is this degree of optionality important?

I implemented this for two reasons:
1. Button takes space away from the title space. Some people may find this annoying for long titles
image

2. When the button is rendered, it may increase the hight of the video item box in grid mode. This will shift all the videos in the subsequent rows. Again people may not like this.

> [**suggestion:** I'd recommend not even showing a "not applicable" icon if the video does not have a DeArrow version. It doesn't seem to provide any utility and is a bit confusing.]

That sound like a good idea. I was following the DeArrow plugin implementation

> [**suggestion:** I'd prefer if the icon did not displace the title and instead showed up quietly like the "Show options" button.]

So render it but not show it? This will take away title space though.

kommunarr:
> So render it but not show it? This will take away title space though.

Not if you show it in a column with the "More Options" button right beneath it.

src/renderer/components/ft-list-video/ft-list-video.vue
JL0000: In the end I just ended up using `<font-awesome-icon>` and creating a custom css class for the button because the main and hover colours are different.

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -206,17 +206,33 @@
...
+        <font-awesome-icon
+          v-if="showDeArrowToggle || deArrowTogglePinned"
+          :title="deArrowToggleTitle"
+          :icon="['far', 'dot-circle']"
+          class="optionsButton deArrowToggleButton"
+          :class="{ alwaysVisible: deArrowTogglePinned }"
+          :style="{
+            padding: 10 + 'px',
+            fontSize: 16 + 'px'
+          }"
```
absidue, line 219: As these values are static/don't change, please move them to the `deArrowToggleButton` class in the SCSS file.

JL0000:
Thank you for the feedback! After some rewriting I have done/fixed:
- Toggle has been changed to look more similar to the official DeArrow logo
- Toggle does not appear if neither title or thumbnail are changed
- Toggle changes to grey colour after being pinned/activated
- Toggle has been moved to sit above the "More Options" button so it does not displace the title
- The`Show "Show Original" Button` setting is disabled if all DeArrow toggles are disabled. Whether this setting should exist, I am not sure.

https://github.com/user-attachments/assets/7ff6f422-0e8b-492a-a010-b82cb7bf2854

kommunarr:
Thanks so much for your work on this!

My bad, I misspoke earlier - I'd prefer if it was below the More Options button, that way it's minimal difference of the placement of things on a video with the icon versus without it. Really sorry about that!

>The Show "Show Original" Button setting is disabled if all DeArrow toggles are disabled. Whether this setting should exist, I am not sure.

With the above in, there should now be very little difference to the user if the button is present or not on hover - in which case I recommend removing the option, to make the logic simpler and reduce # of settings.

Haven't looked at the code yet, but with the above two changes, I'm very happy with this PR's functionality

static/locales/en-US.yaml
```diff
@@ -830,6 +830,8 @@ Video:
   Streamed on: Streamed on
   Started streaming on: Started streaming on
   Sponsor Block category:
+    Show Original Details: Show Original Details
+    Show Modified Details: Show Modified Details
```
absidue, line 834: Could you maybe move these strings to `Video.DeArrow` or something like that instead, as they aren't SponsorBlock segment categories. Sorry that I didn't notice this earlier, as I could have suggested it with my other code suggestions and saved you time 🫣.

JL0000: No worries

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -206,17 +206,29 @@
...
+        />
+        <font-awesome-icon
+          v-if="deArrowChangedContent || deArrowTogglePinned"
+          :title="deArrowToggleTitle"
+          :icon="['far', 'dot-circle']"
+          class="optionsButton deArrowToggleButton"
+          :class="{ alwaysVisible: deArrowTogglePinned }"
+          tabindex="0"
+          role="button"
+          @click="toggleDeArrow"
```
kommunarr, line 229:
```suggestion
          @click="toggleDeArrow"
          @keydown.enter.prevent="toggleDeArrow"
          @keydown.space.prevent="toggleDeArrow"
```

**suggestion (blocking):** Add space / Enter key handling to be accessible for users using keyboard navigation

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -408,7 +412,7 @@ export default defineComponent({
     },
 
     displayDuration: function () {
-      if (this.useDeArrowTitles && (this.duration === '' || this.duration === '0:00') && this.deArrowCache?.videoDuration) {
+      if (this.showDeArrowTitle && (this.duration === '' || this.duration === '0:00') && this.deArrowCache?.videoDuration) {
         return formatDurationAsTimestamp(this.deArrowCache.videoDuration)
       }
       return this.duration
```
absidue, line 418: With this change the duration would vanish when you click on the `Show Original Details` button, which is undesirable. This change should be reverted to use the DeArrow setting, as this code uses the video duration from the DeArrow API response if it is otherwise unknown (e.g. unwatched videos on the shorts tab on the subscriptions page).

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -559,14 +570,31 @@ export default defineComponent({
...
     },
+    toggleDeArrow() {
+      if (!this.deArrowChangedContent) {
+        return
+      }
+
+      this.deArrowTogglePinned = !this.deArrowTogglePinned
+      this.deArrowToggleTitle = this.deArrowTogglePinned
+        ? this.$t('Video.DeArrow.Show Modified Details')
+        : this.$t('Video.DeArrow.Show Original Details')
```
absidue, line 589: Please create a computed prop for this instead, that way when the user changes the language the text will update automatically, whereas at the moment you would have to manually click the button for it to switch to the correct language. The use of translations strings in FreeTube should be reactive wherever possible.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM! Really nicely put together I hope we will see more wonderful contributions from you :)

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Sorry for this :(

Noticed this one when spamming the DeArrow icon. Characters from the video next to it will highlight when spamming the icon.


https://github.com/user-attachments/assets/e9fec810-b7a5-4b9e-8bbf-7995b4b7584a

PikachuEXE:
That's just same behaviour on other text (and other non button places)
Unless it doesn't happen in dev

https://github.com/user-attachments/assets/a0ac1a98-4806-49ae-998a-bbb78bcf10be

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm im a bit confused as it doesnt happen here


https://github.com/user-attachments/assets/43446ee0-af67-4d1e-96a8-449b25084740

PikachuEXE:
I confirm mouse clicks would select nearby test for some reason in dev
Not this PR specific

JL0000:
I think this happens because of `v-if` and if you click on the button again too quickly, it clicks on the background which selects the text?

kommunarr (approved):
LGTM. Above issue appears to be pretty minor and can be addressed separately if there's no apparent clean fix at the moment.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
This minor issue isnt related to this PR so, LGTM!

Thank you for putting so much time and effort into this. I hope we will see more nice contributions from you :)

PR #6180, opened 2024-11-17
Better player error handling

### LINKED ISSUE #6028: [Bug]: Windows freeze when Tor service restarts
### Guidelines

### Describe the bug

1. Use Local API and Tor proxy
2. Restart Tor service while watching a video
3. Windows freeze and must be force-closed, can't open console or click anything

### Expected Behavior

Video should just stutter or rebuffer at most

### Issue Labels

causes crash

### FreeTube Version

0.22.0

### Operating System Version

Debian 12

### Installation Method

Flathub

### Primary API used

Local API

0.21.0

### Additional Information

_No response_

### Nightly Build

absidue:
# Better player error handling

## Pull Request Type

- [x] Bugfix

## Related issue

closes #6028

## Description
This pull request makes various changes to the player error handling.

1. When an error happens that we consider to be critical enough to trigger a format fallback, we now ignore any other errors that occur because of the previous format, as the legacy formats can throw multiple errors in a row, which ends up interrupting the fallback to the audio formats, which then throws an error because it got interrupted and we end up in an extremely quick loop piling up thousands of errors in just a few seconds as it attempts to fallback and keeps interrupting itself. This results in it being impossible to do anything with the window as it is stuck in that endless loop.
2. We now fully unload the previous format during a format switch before loading the next one, this helps with number 1 as we can separate which errors happened when and helps slow down the fallbacks and errors, giving the user more time to leave the page if they do end up in an error loop
3. We now ignore errors when destroying the player, while I haven't found any direct problems caused by that, it's just a precaution to hopefully allow the user to always leave the watch page.
4. When an error happens in the response filter, shaka-player throws a RESPONSE_FILTER_ERROR which is a wrapper around the original error, previously we weren't unwrapping the nested errors so we didn't actually log or look at the original error.
  This can happen if the user is using a blocked IP and uses an Invidious instance far away from them without having `Proxy Videos Through Invidious` enabled. In that scenario YouTube will redirect you multiple times to give you a googlevideo server that is closer to your IP address and the last redirect will get a 403 from YouTube because of your blocked IP, this then results in RESPONSE_FILTER_ERROR  -> RESPONSE_FILTER_ERROR -> RESPONSE_FILTER_ERROR  -> BAD_HTTP_STATUS error chain. We have handling for BAD_HTTP_STATUS errors but we weren't getting to that before because we weren't looking at the original error.

## Testing

1. Check that changing formats still works
2. If you have access to a blocked IP address: Use the Invidious API with Proxy Videos Through Invidious enabled and check that the 403 IP block message shows up

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** f9736fb11af9a1015803d5e1bb3505d45510ca22

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -2147,17 +2147,28 @@ export default defineComponent({
 
     // #endregion keyboard shortcuts
 
+    let ignoreErrors = false
+
     /**
      * @param {shaka.util.Error} error
      * @param {string} context
      * @param {object?} details
      */
     function handleError(error, context, details) {
+      // These two errors are just wrappers around another error, so use the original error instead
+      // As they can be nested (e.g. multiple googlevideo redirects because the Invidious server was far away from the user) we should pick the inner most one
+      while (error.code === shaka.util.Error.Code.REQUEST_FILTER_ERROR || error.code === shaka.util.Error.Code.RESPONSE_FILTER_ERROR) {
+        error = error.data[0]
+      }
+
       logShakaError(error, context, props.videoId, details)
 
       // text related errors aren't serious (captions and seek bar thumbnails), so we should just log them
       // TODO: consider only emitting when the severity is crititcal?
-      if (error.category !== shaka.util.Error.Category.TEXT) {
+      if (!ignoreErrors && error.category !== shaka.util.Error.Category.TEXT) {
+        // don't react to multiple consecutive errors, otherwise we don't give the format fallback from the previous error a chance to work
+        ignoreErrors = true
+
         emit('error', error)
 
         stopPowerSaveBlocker()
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
> 1. Check that changing formats still works

Error :( 
 
https://github.com/user-attachments/assets/f8be7911-fada-44cc-a547-00fea7910648


> 2. If you have access to a blocked IP address: Use the Invidious API with Proxy Videos Through Invidious disabled and check that the 403 IP block message shows up, instead of it switching formats.

Looks to be working correctly 

https://github.com/user-attachments/assets/e86577b6-16a2-4132-85c7-3c4bb8cd4487

absidue:
Should be fixed now.

efb4f5ff-1298-471a-8973-3d47447115dc:
still doesnt work :(

https://github.com/user-attachments/assets/ea5710ec-250a-4049-99ed-1f848202f305

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

PR #6221, opened 2024-11-24
Use ft-input-tags for external player custom args

absidue:
# Use ft-input-tags for external player custom args

## Pull Request Type

- [x] Feature Implementation

## Description
It's not uncommon for users to struggle to configure the external player settings correctly. One of the known stumbling blocks is that people don't read the tooltips, so don't realise that they have to separate the custom arguments with a semicolon (`;`). This pull request changes the custom arguments to use the ft-input-tags component, that way users can enter each argument separately, which should hopefully make it easier to configure the custom arguments correctly. Interally we now store the args in a JSON array.

I updated the tooltip text to remove the ` separated by semicolons (';'),` part in most languages where it was obvious which part of the text it is, in other languages where it was less obvious to me (as I don't know many of FreeTube's languages) I left it to be changed by the translators. At least some of them will have less work to do :).

## Screenshots
external-player-custom-args

## Testing
So that you don't have to install various external players, I would suggest editing the `openInExternalPlayer` function in `src/renderer/store/modules/utils.js` to log what it would send to the external player instead of launching it e.g. 

```diff
-    if (process.env.IS_ELECTRON) {
-      const { ipcRenderer } = require('electron')
-      ipcRenderer.send(IpcChannels.OPEN_IN_EXTERNAL_PLAYER, { executable, args })
-    }
+    console.log(executable, args)
```

### Testing the migration
Check that the migration of the old args string to the new JSON string works correctly

1. While FreeTube is not running open your `settings.db` file (data location but in the `Electron` instead of `FreeTube` folder)
2. If you didn't have any args configured previously add this line, otherwise edit the existing one: `{"_id":"externalPlayerCustomArgs","value":"--test1;--test2"}`
3. Run `yarn dev` on this branch
4. Check in the external player settings that the two args show up as two separate tags

### Other tests
1. Set some custom args
2. Click on the open in external player button on the watch page or in a video list, in the console your custom args should be correctly mixed in with the one that FreeTube sets

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** a3823c54ab970614c119eabe57446090c091386f


static/locales/en-US.yaml
```diff
@@ -1017,7 +1017,7 @@ Tooltips:
     Ignore Default Arguments: Do not send any default arguments to the external player
       aside from the video URL (e.g. playback rate, playlist URL, etc.).
       Custom arguments will still be passed on.
-    Custom External Player Arguments: Any custom command line arguments, separated by semicolons (';'),
+    Custom External Player Arguments: Any custom command line arguments,
       you want to be passed to the external player.
```
kommunarr, line 1021:
```suggestion
    Custom External Player Arguments: Any custom command line arguments you want to be passed to the external player.
```

static/locales/en-US.yaml
```diff
@@ -1017,8 +1017,7 @@ Tooltips:
     Ignore Default Arguments: Do not send any default arguments to the external player
       aside from the video URL (e.g. playback rate, playlist URL, etc.).
       Custom arguments will still be passed on.
-    Custom External Player Arguments: Any custom command line arguments, separated by semicolons (';'),
-      you want to be passed to the external player.
+    Custom External Player Arguments: Any custom command line arguments you want to be passed to the external player.
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 1020: Sound better in my head

```suggestion
    Custom External Player Arguments: Any custom command line arguments you want to be passed on to the external player.
```

absidue: Done :)

static/external-player-map.json
```diff
@@ -43,7 +43,7 @@
         "value": "iina",
         "cmdArguments": {
             "defaultExecutable": "iina",
-            "defaultCustomArguments": "--no-stdin",
+            "defaultCustomArguments": ["--no-stdin"],
```
ChunkyProgrammer, line 46: Havent tested but since this change is made, I think this line needs to be updated: https://github.com/FreeTubeApp/FreeTube/blob/3f430c0b0dfca3a09e7478fa8cce1e921b72a518/src/renderer/components/external-player-settings/external-player-settings.js#L55

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
I never got external player working so did not test arguments applied

PR #6222, opened 2024-11-24
Switch the data import to the Web File System API

absidue:
# Switch the data import to the Web File System API

## Pull Request Type

- [x] Refactoring

## Description

Currently the data import uses the Electron dialog API and the Node.js `fs` API in the renderer to import user data, this pull request switches it over to the Web File System API. That lets us get rid of one more use of a Node.js API in the renderer, which gets us one step closer to turning off `nodeIntegration` in the renderer, this time without needing to build an IPC call as a replacement.

I haven't removed the `SHOW_OPEN_DIALOG` IPC call yet, as that is currently still used for the default download and screenshot directory pickers.

## Testing

1. `yarn dev`
2. Test importing your history, subscriptions and playlists
3. `yarn dev:web`
4. Open the dev server URL in a web browser that doesn't support the Web File System API e.g. Firefox or Safari or if you have access to a FreeTubeAndroid compatible device you can test this pull request with that too.
5. Test importing your history, subscriptions and playlists

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 742cd22c315eacb21543eadcdb453bd5dbe7ebea

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -12,12 +12,14 @@ import {
   deepCopy,
   escapeHTML,
   getTodayDateStrLocalTimezone,
-  readFileFromDialog,
-  showOpenDialog,
+  readFileWithPicker,
   showSaveDialog,
   showToast,
   writeFileFromDialog,
 } from '../../helpers/utils'
+
+const IMPORT_DIRECTORY_ID = 'data-settings-import'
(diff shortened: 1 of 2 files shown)


PikachuEXE (approved):
Tested web one w/ Firefox 132.0.2

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PR #6242, opened 2024-11-29
Add setting for opening freetube:// links in new window

### LINKED ISSUE #1501: Add option to always open a redirected video in a new window / instance

OothecaPickle:
# Add setting for opening freetube:// links in new window

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #1501

## Description

Adds a toggle under General Settings to open `freetube://` links in a new window instead of in the replacing the contents of an existing window.

## Screenshots 

openlinksinnewwindow

## Testing 

Test that freetube:// links opened from other applications are opened in a new window when the added setting is toggled on.

## Desktop

- **OS:** macOS
- **OS Version:** 10.15.7
- **FreeTube version:** e646a0f2befebc6373c527a9a0b23086f485e82c

## Additional context

I used some code from @pakoito's #5592.

### DIFF
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -142,6 +142,10 @@ export default defineComponent({
 
     externalLinkHandling: function () {
       return this.$store.getters.getExternalLinkHandling
+    },
+
+    openLinksInNewWindow: function () {
+      return this.$store.getters.getOpenLinksInNewWindow
     }
   },
   watch: {
@@ -514,7 +518,7 @@ export default defineComponent({
     enableOpenUrl: function () {
       ipcRenderer.on(IpcChannels.OPEN_URL, (event, url) => {
         if (url) {
-          this.handleYoutubeLink(url, { doCreateNewWindow: true })
+          this.handleYoutubeLink(url, { doCreateNewWindow: this.openLinksInNewWindow })
         }
       })
 
--- a/src/renderer/App.js
+++ b/src/renderer/App.js
@@ -514,7 +514,7 @@ export default defineComponent({
     enableOpenUrl: function () {
       ipcRenderer.on(IpcChannels.OPEN_URL, (event, url) => {
         if (url) {
-          this.handleYoutubeLink(url)
+          this.handleYoutubeLink(url, { doCreateNewWindow: true })
         }
       })
 
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -232,6 +232,7 @@ const state = {
   listType: 'grid',
   maxVideoPlaybackRate: 3,
   onlyShowLatestFromChannel: false,
+  openLinksInNewWindow: false,
   playNextVideo: false,
   proxyHostname: '127.0.0.1',
   proxyPort: '9050',
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -283,6 +283,7 @@ Settings:
     Fallback to Non-Preferred Backend on Failure: Fallback to Non-Preferred Backend
       on Failure
     Enable Search Suggestions: Enable Search Suggestions
+    Open Links In New Window: Open freetube:// Links in a New Window
     Auto Load Next Page:
       Label: Auto Load Next Page
       Tooltip: Load additional pages and comments automatically.
--- a/src/renderer/components/general-settings/general-settings.js
+++ b/src/renderer/components/general-settings/general-settings.js
@@ -200,6 +200,10 @@ export default defineComponent({
         this.$t('Settings.General Settings.External Link Handling.Ask Before Opening Link'),
         this.$t('Settings.General Settings.External Link Handling.No Action')
       ]
+    },
+
+    openLinksInNewWindow: function () {
+      return this.$store.getters.getOpenLinksInNewWindow
     }
   },
   created: function () {
@@ -269,6 +273,7 @@ export default defineComponent({
       'updateCurrentLocale',
       'updateExternalLinkHandling',
       'updateGeneralAutoLoadMorePaginatedItemsEnabled',
+      'updateOpenLinksInNewWindow',
     ])
   }
 })
--- a/src/renderer/components/general-settings/general-settings.vue
+++ b/src/renderer/components/general-settings/general-settings.vue
@@ -38,6 +38,12 @@
           :compact="true"
           @change="updateEnableSearchSuggestions"
         />
+        <ft-toggle-switch
+          :label="$t('Settings.General Settings.Open Links In New Window')"
+          :default-value="openLinksInNewWindow"
+          :compact="true"
+          @change="updateOpenLinksInNewWindow"
+        />
       </div>
     </div>
     <div class="switchGrid">


PikachuEXE:
Are you talking about https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app#overview ?
Maybe the preference name should be updated first

Also I am not sure how/if the doc mentions deep links and if users understand what is `freetube://` links

OothecaPickle:
@PikachuEXE 
> Are you talking about https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app#overview ?
Yes.

>Maybe the preference name should be updated first
> Also I am not sure how/if the doc mentions deep links and if users understand what is `freetube://` links
I agree, but I wasn't sure what else to name it. Do you have any ideas?

src/renderer/App.js
```diff
@@ -514,7 +518,7 @@ export default defineComponent({
     enableOpenUrl: function () {
       ipcRenderer.on(IpcChannels.OPEN_URL, (event, url) => {
         if (url) {
-          this.handleYoutubeLink(url)
+          this.handleYoutubeLink(url, { doCreateNewWindow: this.openLinksInNewWindow })
```
pakoito, line 521: It was so simple, just right there. Thanks!

PikachuEXE:
I found one place mentioning extensions but not sure how to call it
https://docs.freetubeapp.io/usage/browser-extension/

It shouldn't be called "YouTube links" on the preference coz there are many "YT" links in video description/comment etc.

Inviting more members to comment on this

pakoito:
In Android land these are called "Intents" rather than "Links". But tbh I'd be sad if this feature is blocked on a label :D

src/main/index.js
```diff
@@ -260,7 +260,7 @@ function runApp() {
 
         const url = getLinkUrl(commandLine)
         if (url) {
-          mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+          mainWindow.webContents.send(IpcChannels.OPEN_URL, { url: url, isLaunchLink: false })
```
absidue, line 263: Please use two separate parameters instead of using a single parameter with an object, that lets you avoid the destructuring on the other end.

OothecaPickle:
Sorry to bother @pakoito, didn't mean to request another review from you.

static/locales/en-US.yaml
```diff
@@ -283,6 +283,7 @@ Settings:
     Fallback to Non-Preferred Backend on Failure: Fallback to Non-Preferred Backend
       on Failure
     Enable Search Suggestions: Enable Search Suggestions
+    Open Links In New Window: Open freetube:// Links in a New Window
```
OothecaPickle, line 286:
```suggestion
    Open Links In New Window: Open Deep Links Into FreeTube in a New Window
```

OothecaPickle: or maybe 'URLs Passed to FreeTube" as this setting also applies to links passed via the command line

efb4f5ff-1298-471a-8973-3d47447115dc: Maybe we should just implement an tooltip message that explains what this does?

efb4f5ff-1298-471a-8973-3d47447115dc:
could someone provide a build for me to test this one :grimacing:

OothecaPickle:
> could someone provide a build for me to test this one 😬

@efb4f5ff-1298-471a-8973-3d47447115dc https://github.com/OothecaPickle/FreeTube/actions/runs/12128622135

efb4f5ff-1298-471a-8973-3d47447115dc:
PR LGTM, only thing that should be looked at is the wording of the setting

src/renderer/store/modules/settings.js
```diff
@@ -232,6 +232,7 @@ const state = {
   listType: 'grid',
   maxVideoPlaybackRate: 3,
   onlyShowLatestFromChannel: false,
+  openLinksInNewWindow: false,
```
PikachuEXE, line 235: This is not for any link shown inside FT
[Electron doc calls it deep link](https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app#overview) if I read that right
```suggestion
  openDeepLinksInNewWindow: false,
```

static/locales/en-US.yaml
PikachuEXE: Or just add tooltip giving examples of redirect extensions (mostly from those right?
Setting `External Link Handling` has one

PikachuEXE (changes requested):

PikachuEXE:
You committed when I am about to comment on it lol

PikachuEXE (approved):
Did not test actual link open on mac but someone else tested it already

static/locales/en-US.yaml
```diff
@@ -1007,6 +1008,9 @@ Tooltips:
     External Link Handling: |
       Choose the default behavior when a link, which cannot be opened in FreeTube, is clicked.
       By default FreeTube will open the clicked link in your default browser.
+    Open Deep Links In New Window: Choose whether URLs passed to FreeTube, such as by
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 1011: Wording makes me think that i can choose the desired behavior if i enable this setting. 

Change it to:

URLs passed to FreeTube, such as by redirect browser extensions or command line arguments, are opened in a new window.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Tooltip wording change

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PikachuEXE (approved):

static/locales/en-US.yaml
pakoito: BTW, I wrote a redirect that doesn't require a browser extension: https://pakoito.github.io/freetube-forwarder/

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @OothecaPickle i noticed some undesired behavior and i would like to know if it possible for you to create a followup PR to fixed these. I noticed this behavior in while testing another PR https://github.com/FreeTubeApp/FreeTube/pull/6915#pullrequestreview-2818058051

- If one instance of FT is minimized LibRedirect will open the video in a new window AND will also unminimize the previously minimized window
- If two instances of FT are minimized LibRedirect will open the video in a new window AND will ONLY unminimize the first instance that was previously minimized. So the Second instance that was minimized will stay minimized but the first instance wont.

PR #6244, opened 2024-12-01
Filter out unsupported watch page only mixes from search results

absidue:
# Filter out unsupported watch page only mixes from search results

## Pull Request Type

- [x] Bugfix

## Related issue
closes #6196

## Description

Watch page only mixes are not supported in FreeTube as they only exist on the YouTube watch page, that means they don't have their own playlist pages. Those unsupported mix playlists used to be filtered out automatically as they had their own `Mix` node which we didn't parse, but now YouTube has switched them to the same `LockupView` with `content_type = 'PLAYLIST'` that they use for normal playlists.

Some mixes such as the ones on `https://youtube.com/music` have actual playlist pages, they can be identified by the fact that their playlist IDs start with `RDCL`, which this pull request explicitly checks for to make sure that those types don't get filtered out.

## Testing

Search for `music` and check that the mix playlist doesn't show up in the local API search results.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -849,6 +849,7 @@ function handleSearchResponse(response) {
       return item.type === 'Video' || item.type === 'Channel' || item.type === 'Playlist' || item.type === 'HashtagTile' || item.type === 'Movie' || item.type === 'LockupView'
     })
     .map((item) => parseListItem(item))
+    .filter((item) => item)
 
   return {
     results,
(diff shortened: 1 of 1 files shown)


PikachuEXE:
What is mixes... (I never use Youtube unless leaving comments which is rare)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

kommunarr (approved):

PR #6245, opened 2024-12-01
Migrate ThemeSettings and FtSponsorBlockCategory to the composition API

absidue:
# Migrate ThemeSettings and FtSponsorBlockCategory to the composition API

## Pull Request Type

## Description

This pull request migrates the ThemeSettings and FtSponsorBlockCategory components to the composition API. I introduced a new `useColorTranslations` composable that returns a computed array to replace the existing `getColorTranslations` helper function. The benefits are that it should now be more obvious that the translations will update automatically when the language changes and it is less verbose as you only have to write `t()` on each line instead of `i18n.t()` or `i18n.global.t()` in Vue 3.

### ThemeSettings

Please test that changing the various settings works as expected. For most of them you don't even have to leave the settings page to see if they work.

### FtSponsorBlockCategory

Check that changing the SponsorBlock category colors and skip options and work as expected. **You DO NOT have to test the entire SponsorBlock settings section; only the categories were affected by this pull request**

Video with some SponsorBlock segments: https://youtu.be/ESyiQ0tfqFQ

## Desktop

- **OS:** Windows
- **OS Version:** 10


src/renderer/components/ThemeSettings.vue
PikachuEXE, line 1: Any reason to put it outside a folder unlike `ParentalControlSettings`?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/ThemeSettings.vue
absidue: In my mind it doesn't make sense to keep a folder for just one file, especially when it is named the same (a simpler file structure should be a bit easier to navigate and result in better editor and linter performance as they have less things to keep track of/scan). I forgot to move `ParentalControlSettings` out of its folder but I'll create a pull request to move that and the `Post` component out of their folders.

If we were doing SFCs properly, we would be placing the CSS and SCSS files inside the Vue file too, but that requires extra webpack config that we haven't set up and some of the files are already quite long and adding the CSS would make them even longer (Unlike with `<script setup>` which also changes how the actual code works, placing the CSS in the same file just gives you a slightly better editor experience as it can auto-complete the class names and let you jump to them).

kommunarr (approved):

PikachuEXE (approved):

PR #6248, opened 2024-12-01
Small ESLint cleanup

absidue:
# Small ESLint cleanup

## Pull Request Type

- [x] Other

## Description

This pull request does various changes to our ESLint setup to clean it up a bit:
1. Remove the unused `eslint-plugin-import` dependency
2. Remove our explicit disabling of the `vuejs-accessibility/no-onchange` rule, as it is disabled by default/not listed in vuejs-accessibility's recommended rules configuration ([vue-a11y/eslint-plugin-vuejs-accessibility `src/configs/flat/recommended.ts`](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/configs/flat/recommended.ts) -> [vue-a11y/eslint-plugin-vuejs-accessibility `src/configs/rules.ts`](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/configs/rules.ts)).
3. Parse the commonjs scripts as commonjs, as ESLint 9 defaults to parsing files as ES modules
4. Import the `neostandard` function directly from the file that it is defined in in neostandard, as the default neostandard import also loads all of the dependencies for the `resolveIgnoresFromGitignore` function that we don't use.

## Testing

Run `yarn eslint-lint`

## Desktop

- **OS:** Windows
- **OS Version:** 10


eslint.config.mjs
```diff
@@ -9,7 +9,10 @@ import jsoncEslintParser from 'jsonc-eslint-parser'
 import eslintPluginJsonc from 'eslint-plugin-jsonc'
 import eslintPluginYml from 'eslint-plugin-yml'
 import yamlEslintParser from 'yaml-eslint-parser'
-import neostandard from 'neostandard'
+// Faster than importing the default import,
+// because the default import imports a lot of other dependencies
+// for the `resolveIgnoresFromGitignore` function that we don't use
+import { neostandard } from 'neostandard/lib/main.js'
```
PikachuEXE, line 15: Extension required?
```suggestion
import { neostandard } from 'neostandard/lib/main'
```

absidue: It seemed to work without it for me, not sure why though as technically the way I wrote it is invalid for ES Module imports, maybe ESLint is doing something weird with the config file (I know they use [jiti](https://github.com/unjs/jiti) on the config file to support TypeScript, maybe that tolerates extensionless imports).

I've added your change now.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue:
ES Modules require passing a real path including the extension, as ES Modules also work in web browsers and making various network requests to guess the real filename is a lot more expensive than Node.js doing a few file system calls with CommonJS modules to guess the file you meant.

As webpack existed before ES Modules did, they had to implement the guessing system needed for CommonJS imports, they apply it to all imports. That is why we are able to use extension-less imports in our app code.

ChunkyProgrammer:
This is why eslint didnt error about the bad import btw: https://github.com/FreeTubeApp/FreeTube/blob/f158611e3e7515a82dab7908f6ff36e06e1e2988/eslint.config.mjs#L26

There were some issues with json import causing eslint to error in the config file so i ignored the eslint config

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

ChunkyProgrammer (approved):

Issue #5975, opened 2024-10-27
[Bug]: Certain Videos glitch out into a white screen


liammcsh2008:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Go to a certain video
2. Click play
3. FreeTube Glitches out and video gets a white screen

### Expected Behavior

Video should play normally

### Issue Labels

causes crash, content not loading

### FreeTube Version

v0.22.0

### Operating System Version

Windows 11 version 23H2

### Installation Method

.exe

### Primary API used

Local API

### Additional Information

Result of new video player update

### Nightly Build

- [x] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
As with your last bug report, please provide a lot more information, ideally in the original message so we don't have to ask you for it. The more information you can provide the more likely we are going to be able to investigate t and find a solution. Using vague language like "certain videos" but not actually providing the videos, doesn't help, because we can't read minds, we have no idea what is happening on your computer unless you provide details.

liammcsh2008:
![image](https://github.com/user-attachments/assets/32527902-f94e-47f7-81a9-ba76f00f707b)
![image](https://github.com/user-attachments/assets/6945975b-04d9-4463-a31e-4198977df70f)

tuxayo:
@liammcsh2008 links please, to make it as easy as possible.

liammcsh2008:
https://www.youtube-nocookie.com/embed/AodJG8o8UA0

liammcsh2008:
https://www.youtube-nocookie.com/embed/mpwYovWYFiE

liammcsh2008:
[...]

Issue #4890, opened 2024-04-06
[Bug]: video glitch since version 0.20


Angedestenebres:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Open Freetube
2. Click on a video
3. The playing starts
4. After few second, there are some glitches.
5. See this video to illustrate the issue: https://video.tedomum.net/w/nf5o9io8pTRB7dbNXvUbMC

### Expected Behavior

Playing video without issue

### Issue Labels

visual bug

### FreeTube Version

0.20

### Operating System Version

Linux Mint 21.3

### Installation Method

Flathub

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

0.19.2

### Additional Information

I have two laptops and both has the same OS (Linux Mint 21.3) and version (flatpak 0.20) but only one laptop has this issue.
Both using X11.

However, the resolution of the laptop with the issue is 1366x768 (max for my screen).
I don't have the issue on the laptop with the resolution 1920x1080.

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Angedestenebres:
I complete the ticket with the PC configuration where I get the issue:

CPU: Intel© Core™ i5-5300U CPU @ 2.30GHz × 2
GPU: Intel Corporation HD Graphics 5500
RAM: 8Gb
Resolution: 1366x768

I tried to remove the app cache in /.var and also all the config, no effect.
Remove and install again the app didn't resolved the issue.

Also, I have no issue to watch a video directly from Youtube website or from a Peertube instance or a local video. The issue is linked to Freetube only.

ssnor:
[...]

PR #6274, opened 2024-12-04
Update hardware acceleration flag for Linux

Revival8697:
# Update hardware acceleration flag for Linux

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

Chromium [changed the flag](https://chromium-review.googlesource.com/c/chromium/src/+/5893615) in v131.

Please merge this after updating Electron to v34, which is [scheduled for release on January 14, 2025](https://www.electronjs.org/docs/latest/tutorial/electron-timelines#timeline).

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -212,7 +212,7 @@ function runApp() {
     // Enable hardware acceleration via VA-API with OpenGL if no other feature flags are found
     // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/gpu/vaapi.md
     if (!app.commandLine.hasSwitch('enable-features')) {
-      app.commandLine.appendSwitch('enable-features', 'VaapiVideoDecodeLinuxGL')
+      app.commandLine.appendSwitch('enable-features', 'AcceleratedVideoDecodeLinuxGL')
     }
   }
 


PikachuEXE:
From https://chromium-review.googlesource.com/c/chromium/src/+/5893615/4/docs/gpu/vaapi.md

2024-12-04 19_17_38-…_vaapi md · Gerrit Code Review - Vivaldi

PikachuEXE:
@Revival8697 
Nope, just letting other reviewers to see the diff without clicking around

Revival8697:
> If you meant to implement `--enable-features=AcceleratedVideoDecodeLinuxZeroCopyGL`, I tried with Brave on both Intel and AMD systems, it didn't work. Feel free to test and report back.

It turns out that you have to use this flag together with `AcceleratedVideoDecodeLinuxGL`.
On my Intel Gen 9 GPU, I can now watch 4K videos with ~0 frame drops at 50% video usage(?) reported by `intel_gpu_top`, compared to 100% usage with ~10 frame drops per second.

This improvement is too significant not to implement.

One issue: FreeTube has to run as a native Wayland application for it to work. It doesn't.
[This](https://github.com/flathub/io.freetubeapp.FreeTube/blob/558ce0eda25d56e29a6a0d188d671b2145f3789e/run.sh#L9) doesn't seem to do anything. Thoughts?

absidue:
@Revival8697 You can try using `--ozone-platform=wayland` to force it to use wayland. Please note that wayland support in Chromium and Electron is still considered WIP, which is why it isn't enabled by default, so you may experience issues that you wouldn't with the X11 backend.

Revival8697:
> @Revival8697 You can try using `--ozone-platform=wayland` to force it to use wayland. Please note that wayland support in Chromium and Electron is still considered WIP, which is why it isn't enabled by default, so you may experience issues that you wouldn't with the X11 backend.

i would consider it mature enough, i use brave in native wayland as my main browser. but if wayland auto switching isn't implemented, i won't implement ZeroCopy. the performance loss though...

absidue:
Difficult to pin down exactly why it isn't working, as you deleted the relevant section of the pull request template instead of filling it in. According to the Arch Linux wiki, auto-detection is known to be broken with Weston, on if you are using flatpak you need to enable the wayland socket and there is also an open PR there to enhance the launch script.

There is also a question if these flags should be in FreeTube by default at all, as they seem to be the cause of the video player crashes, because all of the Linux users that are experiencing it are able to stop it happening by passing an empty enable features flags which overwrites the defaults.

Revival8697:
> Difficult to pin down exactly why it isn't working, as you deleted the relevant section of the pull request template instead of filling it in. According to the Arch Linux wiki, auto-detection is known to be broken with Weston, if you are using flatpak you need to enable the wayland socket and there is also an open PR there to enhance the launch script.

You just pinned it down. FreeTube doesn't have access to the Wayland socket [by default](https://github.com/flathub/io.freetubeapp.FreeTube/blob/558ce0eda25d56e29a6a0d188d671b2145f3789e/io.freetubeapp.FreeTube.yml#L13).

> There is also a question if these flags should be in FreeTube by default at all, as they seem to be the cause of the video player crashes, because all of the Linux users that are experiencing it are able to stop it happening by passing an empty enable features flags which overwrites the defaults.

If you meant https://github.com/FreeTubeApp/FreeTube/issues/4890, I think it is a driver issue, but seems like affected users are happy with just passing the flag.

> The current issue that people have mentioned experiencing with FreeTube when using wayland is not being able to resize the window, which yes is an improvement over the app crashing when you maximise the window but it does still prove that it is still work in progress.

I will check on related issues then.

absidue:
The flatpak not having access to to the wayland socket is intentional due to the problems with wayland and the fact that even Chromium themselves doesn't consider it ready which is why they still use X11 by default instead of auto-detecting wayland, please check the pull requests on the flatpak repo for the exact history.

With the video player issues I was actually talking about https://github.com/FreeTubeApp/FreeTube/issues/5975 where the GPU process terminates itself because of fatal errors while talking to the GPU driver, which go away when removing the VA-API flag (I guess that is why just like wayland support, Chromium still has it behind a flag).

absidue:
If we have to choose between a bunch of people not being able to watch videos at all by default and having to pass flags to make it possible or everyone being able to watch videos but that hardware acceleration is disabled by default and people that have a setup where it actually works can opt-in via flags, I would definitely chose the latter.

Revival8697:
> If we have to choose between a bunch of people not being able to watch videos at all by default and having to pass flags to make it possible or everyone being able to watch videos but that hardware acceleration is disabled by default and people that have a setup where it actually works can opt-in via flags, I would definitely chose the latter.

That is true. I will look into the issue, but please remove this if it is deemed too buggy and breaks for too many users.

I would love to have this implemented in the GUI as a drop-down menu with "Disabled" as an option, but I have no idea how to pass the value from `src/renderer/components/player-settings/player-settings.js` to `src/main/index.js`. 

I honestly don't understand why there are so many issue on certain hardware as it works fine OoTB on my hardware. And in testing, even if I force it to fail, it would just fallback to software rendering instead of breaking like that.

absidue:
As the flag has to be added before the app is ready, we couldn't do it like a normal setting as those are loaded asynchronously in-parallel whereas this setting would have to be loaded synchronously and blocking to prevent the app/electron getting ready while it is happening.

What we could so is do it like the experimental disable HTTP cache setting, which creates/deletes an empty file in the user data directory and FreeTube checks if the file exists at start up. So as VA-API is so buggy I definitely think it qualifies as an experimental setting and should be disabled by default, then we could have a toggle in the experimental settings page to turn it on for people like you who mysteriously don't have the problems that everyone else has.

Or we could just remove it and then people can manually pass the flag if they are lucky enough to have drivers that it actually works with.

Issue #4762, opened 2024-03-13
[Bug]: Crash on Startup - "Object has been destroyed"


malacandrian:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Install Freetube
2. Open Freetube
3. Close Freetube
4. Repeat 2-3 until crash happens

![image](https://github.com/FreeTubeApp/FreeTube/assets/858304/144c3f00-7a9f-4ca4-adf5-cc1c8afe5326)

> Uncaught Exception:
> TypeError: Object has been destroyed at App.<anonymous> (%AppData%\Local\Programs\FreeTube\resources\app.asar\di...:242736) at App.emit (node:events:513:28)

Once this crash has occurred once it will happen every time until Freetube is uninstalled & reinstalled.

### Expected Behavior

Freetube should open as normal

### Issue Labels

causes crash

### FreeTube Version

v0.19.2

### Operating System Version

Windows 10 Pro 22H2 19045.4170

### Installation Method

.exe

### Primary API used

Local API

### Additional Information

I have been experiencing this error sporadically for at least 3 months, possibly closer to 6. So it has persisted across multiple different versions of FreeTube.

While it is also a Javascript error crash-on-start, is a different issue to #2131 because the error is different - this one is "Object has been destroyed".

It could be related to #1250 as the error is the same, but the referenced files and lines appear to be different.

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

vF360:
I'm experiencing the same problem too. I guess it is easier to code a script that auto uninstalls & reinstalls at this point rather than waiting for a fix.

hboetes:
[...]

PR #6278, opened 2024-12-04
Fix 'Object has been destroyed' when FreeTube is passed a URL with no windows open on macOS

OothecaPickle:
# Fix 'Object has been destroyed' when FreeTube is passed a URL with no windows open on macOS

## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Relates to @srolandmarshall's [comment](https://github.com/FreeTubeApp/FreeTube/issues/4762#issuecomment-2464673877) on #4762 (but unrelated to the issue commented on).

## Description

Because FreeTube does not quit automatically when all windows are closed (as on other plaforms), the main process tries to fire an IPC message to a non-existent `mainWindow` when it receives an `open-url` event or a command line URL argument. This results in an `Object has been destroyed` error message.
This PR sets `mainWindow` to a string indicating it doesn't exist when all FreeTube windows have been closed, and, if this is the case, has the main process create a new window and pass it a URL to be opened on `open-url` or `second-instance` (with a command line URL argument) event.

## Testing 

1. Open FreeTube on macOS and close its window without quitting the application (the FreeTube icon should remain visible in the Dock).
2. Pass FreeTube a video URL either as a command line argument or by opening a `freetube://` link (deep link).
3. Check that video opens in a new window without an error.

## Desktop

- **OS:** macOS
- **OS Version:** 10.13.6
- **FreeTube version:** 0e44e6b841d6799af7498d279bf197bc4b8c2e77

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1442,6 +1442,7 @@ function runApp() {
   app.on('window-all-closed', () => {
     // Clean up resources (datastores' compaction + Electron cache and storage data clearing)
     cleanUpResources().finally(() => {
+      mainWindow = 'all-windows-closed'
       if (process.platform !== 'darwin') {
         app.quit()
       }
@@ -1507,7 +1508,14 @@ function runApp() {
   app.on('open-url', (event, url) => {
     event.preventDefault()
 
-    if (mainWindow && mainWindow.webContents) {
+    if (mainWindow === 'all-windows-closed') {
+      app.once('browser-window-created', (_, mainWindow) => {
+        mainWindow.webContents.once('did-finish-load', () => {
+          setTimeout(function() { mainWindow.webContents.send(IpcChannels.OPEN_URL, baseUrl(url), { isLaunchLink: true }) }, 1000)
+        })
+      })
+      createWindow()
+    } else if (mainWindow && mainWindow.webContents) {
       mainWindow.webContents.send(IpcChannels.OPEN_URL, baseUrl(url))
     } else {
       startupUrl = baseUrl(url)
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -254,13 +254,18 @@ function runApp() {
 
     app.on('second-instance', (_, commandLine, __) => {
       // Someone tried to run a second instance, we should focus our window
-      if (mainWindow && typeof commandLine !== 'undefined') {
-        if (mainWindow.isMinimized()) mainWindow.restore()
-        mainWindow.focus()
-
+      if (typeof commandLine !== 'undefined') {
         const url = getLinkUrl(commandLine)
-        if (url) {
-          mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+        if (mainWindow === 'all-windows-closed') {
+          if (url) macOSnewURLWindow(url)
+          createWindow()
+        } else if (mainWindow && mainWindow.webContents) {
+          if (mainWindow.isMinimized()) mainWindow.restore()
+          mainWindow.focus()
+
+          if (url) {
+            mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+          }
         }
       }
     })
@@ -1509,13 +1514,7 @@ function runApp() {
     event.preventDefault()
 
     if (mainWindow === 'all-windows-closed') {
-      app.once('browser-window-created', (_, mainWindow) => {
-        mainWindow.webContents.once('did-finish-load', () => {
-          // A timeout here is necessary or the new window won't receive the URL
-          setTimeout(function() { mainWindow.webContents.send(
-            IpcChannels.OPEN_URL, baseUrl(url), { isLaunchLink: true }) }, 1000)
-        })
-      })
+      macOSnewURLWindow(baseUrl(url))
       createWindow()
     } else if (mainWindow && mainWindow.webContents) {
       mainWindow.webContents.send(IpcChannels.OPEN_URL, baseUrl(url))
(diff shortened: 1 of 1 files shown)


src/main/index.js
```diff
@@ -1550,6 +1559,15 @@ function runApp() {
...
 
+  function macOSnewURLWindow(url) {
+    app.once('browser-window-created', (_, mainWindow) => {
+      mainWindow.webContents.once('did-finish-load', () => {
+        // A timeout here is necessary or the new window won't receive the URL
+        setTimeout(function() { mainWindow.webContents.send(IpcChannels.OPEN_URL, url, { isLaunchLink: true }) }, 1000)
+      })
+    })
+  }
+
```
srolandmarshall, line 1570: Would 

```js
  async function macOSnewURLWindow(url) {
    app.once('browser-window-created', async (_, mainWindow) => {
      await mainWindow.webContents.once('did-finish-load', () => {
        mainWindow.webContents.send(IpcChannels.OPEN_URL, url, { isLaunchLink: true })
      })
    })
  }
```

work better, instead of timeout? Are we just waiting for a promise to resolve?

src/main/index.js
```diff
@@ -254,13 +254,18 @@ function runApp() {
...
-        if (mainWindow.isMinimized()) mainWindow.restore()
-        mainWindow.focus()
-
+      if (typeof commandLine !== 'undefined') {
         const url = getLinkUrl(commandLine)
-        if (url) {
-          mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+        if (mainWindow === 'all-windows-closed') {
+          if (url) macOSnewURLWindow(url)
+          createWindow()
```
absidue, line 261: You need to set this to true, otherwise the mainWindow variable will stay as the string.
```suggestion
          createWindow({ replaceMainWindow: true })
```

src/main/index.js
```diff
@@ -1507,7 +1513,10 @@ function runApp() {
   app.on('open-url', (event, url) => {
     event.preventDefault()
 
-    if (mainWindow && mainWindow.webContents) {
+    if (mainWindow === 'all-windows-closed') {
+      macOSnewURLWindow(baseUrl(url))
+      createWindow()
```
absidue, line 1518: ditto
```suggestion
      createWindow({ replaceMainWindow: true })
```

src/main/index.js
```diff
@@ -1550,6 +1559,15 @@ function runApp() {
     }
   }
 
+  function macOSnewURLWindow(url) {
+    app.once('browser-window-created', (_, mainWindow) => {
+      mainWindow.webContents.once('did-finish-load', () => {
+        // A timeout here is necessary or the new window won't receive the URL
+        setTimeout(function() { mainWindow.webContents.send(IpcChannels.OPEN_URL, url, { isLaunchLink: true }) }, 1000)
+      })
+    })
```
absidue, line 1568: If you reuse the same code that is used at app-startup you avoid having to hardcode a timeout (generally a good idea to check the existing code first before re-inventing the wheel).

https://github.com/FreeTubeApp/FreeTube/blob/9e09ec5dc6d2dacbc1cc8a9143e0ccce71e1a750/src/main/index.js#L832-L836

```suggestion
    ipcMain.once(IpcChannels.APP_READY, () => {
      mainWindow.webContents.send(IpcChannels.OPEN_URL, URL, { isLaunchLink: true })
    })
```

src/main/index.js
absidue: Nope, as you would be sending the URL as soon as the HTML has loaded but the JavaScript is loaded asynchronously and it still needs to load all of your user data such as the settings and actually register the handler for that IPC call. Please see my better solution below, which is the same as what is done when the app originally starts, which is waiting for the window to say it is ready (refer to `src/renderer/App.js` for the window-side code.

src/main/index.js
OothecaPickle:
```js
  async function createWindow(
    {
      replaceMainWindow = true,
      windowStartupUrl = null,
      showWindowNow = false,
      searchQueryText = null
    } = { }) {
```

`replaceMainWindow` is true by default

absidue: Looks like I was focusing too much on the other issue and didn't read this bit correctly.

For the other issue I was originally going to ask you to use modern JavaScript syntax and then realised that the timeout wasn't even necessary, as you could just do the same as when the app is launched and listen for the `APP_READY` event which is ironically sent right below the code in the `OPEN_URL` listener that you modified in your previous PR.

src/main/index.js
OothecaPickle: My bad, I've committed changes so that this callback will be used instead.

src/main/index.js
```diff
@@ -254,13 +254,16 @@ function runApp() {
...
-        if (url) {
-          mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+        if (mainWindow && mainWindow.webContents) {
+          if (mainWindow.isMinimized()) mainWindow.restore()
+          mainWindow.focus()
+
+          if (url) mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+        } else {
+          if (url) startupUrl = url
+          createWindow()
```
OothecaPickle, line 266: This means attempting to launch a new instance via the command line (without a URL) will result in a new window being opened.

Let me know if I should change this so that's not the case.

PikachuEXE (approved):
Tested on macOS 13.7.1
`Bug has been destroyed`

kommunarr:
Don't currently have access to a Mac to validate, but looks good from static analysis of the code.

efb4f5ff-1298-471a-8973-3d47447115dc:
@kommunarr will you have access to it anytime soon or should we go ahead and approve this PR as @PikachuEXE tested it on their machine?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

Issue #6296, opened 2024-12-07
[Bug]: electron window not opening in local development


80avin:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Do a fresh clone of FreeTube
2. Run `yarn install`
3. Run `yarn run dev`

Notice that the script gets stuck at following status
![image](https://github.com/user-attachments/assets/8c92a717-21e8-42b4-851b-fa2e8bba0807)

It also happens if I build the appimage and run it.

### Expected Behavior

Electron window should popup

### Issue Labels

causes crash, only happens in developer mode

### FreeTube Version

0e8db50ed07d0b721104031c1cfc81721cdbbe38

### Operating System Version

Kubuntu 24.04 ( Linux avinash-Nitro-AN515-45 6.8.0-49-generic #49-Ubuntu SMP PREEMPT_DYNAMIC Mon Nov  4 02:06:24 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux )

### Installation Method

other

### Primary API used

Local API

### Additional Information

I think major issue is that [logging is disabled](https://github.com/FreeTubeApp/FreeTube/blob/59590d77316f4777859c1ab4e60d1ab7268be667/_scripts/dev-runner.js#L67-L71) in local, which prevents me from debugging it.

After I enable logging, I can find that the issue is
https://github.com/electron/electron/issues/42510
![image](https://github.com/user-attachments/assets/21f31a2f-a94d-4b8f-9958-5c91f155cfc1)

which can be worked around by adding `--no-sandbox` flag to the electron.

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

[...]

Issue #333, opened 2019-08-14
Implement tabs like in web browsers


morfikov:
Is there a chance to implement tabs like the ones in web browsers? For instance, you have a list of channels, and you want to watch five videos. I would normally use the middle mouse button, and the vids would open in separate tabs. I can switch to individual tab and play its video.

Emaro:
I'd like an option to manage multiple videos for watching soon too. Tabs would be a familiar way.

JonnyHaystack:
Tabs would be nice, but I think multiple windows would be more useful, particularly for those who use workspaces and/or tiling WMs. Ideally having both would be nice.

Archive5:
> Tabs would be nice, but I think multiple windows would be more useful, particularly for those who use workspaces and/or tiling WMs. Ideally having both would be nice.

Expect multiple windows as well.

- allow create new window by drag a tab out
- allow merge existing tabs to new  or old window etc same as browser
- allow pin tabs to left or right from desired area (such features has on vivaldi browser)
- allow close tabs to right and left also selected random tabs
- allow bookmark random tabs easily and bookmark on different categories (check vivaldi)

Agych:
> Tabs would be nice, but I think multiple windows would be more useful, particularly for those who use workspaces and/or tiling WMs. Ideally having both would be nice.

I would disagree and say that tabs would be more useful, I normally have a lot of stuff open in the same time - many tabs with videos, channels, playlists.
Would be just great to have both tabs and windows, with a browser-like behavior like KizunaH explained.

GilgusMaximus:
You can already open a second window by using the middle mouse button. However not everything you do in the second window will be saved, due to issues with how the data is stored

LongJohn-Silver:
I would also love tabs and/or windows.

closewall:
I'd also like to see this. I'd say it's the one feature that would make this app perfect.

refouna:
https://winaero.com/youtube-is-now-available-as-a-pwa/

FunWithUrban:
I too would love tab support. Its the one thing missing for me until I can make the switch.

efb4f5ff-1298-471a-8973-3d47447115dc:
> 
> 
> You can already open a second window by using the middle mouse button. However not everything you do in the second window will be saved, due to issues with how the data is stored

[...]

PR #6297, opened 2024-12-07
chore: enable logs for local development

80avin:
# Title

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/6296

## Description

I'm not sure if everyone is facing issue with sandboxing, that's why I have added that only as a comment. But enabled the logs so that such errors can be easily identified.

## Screenshots 

image

## Testing 

Only tested using eslint ( as part of commit hook )

## Desktop

- **OS:** Kubuntu
- **OS Version:** 24.04
- **FreeTube version:** https://github.com/FreeTubeApp/FreeTube/commit/0e8db50ed07d0b721104031c1cfc81721cdbbe38

## Additional context

No

### DIFF
--- a/_scripts/dev-runner.js
+++ b/_scripts/dev-runner.js
@@ -64,18 +64,23 @@ async function restartElectron() {
 
   electronProcess = spawn(electron, [
     path.join(__dirname, '../dist/main.js'),
-    // '--enable-logging', // Enable to show logs from all electron processes
+    '--enable-logging', // Enable to show logs from all electron processes
     remoteDebugging ? '--inspect=9222' : '',
-    remoteDebugging ? '--remote-debugging-port=9223' : ''
+    remoteDebugging ? '--remote-debugging-port=9223' : '',
+    // fix for "The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /.../FreeTube/node_modules/electron/dist/chrome-sandbox is owned by root and has mode 4755."
+    // '--no-sandbox',
   ],
-    // { stdio: 'inherit' } // required for logs to actually appear in the stdout
+  { stdio: 'inherit' } // required for logs to actually appear in the stdout
   )
 
-  electronProcess.on('exit', (code, _) => {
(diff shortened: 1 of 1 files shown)


absidue:
You need to set up an apparmor profile for Electron. Disabling the sandbox is not the right solution to the problem, the error is telling that it is failing to start because something wrong in your environment (in your case Ubuntu being way too restrictive) and your solution is to just disable Electron's security.

Additionally Electron's logging is disabled by default in dev as 99.98% of the time it is irrelevant and just spams the console. But in the rare case that you do need it all you need to do is uncomment those lines temporarily.

80avin:
@absidue I can understand.

Is your opinion the same about printing the exit status? or, inheriting only stderr of electron process ?

absidue:
You will still get unnecessary logs with just stderr.

80avin:
But exit status will only be printed once per session.
And that is also sufficient to hint the new developers.

If we are disabling logs, I suggest we show the non-zero/non-restart exit status with the message like `Electron exited with code {code} on receiving signal {signal}. Enable electron logging in `_scripts/dev-runner.js` to enable logs`

Or, have an environment variable to enable electron logging as otherwise, unstaged changes would create issues.

80avin:
> if you are the kind of person to commit every changed file every single time without doing a quick check, then it really is your own fault.

I am fine with the closure of this PR and was only seeking your opinion on showing the electron's exit status as without that, we are completely hiding any sign of crash from the electron process.
I originally came to implement https://github.com/FreeTubeApp/FreeTube/issues/333 as I liked the idea and found it doable. But, I'm not sure where this assumption is coming from or where is the argument going.

If someone is capable of hand-picking each hunk for staging, he should be doing it and not finding any efficient process. He should do 2 extra steps of stash push/pop before every rebase, almost every checkout, etc because he can.


Nevertheless... This was just a bad assumption for you to make. I hope it is just an unfortunate situation or timing for me to bother you and we can have better arguments in future.

PR #6300, opened 2024-12-07
Switch the data export to the Web File System API

absidue:
# Switch the data export to the Web File System API

## Pull Request Type

- [x] Refactoring

## Related issue
- https://github.com/FreeTubeApp/FreeTube/pull/6222

## Description
Currently the data export uses the Electron dialog API and the Node.js `fs` API in the renderer to export user data, this pull request switches it over to the [Web File System API](https://developer.mozilla.org/en-US/docs/Web/API/Window/showOpenFilePicker). That lets us get rid of one more use of a Node.js API in the renderer, which gets us one step closer to turning off `nodeIntegration` in the renderer (`src/renderer/helpers/utils.js` no longer imports `fs/promises`), this time without needing to build an IPC call as a replacement.

I haven't removed the `showSaveDialog` function yet, as that is currently still used for the download and screenshot file pickers.
You may have noticed that `writeFileWithPicker` accepts both strings and `Blob`s as the content, that is so that I can use it for the screenshot functionality in a follow up pull request. I decided to do that in a follow up pull request as the save to default directory and save to prompted directory screenshot code are current rather intertwined so it will require more refactoring than just swapping out one function.

## Testing

1. `yarn dev`
2. Test exporting your history, subscriptions and playlists
3. Test exporting a user playlist from its playlist page
4. `yarn dev:web`
5. Open the dev server URL in a web browser that doesn't support the Web File System API e.g. Firefox or Safari or if you have access to a FreeTubeAndroid compatible device you can test this pull request with that too.
6. Test exporting your history, subscriptions and playlists
7. Test exporting a user playlist from its playlist page

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 6e2ec287734800ccb374e0e815c79be3eb9dbad5

### DIFF
--- a/src/renderer/components/data-settings/data-settings.js
+++ b/src/renderer/components/data-settings/data-settings.js
@@ -13,13 +13,12 @@ import {
   escapeHTML,
   getTodayDateStrLocalTimezone,
   readFileWithPicker,
-  showSaveDialog,
   showToast,
-  writeFileFromDialog,
+  writeFileWithPicker,
 } from '../../helpers/utils'
 
 const IMPORT_DIRECTORY_ID = 'data-settings-import'
-const IMPORT_START_IN_DIRECTORY = 'downloads'
+const START_IN_DIRECTORY = 'downloads'
 
 export default defineComponent({
   name: 'DataSettings',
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested w/ FF 132

kommunarr (approved):

Issue #6331, opened 2024-12-09
[Bug]: Player: Tooltips get cut off with long text


MarkusGaugg:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

When you open the player and play any video, longer tooltip texts get cut off automatically. 

![Ashampoo_Snap_Montag, 9  Dezember 2024_20h50m35s](https://github.com/user-attachments/assets/ef4a78e3-129e-47b4-a23a-4147773261fb)

### Expected Behavior

It would be great if, when the tooltips reach the edge, they stop being centered on the mouse, just like how it already works with the video thumbnail preview.

### Issue Labels

visual bug

### FreeTube Version

v0.22.0-nightly-5283 Beta

### Operating System Version

Windows 11 - 23H2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

alexeevrevan:
Hey! I can go try and fix it. Assign this to me

MarkusGaugg:
You can work on it and then publish a pull request for it.
https://github.com/FreeTubeApp/FreeTube/pulls

alexeevrevan:
Hey, I'm ready with it. Now it works properly as it should.  Created a PR with screenshots

PR #6336, opened 2024-12-09
Fix Player: Tooltips get cut off with long text

alexeevrevan:
# Tooltip bug fixed

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/issues/6331

## Description
The application was inspected via devtools to find where the bug happens, I fixed it in the code afterwards. Now Tooltip pops up correctly, even with more symbols in it.

## Screenshots
Before and After bugfix
before
after

## Testing
Tried different options right in the browser devtools, found out that there's a css feature conflict that causes unnecessary overlapping. A little time to inspect the codebase, quite a few simple moves and here we go, it works properly. Tested a different amount of symbols. Works fine until there's one milliion symbols, but there's no need in such an amount. Tested in Russian langauge and in German (as in the following issue), works as it should.

## Desktop
- OS: Windows
- OS Version: 10
- FreeTube version: v0.22.0 Beta

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
@@ -266,8 +266,8 @@
   :deep(.shaka-tooltips-on button:focus-visible::after),
   :deep(.shaka-tooltips-on button:hover::after) {
     margin-bottom: 8px;
-    // Fixes long tooltips on the right got cut off
-    // Assumes tooltips on the left are all short enough
+    /*Fixes long tooltips on the right got cut off*/
+    /*Assumes tooltips on the left are all short enough*/
     left: 1px;
   }
 }
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
@@ -266,8 +266,8 @@
   :deep(.shaka-tooltips-on button:focus-visible::after),
   :deep(.shaka-tooltips-on button:hover::after) {
     margin-bottom: 8px;
-    /*Fixes long tooltips on the right got cut off*/
-    /*Assumes tooltips on the left are all short enough*/
+    
+    /* Fixes long tooltips on the right got cut off by adjusting their positioning */
     left: 1px;
   }
 }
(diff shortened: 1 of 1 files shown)


src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
```diff
@@ -266,6 +266,7 @@
   :deep(.shaka-tooltips-on button:focus-visible::after),
   :deep(.shaka-tooltips-on button:hover::after) {
     margin-bottom: 8px;
+    left: 1px;
```
PikachuEXE, line 269:
```suggestion
    // Fixes long tooltips on the right got cut off
    // Assumes tooltips on the left are all short enough
    left: 1px;
```

PikachuEXE (commented):
Some comments would be great

PikachuEXE:
Also there is a PR template please follow it

alexeevrevan:
> Also there is a PR template please follow it

you mean text formatting? OK I'll try to

alexeevrevan:
> Also there is a PR template please follow it

Done editing

PikachuEXE (approved):

alexeevrevan:
@absidue @efb4f5ff-1298-471a-8973-3d47447115dc  @ChunkyProgrammer

absidue:
@alexeevrevan Please don't unnecessarily ping people, our reviews were already requested and we will get to them when we have time to look at this PR. We are volunteers that work on FreeTube in our spare time, we all have lives outside of it, so yes we don't work on it 24/7. Please be patient, instead of annoying us with extra pings.

kommunarr:
This particular fix improves the situation for the right-side buttons but worsens it for left-side buttons. An ideal fix would make the title text work like the thumbnail image does currently, which is also how title text works in YT currently.

Screenshot_20241212_140625

alexeevrevan:
> This particular fix improves the situation for the right-side buttons but worsens it for left-side buttons. An ideal fix would make the title text work like the thumbnail image does currently, which is also how title text works in YT currently.
> 
> Screenshot_20241212_140625

thats the good point

alexeevrevan:
> This particular fix improves the situation for the right-side buttons but worsens it for left-side buttons. An ideal fix would make the title text work like the thumbnail image does currently, which is also how title text works in YT currently.
> 
> Screenshot_20241212_140625

when i was about to come to a commit I made I found out that word-break property works just fine avoiding border overlapping by stacking words. but i guess its not youtube-ish enough, hahah.

PR #6339, opened 2024-12-10
improve jsdoc comments for invidious

ChunkyProgrammer:
# improve jsdoc comments for invidious

## Pull Request Type
- [x] Other - better type support when using Invidious functions

## Description
Improves jsdoc comments by referencing some of the types on the invidious documentation page: https://docs.invidious.io/api/

## Desktop
- **OS:** Fedora Linux
- **OS Version:** 41
- **FreeTube version:** 0.22.0 (latest nightly)


src/renderer/views/Channel/Channel.js
```diff
@@ -1048,7 +1048,7 @@ export default defineComponent({
       }
     },
 
-    getChannelInfoInvidious: function () {
+    getChannelInfoInvidious: async function () {
```
absidue, line 1051: This function doesn't use await, so it could probably be switched back to a normal sync function.

absidue (commented):
We should probably base these types off real responses, instead of the docs, as in my experience Invidious' docs don't match the real response data.

ChunkyProgrammer:
> We should probably base these types off real responses, instead of the docs, as in my experience Invidious' docs don't match the real response data.

I mainly looked at the docs but I did verify by looking at our code, the invidious api code and the actual responses so it should be pretty accurate. I also opened a PR to the Invidious documentation repo to add some of the missing data:
https://github.com/iv-org/documentation/pull/616

absidue (approved):

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #6366, opened 2024-12-12
Make a single IPC call for the navigation history

absidue:
# Make a single IPC call for the navigation history

## Pull Request Type

- [x] Performance improvement

## Related issue

Follow up to #5227

## Description

The current implemenation of the navigation history dropdowns makes a lot of IPC calls, one to get the active index, a second one to get the total number of history entries and then one IPC call for each history entry that it wants to get the title for. This pull request moves the construction of that navigation history list into the main process, that way we just need a single IPC call that can directly return the list, avoiding all of the overhead from the extra IPC calls.

While I was at it, I also replaced the IPC call to navigate to a history offset with `$router.go(offset)` as it does the same thing without needing to do the IPC call.

## Testing

Right click on the navigation arrows and navigate to various history entries, it should behave the same as before.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** c686b84bcaf1589549762120e7bd802cbe633200


kommunarr (commented):
This is great! Thank you for doing this!

**issue (minor):** I am seeing a new bug that I can confirm I do not see in `development`: when making two consecutive page searches, the title displayed in the dropdown for the active entry is now incorrectly the title for the previous search, not the current one.

Screenshot_20241212_174245

absidue:
Okay this is an interesting one, as I can't spot any changes to the logic, I think it is actually a pre-exiting race-condition. Both the title setting code and the fetching the menu are async and I think by messing with one of them, I've changed the order in which they finish triggering changing the outcome.

I'll try experimenting a bit with the title setting code, e.g. at the moment we call a store action that is just a thin wrapper around a store mutation, so if we directly call the mutation which is sync, instead of the action which is asnyc it might fix it.

absidue:
Fixed the issue.

efb4f5ff-1298-471a-8973-3d47447115dc:
Im unsure if this is caused by the PR but will mention it anyway. The history will list the previous page for a split second when navigating to a route that has to load stuff e.g. watch page

https://github.com/user-attachments/assets/78fc0718-e3d9-4053-a538-7db2c8aa96f2

efb4f5ff-1298-471a-8973-3d47447115dc:
Should that be addressed in this PR or a separate one?

absidue:
I would say introducing a placeholder should happen in a follow up, as the main goal of this PR was to speed it up and there not being anything to show while it is loading is sort of unrelated to that, at least in my mind.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

kommunarr (approved):

Issue #5500, opened 2024-07-30
[Feature Request]: Hide/sort channel videos based on viewed or not


jojo2357:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

When viewing a channel page, I would like to see only videos that I have not seen. Another place I would like this is on the subscription feed.

### Proposed Solution

Add a local toggle to show/hide watched/unwatched videos,

### Alternatives Considered

Make a global setting; however this has the obvious drawback of being cumbersome to toggle off.

### Issue Labels

ease of use improvement, new feature

### Additional Information

Example of videos I would like to hide:

![image](https://github.com/user-attachments/assets/07260c8d-1cec-4b78-ad66-260a7cad30e1)

efb4f5ff-1298-471a-8973-3d47447115dc:
> Another place I would like this is on the subscription feed.

Go to: Settings -> Subscription Settings -> Enable `Hide Videos on Watch`

> When viewing a channel page, I would like to see only videos that I have not seen.

duplicate of https://github.com/FreeTubeApp/FreeTube/issues/4497

jojo2357:
> Go to: Settings -> Subscription Settings -> Enable Hide Videos on Watch

This is a rather cumbersome solution, would prefer a simpler check similar to what was asked for in 4497. 

Sorry for partial duplicate, not sure how i missed it

Issue #5261, opened 2024-06-13
[Feature Request]: Hide individual videos from Subscription feed (NOT watched)


luke-bravenboer:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Youtube's website has the ability to 'hide' individual videos from the subscriptions page, which helps me curate my feed (eg. I use it frequently for channels that produce multiple streams of content but I'm only interested in one of them, such as Critical Role).

I know there is an option to hide **watched** videos from the subs feed, but that's not suitable for my use case - I often watch very long videos that I stop and resume later, so I do not want things disappearing from the subs feed as soon as they're marked as 'watched'

### Proposed Solution

A 'hide video' button in the drop-down menu for videos on the subscriptions feed

### Alternatives Considered

- The setting to hide watched videos does not suit my needs - I would still like to be able to see videos that are partially watched
- Using the 'mark as watched' button to grey them out is a reasonable workaround, but a true hide button would just be a nice-to-have (and improve feature parity with Youtube)

### Issue Labels

improvement to existing feature

efb4f5ff-1298-471a-8973-3d47447115dc:
duplicate of https://github.com/FreeTubeApp/FreeTube/issues/1362

Issue #5189, opened 2024-05-29
[Feature Request]: Hide "Watched" videos from recommended, channels and playlists


Jeefongithub:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The watched videos feature is good for the homepage but doesn't hide videos in other aspects of the program

### Proposed Solution

Similarly how there is a toggle to hide watched videos from homepage I'd like to suggest a toggle to also hide videos that are in the watched state from recommended or sidebar when in a video, from the channel pages and from playlists. That way only videos that are unscene or unwatched are shown as options. 

### Alternatives Considered

alternatively watched videos could be less prioritized in the list of videos in recommended pushing unwatched videos to the top of the list but I assume the ordering is coming from youtube so that might not be possible

### Issue Labels

new feature

efb4f5ff-1298-471a-8973-3d47447115dc:
basically a duplicate of https://github.com/FreeTubeApp/FreeTube/issues/4497 with an expanded scope. Please put the scope expansion you like to see in the comment section of that issue

Issue #4863, opened 2024-04-04
[Feature Request]: Hide only fully-watched videos


terminal-root:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The `Hide Videos on Watch` setting under `Subscription Settings` hides played videos from the `Subscriptions` page no matter what portion of the video is watched. This is desirable sometimes but not so for videos you're just taking a break on.

### Proposed Solution

Hide or provide and option to hide only fully-watched videos.

### Alternatives Considered

 - option to hide X%-watched videos
 - option to hide X-minutes-left videos

### Issue Labels

improvement to existing feature

### Additional Information

FreeTube v.20.0 Beta

efb4f5ff-1298-471a-8973-3d47447115dc:
Duplicate of https://github.com/FreeTubeApp/FreeTube/issues/609

This is basically the root issue that needs to be addressed.

PR #6382, opened 2024-12-14
Better description for hiding watched vids

### LINKED ISSUE #6389: [Feature Request]: Better description for hiding watched vids
### Guidelines

### Problem Description

`Hide Videos on Watch` confused me and very likely others too so I made some minor changes to the text.
Everyone filing for an issue/feature request calls this feature as `Hide watched...` or something similar, unlike the existing UI description.

### Proposed Solution

The reason why it should be `Hide Videos marked as WATCHED` and not simply `Hide watched Videos` is because of potential future features. If there is an option for users to mark a video manually as `WATCHED` (probably watched somewhere else) then this button should also apply.

### Alternatives Considered

A more concise alternative is "Hide WATCHED Videos" because it also implies the `WATCHED` tag but of course less user friendly.

### Issue Labels

### Additional Information

_No response_

lamteteeow:
# Title

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

closes #6389

## Description

**Hide Videos on Watch** confused me and very likely others too so I made some minor changes to the text.

The reason why it should be **Videos marked as WATCHED** and not simply **watched Videos** is because of future features. If there is an option for user to mark a video manually as `WATCHED` (probably watched somewhere else) then this button should also apply.

## Desktop

- **OS:** Windows
- **OS Version:** 23H2
- **FreeTube version:** v0.22.1 Beta

### DIFF
--- a/src/renderer/components/subscription-settings/subscription-settings.vue
+++ b/src/renderer/components/subscription-settings/subscription-settings.vue
@@ -27,7 +27,7 @@
       </div>
       <div class="switchColumn">
         <ft-toggle-switch
-          :label="$t('Settings.Subscription Settings.Hide Videos on Watch')"
+          :label="$t('Settings.Subscription Settings.Hide Videos marked as WATCHED')"
           :default-value="hideWatchedSubs"
           :compact="true"
           @change="updateHideWatchedSubs"


ChunkyProgrammer:
I'd recommend keeping the label the same and only updating the value here (otherwise you'll have to update the label for all languages):
https://github.com/FreeTubeApp/FreeTube/blob/96bcdc306181f8e493d4c9448fabb4c0613aff8f/static/locales/en-US.yaml#L489

kommunarr:
I also do not think we need anything in all-caps or otherwise. I would contend that the existing label is good at communicating the constancy of the setting, and also that the use case of manually marking videos as watched (as is currently possible) is implied enough such as to not necessitate special wording.

efb4f5ff-1298-471a-8973-3d47447115dc:
I agree with @kommunarr here i dont see a reason to change this.

> confused me and very likely others too

If you're confused about it, that's fine. However, trying to strengthen your argument by making invalid claims about what others are thinking doesn't hold up

lamteteeow:
> If you're confused about it, that's fine. However, trying to strengthen your argument by making invalid claims about what others are thinking doesn't hold up.

- I base my argument on this #5500 #4497
- I got this issue while sitting together with a German friend, hence the statement. Changed language to German, same result for both.
- Calling me out for something I carefully wrote `very likely` seems pedantic to me.

So far searching through the Issue section, the term `Hide Videos on Watch` is mostly used by developers/maintainers/members. Others have created their issues using different ways (to me more intuitive) to call it, which exist in #1362 #5261 #4863 #4497 #5500, etc. Most of them are `Hide watched videos` or related.

I agree with @kommunarr, `Hide watched videos` seems more concise and accurately reflects the term users are using.

I love the app, played around with it and only then figured out what that button does. If this seems more like a "me" problem then feel free to close the PR. Otherwise let me know if I should continue as @ChunkyProgrammer recommended.

efb4f5ff-1298-471a-8973-3d47447115dc:
> I got this issue while sitting together with a German friend, hence the statement. Changed language to German, same result for both.
> 
> Calling me out for something I carefully wrote `very likely` seems pedantic to me.

Over the past years i have seen allot of people making claims about others when they want something implemented. So i made an assumption here and that was wrong of me but i do stand firm on the argument that you should have made it clear why you said that in the first place. You cannot just say something without proper argumentation 
 
> 
> So far searching through the Issue section, the term `Hide Videos on Watch` is mostly used by developers/maintainers/members. Others have created their issues using different ways (to me more intuitive) to call it, which exist in #1362 #5261 #4863 #5189 #4497 #5500, etc. Most of them are `Hide watched videos` or related.

This boils down to a subjective change and still dont see a reason to change this

PR #6400, opened 2024-12-17
Update play next recommended video setting to be "by default"

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

This PR does:
- Setting `Play next video` -> `Autoplay Recommended Videos by Default` (modified from https://github.com/FreeTubeApp/FreeTube/pull/5866
- Setting now affects the video playing session's initial value only, where a "video playing session" is the period that user stays on the watch page (for N videos) 
  - Changing the setting in settings page won't affect existing "video playing sessions" value
  - Changing the autoplay value in a "video playing session" won't affect global setting

## Screenshots 

image

https://github.com/user-attachments/assets/56a17453-5965-4344-a129-0e203ebf2860

## Testing 

(A) Auto play disabled by default
- Update auto play in settings to false
- Open 2 videos without playlist, confirm both got auto play disabled
- Update global setting to true, confirm both still got auto play disabled
- Update global setting to false, update one/both auto play to true, confirm global setting still false

(B)
- Steps of (A) but value reversed

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Existing settings which can be changed in watch page and acts default value
Screenshot 2024-12-17 at 10 33 17

### DIFF
--- a/src/renderer/components/WatchVideoRecommendations/WatchVideoRecommendations.vue
+++ b/src/renderer/components/WatchVideoRecommendations/WatchVideoRecommendations.vue
@@ -11,7 +11,7 @@
         class="autoPlayToggle"
         :label="$t('Video.Autoplay')"
         :compact="true"
-        :default-value="playNextVideo"
+        :default-value="playNextRecommendedVideo"
         @change="updatePlayNextVideo"
       />
     </div>
@@ -27,13 +27,9 @@
 </template>
 
 <script setup>
-import { computed } from 'vue'
-
 import FtCard from '../ft-card/ft-card.vue'
 import FtListVideoLazy from '../ft-list-video-lazy/ft-list-video-lazy.vue'
 import FtToggleSwitch from '../ft-toggle-switch/ft-toggle-switch.vue'
-
-import store from '../../store/index'
 
 defineProps({
   data: {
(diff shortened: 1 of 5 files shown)


kommunarr (commented):
Ooh, I like it! I have three thoughts:

1. **issue (minor, non-blocking):** If we're doing it one way for recommended video autoplay, we should be doing it the same way for playlist autoplay as well, otherwise it's a bit jarring / confusing as a user to read. I get that that's done because playlist autoplay is currently not toggleable from the player page (until #5866), but I'd recommend only merging this if we're confident we can get that other part included in the same release.
2. **nitpick:** Maybe we should colocate the autoplay buttons on the left side now so the "by Default"s are by one another.
3. **nitpick:** The "by Default" phrasing is reasonable to use here and consistent, but in general, I do think it makes our labels more verbose. Maybe in this or a separate PR we move all of these to its own sub-section of "Default Settings" or something like that?

efb4f5ff-1298-471a-8973-3d47447115dc:
@kommunarr i think that pika is waiting on you to respond to their latest message

kommunarr:
>  (2) Not quite sure what you mean. Autoplay Videos probably isn't included, so playlist + recommended videos?

Yes, if we are to update the logic and names of both, all the "by Default"s should be in the same column. Although like I said in 3, I'm still not a fan of the "by Default" being there in our setting names, as I feel like it adds more wordiness than improved user comprehension

PikachuEXE:
So just remove `by Default` label text change? Fine for me
Pending https://github.com/FreeTubeApp/FreeTube/pull/5866 for merge conflict storm

PikachuEXE:
Version based on https://github.com/FreeTubeApp/FreeTube/pull/5866 actually already done and being used by myself
Just waiting for that to be merged before I push those changes...

PikachuEXE:
PR title, body, code updated to be based on https://github.com/FreeTubeApp/FreeTube/pull/5866

kommunarr (approved):

absidue (approved):

PikachuEXE:
@efb4f5ff-1298-471a-8973-3d47447115dc @ChunkyProgrammer 

image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #6414, opened 2024-12-19
Keyword search history

### LINKED ISSUE #1492: Keyword search history

kommunarr:
# Keyword search history

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #1492

## Description
Adopting some pieces from #5003, this PR implements search history with behavior similar to that of YT. When `Enable Search Suggestions` is enabled and the search bar is focused with no non-whitespace text has been entered, up to 14 of the most recent searches will display. See Testing section for more specifics on the exact intended behaviors.

## Testing 
- Test that clicking a history entry preserves the original search's filter parameters without enacting them for the next search
- Test that repeating a search of a keyword (e.g., by clicking on it in history) moves it to the top of the dropdown
- Test that multiple searches of the same keyword but with different filter parameters does not make multiple search entries appear in dropdown
- For a keyword that has been searched multiple times with different filter parameters, ensure that clicking on that history entry uses the most recent search's filter parameters
- Test that search history suggestions do not appear when `Enable Search Suggestions` is disabled
- Test that deleting the search history and cache under Privacy Settings permanently deletes the search history (and cache)

## Desktop

- **OS:** OpenSUSE
- **OS Version:** TW

## Additional context
The `src/renderer/modules/search-history.js` implementation is made in a certain way to accommodate possibilities of features like #5003 while still giving space for further optimization if other routes are taken. That's part of why this uses `route` as the `_id` instead of `name`, and keeping >14 search history entries at a time even if the older entries are not currently accessible through the UI yet, to name a couple of examples.

I would also contend that separating search history by profile is too arguable and out of scope to include in this PR.


static/locales/en-US.yaml
```diff
@@ -469,10 +469,10 @@ Settings:
     Remember History: Remember History
     Save Watched Progress: Save Watched Progress
     Save Watched Videos With Last Viewed Playlist: Save Watched Videos With Last Viewed Playlist
-    Clear Search Cache: Clear Search Cache
-    Are you sure you want to clear out your search cache?: Are you sure you want to
-      clear out your search cache?
-    Search cache has been cleared: Search cache has been cleared
+    Clear Search Cache and History: Clear Search Cache and History
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 472: I could be a dumbass but when i looked at this in the FT UI i thought, "why would this remove my History". I linked this somehow to Watch History even though the button for that is just beside that. Maybe could be changed to `Clear Search Cache and Search History` or `Search Cache / History` (the latter way would be the same as with the button `Remove All Profiles / Subscription`)

kommunarr: I'll try flipping it if that helps!

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Manually removing all the characters does result in seeing the Recent Searches but using the Clear button doesnt.

https://github.com/user-attachments/assets/e733465f-a674-44d8-a384-07538e68be0e

Pure visually but it looks kind of weird that the search bar removes the Clear button for a sec causing the elements to move but makes the Clear button reappear after 1ms. Maybe it should behave more like 2nd video.

https://github.com/user-attachments/assets/b3b7ec17-7040-401c-a57d-66c5191b4ebf


https://github.com/user-attachments/assets/38f43d3f-8cfd-4ffe-95dd-fedfbfd71a66

pressing spacebar throws errors

https://github.com/user-attachments/assets/5be3bcb6-6a30-4d23-a7ea-1170a4561972


Like seen in the vids above Linus Tech Tips was in the recent searches but sometimes i doesnt appear and i dont know why that happens

image

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc: LGTM thanks for unbreaking my brain

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc Thank you for your speedy and attentive testing! Fixed above issues now - the not showing of the current history entry when you're on that same page is actually intentional, but I can change that if it's confusing from a user perspective.

efb4f5ff-1298-471a-8973-3d47447115dc:
Pressing spacebar gives me 2 things from my history. I think i expect all of them or none as i didnt actually type something yet. I think showing all would be the better way to go

https://github.com/user-attachments/assets/9477a337-fb32-42b8-829a-93d80359c081

Press and Hold moves searchbar. This doesnt happen with search suggestions

https://github.com/user-attachments/assets/1c15ea93-0e1f-4114-b520-b960ea8de76b

> Pure visually but it looks kind of weird that the search bar removes the Clear button for a sec causing the elements to move but makes the Clear button reappear after 1ms. Maybe it should behave more like 2nd video.

This one doesnt seem to be addressed

> the not showing of the current history entry when you're on that same page is actually intentional, but I can change that if it's confusing from a user perspective.

I think it was quite confusing as i just searched for it

kommunarr:
Fixed all of these issues now - now that I have, will also work on refactoring some of the logic to make this much easier. Stand by

absidue:
Could you please rebase your branch on top of the development branch, as GitHub's diff seems to be confused about which changes actually belong to this branch and is showing changes from #6398 in the `src/main/index.js` file. This is likely due to the fact that both modified the same file and the branch used for this pull request is 39 commits behind (about 1 week looking at the commits that are missing) at the time of writing. In the future it is probably a good idea to make sure that your branch is up-to-date when you submit a pull request to avoid confusing GitHub.

branch-behind-screenshot
screenshot-of-diff

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
This one seems to be back. If i press clear button i dont see the history

https://github.com/user-attachments/assets/ac84912f-9ba7-430e-872c-bffc2520c839

kommunarr:
🤦 Accidentally removed this line in the refactor, my apologies

efb4f5ff-1298-471a-8973-3d47447115dc:
Im not sure what triggered this but if i click to open filter or select them it is very slow at doing that


https://github.com/user-attachments/assets/8a19f75c-da47-49ef-9840-4b02b4c23d74

kommunarr:
Hm, I don't think it's connected to this PR specifically from what I can tell, but something to keep track of for sure

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc Can you reproduce it on the development branch?

efb4f5ff-1298-471a-8973-3d47447115dc:
Indeed not specific to this PR as i can reproduce it on the dev branch

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc Fixed this and also added a remove option.

> Purely visual, if i clear by removing all the characters manually then the search suggestion box stays visible when. When i use the clear button then the search suggestion box blinks for a second

This one is somewhat intractable to fix and appears to also happen for the YT search box.

On another note, it does appear that the YT search history results do show up as you keep typing, but it's very janky. For some reason, one of three things keep happening when I try to test with their implementation:
1. no search history results come up at all
2. a single search history result shows up if it's a full match (i.e., entire search string == entire query string)
3. all matching search results show up as you type (what you would expect)

This PR is currently doing 1, but I'm open to 3 or 2 as well.

efb4f5ff-1298-471a-8973-3d47447115dc:
yeah lets do It I think adding 2 and 3 would be a nice to have.

kommunarr:
> all matching search results show up as you type (what you would expect)

Implemented dynamic search history entries in the results through this proposal and have updated the description & testing criteria accordingly.

efb4f5ff-1298-471a-8973-3d47447115dc:
Cant use left and right arrow on suggestions

https://github.com/user-attachments/assets/521c5d8c-1dff-4624-a208-ef29d68d7418


Unable to see full search suggestion 

before

Screenshot (356)

now

https://github.com/user-attachments/assets/6015fd8d-c662-413a-9da5-bc43b151b813

> Test that search history suggestions do not appear when Enable Search Suggestions is disabled

I expect none of my searches to be added when the setting is disabled

https://github.com/user-attachments/assets/990cd1d6-3d17-4b62-9b58-b149b3440835

> Test that search queries with matching search history entry names have up to 4 of those matching search history entries appear in the results, and in any case, the total # of search results does not exceed 14

It kinda acts that i've never searched "Linus Tech Tips" because it is capped at 4 so the suggestion are like hey maybe you want "Linus Tech Tips" but i already have that one in my history?! Maybe just remove the limit?

https://github.com/user-attachments/assets/0afff488-160f-4440-8bff-21b1e12e9166

There seems to be a performance hit all over the place. This vid is just an example:

Normally

https://github.com/user-attachments/assets/98b6dba6-a9cc-45e2-a1e1-dfb7cf0e99cf

Now

https://github.com/user-attachments/assets/3e2f36bf-f491-4b78-8ba3-4ac780bd4a6c

kommunarr:
> Cant use left and right arrow on suggestions

Fixed!

> Unable to see full search suggestion

YT Web appears to just extend the search box infinitely instead:

Screenshot_20241228_134227

I just reverted now to the old way of having search results be able to cover multiple lines, although I'm not much of a fan. I guess we can keep any changes to this out of scope for this PR.

> I expect none of my searches to be added when the setting is disabled

I'm not sure if that is necessarily true (`Enable Search Suggestions` is a General setting related to preferred app usage, whereas what you're suggesting is more of a Privacy setting), but added now regardless. I suppose we'll let users chime in following the release that includes these changes on if they want more granular privacy settings related to this.

> It kinda acts that i've never searched "Linus Tech Tips" because it is capped at 4 so the suggestion are like hey maybe you want "Linus Tech Tips" but i already have that one in my history?! Maybe just remove the limit?

This was an issue with the implementation being heuristic - was previously getting up to 4 matching results and then sorted by length ascending. Updated now to do it the other way around to ensure it will always get the best match even in edge cases, just a bit less performant now in the average case.

> There seems to be a performance hit all over the place. This vid is just an example:

Hm, can't replicate, at least with my latest changes.

efb4f5ff-1298-471a-8973-3d47447115dc:
Could you provide a build for me. I have a hunch that these issues are maybe related to my dev environment

kommunarr:
https://github.com/kommunarr/FreeTube/actions/runs/12529492523

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM🚀🚀🚀

Edit: performance issues were related to dev mode, see https://github.com/FreeTubeApp/FreeTube/pull/6463#issuecomment-2564429129

PikachuEXE:
Pending https://github.com/FreeTubeApp/FreeTube/pull/6469 to test in dev

PikachuEXE:
Long text = boom
image

kommunarr:
@PikachuEXE Good catch! Missed this when I reverted the result truncation code.

Screenshot_20241228_200237

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

kommunarr:
Renamed existing `Remember History` setting to `Remember Watch History` and added `Remember Search History` setting as per suggestions in the Matrix chat.

efb4f5ff-1298-471a-8973-3d47447115dc:
>  I still allow pre-existing search history suggestions to show up if Enable Search Suggestions is enabled to prevent entangling the logic with a hidden rule, but I can change this if desired.

I dont think this is desired as we kind of do the same for Watch history. If Remember Watch History is disabled the pre-existing history can still be shown in the History tab. The user can decide themselves if they want to delete it or not

efb4f5ff-1298-471a-8973-3d47447115dc:
Remember Search History -> Enabled
Enable Search Suggestions -> Disabled

If i search for something i expect for my search query to be saved. It shouldnt be dependent on search suggestions anymore

kommunarr:
@efb4f5ff-1298-471a-8973-3d47447115dc Implemented now in [c27b89b](https://github.com/FreeTubeApp/FreeTube/pull/6414/commits/c27b89b502d10dd061c15fea4825719f28f312d2)

efb4f5ff-1298-471a-8973-3d47447115dc:
> Remember Search History -> Enabled
Enable Search Suggestions -> Disabled

> If i search for something i expect for my search query to be saved. It shouldnt be dependent on search suggestions anymore

This isnt implemented

kommunarr:
> Remember Search History -> Enabled
    Enable Search Suggestions -> Disabled

>  If i search for something i expect for my search query to be saved. It shouldnt be dependent on search suggestions anymore


This was added in https://github.com/FreeTubeApp/FreeTube/pull/6414/commits/613fd44d980c1d06ca19d0577c676beea9f9ad9b, unless I'm misreading:
https://github.com/FreeTubeApp/FreeTube/blob/c27b89b502d10dd061c15fea4825719f28f312d2/src/renderer/views/Search/Search.js#L151-L153

efb4f5ff-1298-471a-8973-3d47447115dc:
https://github.com/user-attachments/assets/05c6db53-40fe-4a8e-afde-01a07b7a8417

kommunarr:
Updated now as per John's suggestions and modified the testing criteria in the OP:
> - Test that new search histories are not saved, and old search history entries do not appear in the search results, when Remember Search History is disabled and Enable Search Suggestions is enabled
> - Test that search history entries appear in the search results when Remember Search History is enabled and Enable Search Suggestions is disabled

efb4f5ff-1298-471a-8973-3d47447115dc:
It should still show the previously made searches

https://github.com/user-attachments/assets/0460c6c7-0e74-402e-a77a-d2a659e2e0c2


Search suggestions on and search history turned off. Removing all characters results in seeing the previously inserted query

https://github.com/user-attachments/assets/b571a91d-95e6-4c90-ac3f-919c898fe0a7

kommunarr:
Updated now - I misread this prior comment of yours as disagreeing with my statement and implemented the exact opposite of what you were suggesting:

> > I still allow pre-existing search history suggestions to show up if Enable Search Suggestions is enabled to prevent entangling the logic with a hidden rule, but I can change this if desired.

> I dont think this is desired as we kind of do the same for Watch history.

kommunarr:
Discussed in Matrix chat with another contributor some pre-existing issues we discovered with the main search bar:
1. Clicking near the top of the search bar moves the cursor all the way to the left. Thinking about it now, I do remember that this might be an intended feature, although I can't seem to find the original feature request.
2. Pressing the down arrow through search results enough times locks the search bar input value, such that backspacing doesn't work until arrowing up or down again. Image of the bug below:

Screenshot_20250102_172443

3. When a dropdown option is selected using the arrow keys, Ctrl+A causes the letter "a" to be typed rather than highlighting the input text.
4.  When a dropdown option is selected using the arrow keys, click to highlight + Backspace doesn't appear to work properly.

These can be addressed in a separate PR.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/views/Search/Search.js
```diff
@@ -108,6 +110,15 @@ export default defineComponent({
     this.checkSearchCache(payload)
   },
   methods: {
+    updateSearchHistoryEntry: function () {
+      const persistentSearchHistoryPayload = {
+        _id: this.query,
+        lastUpdatedAt: new Date()
```
absidue, line 116: As this field is just used for sorting, we might as well save it as a number directly.
```suggestion
        lastUpdatedAt: Date.now()
```

absidue (commented):
As the left and right arrow keys are already used for moving left and right in your search query and the enter key is used to trigger a search, it is now really easy to accidentally delete search history entries when using the arrow keys to adjust your search query and enter to search, especially because the highlighting of the remove link is very subtle, so most of the time you don't even realise that it is activated. Which I can say from personal experience as it happened multiple times while I was testing this pull request.

kommunarr:
> As the left and right arrow keys are already used for moving left and right in your search query and the enter key is used to trigger a search, it is now really easy to accidentally delete search history entries when using the arrow keys to adjust your search query and enter to search, especially because the highlighting of the remove link is very subtle, so most of the time you don't even realise that it is activated. Which I can say from personal experience as it happened multiple times while I was testing this pull request.

Thanks for the feedback - is this any different from YT's implementation, and/or do you have a suggestion on how to improve this?

absidue:
I haven't personally experienced YouTube's implementation of it, but here are some suggestions off the top of my head:
1. It would be to call `.preventDefault()` on the arrow key events when you are using them to change the focus so it doesn't move the cursor at the same time, that'll make it more obvious that something else is happening, you'll need to make sure that they still work normally when it isn't being used to change the focus (e.g. no entry selected or a search suggestion).
2. To make the highlight more obvious we could potentially change the colour when it is focused or do something like the outline highlighting that we do for tabs (channels, trending, subscription page) when you focus them with the arrow keys.

kommunarr:
1 is very different from YT's implementation and similar apps, so it might be confusing to familiar users if we change it. Added bolding on selection of the entry now.

absidue:
At the moment when you press the right arrow key while a suggestion is "focused" two things happen at the same time: the cursor in the text field moves one character to the right and the focus changes to the remove button.
Number 1 is about making it only do one thing at a time, either move to the remove button (call preventDefault like we do for the up and down arrow keys to prevent to cursor in the text box moving) or move the cursor in the text box when no remove button is available e.g. youtube suggestion or focus is in the text box.

kommunarr:
[recording.webm](https://github.com/user-attachments/assets/2d8ff550-b849-49af-b9ee-faaa0c7a9170)

absidue:
Okay wow that is weirdly unexpected that YouTube would have it like that, feels like an oversight on their part to me.
I agree that we can leave it like this as users are likely be used to that behaviour and change it if users dislike it, maybe I'm the only person that thinks it's weird and we'll hear nothing.

absidue (approved):
Please create a PR to the docs repo to add the new database file to the data location page, after this has been merged.

PikachuEXE:
What's changed since last review...
This?

> Test that new search histories are not saved, but old search history entries still appear in the search results, when Remember Search History is disabled and Enable Search Suggestions is enabled

kommunarr:
[That's correct](https://github.com/FreeTubeApp/FreeTube/pull/6414/files/ea73ad13e97fbe5f17e4132c0662b31c624ac368..c8b08fe7c78b5cab2be87d9a5620d1fc09092dee)

PikachuEXE (approved):

static/locales/en-US.yaml
```diff
@@ -481,13 +482,14 @@ Settings:
...
-    Remember History: Remember History
+    Remember History: Remember Watch History
+    Remember Search History: Remember Search History
     Save Watched Progress: Save Watched Progress
     Save Watched Videos With Last Viewed Playlist: Save Watched Videos With Last Viewed Playlist
-    Clear Search Cache: Clear Search Cache
-    Are you sure you want to clear out your search cache?: Are you sure you want to
-      clear out your search cache?
-    Search cache has been cleared: Search cache has been cleared
+    Clear Search History and Cache: Clear Search History and Cache
```
ChunkyProgrammer, line 489: Did we want to remove the old key for the other locales? Or is the plan to keep it for now so it's easier to update the translation?

kommunarr: Yes, I was thinking the latter.

ChunkyProgrammer: Sounds good, approving now

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
A thought that popped up that i wanted to share: maybe allow importing exporting search history and importing from YT

PikachuEXE:
Test that search history suggestions do not appear when `Enable Search Suggestions` is disabled

This seems not true now (and I remember I tested this before approving
Something changed?	

image

efb4f5ff-1298-471a-8973-3d47447115dc:
When `Search Suggestions` is disabled it only shows you the Search History. If user makes a search then it will still be added to the History. If you would not like future searches to be added to the history go to Privacy Settings and disable Remember Search History.

PR #6157, opened 2024-11-13
Update build workflow to add back osx-arm64

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Description

> "In macOS Sequoia, users will no longer be able to Control-click to override Gatekeeper when opening software that isn’t signed correctly or notarized," the brief note reads. "They’ll need to visit System Settings > Privacy & Security to review security information for software before allowing it to run."

https://arstechnica.com/gadgets/2024/09/macos-15-sequoia-the-ars-technica-review/#page-17

x64 would provide same UX as arm64 build
So add arm64 back for macOS

## Screenshots 

image

## Testing 

If you have macOS you can download from
https://github.com/PikachuEXE/FreeTube/actions/runs/11808637281
Run `xattr -dr com.apple.quarantine <path>`

## Desktop
I tested in

- **OS:** macOS
- **OS Version:** 13.7.1 (22H221)
- **FreeTube version:**

PR #6415, opened 2024-12-19
Add missing release builds

efb4f5ff-1298-471a-8973-3d47447115dc:
# Add missing release builds

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/pull/6157

## Description

This PR adds the following release builds:
- AppImage ARMv7l
- AppImage ARM64
- Alpine .apk x64
- Alpine .apk ARMv7l
- Alpine .apk ARM64
- Pacman .pacman x64

## Testing 

As this is a release workflow im not sure how to properly test it. This should just work out of the box. Almost everything could be taken over from the nightly build workflow. The only thing im worried about is the `asset_content_type` for the Pacman build, that was the only real thing i had to figure out as there were varying answers online.

## Additional context

This journey is only part one of many PR's (well i hope not too many) to come on this subject and has been triggered by @PikachuEXE PR. I noticed that we had more nightly builds than release builds so i went on a quest to provide these missing builds to users so they wouldnt need to use the nightly builds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,6 +95,28 @@ jobs:
         asset_path: build/FreeTube-${{ steps.getPackageInfo.outputs.version }}.AppImage
         asset_content_type: application/vnd.appimage
 
+    - name: Upload AppImage ARMv7l Release
+      uses: actions/upload-release-asset@v1
+      if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-armv7l')
+      env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: https://uploads.github.com/repos/FreeTubeApp/FreeTube/releases/${{ secrets.UPLOAD_ID }}/assets{?name,label}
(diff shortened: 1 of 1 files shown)


PikachuEXE:
Can we put linux entries together (instead of some below MacOS entries
Also no entries for `Upload .rpm`?

efb4f5ff-1298-471a-8973-3d47447115dc:
> Can we put linux entries together (instead of some below MacOS entries

I would like to do the grouping of the builds in a separate PR for both the release and build workflow as there is some randomness going on

>  Also no entries for `Upload .rpm`? (Compared with https://github.com/FreeTubeApp/FreeTube/blob/development/.github/workflows/build.yml

I added no .rpm because the release workflow already provides the same ones as the build workflow so im unsure what you mean by this.

PikachuEXE:
`.rpm` is above windows one I missed it...

PikachuEXE (approved):

kommunarr (approved):

absidue (approved):

PR #6423, opened 2024-12-20
Build name convention standardization

efb4f5ff-1298-471a-8973-3d47447115dc:
# Build name convention standardization

## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/pull/6415

## Description

This PR changes the following:
-  We were using a mix match of hyphens and underscores so this changes all builds to use hyphens instead of underscores. Why hyphens? Most release builds were using hyphens and for nightly builds it was about 50/50 so i opted for hyphens as most normal users (the ones that use release builds) are used to seeing hyphens.
- Some build names didnt include the .zip extension so i added them
- if file name contains portable then the following structure is used, platform-architecture-portable. There was a mix of this structure and platform-portable-architecture being used.

## Additional context

I will create a PR for our website that will pull the now adjusted release builds. That PR will be drafted as it should be merged right before release.

This is part 2 of making changes in these workflow. 
Upcoming parts:
- Part 3: Group all builds based on platform
- Part 4: Investigate if all builds have .zip and .7z equivalents

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -117,91 +117,91 @@ jobs:
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-x64')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_linux_portable_x64
+        name: freetube-${{ steps.versionNumber.outputs.result }}-linux-x64-portable.zip
         path: build/freetube-${{ steps.versionNumber.outputs.result }}.zip
 
     - name: Upload Linux .7z x64 Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-x64')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_linux_portable_x64.7z
+        name: freetube-${{ steps.versionNumber.outputs.result }}-linux-x64-portable.7z
         path: build/freetube-${{ steps.versionNumber.outputs.result }}.7z
 
     - name: Upload Linux .zip ARMv7l Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-armv7l')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_linux_portable_armv7l
+        name: freetube-${{ steps.versionNumber.outputs.result }}-linux-armv7l-portable.zip
         path: build/freetube-${{ steps.versionNumber.outputs.result }}-armv7l.zip
 
     - name: Upload Linux .7z ARMv7l Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-armv7l')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_linux_portable_armv7l.7z
+        name: freetube-${{ steps.versionNumber.outputs.result }}-linux-armv7l-portable.7z
         path: build/freetube-${{ steps.versionNumber.outputs.result }}-armv7l.7z
 
     - name: Upload Linux .zip ARM64 Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-arm64')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_linux_portable_arm64
+        name: freetube-${{ steps.versionNumber.outputs.result }}-linux-arm64-portable.zip
         path: build/freetube-${{ steps.versionNumber.outputs.result }}-arm64.zip
 
     - name: Upload Linux .7z ARM64 Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-arm64')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_linux_portable_arm64.7z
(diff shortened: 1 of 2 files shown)


absidue:
We should probably stick with underscores for the artifacts that electron-builder also uses underscores e.g. the deb files where using underscores to separate the name, version and architecture in the file name is a convention.

efb4f5ff-1298-471a-8973-3d47447115dc:
Applied standard naming conventions where it wasnt being applied before.

efb4f5ff-1298-471a-8973-3d47447115dc:
applied WIP as i still need some more changes

efb4f5ff-1298-471a-8973-3d47447115dc:
ready for review!

.github/workflows/build.yml
```diff
@@ -180,28 +180,28 @@ jobs:
...
-        name: freetube_${{ steps.versionNumber.outputs.result }}_arm64.AppImage
+        name: freetube-${{ steps.versionNumber.outputs.result }}-arm64.AppImage
         path: build/FreeTube-${{ steps.versionNumber.outputs.result }}-arm64.AppImage
 
     - name: Upload .rpm x64 Artifact
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-x64')
       with:
-        name: freetube_${{ steps.versionNumber.outputs.result }}_amd64.rpm
+        name: freetube-${{ steps.versionNumber.outputs.result }}.amd64.rpm
```
PikachuEXE, line 204: I know nothing about rpm files so just want to know why dot is used instead of `-`

.github/workflows/build.yml
```diff
@@ -322,26 +322,26 @@ jobs:
       uses: actions/upload-artifact@v4
       if: startsWith(matrix.os, 'macos') && startsWith(matrix.runtime, 'osx-x64')
       with:
-        name: freetube-${{ steps.versionNumber.outputs.result }}-mac-x64.zip
+        name: freetube-${{ steps.versionNumber.outputs.result }}-mac-x64-portable.zip
```
PikachuEXE, line 325: Not objecting this but I found that projects rarely got portable suffix to zips & dmg for mac
But it's still fine to be extra clear

.github/workflows/build.yml
efb4f5ff-1298-471a-8973-3d47447115dc: Its the name convention that is used by redhead

https://kerneltalks.com/tools/understanding-package-naming-convention-rpm-deb/

.github/workflows/build.yml
efb4f5ff-1298-471a-8973-3d47447115dc: I can revert it if you want. I only added this because some of our portable builds did have it and others not

PikachuEXE: Ya I think it's better to revert.
There is no portable concept for macOS apps (mostly). The data is stored in user directory anyway.
I would even argue that the windows one isn't portable as well. Since the data isn't stored inside the same folder as the app folder (and no way to change that). It's just another way of installation but the data isn't "portable".

efb4f5ff-1298-471a-8973-3d47447115dc: > I would even argue that the windows one isn't portable as well. Since the data isn't stored inside the same folder as the app folder (and no way to change that). It's just another way of installation but the data isn't "portable".

A conversation i had before opening this PR

image

.github/workflows/release.yml
```diff
@@ -124,7 +124,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         upload_url: https://uploads.github.com/repos/FreeTubeApp/FreeTube/releases/${{ secrets.UPLOAD_ID }}/assets{?name,label}
-        asset_name: freetube-${{ steps.getPackageInfo.outputs.version }}-linux-portable-x64.zip
+        asset_name: freetube-${{ steps.getPackageInfo.outputs.version }}-linux-x64-portable.zip
```
ChunkyProgrammer, line 127: I think the flatpak.yml file would need to be updated in a few places.

Ex:
https://github.com/FreeTubeApp/FreeTube/blob/550ce7739bbee8662042cae148cd8e84ebb5542d/.github/workflows/flatpak.yml#L54

ChunkyProgrammer (commented):
Note: these changes could break some workflows for some of the unofficial downloads.

We should probably notify the maintainer of the PortableApps version of FreeTube that this change was made as well as other maintainers of unofficial downloads if we think their workflow might break from this.

efb4f5ff-1298-471a-8973-3d47447115dc:
> Note: these changes could break some workflows for some of the unofficial downloads.
> 
> We should probably notify the maintainer of the PortableApps version of FreeTube that this change was made as well as other maintainers of unofficial downloads if we think their workflow might need to be modified for these changes

@ChunkyProgrammer i have notified all maintainers of the unofficial downloads except for WAPT and Homebrew (i think this is pika though)

ryand56:
Thanks for letting us know about the changes, though no changes will be needed on our end as we are building from source now.

PikachuEXE:
My homebrew tab also builds from source = not relying on those names
https://github.com/PikachuEXE/homebrew-FreeTube/blob/master/.github/workflows/release-manual-trigger.yml

rddim:
@ChunkyProgrammer and @efb4f5ff-1298-471a-8973-3d47447115dc 

Thank you very much for the notification. I'm not affected from the new changes because I use the only the final releases.

efb4f5ff-1298-471a-8973-3d47447115dc:
Are there still things that we need to consider? Are there maybe some reservations about this PR that i can address?

absidue (approved):
We'll need to remember to update the files in the flathub repo that reference the linux zip files.

PikachuEXE (approved):

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Im on it!

PR #6436, opened 2024-12-22
Switch back to the WEB client for the local API with potokens

absidue:
# Switch back to the WEB client for the local API with potokens

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Description

This pull request switches back to the WEB client as a way to hopefully avoid the 403s that have started to appear for some people. As an added bonus this means we can use the DASH manifests for live streams which have 4 hours of seeking, the alternative would be to use the HLS manifest but as the WEB ones don't support audio-only or seeking, using the DASH ones seemed like a better idea.

To make this possible, I added the `bgutils-js` dependency to help generate the poTokens required for using the WEB client these days. As the BotGuard code accesses the global `window` and `document` objects, I decided it would be best to isolate it to its own webview with a separate temporary session and reduced permissions compared to our normal webviews that the user interacts with. That also required introducing a separate webpack config to be able to generate the script to run inside of that webview.

## Testing

Test video playback with the local API, both normal videos and live streams. If you can find a movie with a trailer that isn't dubbed you can test that too (dubbed trailers are known to be an issue, i'll be creating a PR to youtube.js to fix the DASH manifest generation).

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** f419aa56c4c3a78097fe11d72ab8ea5233b2996b


PikachuEXE (dismissed):

PikachuEXE:
Post merge action: merge https://github.com/FreeTubeApp/FreeTube-Docs/pull/157

efb4f5ff-1298-471a-8973-3d47447115dc:
ofcourse...

absidue:
It's because autolinker and bgutils-js are next to each other in the package.json file :/

PikachuEXE (approved):
Approve before another conflict appears~
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

Issue #2295, opened 2022-06-05
[Feature Request]: Override default video player settings


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have searched the [issue tracker](https://github.com/FreeTubeApp/FreeTube/issues) for a feature request that matches the one I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.

### Problem Description

When https://github.com/FreeTubeApp/FreeTube/issues/2263 was opened up. I noticed how many issues were actually opened for the same request. So here we are making a big summary.

Why do we want this? Let say in the settings i have set my default video quality to 1080p. My internet provider is giving me a hard time today so while im watching a video i switch from 1080p to 360p in the video player. The next video should also run in 360p because i just have overwriting the default quality because my provider giving me a hard time. 

I could write usecases for the rest but i assume u get my point. U change it in the player for a reason.

### Proposed Solution

Player buttons from the left to right, that already do this.

- [x] Volume
- [ ] Playback speed
- [ ] Captions
- [ ] PIP
- [x] Theatre mode*
- [ ] Fullwindow (can only be implemented with autoplay because u have to go out of this mode to select another video when autoplay is disabled)
- [ ] Video Quality
- [ ] Fullscreen

* There is a setting called ```Enable Theatre Mode by Default```. When this is enabled Theatre mode will always get forced on u on the next video. If this is disabled and u switch from default view to theatre mode view in the video player. It will retain these setting.

### Alternatives Considered

I do understand that some people like it that the default settings overwrite the player settings when a next video is being played. We maybe could have 2 modes?

1. Player settings is always overwritten by default settings when another video is being player.
2. Player settings dont get overwritten and stay like u set them to be for that session.

### Issue Labels

ease of use improvement

### Additional Information

I understand that this will be a difficult and long discussion. The point of this issue is to centralize the discussion to this issue instead of 8 other issues.

Allot of discussion has already taken place [here](https://github.com/FreeTubeApp/FreeTube/pull/1281).

anonaddict:
Makes sense to keep everything for the session, if someone is watching a video playlist in full screen odds are they want to continue watching in fullscreen instead of having to focus freetube and manually fullscreen every video. Especially not if it is shorter videos like vines (if those still exist) or music (2-4 minute long videos generally)

[...]

PR #6449, opened 2024-12-24
Feat remember playbackrate setting

### LINKED ISSUE #6448: [Feature Request]: setting to remember playback rate between videos
### Guidelines

### Problem Description

My preferred playback rate is 3.0. But sometimes I want to listen to a playlist of music. And I prefer to listen to the music with playback rate of 1.0. But on every new song in a playlist, playback rate will be reset to 3.0. Current solution is to go to settings, change default playback rate to 1, go to playlist, listen, then don't forget to get back to settings and change rate back to 3.0. 

### Proposed Solution

Add a setting that will remember playback rate, when I change it on a video.

### Alternatives Considered

The alternative is to make this behavior by default.

### Issue Labels

ease of use improvement

### Additional Information

Idea is from "Video Speed controller" plugin for Chrome/Firefox

ikizey:
# Title

## Pull Request Type
- [x] Feature Implementation

## Related issue
closes #6448 

## Description
Pull request adds new setting "remember playback rate". If the setting is enabled, when a user changes playback rate on a video, it will be saved as the default playback rate.
The setting is off by default, to preserve current configuration.

## Testing 
Manually tested changing playback rate using ui, keyboard shortcut and mouse scroll.

## Additional context
Closing issue created by me.


kommunarr:
Hi @ikizey, thank you for working on this. I do think this idea needs more discussion before deciding on an implementation. Here are my two cents:

- This does not needs a new setting; reasonable defaults are better than total configurability, the latter of which can lead to "settings overload".
- Instead of permanent changes to the value, people are most likely expecting it to work like theater mode does currently, where changing the value affects it just for the session instead.
- Someone can correct me if I'm wrong on this, but I believe that we require that all non-English translations are to be done through [Weblate](https://hosted.weblate.org/engage/free-tube/). If we wanted to use an automated tool for generating these strings when no manual translation existed, we would configure that through [Weblate](https://docs.weblate.org/en/latest/admin/machine.html).

efb4f5ff-1298-471a-8973-3d47447115dc:
I agree with @kommunarr. There is no extra setting needed here. Users sets default playback speed to 1x -> user watches video -> changes playback speed to 2x -> next video still has 2x speed -> User closes the application -> Open application -> watch video -> video uses default playback speed.

PikachuEXE:
I would argue that it should be like https://github.com/FreeTubeApp/FreeTube/pull/6400: exit watch page = reset to default value
Not sure what's the current behaviour (never use it

ikizey:
@kommunarr @efb4f5ff-1298-471a-8973-3d47447115dc 
Right now, theater mode works this way:
you open video (watch view), theater mode sets to default; you change mode; you change video; setting persists; you leave watch view to subscriptions for example, you open another video, mode resets (sets to default). That is not, exactly, what you are saying.

I updated the code to do the same thing with playback rate. (PR description updated)

kommunarr (approved):
Great work!

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -123,13 +123,18 @@ export default defineComponent({
...
+    },
   },
   emits: [
     'error',
     'loaded',
     'ended',
     'timeupdate',
-    'toggle-theatre-mode'
+    'toggle-theatre-mode',
+    'current-playback-rate-update'
```
absidue, line 137: Slightly shorter event name with the same meaning.
```suggestion
    'playback-rate-updated'
```

src/renderer/views/Watch/Watch.js
```diff
@@ -133,7 +133,8 @@ export default defineComponent({
       customErrorIcon: null,
       videoGenreIsMusic: false,
       /** @type {Date|null} */
-      streamingDataExpiryDate: null
+      streamingDataExpiryDate: null,
+      currentPlaybackRate: parseFloat(this.$store.getters.getDefaultPlayback)
```
absidue, line 137: Can you please access the store in the created lifecycle hook, instead of in data. In the options API the only thing you are supposed to access at that point is the props, everything else is usually only ready when the created lifecycle hook fires.

I know we are slowly moving away from the options API but while we still have options API components we should still write their code correctly.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2413,6 +2413,10 @@ export default defineComponent({
       container.value.classList.add('no-cursor')
 
       await performFirstLoad()
+
+      player.addEventListener('ratechange', () => {
+        emit('current-playback-rate-update', video.value.playbackRate)
```
absidue, line 2418: As you are listening to the player and not the video element you should also read the value from the player (as you are combining two separate things the values can differ).
```suggestion
        emit('current-playback-rate-update', player.getPlaybackRate())
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -123,13 +123,18 @@ export default defineComponent({
       type: String,
       default: null
     },
+    currentPlaybackRate: {
+      type: Number,
+      default: () => 1.0
```
absidue, line 128: As JavaScript doesn't differentiate between integers and floating point numbers, 1.0 = 1, so we might as well write it like that (saves terser having to do it for us at build time).
```suggestion
      default: () => 1
```

absidue:
I have fixed the issue linked in the pull request description to point to the correct one for you (the one you had originally linked was your own issue that was closed as a duplicate of an existing one).

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -123,13 +123,18 @@ export default defineComponent({
       type: String,
       default: null
     },
+    currentPlaybackRate: {
+      type: Number,
+      default: () => 1
```
absidue, line 128: Numbers are primitives and immutable so they get copied every time, you only need the function when you need to create a new instance of something every time e.g. objects and arrays.
```suggestion
      default: 1
```

src/renderer/views/Watch/Watch.js
```diff
@@ -304,6 +305,7 @@ export default defineComponent({
     this.activeFormat = this.defaultVideoFormat
 
     this.checkIfTimestamp()
+    this.currentPlaybackRate = parseFloat(this.$store.getters.getDefaultPlayback)
```
absidue, line 308: The value is already a number (as `parseFloat` takes a string, this would convert it from a number to a string and then back to a number).
```suggestion
    this.currentPlaybackRate = this.$store.getters.getDefaultPlayback
```

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
one step closer to #2295. Feel free to take a stab at more player controls

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @ikizey , we are currently fixing https://github.com/FreeTubeApp/FreeTube/issues/6668 in PR https://github.com/FreeTubeApp/FreeTube/pull/6943 but i noticed something weird and wanted to ask you if this is the originally intended behavior.

Default playback rate is set to 1x in the setting -> playback rate has been changed while watching the video, 9x -> next video uses playback rate of previous video so still 9x-> ctrl click resets to 9x instead of the default playback rate predefined in the settings


https://github.com/user-attachments/assets/62759840-7677-4335-9ed4-8b38251f6c99


Edit:

The wording in the tooltip seems to indicate that it should always reset to the value set in the settings

135803944-907060fe-41c6-4a18-a1e8-5c86ecb7b585

Issue #4446, opened 2023-12-12
[Feature Request]: More Command Line Arguments


ChunkyProgrammer:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

I'd like to be able to use command line arguments with FreeTube

### Proposed Solution

Add the following comnands/arguments
<table>
<tr><td><a href="https://github.com/FreeTubeApp/FreeTube/discussions/2100#discussioncomment-7776806">open new window</a></td>
<td>an argument that opens a new window</td>
</tr>
<tr><td><a href="https://github.com/FreeTubeApp/FreeTube/issues/1619">search freetube</a></td>
<td>a command that searches a given query</td>
</tr>
<tr><td><a href="https://github.com/FreeTubeApp/FreeTube/issues/192">export subscriptions</a></td>
<td>a command that exports subscriptions. (Should it export to different files, ex: newpipe subscriptions?)</td>
</tr>
<tr><td><a href="https://github.com/FreeTubeApp/FreeTube/issues/1397">download video</a></td>
<td>a command that downloads a video (would we want to implement this since ytdl already exists?)</td>
</tr>
<tr><td>help</td><td>display different commands/arguments available</td>
</tr>
</table>

### Alternatives Considered

Dont add more command line arguments

### Issue Labels

ease of use improvement, improvement to existing feature

absidue:
> a command that downloads a video (would we want to implement this since ytdl already exists?)

Probably not, as youtube-dl and yt-dlp already exist and have many more options and features, as they are dedicated download tools.

rhysday:
I requested command line search in 2021. Hopefully this gets some traction as launching YT searches straight from dmenu that open up freetube would be great.

Salamimonster:
To make FreeTube more usable (which in almost all circumstances means: faster), I use a **Autohotkey** script as a workaround, to start a search.

```
LALT & y::
Run, <yourpathhere>FreeTube.exe
WinWait, Abonnements - FreeTube ;window name may be different in other languages
[...]

PR #6455, opened 2024-12-26
Add --help and --new-window flag and fix --version flag

Zander671:
# Add --help and --new-window flag and fix --version flag

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

#4446 

## Description

This contains three commits. The first adds a `--new-window` argument. This argument causes a new window to be opened in an existing instance. If there is no existing instance, it creates a new instance. The second fixes the `--version` flag that got broken by #2606. The third adds a `--help` flag.

## Screenshots 

screenshot-2024-12-25--21-12-33
screenshot-2024-12-25--21-12-46

## Testing 

The `--help` and `--version` flags were (hopefully) trivial to test. Their output can be seen above. The scope of the modifications required for the `--new-window` flag is also fairly small. I tested starting freetube both with and without the `--new-window`  flag and both with and without a url. It worked in all four cases.

## Desktop

- **OS:** ArchLinux
- **OS Version:** 6.12.4-zen1-1-zen
- **FreeTube version:** v0.22.1

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -253,14 +253,24 @@ function runApp() {
     }
 
     app.on('second-instance', (_, commandLine, __) => {
-      // Someone tried to run a second instance, we should focus our window
+      // Someone tried to run a second instance
       if (typeof commandLine !== 'undefined') {
         const url = getLinkUrl(commandLine)
         if (mainWindow && mainWindow.webContents) {
-          if (mainWindow.isMinimized()) mainWindow.restore()
-          mainWindow.focus()
+          if (commandLine.includes('--new-window')) {
+            // The user wants to create a new window in the existing instance
+            if (url) startupUrl = url
+            createWindow({
+              showWindowNow: true,
+              replaceMainWindow: true,
+            })
+          } else {
+            // Just focus the main window (instead of starting a new instance)
+            if (mainWindow.isMinimized()) mainWindow.restore()
+            mainWindow.focus()
 
-          if (url) mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+            if (url) mainWindow.webContents.send(IpcChannels.OPEN_URL, url)
+          }
         } else {
           if (url) startupUrl = url
           createWindow()
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -28,8 +28,21 @@ const brotliDecompressAsync = promisify(brotliDecompress)
 if (process.argv.includes('--version')) {
   console.log(`v${app.getVersion()}`) // eslint-disable-line no-console
   app.exit()
+} else if (process.argv.includes('--help') || process.argv.includes('-h')) {
+  printHelp()
+  app.exit()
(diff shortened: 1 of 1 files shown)


kommunarr (commented):
This is working great for me!

**suggestion (non-blocking):** A non-recognized argument error might be a good idea for preventing an accidental app load from a typo if we don't mind slightly expanding the scope. It could be frustrating to type `-version` and accidentally launch a window. E.g., Here's a similar error from `nano` for reference:

```
nano: unrecognized option '--potato'
Type 'nano -h' for a list of available options.
```

Zander671:
The problem (at least on Linux) with validating arguments in this way is that Electron parses arguments like `--enable-features=UseOzonePlatform` and `--ozone-platform=wayland`. While it would be possible to check against a list of all Electron's supported flags, this seems like it might be hard to maintain as new flags are added (unless there is a way to get all supported arguments programmatically, but a quick search yielded no results).

kommunarr (approved):
Ah, I completely forgot about that. Here's how Codium handles this for a more similar example, where they show a warning but still continue to launch. Will approve now, as I'm ~not sure of~ [not a fan of](https://github.com/microsoft/vscode/blob/0a66dc39ff0a3476dd5f711df232cba4d713a57d/src/vs/platform/environment/node/argv.ts#L49-L216) the effort to get this list of known options:
> Warning: 'potato' is not in the list of known options, but still passed to Electron/Chromium.

ChunkyProgrammer (dismissed):

src/main/index.js
```diff
@@ -26,11 +26,25 @@ import packageDetails from '../../package.json'
 const brotliDecompressAsync = promisify(brotliDecompress)
 
 if (process.argv.includes('--version')) {
+  console.log(`v${app.getVersion()}`) // eslint-disable-line no-console
```
absidue, line 29: To avoid confusion about version mismatches, it is probably best to get it as close to the one shown on the About page as possible (it won't be translated like on the about page but at least it will have the beta text).
```suggestion
  console.log(`v${packageDetails.version} Beta`) // eslint-disable-line no-console
```

absidue (commented):
I haven't got around to testing this yet, but at least from the code it looks like `--version` and `--help` will only be usable if FreeTube is not currently running.

Zander671:
> I haven't got around to testing this yet, but at least from the code it looks like `--version` and `--help` will only be usable if FreeTube is not currently running.

FreeTube calls `app.requestSingleInstanceLock()` and checks the return value in `runApp()`. The `--help` and `--version` flags are checked before `runApp()` ever gets called. This seems to work fine in my testing.

kommunarr (approved):
Can confirm these commands work even with a running window.

MarmadileManteater (approved):

absidue (approved):

Issue #6419, opened 2024-12-20
[Bug]: Expanding description automatically scrolls to the bottom of the description


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

When expanding the description it scrolls all the way to the bottom

https://github.com/user-attachments/assets/f83e2faf-dcba-4e03-ab68-cc01c5884046

### Expected Behavior

Do not scroll down to the bottom. Position shouldnt change.

### Issue Labels

usability issue, visual bug

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/0fcf1bec04ce80d6f39e425e32d10f94d342ce84

### Operating System Version

Win10 22H2, Ubuntu dev

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

PR #6461, opened 2024-12-27
Fix a few issues with collapsable video description

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/pull/5665#issuecomment-2528890790

## Description

Fixes for https://github.com/FreeTubeApp/FreeTube/pull/5665

## Screenshots 

https://github.com/user-attachments/assets/cc826646-6aec-43ef-92a1-d3c6b8182944

## Testing 

- Test like https://github.com/FreeTubeApp/FreeTube/pull/5665
- Then ensure issues in https://github.com/FreeTubeApp/FreeTube/pull/5665 fixed

Example videos:
- Short description - https://youtu.be/27iu423Yjtc
- No description - https://youtu.be/ASeJV9KGAOc
- Long description - any LTT video e.g. https://youtu.be/1arS0zAyJWM

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

- No idea why `flex-direction: column-reverse` is used which is the cause of scroll to bottom on expand. Let me know if that shouldn't be changed
- Overlay element removed to solve 2nd issue

### DIFF
--- a/src/renderer/components/FtTimestampCatcher.vue
+++ b/src/renderer/components/FtTimestampCatcher.vue
@@ -1,5 +1,7 @@
 <template>
   <p
+    @click="onClick"
+    @keydown.space.prevent="onClick"
     @timestamp-clicked="catchTimestampClick"
     v-html="displayText"
   />
@@ -37,7 +39,13 @@ const displayText = props.inputHtml.replaceAll(/(?:(\d+):)?(\d+):(\d+)/g, (times
   return `<a href="${url}" onclick="event.preventDefault();this.dispatchEvent(new CustomEvent('timestamp-clicked',{bubbles:true,detail:${time}}));window.scrollTo(0,0)">${timestamp}</a>`
 })
 
-const emit = defineEmits(['timestamp-event'])
+const emit = defineEmits(['text-click', 'timestamp-event'])
+
+function onClick(e) {
+  // Ignore link clicks
+  if (e.target.tagName === 'A') { return }
+  emit('text-click')
+}
 
 /**
  * @param {CustomEvent} event
--- a/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
+++ b/src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
@@ -25,19 +25,13 @@
       >
         {{ $t("Description.Expand Description") }}
       </span>
-      <div
-        v-if="!showFullDescription"
-        class="overlay"
-        @click="expandDescription"
-        @keydown.space.prevent="expandDescription"
-        @keydown.enter.prevent="expandDescription"
-      />
     </template>
     <FtTimestampCatcher
       ref="descriptionContainer"
       class="description"
       :input-html="shownDescription"
       @timestamp-event="onTimestamp"
+      @text-click="expandDescription"
     />
   </FtCard>
 </template>
--- a/src/renderer/components/WatchVideoDescription/WatchVideoDescription.css
+++ b/src/renderer/components/WatchVideoDescription/WatchVideoDescription.css
@@ -1,7 +1,7 @@
 .videoDescription {
   position: relative;
   display: flex;
-  flex-direction: column-reverse;
+  flex-direction: column;
   overflow-y: auto;
   max-block-size: 20lh;
 }
@@ -20,6 +20,10 @@
   color: var(--title-color);
 }
 
+.videoDescription.short {
+  cursor: pointer;
+}
+
 .videoDescription.short .descriptionStatus {
   position: absolute;
   inset-block-end: calc(1lh + 12px);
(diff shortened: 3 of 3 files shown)


kommunarr:
**suggestion:** I recommend keeping the tab order the same as the visual. It's a bit unexpected to have `Show less` not appear in the tab order at the end.

**issue (blocking):** Clicking on the `...more` button doesn't seem to work.

efb4f5ff-1298-471a-8973-3d47447115dc:
Closes https://github.com/FreeTubeApp/FreeTube/issues/6419?

efb4f5ff-1298-471a-8973-3d47447115dc:
I think the users shouldnt be able to tab to links within the hidden description part

https://github.com/user-attachments/assets/9b58cecb-bb7d-47c9-abb2-bd37bf11b4d7

PikachuEXE:
I just follow youtube ...

https://github.com/user-attachments/assets/d00381c0-468b-4212-ac80-1d07de28c3ee

PikachuEXE:
The best I can do now is to prevent tabbing into any link inside description (more button can still be tabbed)

https://github.com/user-attachments/assets/9d4c4534-1848-4158-9a89-cba0a5ca8260

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

kommunarr (dismissed):

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -5,47 +5,45 @@
...
+        class="descriptionStatus"
+        role="button"
+        tabindex="0"
+        @click="collapseDescription"
+        @keydown.space.prevent="collapseDescription"
+        @keydown.enter.prevent="collapseDescription"
+      >
+        {{ $t("Description.Collapse Description") }}
+      </span>
+    </template>
```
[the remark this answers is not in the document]
PikachuEXE: Done and did the same to expand button

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -86,6 +80,16 @@ if (props.descriptionHtml !== '') {
...
 }
 
+const processedShownDescription = computed(() => {
+  if (shownDescription === '') { return shownDescription }
+
+  return processDescriptionHtml(shownDescription, linkTabIndex.value)
+})
+
+const linkTabIndex = computed(() => {
+  return showFullDescription.value ? 0 : -1
```
absidue, line 90: **Non-blocking**: As the values are only used inside of strings, it makes sense to write them as string directly to avoid them needing to be converted from a number into a string at runtime.
```suggestion
  return showFullDescription.value ? '0' : '-1'
```

src/renderer/components/FtTimestampCatcher.vue
```diff
@@ -1,25 +1,32 @@
...
 
 const props = defineProps({
   inputHtml: {
     type: String,
     default: ''
-  }
+  },
+  linkTabIndex: {
+    type: Number,
+    default: 0
```
absidue, line 21:
```suggestion
    type: String,
    default: '0'
```

src/renderer/components/FtTimestampCatcher.vue
```diff
@@ -1,25 +1,32 @@
 <template>
   <p
+    @click="onClick"
+    @keydown.space.prevent="onClick"
```
absidue, line 4: **Non-blocking:** As this component is also used for every single comment under videos, where these event listeners would be unnecessary, it might make more sense to move them to the WatchVideoDescription component with the `.native` modifier added to them so that Vue applies them to the `<p>` element instead of the component instance.

PikachuEXE:
Both done
Forgot `.native` exists (I have never used it)

PikachuEXE:
Though now I wonder how to implement this with vue 3

https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed

ChunkyProgrammer:
> Though now I wonder how to implement this with vue 3
> 
> https://v3-migration.vuejs.org/breaking-changes/v-on-native-modifier-removed

You'd just need to remove the ".native" for vue3 and it should work.

Since we dont have a "click* in the emits of the FtTimestampCatcher, vue3 will use a native event. 
https://github.com/FreeTubeApp/FreeTube/blob/8a771f1f8496d26cf4694c8c1459fe29fc34aeea/src/renderer/components/FtTimestampCatcher.vue#L40

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

Issue #5966, opened 2024-10-27
[Feature Request]: Add back and forward keyboard shortcuts to Mac client


chriscjcj:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

The standard Mac keyboard shortcut for navigating back and forward are Command-[ and Command-] respectively. Currently the Mac Freetube app uses option-[ and option-]. These keyboard shortcuts seem to be trapped by shaka-player and don't function when the player is in focus.

### Proposed Solution

On the Mac client, please add Command-[ as a keyboard shortcut to navigate back to the previous page. And add Command-] as a keyboard shortcut to navigate forward. Not only are these the standard keyboard shortcuts for the Finder and for all browsers, they don't seem to conflict with the shaka-player.

### Alternatives Considered

Allow users to set the keystroke in preferences.

### Issue Labels

ease of use improvement

### Additional Information

https://support.apple.com/en-us/102650

chriscjcj:
@PikachuEXE Thank you for this. And thank you for the Apple Silicon port on homebrew. 🙏🏻 I am grateful!

PR #6466, opened 2024-12-28
Make Cmd left/right go back/forward

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes https://github.com/FreeTubeApp/FreeTube/issues/5966

## Description

For **MacOS only**, make Cmd left/right as an additional keyboard shortcut to go back/forward in history

## Screenshots 

image
image

## Testing 

- Ensure existing back/forward shortcuts working (All OS) (Alt Left/Right)
- Ensure **new** back/forward shortcuts working (macOS only) (Cmd Left/Right)
- Ensure tooltip on back/forward buttons unchanged (non macOS only)
- Ensure tooltip on back/forward buttons updated (macOS only)
- Open keyboard modal, ensure shortcuts unchanged (non macOS only)
- Open keyboard modal, ensure shortcut list updated (MacOS only)

## Desktop

- **OS:** macOS
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
+++ b/src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
@@ -115,6 +115,14 @@ const localizedShortcutNameDictionary = computed(() => {
     ['SHOW_SHORTCUTS', t('KeyboardShortcutPrompt.Show Keyboard Shortcuts')],
     ['HISTORY_BACKWARD', t('KeyboardShortcutPrompt.History Backward')],
     ['HISTORY_FORWARD', t('KeyboardShortcutPrompt.History Forward')],
+    ...(
+      isMac
+        ? [
+            ['HISTORY_BACKWARD_ALT_MAC', t('KeyboardShortcutPrompt.History Backward (Mac only)')],
+            ['HISTORY_FORWARD_ALT_MAC', t('KeyboardShortcutPrompt.History Forward (Mac only)')],
+          ]
+        : []
+    ),
     ['FULLSCREEN', t('KeyboardShortcutPrompt.Fullscreen')],
     ['NAVIGATE_TO_SETTINGS', t('KeyboardShortcutPrompt.Navigate to Settings')],
     (
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -1021,10 +1021,14 @@ export function addKeyboardShortcutToActionTitle(actionTitle, shortcut) {
 
 /**
  * @param {string} localizedActionTitle
- * @param {string} unlocalizedShortcut
+ * @param {string|string[]} sometimesManyUnlocalizedShortcuts
  * @returns {string} the localized action title with keyboard shortcut
  */
-export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle, unlocalizedShortcut) {
-  const localizedShortcut = getLocalizedShortcut(unlocalizedShortcut)
-  return addKeyboardShortcutToActionTitle(localizedActionTitle, localizedShortcut)
+export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle, sometimesManyUnlocalizedShortcuts) {
+  let unlocalizedShortcuts = sometimesManyUnlocalizedShortcuts
+  if (!Array.isArray(sometimesManyUnlocalizedShortcuts)) {
+    unlocalizedShortcuts = [unlocalizedShortcuts]
+  }
+  const localizedShortcuts = unlocalizedShortcuts.map((s) => getLocalizedShortcut(s))
+  return addKeyboardShortcutToActionTitle(localizedActionTitle, localizedShortcuts.join('/'))
 }
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -1146,6 +1146,8 @@ KeyboardShortcutPrompt:
   Show Keyboard Shortcuts: Show keyboard shortcuts
   History Backward: Go back one page
   History Forward: Go forward one page
+  History Backward (Mac only): Go back one page (Mac only)
+  History Forward (Mac only): Go forward one page (Mac only)
   New Window: Create a new window
   Navigate to Settings: Navigate to the Settings page
   Navigate to History: Navigate to the History page
--- a/src/constants.js
+++ b/src/constants.js
@@ -127,6 +127,8 @@ const KeyboardShortcuts = {
       SHOW_SHORTCUTS: 'shift+?',
       HISTORY_BACKWARD: 'alt+arrowleft',
       HISTORY_FORWARD: 'alt+arrowright',
+      HISTORY_BACKWARD_ALT_MAC: 'cmd+arrowleft',
+      HISTORY_FORWARD_ALT_MAC: 'cmd+arrowright',
       FULLSCREEN: 'f11',
       NAVIGATE_TO_SETTINGS: 'ctrl+,',
       NAVIGATE_TO_HISTORY: 'ctrl+H',
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -1702,6 +1702,20 @@ function runApp() {
             },
             type: 'normal',
           },
+          ...(process.platform === 'darwin'
+            ? [
+                {
+                  label: 'Back',
+                  accelerator: 'Cmd+Left',
+                  click: (_menuItem, browserWindow, _event) => {
+                    if (browserWindow == null) { return }
+
+                    browserWindow.webContents.navigationHistory.goBack()
+                  },
+                  visible: false,
+                },
+              ]
+            : []),
           {
             label: 'Forward',
             accelerator: 'Alt+Right',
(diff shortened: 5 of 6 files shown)


src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
```diff
@@ -115,6 +115,14 @@ const localizedShortcutNameDictionary = computed(() => {
...
     ['HISTORY_BACKWARD', t('KeyboardShortcutPrompt.History Backward')],
     ['HISTORY_FORWARD', t('KeyboardShortcutPrompt.History Forward')],
+    ...(
+      isMac
+        ? [
+            ['HISTORY_BACKWARD_ALT_MAC', t('KeyboardShortcutPrompt.History Backward (Mac only)')],
+            ['HISTORY_FORWARD_ALT_MAC', t('KeyboardShortcutPrompt.History Forward (Mac only)')],
+          ]
+        : []
+    ),
```
kommunarr, line 125: **suggestion (non-blocking):** I'm curious if showing a non-monospace slash character joining the different alts would be more elegant than separate entries for alts. I was thinking about storing alts by storing the shortcut constant as `A|B`, which is then parsed into `A<span>/</span>B`, where we can then apply a separate styling rule (no monospace) to the nested `span`s. But that would need some special logic in the caller as well for Mac-only additional shortcut cases like this one, or a separate character for representing non-Mac and Mac-exclusive shortcut alts (e.g., `A┛B┗C|D for Windows/Mac-specific alt B, Mac-specific alt C, and universal alt D).

absidue:
This pull request doesn't actually close the linked issue, this adds Command+Arrow Left and Command+Arrow Right but the feature request is asking for <kbd>Cmd</kbd>+<kbd>[</kbd> and <kbd>Cmd</kbd>+<kbd>]</kbd>.

Would you prefer to unlink the issue so that it stays open for someone else to implement in the future or change the code to use the keyboard shortcuts that that person asked for?

src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
PikachuEXE: Way too complicated to understand
Left for another PR at least -_-

PikachuEXE:
@absidue Yup my Vivaldi uses Cmd+arrow key so I got it wrong. Fixed now

src/renderer/helpers/utils.js
```diff
@@ -1021,10 +1021,14 @@ export function addKeyboardShortcutToActionTitle(actionTitle, shortcut) {
...
-export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle, unlocalizedShortcut) {
-  const localizedShortcut = getLocalizedShortcut(unlocalizedShortcut)
-  return addKeyboardShortcutToActionTitle(localizedActionTitle, localizedShortcut)
+export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle, sometimesManyUnlocalizedShortcuts) {
+  let unlocalizedShortcuts = sometimesManyUnlocalizedShortcuts
+  if (!Array.isArray(sometimesManyUnlocalizedShortcuts)) {
+    unlocalizedShortcuts = [unlocalizedShortcuts]
+  }
+  const localizedShortcuts = unlocalizedShortcuts.map((s) => getLocalizedShortcut(s))
+  return addKeyboardShortcutToActionTitle(localizedActionTitle, localizedShortcuts.join('/'))
```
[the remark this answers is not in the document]
PikachuEXE: Updated

src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
kommunarr: Sounds good - this will currently be the only action with two displayed shortcuts we have in the modal (I excluded the dupe alt shortcuts in the original PR), so I suggest we move to a more elegant solution when we want to add more of them

kommunarr (dismissed):

src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
absidue: As this extra shortcut is only displayed to MacOS users, do we really need to specify in the text that it is macOS only? It also uses the command key which doesn't exist on Windows or Linux, so it seems unlikely that someone would try to use it on Windows or Linux anyway if they happen to use macOS + Windows or Linux.

PikachuEXE: The best way is to have 1 action to N shortcut(s)
Not sure how to do that yet...
Let me take a look ~_~

PikachuEXE: Updated
image

Also in PR description

static/locales/en-US.yaml
```diff
@@ -1146,6 +1147,8 @@ KeyboardShortcutPrompt:
   Show Keyboard Shortcuts: Show keyboard shortcuts
   History Backward: Go back one page
   History Forward: Go forward one page
+  History Backward (Mac only): Go back one page (Mac only)
+  History Forward (Mac only): Go forward one page (Mac only)
```
absidue, line 1151: As these new strings are no longer used please remove them.

absidue (commented):
Would hardcoding a wider separator be an option, just the slash without any whitespace makes it difficult to easily tell that they are two separate shortcuts.

src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
```diff
@@ -110,56 +110,60 @@ const primarySections = computed(() => [
 
 const isMac = process.platform === 'darwin'
 
-const localizedShortcutNameDictionary = computed(() => {
+const localizedShortcutNameToShortcutsMap = computed(() => {
   return new Map([
```
absidue, line 114:
```suggestion
  return [
```

src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
```diff
@@ -110,56 +110,60 @@ const primarySections = computed(() => [
...
+    [t('KeyboardShortcutPrompt.Small Rewind'), ['SMALL_REWIND']],
+    [t('KeyboardShortcutPrompt.Small Fast Forward'), ['SMALL_FAST_FORWARD']],
+    [t('KeyboardShortcutPrompt.Decrease Video Speed'), ['DECREASE_VIDEO_SPEED']],
+    [t('KeyboardShortcutPrompt.Increase Video Speed'), ['INCREASE_VIDEO_SPEED']],
+    [t('KeyboardShortcutPrompt.Skip by Tenths'), ['SKIP_N_TENTHS']],
+    [t('KeyboardShortcutPrompt.Last Chapter'), ['LAST_CHAPTER']],
+    [t('KeyboardShortcutPrompt.Next Chapter'), ['NEXT_CHAPTER']],
+    [t('KeyboardShortcutPrompt.Last Frame'), ['LAST_FRAME']],
+    [t('KeyboardShortcutPrompt.Next Frame'), ['NEXT_FRAME']],
   ])
```
absidue, line 167:
```suggestion
  ]
```

src/renderer/components/FtKeyboardShortcutPrompt/FtKeyboardShortcutPrompt.vue
```diff
@@ -168,15 +172,16 @@ function hideKeyboardShortcutPrompt() {
...
 
 function getLocalizedShortcutNamesAndValues(dictionary) {
-  const localizedDictionary = localizedShortcutNameDictionary.value
-  return Object.entries(dictionary)
-    .filter(([key]) =>
-      localizedDictionary.has(key)
+  const shortcutNameToShortcutsMap = localizedShortcutNameToShortcutsMap.value
+  const shortcutLabelSeparator = t('shortcutLabelSeparator')
+
+  return shortcutNameToShortcutsMap.entries()
```
absidue, line 178:
```suggestion
  return shortcutNameToShortcutsMap
```

absidue (commented):
As we now just use it as a list and don't actually use it for the look ups, we can just return an array of arrays.

PikachuEXE:
If using ` / `
image
image

If using `｜` (https://www.amp-what.com/unicode/search/vertical%20line
image
image

Going for later assuming that's good enough

absidue (approved):
I don't have a mac so I wasn't able to test this properly but the code looks like it should work and the keyboard shortcuts prompt seems to work the same as it did before on Windows.

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #6473, opened 2024-12-29
feat: add volume and playback rate change notification feature to vid…

ikizey:
## Pull Request Type
- [x] Feature Implementation

## Related issue
... searching numbers

## Description
it adds small popup in bottom right corner of the player, when volume or playback rate is changed using shortcuts (but not ui)

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
@@ -91,6 +91,24 @@
   /* user-select is intentionally not disabled, so that you can select and copy the stats */
 }
 
+.valueChangePopup {
+  position: absolute;
+  right: 20px;
+  bottom: calc(20px + 5%);
+  padding: 10px;
+  border-radius: 10px;
+  background-color: rgb(0 0 0 / 80%);
+  color: #fff;
+  pointer-events: none;
+  width: 85px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 4px;
+  font-size: 1.1em;
+  opacity: 0.75;
+}
+
 .offlineWrapper {
   position: absolute;
   top: 20px;
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -1808,6 +1808,8 @@ export default defineComponent({
       }
 
       volumeBar.dispatchEvent(new Event('input', { bubbles: true, cancelable: true }))
+
+      showValueChange(`${Math.round(video.value.volume * 100)}%`, 'volume-high')
     }
 
     /**
@@ -1823,6 +1825,8 @@ export default defineComponent({
         video_.playbackRate = newPlaybackRate
         video_.defaultPlaybackRate = newPlaybackRate
       }
+
+      showValueChange(`${newPlaybackRate.toFixed(2)}x`)
     }
 
     function canSeek() {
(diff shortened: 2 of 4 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Maybe also use a icon for playbackspeed the same one that is used in the player controls should be fine.

Not sure about the placement would like others to chime in. Does YouTube this with volume and playback speed



https://github.com/user-attachments/assets/3ebe84d6-ad50-4a37-9dc9-24e590f6dfd7

ikizey:
I wonder about placement too. I'd like to put it in top left corner, but stats are already there.
Not sure about icons. Idea, is '%' in the end indicate volume, and 'x' in the end indicate playback rate.
It's possible to use volumehigh/volumelow icons, but it's confusing, does it mean volume now is high/low, or volume now is higher/lower. So I made a static icon, just to indicate volume is changed.
There is no speed icon in fontawesome, that player uses. So I don't show any.

kommunarr:
I think the volume low/high icons make sense to indicate down/up. Top middle placement similar to YT here makes sense because the position of these buttons is subject to change, and top middle ensures that it can catch users' attention.

efb4f5ff-1298-471a-8973-3d47447115dc:
- I think you can put Closes behind the issue in the PR body because this does fully resolve it
- Speed being displayed below actual minimum and above actual maximum 

https://github.com/user-attachments/assets/c35bf530-1f64-4886-b8d0-de5d1d908c22

ikizey:
I placed message in top middle.
Fixed value shown for playback rate.

ikizey:
Volume icon now shows if volume become higher/lower or muted.

kommunarr:
This is awesome! Minor notes:

**suggestion (blocking):** Show mute / current volume popup on mute/unmute with <kbd>m</kbd> key.

**suggestion (blocking):** Show mute icon in popup instead of the volume low icon when using volume down shortcut and hitting 0. Currently, it shows volume low icon when it reaches zero, and you have to press it again to show the mute icon.

**nitpick:** Don't show trailing zeroes in playback rate popup (e.g., `2.00x`). This can be done by calling `toString()` on it, IIRC.

ikizey:
> **suggestion (blocking):** Show mute / current volume popup on mute/unmute with m key.
done

> **suggestion (blocking):** Show mute icon in popup instead of the volume low icon when using volume down shortcut and hitting 0. Currently, it shows volume low icon when it reaches zero, and you have to press it again to show the mute icon.
fixed

> **nitpick:** Don't show trailing zeroes in playback rate popup (e.g., `2.00x`). This can be done by calling `toString()` on it, IIRC.
I did this on purpose. Otherwise if you have step eq. 0.25 and hold <kbd>o</kbd>/<kbd>p</kbd> values will "jump" inside message.

something like this
|  1.25x  |
|   1x    |
|  0.75x  |
|  0.5x   |
now it's like this
|  1.25x  |
|  1.00x  |
|  0.75x  |
|  0.50x  |

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
nitpick: It doesn't fade out elegantly like the player controls. Is it possible to make it do that


https://github.com/user-attachments/assets/b7cf500e-1e9e-4a46-9c51-017b308bc4e0

kommunarr:
FWIW, YT also has no transition for it.

kommunarr (approved):
Would love to have something like this for toasts while in FS / FW 👀

PikachuEXE (approved):

absidue (approved):

checktext00:
I also would like to say it would be nice for every single keyboard shortcut affecting the video to have HUD/OSD feedback (like for example pressing left arrow to jump backwards 5 seconds, YouTube also has this)
YT_5sec_back

also it would be nice if the PIP mode supported these HUD notifications as well

PR #6487, opened 2024-12-30
Migrate FtSubscribeButton to the composition API

absidue:
# Migrate FtSubscribeButton to the composition API

## Pull Request Type

## Description

This pull request migrates the FtSubscribeButton component to the composition API. I made a few minor tweaks to improve the performance, such as changing the profileInitials computed to return an object with the profile IDs as the keys, based on the profile list, that way it only updates when necessary (profile added/deleted or existing profile renamed), instead of every time you subscribe and unsubscribe.

## Testing

Test subscribing and unsubscribing with just one profile (no dropdown) and multiple (has a dropdown).

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** b01abad5c35097d80a2088e66ababbc8f14bc76a


kommunarr (commented):
I'm seeing some weird jankiness when trying to subscribe/unsubscribe on a given channel's page within the first 3 seconds of loading both in `development` and this PR in dev, but not from my up-to-date `development` prod build. Could you share the prod builds for this for convenience?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue:
Just triggered it but is should be done in about 10 minutes: https://github.com/absidue/FreeTube/actions/runs/12558709764

kommunarr (approved):

PR #6495, opened 2025-01-01
Deduplicate "Sort By" strings into a single "Global.Sort By" string

absidue:
# Deduplicate "Sort By" strings into a single "Global.Sort By" string

## Pull Request Type

- [x] Other

## Description

We currently have five "Sort By" strings in FreeTube, they are all used for the same purpose: the label for a select box with sort orders for lists. This pull request deduplicates them into a single string in the `Global` namespace.

For locales where the same string was used, I just used that one, but some locales had different translations, so for those I picked the most recently updated one according to git blame.

The reasons for such a change are as follows:
- As these strings all have the same meaning and use, it makes sense to use the same one throughout the app.
- For languages that might not have all 5 of them translated yet, this pull request will result in more parts of the app being translated without the translators having to do anything.
- Less strings means less work for translators, especially when it is duplicates like these.

## Testing

Test a few locales and check that the sort by text shows up correctly.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 367781a3bb7e68adfb179a66e556c5e9a970a1cf


PikachuEXE (approved):
Tested en/jp/zh-TW

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Issue #6138, opened 2024-11-10
[Bug]: Strong audio compression artifacts audible since version 0.22.0


MegadoomerX:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

Since updating to version 0.22.0 on Debian 12/Flatpak, I notice strong audio artifacts in videos using FreeTube that were not there before. The artifacts sound like a form of heavy audio compression ("aquarium-effect") and appear in every quality setting using the DASH video format. I can only compare legacy formats in 360p, however, even on this lower quality setting, the audio artifacts are less noticeable (or different) than using the DASH format. 
 
The artifacts are much more noticeable with music, rather than speech. To reproduce, play the following video in both FreeTube and the native YouTube website:
[https://youtu.be/G4qVG_FL4_s](https://youtu.be/G4qVG_FL4_s)

Already in the beginning, the artifacts can be heard in the "s"-sounds of the voice and the clarity of the precussion.

### Expected Behavior

FreeTube should not introduce audio artifacts and did not in past versions (at least not audibly).

### Issue Labels

feature stopped working

### FreeTube Version

v0.22.0 Beta

### Operating System Version

Debian 12

### Installation Method

Flathub

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

previous version

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

mahidzzr:
I have encountered the same bug. I know exactly why it is happening. After the update the default audio quality of DASH format become 54 kbits/s. And there is no option to see or change it in the UI. it should have been max which is 140 kbits/s.

[...]

PR #6504, opened 2025-01-03
Improve audio quality by selecting highest audio bitrate

AlyoshaVasilieva:
# Improve audio quality by selecting stream with highest audio bitrate

## Pull Request Type

- [x] Bugfix

## Related issue
#6138 

## Description
Select the stream with the highest audio bitrate. Improves the issue where FreeTube usually uses the lowest quality audio available, but does not fix it. This patch seems to work as long as you do not select a quality while watching a video. Selecting a quality (e.g. 1080p) specifically while watching a video will usually give you bad audio. See #6138 for more info. I don't know how to fix this issue in the general case, but at least this improves things.

## Testing 
Try watching videos with/without this patch, e.g. https://youtu.be/G4qVG_FL4_s

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.22.1 (1c3728ec699777c16b565b022753a255b1e5cc34 + this patch)

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -1292,7 +1292,9 @@ export default defineComponent({
 
         chosenVariant = findMostSimilarAudioBandwidth(matches, audioBandwidth)
       } else {
-        chosenVariant = matches[0]
+        chosenVariant = matches.reduce((prev, current) => {
+          return (current.audioBandwidth > prev.audioBandwidth) ? current : prev
+        }, matches[0])
       }
 
       player.selectVariantTrack(chosenVariant)


absidue (approved):
Accepting this as a temporary workaround until a proper solution is implemented. Worth noting that this increases bandwidth use and will slow down video loading.

kommunarr (approved):

PikachuEXE (changes requested):
Quality selector display broken
image

Using commit before this PR
image

AlyoshaVasilieva:
I don't know how to fix the quality selector display

absidue:
@PikachuEXE @AlyoshaVasilieva The quality selector being broken is expected, shaka-player quality selector chooses the first variants in the array, but as we are now intentionally selecting something else, the quality selector doesn't know what to show. The solutions are don't merge this PR or build an entirely custom quality selector that knows how to cope with our non-standard behaviour. The latter is definitely out of scope of this pull request.

PikachuEXE:
Even though I might be fine with using this personally
I think it would appears broken to normal users unaware of this workaround and would report it as an issue
This should not be included in a release IMO

maep--:
> Accepting this as a temporary workaround until a proper solution is implemented. Worth noting that this increases bandwidth use and will slow down video loading.

Peanuts compared to video data. This would go fom 52 kb/s to 134 kb/s. For context 1080p video runs between 1000 to 5000 kb/s.

I looked at what youtube player does, they use HQ audio even for 480p video.

PietJanHein:
Considering that I have used the combination of yt-dlp and mpv for years I am very much familiar with the available streams of both audio and video and the impact on the quality. 

I upload 2 screenshots to show the problem of a low audio quality. 
image
image
In the first screenshot you see which audio-quality gets selected: 249. That is a 57 kbps audio-track. That is quite atrocious, even with Opus being better than mp3 in general. 
In the 2nd screenshot you see the available audio-quality options. As you can see the highest audio-quality available is stream 251,  147 kbps Opus.

Now the relevant question: why would you select a lower audio quality? Especially when the viewer selects a 'higher' (1080p or higher)  resolution.  The 1080p video stream is 33.11 MB  large. The highest audiostream is only 5.51 MB large. Note that for many videos the ratio between the video-data and audio-data is much higher than that. In this case you spare 3 MB of audio-data while getting a vastly inferior sound quality, you spare less than 10% of the video-stream. Note that in most cases (1440p or 4k resolution) the ratio is much higher! Often the video has more than 10 times the data compared to the highest audio stream. To substantiate this claim I will add another screenshot for this video from Rick Beato, it has a 1440p resolution. which is my display resolution. I picked this video randomly, I just looked for any video which has that resolution. 
https://youtu.be/JYl-U2NYjNw
image
2.89 GB for the 1440p video stream, the highest quality audio-stream (130 kbps) is only 29.8 MB large! That is only 1% of the total data. Why waste the quality of the sound to spare less than 1% data? It doesn't make sense, right?  

**Suggested solution.** 
Select the highest audio-stream by default, if possible. 

In case anyone thinks that you can't hear the difference. You can with modest headphones or boxes. My headphone costs around $500 and I use it with a simple Soundblaster USB-soundcard (X G6). So it does matter.

PietJanHein:
> # Improve audio quality by selecting stream with highest audio bitrate
> ## Pull Request Type
> 
>     * [x]  Bugfix
> 
> 
> ## Related issue
> 
> #6138
> ## Description
> 
> Select the stream with the highest audio bitrate. Improves the issue where FreeTube usually uses the lowest quality audio available, but does not fix it. This patch seems to work as long as you do not select a quality while watching a video. Selecting a quality (e.g. 1080p) specifically while watching a video will usually give you bad audio. See #6138 for more info. I don't know how to fix this issue in the general case, but at least this improves things.
> ## Testing
> 
> Try watching videos with/without this patch, e.g. https://youtu.be/G4qVG_FL4_s
> ## Desktop
> 
>     * **OS:** Windows
> 
>     * **OS Version:** 10
> 
>     * **FreeTube version:** 0.22.1 ([1c3728e](https://github.com/FreeTubeApp/FreeTube/commit/1c3728ec699777c16b565b022753a255b1e5cc34) + this patch)

Based on experience and my hearing (I can differentiate 128 kbps from 320 kbps MP3 in a blind test with 100% accuracy by focusing on the voices, snares including piano and percussion) I know that YouTube lowers the audio-bitrate together with the video-bitrate. I don't know about 1440p vs 1080p but I do know that for anything below 1440p with ever step lower for the resolution YouTube also lowers the audio bitrate. I don't know why they do that.

PietJanHein:
> Accepting this as a temporary workaround until a proper solution is implemented. Worth noting that this increases bandwidth use and will slow down video loading.

Respectully (and grateful for the great work of the developers :wink: ), for most users this bandwidth-argument has little relevance because 1080p and 1440p are the most common display-resolutions on desktops, laptops (unless the laptop is quite old) and even for phones 1080p or higher is the most common. At those resolutions - as I showed - the higher required bandwidht for the audio is negligible or almost negligible compared to the required bandwidth for the video. Anywhere in between around 1% ('4k') and 10% (1080p). Thanks for having worked on fixing this, but it doesn't work right now in the latest version (23.1).

absidue:
You can leave as many comments as you want, but it won't change the fact that this is currently on hold until a solution is found for the quality selector. Fixing one thing just to break something else in the process is not great.

PietJanHein:
> You can leave as many comments as you want, but it won't change the fact that this is currently on hold until a solution is found for the quality selector. Fixing one thing just to break something else in the process is not great.

That is understandable, It is not my intention to put extra pressure on the developers, I just wanted to make clear that more people would like it if that would/could get fixed. Until then FreeTube also is nice to get the videolinks and then we can simply play it in mpv.

PikachuEXE:
Please use reaction to "vote" / express your support for this change
Devs are all getting notifications for each new comment. 
Unless you have helpful info to push this forward (towards merge / close) attention consuming comments should be restrained.

absidue:
Closing this as the author has opened a new PR with a better (not complete just yet) implementation, so this PR is no longer necessary.

PR #6468, opened 2024-12-28
Add Everforest color themes

DontBlameMe99:
# Add the everforest color themes

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #4650 

## Description
This pull request adds the everforest color themes (found at https://github.com/sainnhe/everforest)
This means that this adds:
- Six base theme variants (Everforest Dark Hard, Everforest Dark Medium, Everforest Dark Low, Everforest Light Hard, Everforest Light Medium, Everforest Light Low)
- Fourteen main/secondary color themes (Everforest Dark Red, Everforest Dark Orange, Everforest Dark Yellow, Everforest Dark Green, Everforest Dark Turquoise, Everforest Dark Blue, Everforest Dark Purple, Everforest Light Red, Everforest Light Orange, Everforest Light Yellow, Everforest Light Green, Everforest Light Turquoise, Everforest Light Blue, Everforest Light Purple)

I have also already compressed the SVG files on the website https://svgoptimizer.com

## Screenshots
image
image
image
image

## Testing
I have checked every screen I can find (if there are any things missing etc.) I have also checked every color contrast I could think of.

PR #6505, opened 2025-01-03
Reduce repeated code in secondary theme colours

absidue:
# Reduce repeated code in secondary theme colours

## Pull Request Type

- [x] Other

## Related issue
- #6468 (waiting until this is merged and will pull in the changes from there, hence the draft status)

## Description

For each secondary theme colour we specify 5 variables `--accent-color` and `--accent-color-opacity{1,2,3,4}`, where the opacity ones are just the `--accent-color` with various levels of opacity. At the moment that is done in the classes of the secondary theme colours. This pull request instead specifies just a single variable `--accent-color-rgb` in each class and then uses that in the body rule to construct the 5 variables with `rgb()`.

This reduces the amount of work for people adding new themes and also shaves about 10kb (10333 bytes) off the output CSS file compared to the development branch.

## Testing

Test a few theme colours to check they still work and that the Hot Pink theme is still able to override them.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 1c3728ec699777c16b565b022753a255b1e5cc34

### DIFF
--- a/src/renderer/themes.css
+++ b/src/renderer/themes.css
@@ -4,6 +4,18 @@ body {
   color: var(--primary-text-color);
   background-color: var(--bg-color);
 
+  /*
+    This works because CSS variables are just a find and replace and
+    are replaced before the expressions are evaluated
+  */
+  --accent-color: rgb(var(--accent-color-rgb));
+  --accent-color-opacity1: rgb(var(--accent-color-rgb) / 4%);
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
I also tested that the Hot Pink makes me blind

kommunarr (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #6595, opened 2025-01-16
Change CREATE_NEW_WINDOW IPC call to accept a path and query

absidue:
# Change CREATE_NEW_WINDOW IPC call to accept a path and query

## Pull Request Type

- [x] Security improvement

## Description

Currently the `CREATE_NEW_WINDOW` IPC call accepts a URL and performs no validation on it, which means it can be used to open new windows to arbitrary URLs. This pull request locks it down by making it accept a path and query parameters instead, so that it can only be used to create windows with in-app URLs and adds some parameter validation.

## Testing

- Please check that the new window button in the top bar still works (calls `CREATE_NEW_WINDOW` with no parameters)
- Please test opening various queries in new windows from the search bar (SHIFT+ENTER or SHIFT+click on the search button)
  - https://youtu.be/jNQXAC9IVRw (just path)
  - https://youtu.be/jNQXAC9IVRw?t=10 (path and query)
  - `test` (path, query and searchQueryText) -> the search query text should appear in the search bar in the new window

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 7ad5873bec7754e757c9d2ddf0ece3f6df8ad337

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -631,6 +631,8 @@ function runApp() {
     }
   }
 
+  const ROOT_APP_URL = process.env.NODE_ENV === 'development' ? 'http://localhost:9080' : 'app://bundle/index.html'
+
   async function createWindow(
     {
       replaceMainWindow = true,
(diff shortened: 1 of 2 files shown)


src/main/index.js
```diff
@@ -1011,12 +1005,39 @@ function runApp() {
...
+
+    if (path == null && query == null && searchQueryText == null) {
+      createWindow({ replaceMainWindow: false, showWindowNow: true })
+      return
+    }
+
+    if (
+      typeof path !== 'string' ||
+      (typeof query !== 'object' && query != null) ||
+      (typeof searchQueryText !== 'string' && searchQueryText != null)
```
PikachuEXE, line 1021: I think to make this easier to read it should be reversed
If query/searchQueryText is not null/undefined and not a value of expected type
```suggestion
      (query != null && typeof query !== 'object') ||
      (searchQueryText != null && typeof searchQueryText !== 'string')
```

PikachuEXE (changes requested):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

ChunkyProgrammer (approved):

Issue #6568, opened 2025-01-13
[Bug]: Videos with no views say "NaN views" instead of "0 views"


satohshi:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

When a video has 0 views, it says "NaN views".

### Expected Behavior

It should say "0 views" instead of "NaN views".

### Issue Labels

text/string issue

### FreeTube Version

v0.22.1 Beta

### Operating System Version

Windows 10

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [X] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
Please provide details if you would like us to actually be able to do anything with this bug report. Which video? Where in the app did you see it? Do you have a screenshot or screen recording of it, if yes please provide it? How can we reproduce it/make the same problem happen on our end?

You also said the version you were using is 0.21.1 but then ticked the latest nightly box too, please clarify which version you are actually using as they are two very different things.

satohshi:
Oops. Sorry, I didn't mean to check the last box.
It's v0.22.1 Beta.

I originally saw this bug in the list of videos under "Subscriptions", but it appears to be the case for the "Search Results" page and the channel page too.
The exact video now has views, but I found this channels with a video with 0 views.
https://www.youtube.com/@animallover2338/videos

ArthurKun21:
[Build 5472](https://github.com/FreeTubeApp/FreeTube/actions/runs/12782525148) 2a683a874af75d14200fbea82aacedd17cc810c9

can confirm the `search page`

I was using `sort by upload date` since I was searching for a new video upload and came into this

[...]

PR #6596, opened 2025-01-16
Do Not Views for Video when there is Zero Views

SudharakaP:
# Title

Do Not Views for Video when there is Zero Views

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Resolve https://github.com/FreeTubeApp/FreeTube/issues/6568

## Description

This PR solve the issue of showing `NaN Views` when there's zero view count for a video

## Screenshots 

image


absidue:
Please change it in the local API parsing not in the display code. If we know that there are zero views we should show 0 views, we only hide stuff when the information is not available (e.g. published dates on the channel shorts tab).

SudharakaP:
> Please change it in the local API parsing not in the display code. If we know that there are zero views we should show 0 views, we only hide stuff when the information is not available (e.g. published dates on the channel shorts tab).

Done, let me know if you see any issues

PR #6599, opened 2025-01-17
Disable spell checker on the `ft-input` Component

### LINKED ISSUE #6548: [Feature Request]: disabling spell check or building in an option in the GUI to either disable or enable it and change the language
### Guidelines

### Problem Description

As far as I can see there is no option in FreeTube yet to disable the spell check from Chromium. This spell check is not active when you use the search bar in the top, but unfortunately it is active when you search within a YouTube channel itself. See the screenshot. In my experience spell check mostly gets in the way while it doesn't help. For example, it will point out that "linux" is wrong because Linux is a name. Fair enough, but the user doesn't care at that moment because it doesn't make a difference for the search result (case-insensitive search) and it takes less effort to not capitalize the name. In other cases the spellcheck simply does't know a name or word, or the user uses another language which might it might not have the dictionary for. Red lines reduce the readability and in my opinion it looks ugly. 

### Proposed Solution

2 options: 
1. fully disable the spellcheck for this Electron-app
2. build in a toggle to turn it on or off, in case this option is chosen and it is no being done automatically based on the set language of the user then it should also allow the user to switch to another language which makes the app more bloated because dictionaries have to be added for those languages (if this is not already the case). 

### Alternatives Considered

I looked up if I could disable spellcheck for electron in general, I haven't found a solution. 

### Issue Labels

ease of use improvement, new feature, visual improvement

### Additional Information

image

SudharakaP:
# Title

Disable spell checker in the `ft-input` component

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Resolve #6548

## Description

This disables the spell checker on the `ft-input` component so that we don't get any spell checks on the search elements. 

## Screenshots 

image

### DIFF
--- a/src/renderer/components/ft-input/ft-input.vue
+++ b/src/renderer/components/ft-input/ft-input.vue
@@ -50,7 +50,7 @@
         :type="inputType"
         :placeholder="placeholder"
         :disabled="disabled"
-        :spellcheck="spellcheck"
+        :spellcheck="false"
         :aria-label="!showLabel ? placeholder : null"
         @input="e => handleInput(e.target.value)"
         @focus="handleFocus"
--- a/src/renderer/components/ft-input/ft-input.js
+++ b/src/renderer/components/ft-input/ft-input.js
@@ -55,10 +55,6 @@ export default defineComponent({
       type: Boolean,
       default: false
     },
-    spellcheck: {
-      type: Boolean,
-      default: true
-    },
     dataList: {
       type: Array,
       default: () => { return [] }
--- a/src/renderer/components/player-settings/player-settings.vue
+++ b/src/renderer/components/player-settings/player-settings.vue
@@ -235,7 +235,6 @@
           class="screenshotFilenamePatternInput"
           placeholder=""
           :value="screenshotFilenamePattern"
-          :spellcheck="false"
           :show-action-button="false"
           :show-label="false"
           @input="handleScreenshotFilenamePatternChanged"
--- a/src/renderer/components/top-nav/top-nav.vue
+++ b/src/renderer/components/top-nav/top-nav.vue
@@ -94,7 +94,6 @@
           :is-search="true"
           :data-list="activeDataList"
           :data-list-properties="activeDataListProperties"
-          :spellcheck="false"
           :show-clear-text-button="true"
           :show-data-when-empty="true"
           @input="getSearchSuggestionsDebounce"
--- a/src/renderer/views/SubscribedChannels/SubscribedChannels.vue
+++ b/src/renderer/views/SubscribedChannels/SubscribedChannels.vue
@@ -9,7 +9,6 @@
         :value="query"
         :show-clear-text-button="true"
         :show-action-button="false"
-        :spellcheck="false"
         :maxlength="255"
         @input="(input) => handleQueryChange(input)"
         @clear="() => handleQueryChange('')"


kommunarr (approved):
Tested by editing a playlist name. LGTM!

absidue:
Personally I would have preferred this to be solved properly by setting the correct attribute on the input field in ft-input so that this will also work outside of Electron. Disabling the spellchecker globally feels like a quick and dirty workaround (similar to your other pull request, even the second solution there isn't ideal).

kommunarr:
You can blame me for that as I was the one who suggested that solution. I thought it might have some kind of subtle performance, live memory usage benefit, or something like that in comparison, but that wasn't informed on any actual evidence, so we can go with whatever solution is best.

SudharakaP:
@absidue @kommunarr : No worries. I've adjusted the PR with new changes.

kommunarr:
Thanks @SudharakaP - could you also remove [the `spellcheck`  `ft-input` property](https://github.com/FreeTubeApp/FreeTube/blob/5e8118ecd1f2fa3ef49bfec7d9f90bdaa6aa40a5/src/renderer/components/ft-input/ft-input.js#L58-L61) and its usages in the code as well?

SudharakaP:
> Thanks @SudharakaP - could you also remove [the `spellcheck` `ft-input` property](https://github.com/FreeTubeApp/FreeTube/blob/5e8118ecd1f2fa3ef49bfec7d9f90bdaa6aa40a5/src/renderer/components/ft-input/ft-input.js#L58-L61) and its usages in the code as well?

Done, didn't see any usages of this one. But let me know if you have any further comments 👍🏽

kommunarr:
You can search for the string `:spellcheck` in the code, e.g., through VSCode - I currently see five usages

SudharakaP:
> You can search for the string `:spellcheck` in the code, e.g., through VSCode - I currently see five usages

I think I know what you mean now. I am pretty new to Vue so still learning the ropes 😯

kommunarr (dismissed):
No worries, we were all at just starting at one point! LGTM!

src/renderer/components/ft-input/ft-input.vue
```diff
@@ -50,7 +50,6 @@
         :type="inputType"
         :placeholder="placeholder"
         :disabled="disabled"
```
kommunarr, line 52: **nitpick:** Probably best to do `:spellcheck="false"` to take it out of the hands of the default
```suggestion
        :disabled="disabled"
        :spellcheck="false"
```

SudharakaP: Changed back. But I thought that was what I did initially. So at other places it's set to `:spellcheck="false"` originally

kommunarr: Only change back this line please. What's happening is that the HTML `input` element has a `spellcheck` property, but as of this PR, the custom `ft-input` component doesn't. So setting `spellcheck` on the `ft-input`s does nothing, but setting it on `input` does.

absidue: It's actually worse than that. Leaving it in the other places now that the prop has been removed, means it will be treated as a fall-through attribute instead, so those other components will be adding an extra `spellcheck` attribute to the root `<div>` in the `ft-input` component.

SudharakaP: Oh now I understand what's going on a bit more 😸 . Thanks for the explanations. Corrected.

kommunarr (approved):

absidue (approved):

PikachuEXE (approved):

Issue #4665, opened 2024-02-11
[Feature Request]: configure text size separately from thumbnail size


aleksejrs:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

FreeTube has only a full page zoom, which kind of assumes that there is an optimal relation between the fonts and thumbnail size that works for all or most users. As a result, thumbnails in a grid of playlists are huge and annoying compared to their captions, and zooming does nothing to change that. https://github.com/FreeTubeApp/FreeTube/issues/4605

### Proposed Solution

Separate the text and thumbnail size controls, at least in the settings. Since everything is controlled by CSS, it should be easy to specify a base font size in pixels and a thumbnail size in pixels or ms or ns.

### Alternatives Considered

Add buttons to +/- font size and thumbnails on the top panel.

### Issue Labels

ease of use improvement, visual improvement

kommunarr:
This is something that is woefully inaccessible with FreeTube's current design that would be great if it could be fixed. The main problem is that we're using absolute pixel units rather than defining everything in terms of `rem`, and not using set CSS variables corresponding to different size classifications. Whichever contributor on here who wants to fix this one will most likely have to implement this structural change.

ACR-Jeff:
Here is just an example of what I am looking at with the UI Scaled at 100% 
![FreeTube-CommentsText](https://github.com/FreeTubeApp/FreeTube/assets/61212551/a6f8230a-f0a9-44a9-8210-83ee58f341f3)
The top half of that image is the video description area, Bottom obviously is the comments, Scaling the UI to get a decent font size in comments would scale everything on the UI far to large for normal use.

PR #6615, opened 2025-01-20
Add a setting to change chat font size

### LINKED ISSUE #6613: [Feature Request]: Add a setting to change chat font size
### Guidelines

- [x] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.

### Problem Description

In a livestream, the chat can feel small.
There is the UI Scale option, but I only want to change the chat text size.

### Proposed Solution

Add an option to change the chat font size.

### Alternatives Considered

None

### Issue Labels

new feature, new optional setting, visual improvement, improvement to existing feature

### Additional Information

_No response_

Pecomare:
# Add a setting to change chat font size

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #6613 

## Description
This PR adds a setting to change the font size used in a livestream chat.

I also took the liberty of changing some things relating to superchats:
- I moved the donation amount to the top of the block so it does not overlap with the message
- I put a max width of 90% to the superchat block so there is no horizontal bar
- I moved the profile picture inside the block (I think it being a bit outside, like a tab, does not fit in the chat, but I'm open to suggestions)
These are minor modifications in the code, I didn't want to create multiple PRs for such simple changes. If they should be separated, I'd be glad to create other PRs.

## Screenshots
Before the change
image

After the change (example uses a 32px font size)
image

The new setting is in the Theme section
image

## Testing
1. Open a livestream
2. Check the size of the text in the chat box changes
3. Pray there is also a superchat

## Desktop
- **OS:** Manjaro Linux
- **OS Version:** 24.2.1

### DIFF
--- a/src/renderer/components/ThemeSettings.vue
+++ b/src/renderer/components/ThemeSettings.vue
@@ -39,8 +39,8 @@
         />
       </div>
     </div>
-    <template v-if="usingElectron">
-      <FtFlexBox>
+    <FtFlexBox>
+      <template v-if="usingElectron">
         <FtSlider
           :label="$t('Settings.Theme Settings.UI Scale')"
           :default-value="uiScale"
@@ -50,9 +50,17 @@
           value-extension="%"
           @change="updateUiScale"
         />
-      </FtFlexBox>
-      <br>
-    </template>
+      </template>
+      <FtSlider
+        :label="$t('Settings.Theme Settings.Chat Font Size')"
+        :default-value="chatFontSize"
+        :min-value="6"
+        :max-value="96"
+        :step="1"
+        value-extension="px"
+        @change="updateChatFontSize"
+      />
+    </FtFlexBox>
     <FtFlexBox>
       <FtSelect
         :placeholder="$t('Settings.Theme Settings.Base Theme.Base Theme')"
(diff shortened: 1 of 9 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Tbh we should just use the same font size as YT if we arent doing that already instead of making it configurable

Pecomare:
What if the text as used by Youtube is too small/big to my liking but the layout is fine ?
I just checked, seems like the chat's font size is 13px.
Compared to Freetube's 12px, this is a tiny increase in size, which I find insufficient.
As seen in #4665, I am not the only one asking for the text size to be configurable. (I found it after writing the PR)
Maybe there should be a single option relating to the font sizes and we should find the optimal ones for every text, but that would be a big undertaking.

efb4f5ff-1298-471a-8973-3d47447115dc:
@kommunarr i would like to know your thoughts on this one

kommunarr:
> Maybe there should be a single option relating to the font sizes and we should find the optimal ones for every text, but that would be a big undertaking.

This is in fact one of the main problems with having a setting for the live chat text, as if we have it for this arbitrary section of the app, we should have it for others. There are a few others, like more variability that makes testing future changes to this feature harder and more involved, settings clutter, etc. If text is too small to read for most, we should have a better default for this section. If someone has particular use cases like wanting exceptionally large text, they should use the UI Scale setting for that instead.

Pecomare:
I see.
I thought about adding font size settings over time, but I understand how it will complexify the solution, moreso from the testing viewpoint.
Since it will prove to be a bad path to take, I will drop the PR.

kommunarr:
Thanks @Pecomare - we're definitely open to the idea of increasing the default font size of the live chat and any other sections if they are problematic.

PR #6638, opened 2025-01-25
Replace rimraf dev dependency with clean script

absidue:
# Replace rimraf dev dependency with clean script

## Pull Request Type

- [x] Other

## Description

Everything we need from the `rimraf` package can be easily achieved with a simple script that uses Node's built-in `rm` function. From my testing this also seems to be a bit faster, not that it was needed but it's just a nice bonus. As this is just a build dependency it won't change anything about the app itself but it does mean both our `yarn.lock` file and `node_modules` folder can be marginally smaller.

## Testing

- Run `yarn dev` to create the `dist` folder and then run `yarn clean` to remove it with this new script
- Test that it doesn't error when the build and dist folders don't exist (e.g. by running `yarn clean` two times)

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 89d72962c8c31677b66cbc5177ba3bccc867e73a


PikachuEXE (approved):
image

MarmadileManteater (approved):

kommunarr (approved):

PR #6657, opened 2025-01-27
Add total duration display to playlists

### LINKED ISSUE #6152: [Feature Request]: Let the user know how much time it would take to watch an entire playlist
### Guidelines

### Problem Description

Currently there is no way to know how much time it would take to finish a playlist.

### Proposed Solution

Let's say you're in a playlist; I think it would nice if it said at the bottom or top something like this:

Total time: 76 hours and 31 minutes and 11 seconds to watch

And also there would be a speed adjuster that affected the time for example if you selected x1.4 speed then the time would be 56 hours etc.

### Alternatives Considered

-

### Issue Labels

### Additional Information

_No response_

sullivanlng:
# Title
Add total duration display to playlists

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #6152 
## Description

This pull request adds the total playlist duration display to the playlist info container.

Changes made:
- Implemented a feature to show the total duration of the playlist.
- Ensured the visual style integrates seamlessly with the current UI.
## Screenshots 

After : 
totaltime2

## Testing 

Approach:
- Verified the feature on different screen sizes using the toggle device bar.
Behavior Confirmed:
- The total duration appears correctly and adapts visually to the playlist info container.
## Desktop

- **OS:** Windows 11
- **OS Version:** 23H2
- **FreeTube version:** 0.22.1

### DIFF
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -243,6 +243,15 @@ export default defineComponent({
     sortBySelectValues() {
       return this.sortByValues
     },
+    totalPlaylistDuration() {
+      if (!this.playlistItems || this.playlistItems.length === 0) {
+        return '0h 0m 0s'
+      }
+      const totalSeconds = this.playlistItems.reduce((acc, video) => {
+        return acc + (video.lengthSeconds || 0)
+      }, 0)
+      return this.formatDuration(totalSeconds)
+    },
   },
   watch: {
     $route() {
@@ -677,6 +686,13 @@ export default defineComponent({
 
         throw failure
       }
+    },
+    formatDuration(totalSeconds) {
+      const hours = Math.floor(totalSeconds / 3600)
+      const minutes = Math.floor((totalSeconds % 3600) / 60)
+      const seconds = totalSeconds % 60
+
+      return `${hours}h ${minutes}m ${seconds}s`
     },
 
     getIconForSortPreference: (s) => getIconForSortPreference(s),
--- a/src/renderer/views/Playlist/Playlist.js
+++ b/src/renderer/views/Playlist/Playlist.js
@@ -244,9 +244,6 @@ export default defineComponent({
       return this.sortByValues
     },
     totalPlaylistDuration() {
-      if (!this.playlistItems || this.playlistItems.length === 0) {
-        return '0h 0m 0s'
-      }
       const totalSeconds = this.playlistItems.reduce((acc, video) => {
         return acc + (video.lengthSeconds || 0)
       }, 0)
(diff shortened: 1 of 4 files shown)


smm-h (approved):
Everything looks standard, I approve.

src/renderer/views/Playlist/Playlist.js
```diff
@@ -243,6 +243,15 @@ export default defineComponent({
     sortBySelectValues() {
       return this.sortByValues
     },
+    totalPlaylistDuration() {
+      if (!this.playlistItems || this.playlistItems.length === 0) {
```
ChunkyProgrammer, line 247: This function should only be called if the length is greater than 0 so this shouldn't be necessary (247-249)

src/renderer/views/Playlist/Playlist.js
```diff
@@ -678,6 +687,13 @@ export default defineComponent({
         throw failure
       }
     },
+    formatDuration(totalSeconds) {
+      const hours = Math.floor(totalSeconds / 3600)
+      const minutes = Math.floor((totalSeconds % 3600) / 60)
+      const seconds = totalSeconds % 60
+
+      return `${hours}h ${minutes}m ${seconds}s`
```
ChunkyProgrammer, line 695: Something like this should be used instead so we get proper localization support:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat#using_intl.durationformat

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
When navigating to a creators playlist this is what i see,

https://github.com/user-attachments/assets/c46687eb-c556-4431-99d5-cea738c24fb3

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
I dislike the placement. This is playlist info and should belong next to the other data points. Place it in between video count and Last updated on

VirtualBoxVM_G3uHfTuzVv

Place it between video count and views

VirtualBoxVM_9wjOglp8LN

src/renderer/views/Playlist/Playlist.js
```diff
@@ -679,6 +685,16 @@ export default defineComponent({
...
     },
 
+    formatDuration(totalSeconds, locale = 'en') {
+      const duration = {
+        hours: Math.floor(totalSeconds / 3600),
+        minutes: Math.floor((totalSeconds % 3600) / 60),
+        seconds: totalSeconds % 60,
+      }
+
+      return new Intl.DurationFormat(locale, { style: 'short' }).format(duration)
```
ChunkyProgrammer, line 695: This should use the user's locale and fallback to english
```suggestion
      return new Intl.DurationFormat([this.currentLocale, 'en'], { style: 'short' }).format(duration)
```

Issue #451, opened 2020-01-30
[Feature] Use YouTube/Invidious embedded player


Atrate:
Since we have an option to set player type to Youtube embed, why not have an Invidious embed player, too?

PrestonN:
Because the Invidious embedded player is no different than the Dash / Legacy player when it comes to compatibility.  If there were problems with the Dash and Legacy player working, then the Invidious player likely won't work either.

The YouTube embedded player is there as an absolute last resort to get a video to play.  It can also be used if you'd purposely want ads to play for certain videos in order to support certain creators.

ChunkyProgrammer:
I think that this is something we can look at again in the future (especially since there's plans to migrate to a new video player).

PR #6662, opened 2025-01-28
Support for switching to embedded players

ChunkyProgrammer:
# Support for switching to embedded players

## Pull Request Type
- [x] Feature Implementation

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/451

## Description
A quick proof-of-concept that seems to work surprisingly well

## Screenshots 
FreeTube with a youtube embed (ads are blocked and the iframe is somewhat sandboxed, links to other videos + to the channel are blocked):
image

FreeTube with an invidious embed (dash not enabled):
image

## Testing 
Not including tests because I don't think should be merged unless absolutely necessary (which it isn't right now)

### DIFF
--- a/src/renderer/components/player-settings/player-settings.js
+++ b/src/renderer/components/player-settings/player-settings.js
@@ -212,6 +212,10 @@ export default defineComponent({
         }
       }
 
+      viewingModeNames.push(
+        this.$t('Settings.Player Settings.Default Viewing Mode.Invidious Embed'),
+        this.$t('Settings.Player Settings.Default Viewing Mode.YouTube Embed')
+      )
       return viewingModeNames
     },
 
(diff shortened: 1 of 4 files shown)


openforfuture:
Since YouTube occasionally changes its platform in ways that break FreeTube’s functionality, having embedded players as a fallback would be extremely useful. Currently, the only alternative is to open the embedded video in a browser (or in an external player that supports YouTube links, but when YouTube changes something, they often break as well). This workaround is also quite inconvenient, especially because video views aren’t tracked. If you want the video to appear greyed out and be recorded in your history, you have to do it manually.

So please consider implementing this feature — it would significantly improve the user experience. Nowadays, the embedded player is often the only way to watch content with fewer distractions. Thank you! 🙏

Issue #6539, opened 2025-01-09
[Bug]: two-finger scrolling issues with touchpad


checktext00:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### on macOS the direction of two-finger scrolling on a touchpad is reversed

in the FreeTube app settings > Player section, enable
- Scroll Volume Over Video Player
- Scroll Playback Rate Over Video Player
- Skip by Scrolling Over Video Player (after disabling "Scroll Volume Over Video Player")

in macOS Settings > Trackpad > Scroll direction is set to "Natural" (this is the default)

### Expected Behavior

the two-finger scrolling behavior should be:
- two-fingers move up to increase volume (down to decrease)
- holding the Command key + two-fingers move up to increase playback rate (down to decrease)
- two-fingers move left to scroll backwards in video (right to scroll forwards)

currently FreeTube uses the opposite directions for these actions

- lowered sensitivity (maybe 5 to 10x lower)

### Issue Labels

usability issue

### FreeTube Version

0.22.1

### Operating System Version

macOS 11

### Installation Method

.dmg

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

PikachuEXE:
See if you want to test https://github.com/FreeTubeApp/FreeTube/pull/6666
Custom build in reply

checktext00:
> See if you want to test [#6666](https://github.com/FreeTubeApp/FreeTube/pull/6666) Custom build in reply

I haven't tested your custom build, but it seems it just fixes the scrolling rate and not the wrong direction

PikachuEXE:
[...]

PR #6666, opened 2025-01-29
Fix scroll to seek/change volume/playback rate changing too fast with touchpad

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes https://github.com/FreeTubeApp/FreeTube/issues/6539

## Description

`wheel` triggered N times more than mouse wheel with smaller delta values to have smooth scrolling
This PR throttles event handler on value change side **only** (prevent default/stop propagation still called all the time

## Screenshots 

With 100ms wait time, trackpad

https://github.com/user-attachments/assets/c017cf9a-31bb-4479-aa84-ce8404b595fc

## Testing 

A) Those without trackpad
See https://github.com/FreeTubeApp/FreeTube/issues/6539, test with mouse wheel ensure it works similar to/same as before

B) Those with trackpad
https://github.com/FreeTubeApp/FreeTube/issues/6539

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

I tested 50ms wait and it seems too fast
(Too lazy to record that too

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -22,7 +22,8 @@ import {
 import {
   addKeyboardShortcutToActionTitle,
   showToast,
-  writeFileWithPicker
+  writeFileWithPicker,
+  throttle,
 } from '../../helpers/utils'
 
 /** @typedef {import('../../helpers/sponsorblock').SponsorBlockCategory} SponsorBlockCategory */
@@ -904,13 +905,13 @@ export default defineComponent({
 
         if (event.ctrlKey || event.metaKey) {
           if (videoPlaybackRateMouseScroll.value) {
-            mouseScrollPlaybackRate(event)
+            mouseScrollPlaybackRateHandler(event)
           }
         } else {
           if (videoVolumeMouseScroll.value) {
-            mouseScrollVolume(event)
+            mouseScrollVolumeHandler(event)
           } else if (videoSkipMouseScroll.value) {
-            mouseScrollSkip(event)
+            mouseScrollSkipHandler(event)
           }
         }
       }
@@ -947,7 +948,7 @@ export default defineComponent({
       }
 
       // make scrolling over volume slider change the volume
-      container.value.querySelector('.shaka-volume-bar').addEventListener('wheel', mouseScrollVolume)
+      container.value.querySelector('.shaka-volume-bar').addEventListener('wheel', mouseScrollVolumeHandler)
 
       // title overlay when the video is fullscreened
       // placing this inside the controls container so that we can fade it in and out at the same time as the controls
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Custom build: https://github.com/PikachuEXE/FreeTube/actions/runs/13023462048

efb4f5ff-1298-471a-8973-3d47447115dc:
it seems slow but that could be because im testing on a trackpad

PikachuEXE:
Feel free to change the 100ms values to something else to see how it feels
To me 50ms seems too fast (on macOS trackpad
No idea how other OSes handle trackpad scrolling

efb4f5ff-1298-471a-8973-3d47447115dc:
I let the others chime into this one as i find it difficult to evaluate what a good speed for this is

PikachuEXE:
I rarely use trackpad not to mention scrolling
So same as you

NgigaN:
> 
> 
> 
> 
> ## Pull Request Type
> 
> - [x] Bugfix
> - [ ] Feature Implementation
> - [ ] Documentation
> - [ ] Other
> 
> ## Related issue
> 
> 
> 
> Closes https://github.com/FreeTubeApp/FreeTube/issues/6539
> 
> ## Description
> 
> `wheel` triggered N times more than mouse wheel with smaller delta values to have smooth scrolling
> This PR throttles event handler on value change side **only** (prevent default/stop propagation still called all the time
> 
> ## Screenshots 
> 
> With 100ms wait time, trackpad
> 
> https://github.com/user-attachments/assets/c017cf9a-31bb-4479-aa84-ce8404b595fc
> 
> 
> 
> ## Testing 
> 
> 
> 
> A) Those without trackpad
> See https://github.com/FreeTubeApp/FreeTube/issues/6539, test with mouse wheel ensure it works similar to/same as before
> 
> B) Those with trackpad
> https://github.com/FreeTubeApp/FreeTube/issues/6539
> 
> ## Desktop
> 
> - **OS:**
> - **OS Version:**
> - **FreeTube version:**
> 
> ## Additional context
> 
> I tested 50ms wait and it seems too fast
> (Too lazy to record that too
>

checktext00:
scrolling rate/speed is perfect with the new build,

but, there is a a bit of rubber-banding feeling with the scrolling (like I could scroll the volume from 50% to 100%, but it will settle on 95%), and also it doesn't always scroll smoothly, but that's not a major issue.

also there is still the lack of ability to reverse the scrolling direction for the following 3 settings in the FreeTube app settings > Player section:
- Scroll Volume Over Video Player
- Scroll Playback Rate Over Video Player
- Skip by Scrolling Over Video Player

PikachuEXE:
New custom build: https://github.com/PikachuEXE/FreeTube/actions/runs/14505024327

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> @PikachuEXE can you rebase into dev. Im having playback issues

Probably best to rebase again and provide a new build

PikachuEXE:
Rebased and build started: https://github.com/PikachuEXE/FreeTube/actions/runs/14853765114

absidue:
The way our mouse volume/skip scrolling currently works is by using hardcoded jumps for each event, so one event equals:
- 5% volume
- default skip interval * playback rate
- 0.05x playback rate

On trackpads scrolling on a page is generally smoother whereas with a mouse it tends to jump a bit. Unconfirmed theory: Trackpads send events more frequently but each event is for a smaller "scroll distance", mice send less frequently but each event has a larger "scroll distance".

This pull request currently addresses the too large jumps on touch pads by ignoring various events. If my theory above is correct we may instead be able to base the increases/decreases on the distance that was scrolled (deltaX, deltaY and deltaMode) instead of hardcoded values (we would likely have to normalise the values a bit instead of using the raw "scroll distance"), that way it would hopefully even out/make it more similar for mice and trackpads.

PikachuEXE:
The values in `event.deltaY` are +/-`120` via wheels and +/-`1` in trackpad
Problem is that the no. of events is too big to normalize the value changed (e.g. can't change volume by 0.1%... 1% per event = still too fast

Let me see if I can make it only use `throttle` when delta is small enough (and verify if what I mentioned above is correct, forgot

PikachuEXE:
Update: Only apply scroll throttle when delta value is small (assumed to be from touchpad), update throttle wait 100 > 200ms

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Issue #6267, opened 2024-12-03
[Feature Request]: Add an respective icon to the title of every tab


MarkusGaugg:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Recently, icons were added to the Trending Tab.
https://github.com/FreeTubeApp/FreeTube/pull/6239

This pull request also added the respective icon to the title of the Trend tab. 

![Ashampoo_Snap_Dienstag, 3  Dezember 2024_0h57m16s](https://github.com/user-attachments/assets/e3663b78-e8cc-458b-adc7-f111e2f346d0)

### Proposed Solution

Is it also planned for the other tabs? It would make sense for visual consistency.
![Ashampoo_Snap_Dienstag, 3  Dezember 2024_0h57m33s](https://github.com/user-attachments/assets/4ee9a29c-3a4b-4fea-940f-030717262e47)

### Alternatives Considered

Nothing

### Issue Labels

visual improvement

PR #6669, opened 2025-01-30
Add icons to page headers

ChunkyProgrammer:
# Add icons to page headers

## Pull Request Type
- [x] Other - UX

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/issues/6267

## Description
Adds icons to page headers when it makes sense to do so + adds some icons to subscription tabs (similar to trending tabs)

## Screenshots 
image
image
image
image
image
image

## Testing
- go to pages: subscriptions, subscribed channels, settings, user playlists, history, about (Most Popular was updated as well but I'm not aware of any public instances with api enabled + most popular enabled)
- see new icons

## Desktop

- **OS:** Fedore Linux
- **OS Version:** 41 KDE


efb4f5ff-1298-471a-8973-3d47447115dc:
Most Popular icon is white instead of based on the color that is defined in the settings 

https://github.com/user-attachments/assets/138bf34b-0ba8-44f7-91e9-dd693ab35201

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

absidue:
On mobile the settings menu looks odd with the icon in the title, as other than a minor text colour difference it now looks the same as the labels for the sections. It might be worth removing the icon on the settings page.

I also noticed that no icon was added for the search results page.

ChunkyProgrammer:
@absidue I've updated it to hide the icon on the settings page on mobile and added an icon for the search results page

ChunkyProgrammer:
> Should there also be an icon here? Im not sure tbh.
> VirtualBoxVM_hlyT7lZpc3.mp4

It would look like this:
image

efb4f5ff-1298-471a-8973-3d47447115dc:
idk if you should go for that for the sake of consistency or if we shouldn't do it because it would look weird. I need the others to chime in on this one

PikachuEXE:
I am fine with having icon on hash tag page
Doesn't look weird to me

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE the icon always has been there. The subject is more if we should make that icon be based on the primary color theme like other icons

PikachuEXE:
Looks fine in my usual theme but not sure about the other themes

absidue:
I like the coloured hashtag icon.

ChunkyProgrammer:
Implemented the hashtag icon with color

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/views/Hashtag/Hashtag.vue
```diff
@@ -198,9 +187,8 @@ async function getLocalHashtagMore() {
     })
     if (backendPreference.value === 'local' && backendFallback.value) {
       showToast(t('Falling back to Invidious API'))
-      const hashtagWithoutSymbol = hashtag.value.substring(1)
       resetData()
-      getInvidiousHashtag(hashtagWithoutSymbol)
+      getInvidiousHashtag(hashtag)
```
absidue, line 191:
```suggestion
      getInvidiousHashtag(hashtag.value)
```

efb4f5ff-1298-471a-8973-3d47447115dc:
Marked as draft by accident?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #6698, opened 2025-01-31
fix: prevent prompt scrolling propagation

### LINKED ISSUE #6695: [Bug]: Prompt Menu Scroll Reaches End, Causing Parent Page to Scroll
### Guidelines

- [x] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.

### Describe the bug

1. Resize the window so that there is a vertical scroll bar
2. Open the shortcuts menu (SHIFT + ?) and check if it also has a scrollbar.
3. Start scrolling down, once you reach the end, if you move the cursor and continue scrolling, the page behind will start scrolling.

### Expected Behavior

When I reach the end of the scroll, it should not propagate to what is in the background.

### Issue Labels

visual bug, inconsistent behavior

### FreeTube Version

v0.23.0, bfa22f4

### Operating System Version

NixOS 24.11

### Installation Method

Nix (Unofficial)

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

antoniosarro:
# Title

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #6695

## Description

This change adds `overscroll-behavior-y: contain;` to the `.promptCard` class.  This prevents the scroll from the Prompt component from propagating to the underlying content, improving the user experience.

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/FtPrompt/FtPrompt.css
+++ b/src/renderer/components/FtPrompt/FtPrompt.css
@@ -17,6 +17,7 @@
 
 .promptCard {
   overflow-y: scroll;
+  overscroll-behavior-y: contain;
   max-block-size: 95%;
 }
 


MarmadileManteater (commented):
It looks good on the shortcuts prompt and on the search filters prompt. Can u also apply this to the a `playlists-container` in the component `ft-playlist-add-video-prompt`?

antoniosarro:
I'll definitely take a look at this too @MarmadileManteater

antoniosarro:
I used a different approach, adding the style overflow: "hidden" to the body when the Prompt is opened, so this should solve all similar bugs resulting from scrolling behind the Prompt because when it is open the back part has no scrollbar. I checked a bit in all circumstances if it works correctly and it seems so. If anyone can take a further look at it, I'd appreciate it.

Also, not being too familiar with Vue, I don't know if it's the best way to write this mechanism that way.

MarmadileManteater:
i would definitely prefer a css based solution myself. i liked what u had before.

antoniosarro:
Yes, I would also prefer a CSS-only approach. I was trying to achieve the same using the :has clause but it was giving me problems because the portal for the prompt is totally separate from the "main" div.

I should try to break down the css of the prompt component a bit to understand what causes problems, given that the css put previously works on some things and on other things it doesn't.

I'll give it a better look tomorrow

MarmadileManteater:
okay, so

```css
body:has(.promptCard) {
  overflow: hidden;
}
```
this css should do what the js u wrote does, but afaik, it needs to be in a global style. that's the tricky part.

what i have done on my fork (when i needed global styles for a component) is that i have added a `global.css` file under the component folder and included it as a non-scoped style in the component file:

```vue
<style src="./global.css" />
```

tbh, someone else on the team might have a better way to place this style, but this should work functionally.

antoniosarro:
This was my idea, but to work .promptCard must be a direct child of body, because :has doesn't look at all the nodes underneath. 
Now I don't want to be wrong but it seems to me there were at least two divs between body and the prompt.

MarmadileManteater:
`.promptCard` only needs to be a direct child of body if u write it like this:

`body:has( > .promptCard)`

the selector inside of `:has()` will work like any valid selector list relative to the parent

here is a js fiddle demonstrating this:
https://jsfiddle.net/a2fxLtmu/

antoniosarro:
I moved this change to the global css and it works the same way as the change I had made in the component. I'm still not entirely convinced about using the :has clause so many times in a row but as long as it isn't modified the class used for the prompt should never break.

MarmadileManteater:
that's not

MarmadileManteater:
u shouldn't need all those extra has's

antoniosarro:
So either I'm not understanding how `:has` actually works or I don't understand how you would like to solve it.
`.promptCard` cannot currently be a direct child of body, also because everything is wrapped by `.app`

image

MarmadileManteater:
u are not understanding what i am saying at all. not to be rude, but i'm not convinced u r reading what i say.

MarmadileManteater (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #6751, opened 2025-02-04
Change the release workflow to accept the release ID as an input

absidue:
# Change the release workflow to accept the release ID as an input

## Pull Request Type

- [x] Other

## Description

Currently the release workflow takes the release ID out of a repository secret. As editing the repository secrets is a cumbersome process I decided to use a workflow input instead, that way you can just paste it into a text box before pressing the `Run workflow` button right below it.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch

## Screenshots

release-id-input

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,11 @@ name: Upload Release
 # or API.
 on:
   workflow_dispatch:
+    inputs:
+      releaseId:
+        type: string
+        required: true
+        description: Release ID
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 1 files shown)


ChunkyProgrammer:
Should this be merged into master instead?

absidue:
This pull request no, if we do that then the branches will get out of sync. I can however create a second pull request that goes into the master branch if we don't want to wait until the next release to do it.

In my mind the development branch is the source of truth and the master branch just contains snapshots of the development branch.

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PR #6807, opened 2025-02-15
Local API: Improve audio quality by sorting streams, highest bitrate first

AlyoshaVasilieva:
# Local API: Improve audio quality by sorting streams, highest bitrate first

## Pull Request Type
- [x] Bugfix

## Related issue
#6138

## Description
The player seems to select the first audio stream it sees (I do not know its precise behavior). YouTube often puts a 40-50kbps audio stream first in the format list, which means FreeTube selects it and gives low quality audio.

Higher bitrate streams consistently coming first due to sorting means the player selects a high-quality audio stream, rather than potentially a very low-quality one. See #6138 for more info.

## Testing
Try watching videos with/without this patch, e.g. https://youtu.be/G4qVG_FL4_s

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 0.23.1

## Additional context
This doesn't break the quality selector.

I didn't touch the Invidious API because there were no working instances, can't test.

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -270,6 +270,7 @@ export async function getLocalVideoInfo(id) {
   }
 
   if (info.streaming_data) {
+    sortStreams(info.streaming_data)
     decipherFormats(info.streaming_data.formats, webInnertube.session.player)
     decipherFormats(info.streaming_data.adaptive_formats, webInnertube.session.player)
 
@@ -287,6 +288,16 @@ export async function getLocalVideoInfo(id) {
   }
 
   return info
+}
+
+/**
+ * Sort the video's streams so that higher bitrate streams appear first.
+ * Workaround to make the player always select high-quality audio.
+ * @param {import('youtubei.js').APIResponseTypes.IStreamingData} data
+ */
+function sortStreams(data) {
+  data.formats.sort((a, b) => b.bitrate - a.bitrate)
+  data.adaptive_formats.sort((a, b) => b.bitrate - a.bitrate)
 }
 
 /**


absidue (changes requested):
This is already a much better approach that the previous pull request, but it won't work for live streams as we use the DASH manifest from YouTube for those and as you mentioned yourself it doesn't cover the Invidious API.

A better implementation would do the sorting inside of `manifestPreprocessorTXml` in the `src/renderer/components/ft-shaka-player/ft-shaka-video-player.js` file. Additionally this time please edit this PR instead of opening a new one again.

AlyoshaVasilieva:
I'll look at `manifestPreprocessorTXml` but I might not be able to implement it; I don't really know JavaScript/Electron/Vue/shaka-player so I'm basically just poking things to see if it works.

>Additionally this time please edit this PR instead of opening a new one again.

I opened a new PR because I considered it a different approach, I didn't consider it an improvement of the old PR. Though I guess they are both sorting things.

absidue:
That functions is passed to shaka-player and runs after shaka-player has parsed the XML into objects but before it starts processing it. The function already has some code to sort the adaptation sets based on their codec. You will want to add some code that will sort the representations inside of their codec adaptation sets, for performance reasons you only need to sort them in the audio adaptation sets (the video, text (subtitles) and image (thumbnail previews)) don't need sorting.

AlyoshaVasilieva:
I think this works. Tested it (local API) on various videos/livestreams until I got temp-banned by YouTube and it was selecting the highest audio bitrate on each (and not crashing or anything).

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -732,6 +733,34 @@ export default defineComponent({
...
+    /**
+     * Sort audio AdaptationSets so that streams with higher bitrates come first.
+     * Workaround that makes the player select high-quality audio.
+     * @param {shaka.extern.xml.Node[]} periods
+     */
+    function sortAudioAdaptationSetsByBitrate(periods) {
+      for (const period of periods) {
+        period.children
+          ?.filter(child => typeof child !== 'string' && child.tagName === 'AdaptationSet' &&
+            (child.attributes.contentType === 'audio' || child.attributes.mimeType.split('/')[0] === 'audio'))
```
absidue, line 745:
```suggestion
            (child.attributes.contentType === 'audio' || child.attributes.mimeType.startsWith('audio/'))
```

absidue (commented):
Small code nit, other than that LGTM.

absidue (approved):

ChunkyProgrammer (approved):

PikachuEXE:
I can play the video but how do I know if I get highest/lowest/random quality audio stream with this PR?
Don't know where to look

AlyoshaVasilieva:
Easier way but less exhaustive: Right-click the video and click "Show Stats", then look for the itag number next to the audio codec and compare it to a list of itags, e.g. yt-dlp's: https://github.com/yt-dlp/yt-dlp/blob/421bc72103d1faed473a451299cd17d6abb433bb/yt_dlp/extractor/youtube.py#L1333 (bitrates are not exact)

Opus 251 is good, it's something like 130kbps. Opus 249 is bad, it's 40-50kbps. mp4a 140 is good too. Need to check multiple videos because some (recently-posted?) videos *only* have high-quality audio.

More exhaustive way: Open devtools with F12 before viewing video, look for the request to `player?prettyPrint=false&alt=json`. Look in `streamingData` object for all the streams available. FreeTube's [audio codec priority](https://github.com/FreeTubeApp/FreeTube/blob/38472672e41118a802a09e163873ae08981aeb5e/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L690-L695) is opus > mp4a > ec-3 > ac-3, so check as above for what codec it is using, and then look at all the audio streams YouTube listed to make sure FreeTube is using the highest quality stream according to priority and bitrate. (So, even if the AAC 140 stream has a higher bitrate, it'll select Opus if it's present since it's a newer codec, but for a given codec it should select highest bitrate with this PR.)

At least, this is how best I know.

kommunarr (approved):

bgreen29:
This seems to have been merged into the development branch but not master, is there a reason for this? Just noticed the latest beta is still playing with the poor-quality 249 codec rather than the good 251 codec.

efb4f5ff-1298-471a-8973-3d47447115dc:
@bgreen29 the master branch only contains patches and hotfixes. This will be available in v0.24.0

PR #6811, opened 2025-02-16
Move the choose default folder logic to the main process

absidue:
# Move the choose default folder logic to the main process

## Pull Request Type

- [x] Security improvement

## Description

From a security perspective the current `SHOW_OPEN_DIALOG` IPC call has two destinct issues, firstly it just wraps the Electron dialog API, giving the renderer processes full control over what options to pass as it doesn't validate any of the options and secondly while yes the app code prompts the user for the folder and then saves the path to the settings, there is nothing stopping you from setting it to any path you want through the devtools, without prompting the user and then using the `WRITE_SCREENSHOT` IPC call to write to that location.

This pull request attempts to resolve both of those issues. It blocks modifications to the `downloadFolderPath` and `screenshotFolderPath` settings from the renderer process and also provides a single IPC call to handle prompting the user and saving the path to the settings. The new `CHOOSE_DEFAULT_FOLDER` IPC call only takes a single parameter which tells it whether it is prompting the user for the screenshots or downloads folder. That means that the `WRITE_SCREENSHOT` IPC call will only be able to write to the folder selected by the user in the folder chooser dialog.

Yes, I am aware that we still have Node integration turned on, so these changes don't make much of a difference just yet, but once we turn it off, the IPC calls will be the way to perform priviledge actions, so we want to lock them down as much as possible.

## Testing

Use the buttons in the screenshot and download settings to check that the folder chooser shows up and works correctly (e.g. try choosing a folder, try cancelling and check that it doesn't overwrite the existing value, etc).

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/datastores/handlers/base.js
+++ b/src/datastores/handlers/base.js
@@ -59,12 +59,8 @@ class Settings {
     })
   }
 
-  static _findBounds() {
-    return db.settings.findOneAsync({ _id: 'bounds' })
-  }
-
-  static _findTheme() {
-    return db.settings.findOneAsync({ _id: 'baseTheme' })
+  static _findOne(_id) {
+    return db.settings.findOneAsync({ _id })
   }
 
   static _findSidenavSettings() {
@@ -75,10 +71,6 @@ class Settings {
       backendPreference: db.settings.findOneAsync({ _id: 'backendPreference' }),
       hidePlaylists: db.settings.findOneAsync({ _id: 'hidePlaylists' }),
     }
-  }
-
-  static _findScreenshotFolderPath() {
-    return db.settings.findOneAsync({ _id: 'screenshotFolderPath' })
   }
 
   static _updateBounds(value) {
--- a/src/constants.js
+++ b/src/constants.js
@@ -6,7 +6,6 @@ const IpcChannels = {
   GET_SYSTEM_LOCALE: 'get-system-locale',
   GET_PICTURES_PATH: 'get-pictures-path',
   GET_NAVIGATION_HISTORY: 'get-navigation-history',
-  SHOW_OPEN_DIALOG: 'show-open-dialog',
   SHOW_SAVE_DIALOG: 'show-save-dialog',
   STOP_POWER_SAVE_BLOCKER: 'stop-power-save-blocker',
   START_POWER_SAVE_BLOCKER: 'start-power-save-blocker',
@@ -49,6 +48,7 @@ const IpcChannels = {
 
   GENERATE_PO_TOKEN: 'generate-po-token',
 
+  CHOOSE_DEFAULT_FOLDER: 'choose-default-folder',
   WRITE_SCREENSHOT: 'write-screenshot',
 }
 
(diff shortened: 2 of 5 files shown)


src/datastores/handlers/base.js
```diff
@@ -59,12 +59,8 @@ class Settings {
     })
   }
 
-  static _findBounds() {
-    return db.settings.findOneAsync({ _id: 'bounds' })
```
PikachuEXE, line 63: I am against this change since it makes the `'bound` code separated
Or just put them into constants somewhere

Also it's not related to screenshot one

PikachuEXE (changes requested):

src/datastores/handlers/base.js
absidue: Could you please elaborate on what you are saying, as the first and last sentence don't make sense to me.

As for using constants, I would strongly advise against that, as the source of truth for the setting names are the object keys in the settings store, so making each key a constant, would make the rest of the app harder to maintain. As you would have to check the constants files every single time, because the variable names would be completely irrelevant as the getter, actions and mutation names are based on the string value, so would need to know the exact string value for the setting to be able to write the getter, action and mutation calls in the rest of the app as the majority of the files wouldn't be able to use the constants (this seems unnecessarily messy: `computed: { ['get' + Settings.BACKEND_FALLBACK.charAt(0).toUpperCase() + Settings.BACKEND_FALLBACK.slice(1)]: function () { ...`

PikachuEXE: > I am against this change since it makes the `bound` code separated

Currently all the `'bounds'` are inside `src/datastores/handlers/base.js`
Now separated in different files
Not suggesting to have getters/setters like the rest of "settings"
Probably not even worth mentioning maybe I am just ranting

> Also it's not related to screenshot one

PR: Move the choose default folder logic to the main process
Nothing mentioned about the changes & related testing and also unrelated to title (the former is bigger issue than the later

absidue: >> Also it's not related to screenshot one
> PR: Move the choose default folder logic to the main process
Nothing mentioned about the changes & related testing and also unrelated to title (the former is bigger issue than the later

It really didn't seem worth mentioning a code cleanup where the code does exactly the same thing as before, just with less duplication (literally just moved strings around), something you can easily check by looking at the code diff. If you really don't like it, I can change it only add the `_findOne` method and the new uses of it in this pull request and then switch the old code over to it in a separate pull request. But going back to the old approach of adding a separate method for every key, which has no advantage other than unnecessarily bloating the app, is not something I want to be doing.

PikachuEXE: The most important point here from me is to at least mention "there are some extra cleanup" in a short sentence.
No problem having minor cleanup with PRs, just mention it briefly to set an example for other new contributors.
I have encountered several times about mixing too much stuff in a PR (worst kind is changes totally not mentioned in PR description

PikachuEXE (approved):

kommunarr (approved):

PR #6814, opened 2025-02-16
Convert FtSubscribeButton SCSS to CSS

absidue:
# Convert FtSubscribeButton SCSS to CSS

## Pull Request Type

- [x] Refactor

## Description

The FtSubscribeButton component was only using SCSS because it was using nesting. It turned out that most of the nesting was not needed and in other places it could be removed by moving the classes from parent elements to the elements that actually needed them in the Vue template. There is now only one instance of a parent being referenced and it uses the more specific child combinator `>` (direct parent) instead of the descendant selector (parent anywhere up the DOM tree) that can match any parent right up to the root node.

In most cases nesting isn't necessary as we use Vue's scoped CSS feature that limits the CSS to the current component, nesting actually breaks the scoped CSS functionality in various cases, resulting in the CSS leaking outside of the component. Less SCSS also means less work for the build pipeline.

## Testing

Check that the subscribe button still looks the same as before.

## Desktop

- **OS:** Windows
- **OS Version:** 10


src/renderer/components/watch-video-info/watch-video-info.css
```diff
@@ -0,0 +1,132 @@
...
+}
+
+.videoOptions {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-end;
+  gap: 4px;
+}
+
+@media screen and (width <=460px) {
```
PikachuEXE, line 97: Slightly better to read
```suggestion
@media screen and (width <= 460px) {
```

src/renderer/components/watch-video-info/watch-video-info.css
```diff
@@ -0,0 +1,132 @@
...
+  .videoOptions {
+    flex-wrap: nowrap;
+  }
+
+  .videoOptions :deep(.iconDropdown) {
+    inset-inline: auto;
+  }
+}
+
+@media screen and (width <=680px) {
```
PikachuEXE, line 123:
```suggestion
@media screen and (width <= 680px) {
```

PikachuEXE (changes requested):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PikachuEXE (approved):

PR #6815, opened 2025-02-16
Use numbers instead of strings for the DBActions and SyncEvents constants

absidue:
# Use numbers instead of strings for the DBActions and SyncEvents constants

## Pull Request Type

- [x] Performance improvement

## Description

Currently we use string values in our DBActions and SyncEvents constants. While there is nothing wrong with doing that, there are definitely a few advantages, such how long some of the strings have become e.g. `sync-subscriptions-update-shorts-with-channel-page-shorts-by-channel`.

It is a well known fact that computers are great at crunching numbers, so switching to number values should not only be more performant but will also lower memory use and the bundle sizes (-1379 bytes from both the `main.js` and `renderer.js` files).

From a development perspective this shouldn't have a big impact as we still have the human readable variable names and there is a code comment explaining that the numbers need to be unique/not overlap, so I don't think this will impact maintainance too much.

If you are wondering why I didn't do the same to the `IpcConstants`, it is because Electron requires string values for the IPC channel names.

## Testing

1. Open multiple windows
2. Check that changing a setting, modifying playlists and refreshing subscriptions for example get synced properly between windows. That tests that they arrived correctly in the main process via the DBActions constants and got forwarded to the other window with the SyncEvents constants.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -53,76 +53,82 @@ const IpcChannels = {
 }
 
 const DBActions = {
+  // The constants in the GENERAL group are usally intermingeled with the ones in other groups, so they need unique values.
+  // The other groups however are usually not mixed (e.g. HISTORY and PROFILES),
+  // so they can have similar values (as long as they don't overlap with the GENERAL group).
   GENERAL: {
-    CREATE: 'db-action-create',
-    FIND: 'db-action-find',
-    UPSERT: 'db-action-upsert',
-    DELETE: 'db-action-delete',
-    DELETE_MULTIPLE: 'db-action-delete-multiple',
-    DELETE_ALL: 'db-action-delete-all'
+    CREATE: 0,
+    FIND: 1,
+    UPSERT: 2,
+    DELETE: 3,
+    DELETE_MULTIPLE: 4,
+    DELETE_ALL: 5
   },
 
   HISTORY: {
-    OVERWRITE: 'db-action-history-overwrite',
-    UPDATE_WATCH_PROGRESS: 'db-action-history-update-watch-progress',
-    UPDATE_PLAYLIST: 'db-action-history-update-playlist',
+    OVERWRITE: 6,
+    UPDATE_WATCH_PROGRESS: 7,
+    UPDATE_PLAYLIST: 8,
   },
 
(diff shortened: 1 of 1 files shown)


src/constants.js
```diff
@@ -53,76 +53,82 @@ const IpcChannels = {
...
   SUBSCRIPTION_CACHE: {
-    UPDATE_VIDEOS_BY_CHANNEL: 'db-action-subscriptions-update-videos-by-channel',
-    UPDATE_LIVE_STREAMS_BY_CHANNEL: 'db-action-subscriptions-update-live-streams-by-channel',
-    UPDATE_SHORTS_BY_CHANNEL: 'db-action-subscriptions-update-shorts-by-channel',
-    UPDATE_SHORTS_WITH_CHANNEL_PAGE_SHORTS_BY_CHANNEL: 'db-action-subscriptions-update-shorts-with-channel-page-shorts-by-channel',
-    UPDATE_COMMUNITY_POSTS_BY_CHANNEL: 'db-action-subscriptions-update-community-posts-by-channel',
+    UPDATE_VIDEOS_BY_CHANNEL: 6,
+    UPDATE_LIVE_STREAMS_BY_CHANNEL: 7,
+    UPDATE_SHORTS_BY_CHANNEL: 8,
+    UPDATE_SHORTS_WITH_CHANNEL_PAGE_SHORTS_BY_CHANNEL: 8,
```
ChunkyProgrammer, line 91: This has the same number as the previous line

absidue: Fixed

ChunkyProgrammer (approved):

PikachuEXE (commented):
I think it's easer to start other groups on `11` or even `101` so that if a few things are added to `GENERAL` there is no need to reassign values for other groups

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #6830, opened 2025-02-19
Remove fallback to Array.from

efb4f5ff-1298-471a-8973-3d47447115dc:
# Remove fallback to Array.from

## Pull Request Type

- [x] Chore/Code Cleanup

## Additional context

Maybe also remove https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/helpers/strings.js#L73?

### DIFF
--- a/src/renderer/helpers/strings.js
+++ b/src/renderer/helpers/strings.js
@@ -71,15 +71,11 @@ export function getFirstCharacter(text, locale) {
   }
 
   // Firefox only received support for Intl.Segmenter support in version 125 (2024-04-16)
-  // so fallback to Array.from just in case.
-  // TODO: Remove fallback in the future
   if (Intl.Segmenter) {
     const segmenter = new Intl.Segmenter([locale, 'en'], { granularity: 'grapheme' })
 
     // Use iterator directly as we only need the first segment
     const firstSegment = segmenter.segment(text)[Symbol.iterator]().next().value
     return firstSegment.segment
-  } else {
-    return Array.from(text)[0]
   }
 }


src/renderer/helpers/strings.js
```diff
@@ -71,15 +71,11 @@ export function getFirstCharacter(text, locale) {
   }
 
   // Firefox only received support for Intl.Segmenter support in version 125 (2024-04-16)
```
ChunkyProgrammer, line 73: We probably don't need this comment anymore either

efb4f5ff-1298-471a-8973-3d47447115dc, line 73:
```suggestion
```

src/renderer/helpers/strings.js
```diff
@@ -70,16 +70,11 @@ export function getFirstCharacter(text, locale) {
     return ''
   }
 
-  // Firefox only received support for Intl.Segmenter support in version 125 (2024-04-16)
-  // so fallback to Array.from just in case.
-  // TODO: Remove fallback in the future
   if (Intl.Segmenter) {
```
ChunkyProgrammer, line 73: Now that I think about it, this if could probably be removed as well

efb4f5ff-1298-471a-8973-3d47447115dc: Uhh im not sure im following. Why can the if statement be removed?

ChunkyProgrammer: The if statement is checking if `Intl.Segmenter` exists so the if statement isn't necessary since it should always exist. (What's inside the if-block should stay, it's just the if-statement itself that isn't needed)

PikachuEXE:
If `Intl.Segmenter` get something
else...?

ChunkyProgrammer (approved):

PikachuEXE (approved):

kommunarr (approved):

PR #6931, opened 2025-02-26
Implement content and session PO tokens

absidue:
# Implement content and session PO tokens

## Pull Request Type

- [x] Bugfix

## Related issue

Will link it after other devs have confirmed that it solves the issue for them too.

## Description

YouTube has switched from just using session based PO tokens that are added to the streaming URLs, which is what we are currently doing, to also introducing content PO tokens which are tied to the video ID added to the `/player` request. This pull request implements that new PO token. Challenges are now fetched from the `https://www.youtube.com/api/jnn` which is a proxy for `jnn-pa.googleapis.com` instead of fetching it directly from the latter URL, which should hopefully help with people's misconfiguring their ad-blocker to block `jnn-pa.googleapis.com`.

This will definitely have code conflicts with the potoken proxy pull request, I will resolve the code conflicts in whichever pull request is merged later.

## Testing

Open various videos and check that you don't get any 403s or `Error: Error invoking remote method 'generate-po-token'` errors.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** e4092aa2c2d1575538159cb7a44a78b238132dc7


src/botGuardScript.js
```diff
@@ -1,37 +1,94 @@
...
-    bgConfig
+  const botGuard = await BG.BotGuardClient.create({
+    program: challengeData.bgChallenge.program,
+    globalName: challengeData.bgChallenge.globalName,
+    globalObj: window
+  })
+
+  // DO NOT REMOVE
+  // Without this timeout we get an "Async snapshot function not found" error
+  await new Promise((resolve) => setTimeout(resolve, 1000))
```
ChunkyProgrammer, line 67: I'm guessing this is due to a bug in how `BG.BotGuardClient.create` returns the promise?

PikachuEXE:
Putting into custom build to test for 1 week+

kommunarr:
I don't experience this bug in my regular usage but will start testing this build.

src/botGuardScript.js
absidue: Removed the timeout after the BgUtils update.

ArthurKun21:
I was testing Pikachu's build 212 and encountered the following error

https://github.com/PikachuEXE/FreeTube/commit/e910c0d3fbdfc21cf41e7a6f55b22f1f08e9cefa

```js
       Failed to load resource: the server responded with a status of 404 ()
renderer.js:2 Local API, poToken generation failed Error: Error invoking remote method 'generate-po-tokens': Error: [BotGuardClient]: Async snapshot function not found
    at IpcRenderer.invoke (node:electron/js2c/renderer_init:2:7076)
    at async eX (renderer.js:2:1836327)
    at async r.getVideoInformationLocal (renderer.js:2:2301068)
eX @ renderer.js:2
```

[bundle.log](https://github.com/user-attachments/files/19088555/bundle.log)

absidue:
@ArthurKun21 That build is based on the older commits that were using a hardcoded timeout which is no longer present in the latest commits. If you want to help test PRs, please make sure you are actually using the latest changes next time before commenting that something doesn't work for you.

PikachuEXE (approved):
Tested > 1 week

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Got a 403 on the first video i tried to watch. Didnt happen on my nightly build so i think its related to the PR but weirdly enough all videos after that did play fine though

VirtualBoxVM_sXiUhLgMql

efb4f5ff-1298-471a-8973-3d47447115dc:
2nd 403 in a short period of testing and i cant replicate this behavior at all on the nightly build which im almost using daily

VirtualBoxVM_CynvPYgPHQ

PikachuEXE:
I got several 403 but not frequently and I did not use normal build for a while so no idea if YT doing stuff or this PR causing it

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM, i think i forced the 403's by switching to the next video over and over again at the speed of light

PR #6936, opened 2025-02-27
CLI re-use existing window even if freetube window is in the background (podcast usage)

flamby:
# FreeTube CLI when main window is in the foreground

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

When invoking the CLI with a new URL, if the current FreeTube window is in the background (out of focus), a new window is created...
This pull request bring the window in the foreground which fix the problem.

The main use-case is when you are on a youtube podcast for which you generally don't have the freetube window in the foreground or as picture-in-picture.

## Desktop

- **OS:** ALL
- **OS Version:** ALL
- **FreeTube version:** ALL

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -50,9 +50,9 @@ function printHelp() {
   console.log(`\
 usage: ${process.argv0} [options...] [url]
 Options:
-  --help, -h           show this message, then exit
-  --version            print the current version, then exit
-  --new-window         reuse an existing instance if possible`)
+  help, -h           show this message, then exit
+  version            print the current version, then exit
+  new-window         reuse an existing instance if possible`)
 }
 
 function runApp() {
@@ -227,7 +227,7 @@ function runApp() {
 
   // disable electron warning
   process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true'
-  const isDebug = process.argv.includes('--debug')
+  const isDebug = process.argv.includes('debug')
 
   let mainWindow
   let startupUrl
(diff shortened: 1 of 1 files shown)


absidue:
Could you please explain these changes in more detail, as they don't make much sense to me. `mainWindow.focus()` is already called by the existing code, so I'm not sure why you are calling it a second time. As we never call `mainWindow.hide()` it doesn't make much sense to call `mainWindow.show()`.

flamby:
> Could you please explain these changes in more detail, as they don't make much sense to me. `mainWindow.focus()` is already called by the existing code, so I'm not sure why you are calling it a second time. As we never call `mainWindow.hide()` it doesn't make much sense to call `mainWindow.show()`.

when you listen a podcast, you don't have the freetube window in the foreground.
on my mac, when the window was in the foreground but not minimized, launching the cli lead to a new window created all the time.
invoking mainWindow.show() fixed the issue.

You can verify that easily.

PikachuEXE:
I need testing instructions for at least macOS and one another OS (linux/windows/both) for 0/1/2+ windows

flamby:
> I need testing instructions for at least macOS and one another OS (linux/windows/both) for 0/1/2+ windows

Simply launch the binary with an url arg:
- macOS: /Applications/FreeTube.app/Contents/MacOS/FreeTube <youtube_url>
- windows: I guess the excutable is under Program Files\FreeTube

absidue:
None of the changes in this PR are necessary on Windows, which is why I am questioning them so much, because you are essentially claiming that macOS itself is calling `.hide()` on the window, which sounds very hard to believe considering you have provided no proof to back up that claim.

flamby:
> None of the changes in this PR are necessary on Windows, which is why I am questioning them so much, because you are essentially claiming that macOS itself is calling `.hide()` on the window, which sounds very hard to believe considering you have provided no proof to back up that claim.

did you test it on mac?

absidue:
I am not going to go and buy a Mac just to confirm your unsubstantiated claims that code that looks completely overkill, is necessary. Hopefully Pikachu will be able to confirm if any of these changes are actually necessary, as it seems unlikely that you are going to provide any proof that the changes are necessary.

Just a reminder that this is your pull request that you would like to be merged in, so being unhelpful and not answering the questions you have been asked just means whole process is going to take a lot longer and is much more frustrating for everyone involved.

PikachuEXE:
An example on how to write Testing section: (and please also update your testing section instead of just putting those in replies
https://github.com/FreeTubeApp/FreeTube/pull/5783

PR #6944, opened 2025-02-27
Update Invidious DASH xtags handling for YouTube.js changes

absidue:
# Update Invidious DASH xtags handling for YouTube.js changes

## Pull Request Type

- [x] Bugfix (well fixing a future bug)

## Related issue
- https://github.com/LuanRT/YouTube.js/pull/909

## Description

In https://github.com/LuanRT/YouTube.js/pull/909 I switched YouTube.js from parsing the xtags from the URL query parameter to using the xtags property that contains a protobuf with the same information, to support SABR only responses that don't have the adaptive format URLs. As Invidious doesn't expose that property through their API (I don't have access to an Invidious instance with the API enabled but I checked their code), we need to keep the old behaviour of using the URL query parameter for our DASH manifest generation for Invidious.

The xtags contain information about the language and the audio content type of the audio track, as well as whether the audio track is stable volume one (DRC).

## Testing

There are barely any public Invidious instances left and those that do exist have the API disabled, so I haven't actually been able to test this but as the code is the same as what YouTube.js used to do, I think it is probably fine 😅.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 68ec1e12bbdfa6b37122f9513ae1a5fc05ddb3e2

### DIFF
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -867,7 +867,9 @@ export function convertInvidiousToLocalFormat(format) {
   const [initStart, initEnd] = format.init.split('-')
   const [indexStart, indexEnd] = format.index.split('-')
 
-  const duration = parseInt(parseFloat(new URL(format.url).searchParams.get('dur')) * 1000)
+  const url = new URL(format.url)
+
+  const duration = parseInt(parseFloat(url.searchParams.get('dur')) * 1000)
 
   // only converts the properties that are needed to generate a DASH manifest with YouTube.js
   // audioQuality and qualityLabel don't go inside the DASH manifest, but are used by YouTube.js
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
Tested with https://youtu.be/snZ9w5xc_ic and

> Before the newly added `if (localFormat.has_audio && url.searchParams.has('xtags')) {` line, you will need to add the following code snippet:

```js
localFormat.language = undefined
localFormat.is_drc = undefined
localFormat.is_dubbed = undefined
localFormat.is_descriptive = undefined
localFormat.is_secondary = undefined
localFormat.is_auto_dubbed = undefined
localFormat.is_original = undefined
```

image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #6949, opened 2025-03-01
fix:(watch.vue) Add refresh functionality  to  live stream starting soon msg

TheMeanBeanMachine:
# Add refresh functionality and icon to  live stream starting soon message

## Pull Request Type

- [X] Bugfix

## Description
adds the ability for user to click the "Video.Starting soon, please refresh the page to check again" and refresh the page. 
also adds small  refresh icon at the end of the text 

## Screenshots 
N/A

## Testing 
very small change to make text clickable and adding refresh method.

## Desktop

- Linux Mint Cinnamon
-  22.1
- v0.23.2 Beta

## Additional context
Just a small quick change i got motivated to try  after i was waiting for the WAN show  and I had to click out and click back in to the live stream a few times to refresh and get the live stream to start (Ctrl + r wasnt working for me).  So I thought i could quickly add this small quality of life improvement. 

If this is a fix you all like i might go in and  add contextual refreshes like this  other places. ive noticed that when some components error out like live stream chat theres no way to refresh without refreshing the whole page which can make you lose your place in a live stream. 

Absolutely Love the app by the way thank you dev team for all the awesome work.

### DIFF
--- a/src/renderer/views/Watch/Watch.vue
+++ b/src/renderer/views/Watch/Watch.vue
@@ -83,10 +83,18 @@
               </span>
             </p>
             <p
-              v-else
-              class="premiereText"
-            >
-              {{ $t("Video.Starting soon, please refresh the page to check again") }}
+             v-else
+             class="premiereText"
+             @click="refreshPage"
+             style="cursor: pointer;"
+           >
+             {{ $t("Video.Starting soon, please refresh the page to check again") }}
+             <font-awesome-icon
+               :icon="['fas', 'sync']"
+               class="refreshIcon"
+               @click.stop="refreshPage"
+               style="margin-left: 8px; cursor: pointer;"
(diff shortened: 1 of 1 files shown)


absidue:
This is a bad in so many ways, this does exactly the same as CTRL+R and you should never even be doing that in FreeTube, there is a reason it is labelled as Reload Window and not Reload Page. FreeTube is a purely client side app, the page is the app, so by reloading the window you are not just reloading the watch view but the entire app, so then it has to waste time setting itself up again, loading all of your user data, caches and various other files from disk again. If you want to reload the page, click the backwards and forwards buttons instead.

Personally I would love to remove CTRL+R and CTRL+SHIFT+R entirely but that would cause outrage with all the users that love to do the things we tell them are not a good idea.

PR #7002, opened 2025-03-13
Move 'Subscriptions' section below 'About' section in the left sections sidebar

MouradMalki:
Type of the issue: Other

 Hello, I moved 'Subscriptions' section below 'About' in the left sections sidebar since it should be located just above the channels that the user subscribed to, not at the top of the bar. I didn't find an issue mentionning that problem, but i still estimated that this change was good.

Freetube version: 0.23.2
OS: Windows 11

)
Here you can find attached two screenshots: one before the change, and one after. 
Capture d’écran 2025-03-13 010400

Capture d’écran 2025-03-13 030720

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,5 +121,6 @@
     "webpack-cli": "^6.0.1",
     "webpack-dev-server": "^5.2.0",
     "yaml-eslint-parser": "^1.3.0"
-  }
+  },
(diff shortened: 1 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
This is opinion based and we aren't interested in this change.

PR #7007, opened 2025-03-14
Add playlist duration to playlist view

MarmadileManteater:
# Add playlist duration to playlist view

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
This is based on this PR: #6657
closes https://github.com/FreeTubeApp/FreeTube/issues/6152

## Description
This PR adds the total playlist duration to the `playlist-info` component which is displayed within the `Playlist` view.

## Screenshots 

|before|after
|--|--|
|2025-03-14-090636_hyprshot|2025-03-14-090720_hyprshot|

## Testing 
1. Check a creator playlist (ex: https://youtube.com/playlist?list=PL2EEDBEDC9C2BDF07 )
2. Check a user playlist
3. Check a playlist where none of the videos have a duration 
_(This can be accomplished by adding only videos from RSS feed subscriptions to a user playlist.)_
4. Check a playlist where at least one of the videos do not have a duration
5. Check that the localization of the duration works as expected

## Desktop

- **OS:** Fedora Linux w/ hyprland
- **OS Version:** 41 (Workstation Edition) x86_64
- **FreeTube version:** b6ba16f5b9124e0b28aadbbb839f6e7f3d467906

### DIFF
--- a/src/renderer/components/playlist-info/playlist-info.vue
+++ b/src/renderer/components/playlist-info/playlist-info.vue
@@ -78,7 +78,7 @@
           </span>
           {{ lastUpdated }}
           <template v-if="durationFormatted !== ''">
-            <br/>
+            <br>
             {{ $t('User Playlists.TotalTimePlaylist', { duration: durationFormatted }) }}
           </template>
         </p>


efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
When one short is inside an user playlist i get this error

VirtualBoxVM_regZZWhCaV

User playlist full of shorts throws an error in the console

https://github.com/user-attachments/assets/979901ff-5063-431e-9f3b-9e9d46eea0e8

Couldnt find a creator playlist that had shorts without timestamp and normal videos mixed

MarmadileManteater:
I'm not able to replicate, but it seems there is a video in that playlist with a non-number for `lengthSeconds` which is also not `undefined`. I pushed a change which may address the issue.

efb4f5ff-1298-471a-8973-3d47447115dc:
Errors are gone but when a short is added to a playlist the it wont show total time anymore.

https://github.com/user-attachments/assets/ef147b29-633d-4a72-8dbe-40bc36f4aa62

MarmadileManteater:
That was the behavior I intended, but it can be changed. My logic is that, if there are any number of videos with no duration in a playlist, the total duration will be wrong.

efb4f5ff-1298-471a-8973-3d47447115dc:
Sounds logical. One thing i came up with is to call it approximately total time so we can still show it but maybe that is just bad UX

PikachuEXE:
I feel like putting it to another line would make it "less busy"/easier to read
image
image
image
image
image

MarmadileManteater:
@PikachuEXE 
Do you have any opinion on whether we should include playlist duration when there are videos without a length in a playlist? (or if that should be displayed as approximate time)

PikachuEXE:
Don't show it seems better UX for now
Otherwise got to show how many videos have/have no length which I guess might look weird/confusing to users

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #7010, opened 2025-03-15
Fix/player focus bug

### LINKED ISSUE #931: Enable window scrolling with keyboard arrows when clicking outside the video controls.
When I'm watching a video, or when it ends, I want to be able to scroll the page to see the comments or the recommended videos. But even when I click outside the video's screen, I still can't scroll the window with the keyboard arrow keys. Then I have to use the mouse wheel or try to click in the tiny scrollbar on the right, but I think it'll be good to be able to scroll the window by using the keyboard.

Just being able to scroll the window if I click in an empty area on the page, using the arrow keys in the keyboard.

Trying to grab the tiny and hard to see scrollbar or by using the mouse wheel.

Razen04:
# Fix Player Focus and Arrow Key Scroll Behavior

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #931 

## Description

This PR fixes the behavior of the player focus detection and scroll handling using the arrow keys. The following changes were made:
- Added click event listener to detect whether the player is in focus or not by checking the click position inside the player.
- Prevented page scrolling when the player is focused and the arrow keys are pressed.
- Added keydown event handling to stop the default action and propagation of the arrow keys for smooth scrolling only when the player is not focused.
- Improved the responsiveness to user interactions by checking focus state and controlling scroll and player interactions.

## Screenshots 

N/A

## Testing 

- Tested by clicking inside and outside the player to see if playerFocussed correctly toggles.
- Verified that when the player is focused, arrow keys no longer scroll the page.
- Checked that when the player is not focused, arrow keys control the page scrolling without affecting the player.

## Desktop

- **OS:** ZorinOS
- **OS Version:** 17.2
- **FreeTube version:** 0.23.2

## Additional context

N/A

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -2044,3 +2044,5 @@ function runApp() {
     Menu.setApplicationMenu(menu)
   }
 }
+
+console.error('hello')
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -363,10 +363,6 @@ export default defineComponent({
     }, true)
   },
 
-  beforeDestroy() {
-    document.removeEventListener('click', this.handleClickEvent)
-    document.removeEventListener('keydown', this.handleKeyEvent, true)
-  },
   methods: {
     onMountedDependOnLocalStateLoading() {
       // Prevent running twice
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -141,6 +141,7 @@ export default defineComponent({
       /** @type {Date|null} */
       streamingDataExpiryDate: null,
       currentPlaybackRate: null,
+      playerFocussed: true
     }
   },
   computed: {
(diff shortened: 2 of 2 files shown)


src/main/index.js
```diff
@@ -2044,3 +2044,5 @@ function runApp() {
     Menu.setApplicationMenu(menu)
   }
 }
+
+console.error('hello')
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 2048: Seems irrelevant

```suggestion
```

src/main/index.js
```diff
@@ -2044,3 +2044,5 @@ function runApp() {
     Menu.setApplicationMenu(menu)
   }
 }
+
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 2047: Seems irrelevant

```suggestion
```

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Im unable to use the up and down keys to adjust the volume. When user clicks on a video to watch these keys are reserved for the player. Arrow keys should only be used for page down/up when when the user clicks outside the player window 

https://github.com/user-attachments/assets/87bf8bc3-c713-4736-8a31-f31996484afb

PR #7035, opened 2025-03-19
Migrate FtChannelBubble to composition api

ChunkyProgrammer:
# Migrate FtChannelBubble to composition api

## Pull Request Type

- [x] Other -  composition api migration

## Description
This PR migraties FtChannelBubble to composition api

### Channel about page:
- ensure that related channels load properly and you are able to interact with them as expected (clicking one goes to that channel). ex channel: https://www.youtube.com/channel/UCX6OQ3DkcsbYNE6H8uQQuVA/about

### Profile
- open profile switcher (top right of page)
- open profile manager (settings icon in profile switcher)
- select a profile in the manager that has subscriptions
- make sure that the subscription list loads and that a subscription can be interacted with as expected.

## Desktop

- **OS:** Fedora Linux KDE
- **OS Version:** 41


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/FtChannelBubble/FtChannelBubble.vue
```diff
@@ -48,13 +48,51 @@
...
+    type: String,
+    default: null
+  },
+  showSelected: {
+    type: Boolean,
+    default: false
+  }
+})
+
+const selected = ref(false)
```
absidue, line 83: As the profile management components access this ref, you need to use defineExpose on it. It's not a problem in Vue 2 as that automatically exposes all variables and functions, but in Vue 3 where the composition API compilation is a lot more optimised, anything that isn't explicitly exposed will be inaccessible from the outside.

I do have a more extensive rewrite in my git stash (I'm sorry I really should have PRed it a while ago) that rewrites the channel bubble and profile management components in the composition API and to use props instead of template refs.

ChunkyProgrammer: Sounds good, I'll close this PR and review yours when it is opened 🙂

PR #7038, opened 2025-03-19
Fix: Remove unnecessary overflow

est-2025:
Title:
Fix: Remove unnecessary overflow: hidden to resolve scrollbar issue in Turkish/German settings

Description:
In settings.css, I removed the overflow: hidden property. This change fixes the issue where an unwanted extra scrollbar appears when the language is set to Turkish or German.
screenshots :
before:
Screenshot 2025-03-19 224606
after:
Screenshot 2025-03-19 224557

After removing overflow: hidden, the layout behaves as expected and no additional scrollbar shows up in these languages. thank you

### DIFF
--- a/src/renderer/views/Settings/Settings.css
+++ b/src/renderer/views/Settings/Settings.css
@@ -5,7 +5,7 @@
 
 .settingsSections,
 .switchRow {
-  overflow-x: hidden;
+  
   max-inline-size: 90%;
 }
 


MarmadileManteater:
I am unable to replicate the issue this PR is solving:
(on development branch)
2025-03-19-211256_hyprshot
What is the window width in the screenshots you provided?

Also, that `overflow-x: hidden` exists to prevent a horizontal scroll bar on smaller display sizes due to the tool tips overflowing.
(on this branch)
2025-03-19-211517_hyprshot

Ideally, the tool tips should float left to prevent them from overflowing, so removing the `overflow-x: hidden` isn't necessarily a bad thing. However, if you remove it, you will need to address the issue that removing it causes.

Issue #5425, opened 2024-07-16
[Feature Request]: Open new window from taskbar icon


Venomtek:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Lack of "Open New Window" option when right clicking taskbar icon when you already have a window open.

### Proposed Solution

Add an "Open New Window" option when right clicking taskbar icon if you already have a window open.

### Alternatives Considered

None.

### Issue Labels

ease of use improvement

rugabunda:
Got you covered  https://github.com/FreeTubeApp/FreeTube/pull/7049

rugabunda:
If its not available by the time you read this, you can download or build a version with it on github via actions here: https://github.com/rugabunda/FreeTube/actions/runs/14629518312

Venomtek:
Noice, thanks!

rugabunda:
> Noice, thanks!

Welcome, the pull request has now been merged, so building from the official branch will work now.

PR #7049, opened 2025-03-22
Feature/add Windows Taskbar jumplist task 'New Window'

rugabunda:
# Title

[Windows: Open New Window from Taskbar Jumplist](https://github.com/FreeTubeApp/FreeTube/commit/a8916e595a0e08d79b3c1b771964e6c2acf9186a)

## Pull Request Type
- [ ] Feature Implementation

## Description

Windows users now have the option to right click and open a new window from the taskbar. If Windows O/S is detected, this feature adds a 'New Window' option to the taskbar context menu. This enhancement aims to simplify the process of opening multiple instances of an application, aligning with user preferences for more intuitive taskbar functionality.

## Screenshots

2GLSh4sVDc

## Testing 

Built & tested, working fine.

## Desktop

Windows 11
26100
23.2 latest dev

## Additional context

This feature is particularly useful for users who frequently need multiple instances of the same application open simultaneously. It streamlines workflow mimicking browser taskbar functionality by eliminating the need for additional clicks or keyboard shortcuts to achieve this functionality. The implementation is designed to be user-friendly and accessible directly from the taskbar, enhancing overall productivity and user experience on Windows systems.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -225,6 +225,22 @@ function runApp() {
     }
   })
 
+   // If in windows, add Windows Jump List task 'New Window'
+   if (process.platform === 'win32') {
+     const iconPath = path.join(path.dirname(app.getPath('exe')), 'freetube.exe');
+ 
+     app.setUserTasks([
+       {
+         program: process.execPath,
+         arguments: '--new-window',
+         iconPath: `${iconPath},0`,
+         iconIndex: 0,
+         title: 'New Window',
+         description: 'Create a new window'
+       },
+     ])
+   }
+
   // disable electron warning
   process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true'
   const isDebug = process.argv.includes('--debug')
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -225,22 +225,21 @@ function runApp() {
     }
   })
 
-   // If in windows, add Windows Jump List task 'New Window'
-   if (process.platform === 'win32') {
-     const iconPath = path.join(path.dirname(app.getPath('exe')), 'freetube.exe');
- 
-     app.setUserTasks([
-       {
-         program: process.execPath,
-         arguments: '--new-window',
-         iconPath: `${iconPath},0`,
-         iconIndex: 0,
-         title: 'New Window',
-         description: 'Create a new window'
-       },
-     ])
-   }
-
+  // If in windows, add Windows Jump List task 'New Window'
+  if (process.platform === 'win32') {
+    app.setJumpList([{
+      type: 'tasks',
+      items: [{
+        type: 'task',
+        program: process.execPath,
+        args: '--new-window',
+        title: 'New Window',
+        description: 'Create a new window',
+        iconPath: process.execPath,
+        iconIndex: 0
+      }]
+    }])
+  }
   // disable electron warning
   process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true'
   const isDebug = process.argv.includes('--debug')


absidue:
Please implement the same thing on macOS e.g. https://www.electronjs.org/docs/latest/tutorial/macos-dock

efb4f5ff-1298-471a-8973-3d47447115dc:
> Please implement the same thing for macOS too e.g. https://www.electronjs.org/docs/latest/tutorial/macos-dock

If possible adding this for Linux would also be good to include in this PR

src/main/index.js
```diff
@@ -225,6 +225,40 @@ function runApp() {
     }
   })
 
+  // If in windows, add Windows Jump List task 'New Window'
+  if (process.platform === 'win32') {
+    app.setJumpList([{
```
absidue, line 230: Please use the more specific `app.setUserTasks()` API.

src/main/index.js
```diff
@@ -225,6 +225,40 @@ function runApp() {
     }
   })
 
+  // If in windows, add Windows Jump List task 'New Window'
```
absidue, line 228: Please remove this code comment, it doesn't add any additional information compared to what the code already says (adding code comments is generally only necessary if the code alone isn't enough, e.g. explaining why some logic is there).

_scripts/ebuilder.config.js
```diff
@@ -48,6 +48,7 @@ const config = {
     category: 'Network',
     icon: '_icons/icon.svg',
     target: ['deb', 'zip', '7z', 'apk', 'rpm', 'AppImage', 'pacman'],
+    executableArgs: ['--new-window']
```
absidue, line 51: This is the wrong approach, this forces it to always open a new window, whereas we were asking you to add a menu entry in the same way that you have done for macOS and Windows.

src/main/index.js
```diff
@@ -225,6 +225,40 @@ function runApp() {
...
+        args: '--new-window',
+        title: 'New Window',
+        description: 'Open New Window',
+        iconPath: process.execPath,
+        iconIndex: 0
+      }]
+    }])
+  }
+
+  // If in Mac, add macOS Dock menu with 'New Window' option
```
absidue, line 244: Ditto about the code comment

src/main/index.js
```diff
@@ -225,6 +225,40 @@ function runApp() {
...
+            replaceMainWindow: false,
+            showWindowNow: true
+          })
+        }
+      }
+    ])
+    app.whenReady().then(() => {
+      app.dock.setMenu(dockMenu)
+    })
+  }
```
absidue, line 260: Please move this inside the existing event handler for the `ready` event instead of adding another one.

_scripts/ebuilder.config.js
rugabunda:
```
linux: {
  category: 'Network',
  icon: '_icons/icon.svg',
  target: ['deb', 'zip', '7z', 'apk', 'rpm', 'AppImage', 'pacman'],
  executableArgs: ['--new-window']
},


else

linux: {
  category: 'Network',
  icon: '_icons/icon.svg',
  target: ['deb', 'zip', '7z', 'apk', 'rpm', 'AppImage', 'pacman'],
  desktop: {
    Actions: "NewWindow;",
    "Desktop Action NewWindow": {
      Name: "New Window",
      Exec: "${execPath} --new-window"
    }
  }
},
```
I tested both versions.

If you add an extra 'new window' button in in the desktop of ubuntu, you get two "New Window" options on right click, one of them brings the window forward, same as a single left click on icon, why waste the extra two clicks; the one I added creates as many new windows as you like... 

Your proposed version in Ubuntu:
Ubuntu 1_2025-03-30_02-54-03

Rather than duplicate buttons and waste clicks I modified eslint.config.mjs and simply added executableArgs: ['--new-window']; Which is universally compatible with all linux distributions. 

Left clicking the icon on the taskbar brings it forward/backward right clicking and new window creates new window
Ubuntu 2_2025-03-30_02-55-09

Confirmed v1 & V2 working in lubuntu 22.10,  Fedora 41. V2 does not. V1 works great.

In fedora both versions offer the same context menu seen here but only one of them actually creates a new window... v1
Fedora 1 execute arg


Mint Cinnamon 22.1 fails with both versions, fault of [Grouped window list (grouped-window-list@cinnamon.org)], Main Cinnamon window list with app grouping... which is incompatible with destkop icon modifications. 

Mint 22 1

Version 2 Works in Mint Mate 22.1, the version you suggest fails, no option for new window & left click doesn't open one either:

2025-03-30_05-40-55_uXKp6j9OwP

 Mint XFCE 22.1 works great with version 2 with a left click right on the taskbar icon.
 Version 1 doesn't show any "New Window" option at all.
 
2025-03-30_05-51-03_TNRFqki3Kq

 

In Lubuntu's case the extra 'new window' icon looks nicer and as one would hope; however left clicking the icon and right clicking the icon and clicking 'freetube' has the exact same function: bringing the app forward. why waste the extra clicks, as so exec=/opt/FreeTube/freetube --new-window %U is clearly the better option.


Lubuntu 1

Right clicking this and clicking freetube creates new window, left clicking primary icon brings it forward
Lubuntu 2

From my experience, and as far as I understand, the version you suggest may be preferable in a uniform linux env where the buttons actually work or are visible all the time, but given that is not the case, having --new-window as the primary execution argument I believe is currently the more convenient, useful, and helpful over all.

rugabunda:
> Please implement the same thing for macOS too e.g. https://www.electronjs.org/docs/latest/tutorial/macos-dock

I did it, but I don't have a mac, so someone who does must test the code

_scripts/ebuilder.config.js
rugabunda: updated comment

rugabunda: If in linux an option to change default launch it in preferences, what do you think of that?

absidue: So the problem here seems to be the usual one with Linux, great for users crap for app developers, despite a standard existing, loads of distros are not actually following it properly, some are even providing their own New Window buttons and ignoring FreeTube's.

I would rather not change anything for Linux at all, forcing it to always create a new window will also change how the URL scheme works (with your change it will always open a new window instead of respecting FreeTube's deep link handling setting).

PikachuEXE:
macOS 13: it works but since code not finalized yet no review given
Before
image
After
image

PikachuEXE:
Not seeing it in win10 (in my custom build
image

PikachuEXE:
It works but now whenever I open FT via pinned icon the new item is at the end instead of just replacing the pinned icon
That's an unwanted side effect

efb4f5ff-1298-471a-8973-3d47447115dc:
Fyi: Closes https://github.com/FreeTubeApp/FreeTube/issues/5425

PikachuEXE:
Linter check failed

rugabunda:
Updated this, @PikachuEXE  if you could test the Mac version would be appreciated

https://github.com/FreeTubeApp/FreeTube/pull/7049/commits/078373bf29be730cf8d468c303018d6bf4a01ac3

PikachuEXE:
I built one myself on Mac, it works

PikachuEXE (approved):

rugabunda:
> I built one myself on Mac, it works Update 1: Works on windwos too

Thank you! 

Just learned how to link VS to Git, and sync visual studio code commits with github last night. This, people, is a MASSIVE breakthrough.  And I never coded before.  I learned lint wasn't just the stuff between my toes.  We can thank AI for my first contribution, which knew precisely what you were asking when I hadn't the slightest clue. 

AI which has opened otherwise illiterate coders to a whole new world.  Need a lawyer? No excuse to be lazy now!  If everyone wants some change, they can build it.

As a follow up to your earlier comment, I noticed that every time I install Freetube in windows it has a tendency to remove the pinned icon from the task bar, so most people shouldn't run into any issue where they need to unpin-pin again.  Might want to mention in in the changelog just in case.

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@rugabunda Thanks for this PR. Hopefully we will see more wonderful contributions of you in the future!

rugabunda:
Happy to be of service for such a kick ass app!

PR #7116, opened 2025-03-30
Minor build speedups

absidue:
# Minor build speedups

## Pull Request Type

- [x] Other

## Description

This pull request does two minor changes which should speed up the builds a bit, on my local machine it didn't make a major difference (1.5 seconds, ~46.5 vs 45.0) but I still think they are worth doing.

The first change is to no longer call `electron-builder install-app-deps`, this scans the dependencies looking for any that contain native `.node` modules and recompiles them to be compatible with Electron (as Electron is Chromium+Node.js, native modules compiled for Node.js generally won't work out of the box for Electron) but as FreeTube doesn't have any dependencies with native node modules, we don't actually need to run that command.

As all of our dependencies are JavaScript, we bundle them with webpack, which means we don't need to package the `node_modules` folder in our packages. We already tell electron-builder to not include them in the `files` array, but electron-builder still spends time scanning the dependencies and building up a list of runtime dependencies and then ignores that list because we tell it to not bundle the `node_modules` folder in the `files` array. So the second change is to tell electron-builder to skip scanning the dependencies to build up that list.

Another change that I have not done yet which would also help with build times, which I haven't done yet is to remove babel from our renderer build as that requires discussion. As we use the latest Electron version which is also pretty much always the latest Chromium version, it seems unlikely that we'll encounter a situation where we are using JavaScript syntax or API that is so experimental that it hasn't even made its way into Chromium yet. Even if we were to encounter such a situation we should probably reconsider, as a JavaScript feature that is that experimental is likely to change multiple times before it is considered stable or potentially not even be added to JavaScript.
Using babel for the main and web builds probably still makes sense as the Node.js version in Electron is generally not always the latest and the web build might be used on a browser that isn't the latest, the same goes for the downstream FreeTubeAndroid build (as it uses the system web view, there is a chance that it might be running on a device that doesn't have the latest web view version).

There is also this PR in webpack which should help quite a bit with the release build performance once it lands in a webpack release: https://github.com/webpack/webpack/pull/19331

## Testing

When I tested this locally it sped up the build by about 1.5 seconds. I was unable to verify if this actually makes a difference in our CI, as the timings seem to vary wildly from run to run (likely depends on which GitHub machine the workflow runs on), so unless this were to halve the build times (which it doesn't), we probably won't notice it too much.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** d9de46da78f4d99c4aa506b632dfe518f5c0089d


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Most time still being spent on pack & dmg/7z/etc. I just make sure I can build locally
image

ChunkyProgrammer (approved):

PR #7140, opened 2025-04-02
Manually apply ad-hoc signature to macOS (ARM) app to fix "app is damaged" errors

### LINKED ISSUE #6691: [Bug]: Mac arm64 version broken
### Guidelines

- [x] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.

### Describe the bug

1. Download the latest Mac AppImage for arm64
2. Click it
3. Get a popup error "FreeTube" is damaged and can't be opened. You should move it to the Trash.

Image

This only happens on the ARM build. The X64 Mac version works normally.

### Expected Behavior

Application starts

### Issue Labels

causes crash

### FreeTube Version

v0.23.0

### Operating System Version

macOS Sequoia 15.3

### Installation Method

.dmg

### Primary API used

Local API

v0.22.1

### Additional Information

The ARM version appears to be broken in general as I've tried the .dmg, the .zip, the .7z, and even from source and all versions were reported damaged. Currently having to run the x64 version through Rosetta 2.

### Nightly Build

kthchew:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #6691

## Description

macOS on ARM shows "app is damaged" errors due to an invalid signature. This appears to be an issue with the way `electron-builder` handles the build, breaking the automatic linker-applied ad-hoc signature. To fix this we can re-sign the affected parts of the app bundle using the `codesign` command after `electron-builder` does this work but before it places it into the DMG/ZIP/7z container.

## Screenshots 
The warning now shows as
Screenshot 2025-04-02 at 4 57 11 PM
And System Settings shows an option to bypass the warning
Screenshot 2025-04-02 at 4 57 22 PM

## Testing 

I tested the build GitHub actions workflow on my fork and verified that the build does not show a damaged warning on my machine. (It still shows a Gatekeeper warning, but this is expected without notarization and can be bypassed without the use of Terminal.)

## Desktop

- **OS:** macOS
- **OS Version:** 15.4 (24E248)
- **FreeTube version:** 0.23.3

## Additional context

Documentation may need updating to move away from the suggestions of `xattr` and towards [the GUI method of bypassing the warning](https://support.apple.com/en-us/102445#openanyway)?

### DIFF
--- a/_scripts/ebuilder.config.js
+++ b/_scripts/ebuilder.config.js
@@ -30,6 +30,9 @@ const config = {
   // electron-builder will however still spend time scanning the `node_modules` folder and building up a list of dependencies,
   // returning `false` from the `beforeBuild` hook skips that.
   beforeBuild: () => Promise.resolve(false),
+  // electron-builder doesn't seem to handle ad-hoc signatures on macOS properly. To work around this, manually call
+  // the `codesign` command after building the app but before packaging it into distributable forms (dmg, etc.).
+  afterPack: './_scripts/signMacBeforePackaging.js',
   dmg: {
     contents: [
       {
--- a/_scripts/signMacBeforePackaging.js
+++ b/_scripts/signMacBeforePackaging.js
@@ -1,0 +1,13 @@
+const { execSync } = require('child_process')
+const builder = require('electron-builder')
+const path = require('path')
+
+exports.default = async function(context) {
+  if (context.electronPlatformName === 'darwin' && context.arch === builder.Arch.arm64) {
+    const appPath = path.join(context.appOutDir, 'FreeTube.app')
+    const frameworksPath = path.join(appPath, 'Contents/Frameworks')
+
+    execSync(`/usr/bin/codesign --force --sign - ${frameworksPath.toString()}/*`)
+    execSync(`/usr/bin/codesign --force --sign - ${appPath.toString()}`)
+  }
+}


absidue:
It would be better to figure out what the actual issue is and fix it upstream, instead of adding a "temporary" (we all know how temporary workarounds usually become permanent technical debt) workaround here.

kthchew:
The Electron build system seems quite complex and it would probably be a complicated task to see exactly what is incompatible with the signature applied by the linker. The most practical solution would likely still be to manually run codesign after the fact, but passing in an empty identity for an ad-hoc signature rather than an actual identity.

Given that Electron seems to have a working code signing system for developers with actual certificates, I'd be willing to bet that if it were modified to accept ad-hoc identities (and automatically use them on ARM) this issue would be resolved. But yes, you're probably right that it would be ideal to fix it over there.

kthchew:
The PR over at `electron-userland/electron-builder` is going well, closing in favor of that one

Ein-Tim:
Link back to the corresponding PR: https://github.com/electron-userland/electron-builder/pull/9007

PR #7152, opened 2025-04-05
Migrate FtChannelBubble, FtProfileChannelList and FtProfileFilterChannelList to the composition API

absidue:
# Migrate FtChannelBubble, FtProfileChannelList and FtProfileFilterChannelList to the composition API

## Pull Request Type

## Description

This pull request migrates the `FtChannelBubble`, `FtProfileChannelList` and `FtProfileFilterChannelList` components to the composition API. I decided to change the way that the channel selection worked, previously the list would toggle the selected state with refs and directly modifying the data in the channel bubble component, the new way is that the list component maintains an array (can be switched to a set in Vue 3) of selected channel IDs and applies the selected state to the bubbles through a prop. This is cleaner (at least in my opinion) and will be nicer with Vue 3, as Vue 3's composition API doesn't expose any data or methods outside of the component by default so we would have had to explicitly mark the `selected` variable in the `FtChannelBubble` component as exposed.

I'm sorry @ChunkyProgrammer for keeping this in my git stash for so long.

## Testing

Test the profile settings page e.g. adding channels from one profile to another, removing channels, check that the selecting and unselecting behaviour works correctly.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** aee841a691aae3af8091657731d54f517cb22883


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
@ChunkyProgrammer :3 ?

ChunkyProgrammer (approved):

PR #7156, opened 2025-04-06
Fix filename tooltip not showing under screenshot in player settings

MarmadileManteater:
# Fix filename tooltip not showing under screenshot settings

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
The tooltip under the screenshot heading in player settings is hidden on small displays because it is actually stuck in the top corner of the page. This causes overflow as well because the tooltip is not contained within the player settings section. This PR addresses this by adding `position: relative` to `.screenshotFilenamePatternTitle` and adding margins to the tooltip to put it in a visually acceptable place as to prevent overflow.

## Screenshots 

|before|after|
|--|--|
|2025-04-06-134813_hyprshot|2025-04-06-134802_hyprshot|

## Testing 
1. Open up player settings
2. Reduce the window width to below `680px`
3. Ensure the tooltip icon remains visible and there is no horizontal scrollbar on the page

## Desktop

- **OS:** Fedora w/ Hyprland
- **OS Version:** 41
- **FreeTube version:** a058ccdab0616e81d54df86818a54fea99bda8ec

### DIFF
--- a/src/renderer/components/player-settings/player-settings.css
+++ b/src/renderer/components/player-settings/player-settings.css
@@ -18,3 +18,12 @@
   flex-grow: 1;
   margin-block-start: 10px;
 }
+
+.screenshotFilenamePatternTitle {
+  position: relative;
+  margin-inline-end: 20px;
+}
+
+.tooltip {
+  margin-block-start: -12px;
+}


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE:
Probably should try to see if the existing style can be fixed first
image

PikachuEXE:
Alternative:
`src/renderer/components/player-settings/player-settings.vue`
image
`src/renderer/components/FtSettingsSection/FtSettingsSection.scss`
image

MarmadileManteater:
The more I look at the Settings view, the more I hate the css. I just don't like this fix because it is too surface level.

PR #7139, opened 2025-04-02
Remember playlist sort

anurag2787:
## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #5008

## Description
This PR migrate the user playlist sort preference from sessionStorage to Vuex store ensuring that the sorting preference saves even after the app reopens. 

## Desktop
- **OS:** Windows 11  
- **OS Version:** 24H2 (OS Build 26100.3476)  
- **FreeTube version:** 0.23.2  

## Additional context
This feature improves user experience by keeping sorting preferences saved permanently, rather than resetting after each session.

PR #7157, opened 2025-04-06
Added sort by date

### LINKED ISSUE #5595: [Feature Request]: Add sort by date/channel button/filter to the history page
### Guidelines

### Problem Description

I'm unsure if I should make a new feature request for this, or not, but the history has no way of switching between old and new videos (I have to scroll all the way down to see older videos)

(there's somewhat similar request but none are quite exactly like this ask) 

### Proposed Solution

a sort by date/channel button in the history? 

similar to the general search at top, the search in history back should have a button for filtering, for this specific request, just a sort by feature is more than enough

### Alternatives Considered

pagination of the history page would also work to get to the bottom, but sorting probably feels a bit better to me

### Issue Labels

### Additional Information

I would imagine a filter button like the one that already exists in the top platform would be more than enough, for now sort by would be the only option and I would imagine the options would be:

viewed date
upload date

just those two options are probably more than enough for this feature

anurag2787:
## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #5595 

## Description
In this PR i have added sort button in the history page where allowing user can organize their watch history either by Latest First or Oldest First where it will shows the watch history in ascending or descending order

## Sort By Oldest First
Screenshot 2025-04-06 233452

## Sort By Latest First
Screenshot 2025-04-06 233431

## Desktop
- **OS:** Windows 11  
- **OS Version:** 24H2 (OS Build 26100.3476)  
- **FreeTube version:** 0.23.2


efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @anurag2787 nice to see a pr from you again. Did you already take a look at https://github.com/FreeTubeApp/FreeTube/pull/6214 and the feedback we left on that

src/renderer/views/History/History.vue
```diff
@@ -36,6 +36,16 @@
...
           @change="doCaseSensitiveSearch = !doCaseSensitiveSearch"
         />
+        <ft-select
+          v-if="fullData.length > 1"
+          class="sortSelect"
+          :placeholder="$t('Global.Sort By')"
+          :value="sortOrder"
+          :select-names="sortByNames"
+          :select-values="sortByValues"
+          :icon="getIconForSortPreference(setHistorySortOrder)"
```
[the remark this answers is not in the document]
efb4f5ff-1298-471a-8973-3d47447115dc: tested how copilot would work on this PR. Please disregard if this is a non issue.

anurag2787:
In getIconForSortPreference function instead of passing setHistorySortOrder i have passed the value of it sortOrder so that it can determine the correct icon

static/locales/en-US.yaml
```diff
@@ -49,6 +49,8 @@ Global:
   Live: Live
   Community: Community
   Sort By: Sort By
+  DateOldestHistory: Oldest First
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 52:
```suggestion
  DateOldestHistory: Earliest Watched First
```

static/locales/en-US.yaml
```diff
@@ -49,6 +49,8 @@ Global:
...
+  DateNewestHistory: Latest First
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 53:
```suggestion
  DateNewestHistory: Latest Watched First
```

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Icon doesnt change like on other pages.

https://github.com/user-attachments/assets/00b7a6b2-1b9b-427e-ae94-267aa97efd89

anurag2787:
Fixed Icon change like on other pages

https://github.com/user-attachments/assets/e0300fdb-e625-4283-81e1-9aa8e83eadb9

Also Updated history sort labels

anurag2787:
Fixed Icon change like on other pages 

https://github.com/user-attachments/assets/aa9d30b3-f4d7-4fd2-89b9-2c962e690bc5

Also Updated history sort labels

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (commented):
It works but I wonder why the new translation keys are added to `Global` not `History`
image

anurag2787:
I have fixed it, the new translation keys now been added to the History instead of Global

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (dismissed):

src/renderer/store/modules/history.js
```diff
@@ -1,8 +1,10 @@
 import { set as vueSet, del as vueDel } from 'vue'
 import { DBHistoryHandlers } from '../../../datastores/handlers/index'
+import { HISTORY_SORT_BY_VALUES } from '../../../constants'
 
 const state = {
   historyCacheSorted: [],
+  useAscendingOrder: HISTORY_SORT_BY_VALUES.DateAddedNewest,
```
kommunarr, line 7:
```suggestion
  sortOrder: HISTORY_SORT_BY_VALUES.DateAddedNewest,
```

**suggestion:** Update variable name and concomitant uses to better reflect its type (sort type, not boolean).

kommunarr (changes requested):
**issue (blocking):** When setting order to `Earliest Watched First` and then new history entries are added, they're shown at the start instead of the end, despite them being the latest entries watched. I believe this is because `upsertToHistoryCache` is not updated to handle non-default sort methods.

anurag2787:
Fixed the issue of new history not showing according to the sort filter applied

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
1. Sort History by Earliest
2. Open new Window
3. Go to History page
4. See that its sorted by Latest

It should be also sorted by Earliest

anurag2787:
Hy, since i have worked on similar feature #7139 where i migrated the playlist sort preference from sessionStorage to Vuex store to persist it across sessions so I'd like to confirm for the history sort preference should i follow same approach and store it in Vuex or prefer to store in sessionStorage instead?

absidue:
In vuex settings store, that way it will also get remembered across app launches. As a general rule unless there is an extremely good reason, you should not be putting anything in session storage in FreeTube.

As for the sorting itself, I feel like this pull request is a bit over engineered. If we break it down the goal is to implement an option to keep the order the the history is already in (newest to oldest) or have it in the reverse order (oldest to newest). That means that the `.sort()` call is actually overkill, as you can either use the original array as is or use `.toReversed()` to get a copy in the reverse order.

The entire sorting could be handled by changing the existing historyCacheSorted computed on the history page, without needing to modify anything in the history store:

```js
historyCacheSorted: function () {
  const historySorted = store.getter.getHistoryCacheSorted

  if (this.historySortOrder === HISTORY_SORT_BY_VALUES.DateAddedOldest) {
    return historySorted.toReversed()
  } else {
    return historySorted
  }
}
```

anurag2787:
I’ve simplified the implementation as per the feedback and removed the over-engineered sorting logic.Additionally i have stored the user history sort preference in the Vuex settings store ensuring it persists across app launches

efb4f5ff-1298-471a-8973-3d47447115dc:
Sorting doesnt change when user uses keywords to filter

https://github.com/user-attachments/assets/7c6dddb3-cd7f-41ae-8cf4-de3c9f0238df

anurag2787:
I have fixed the issue. This was happening because of filterVideosWithQuery has a hardcoded .sort((a, b) => b.timeWatched - a.timeWatched) which overrode the selected sort order.

efb4f5ff-1298-471a-8973-3d47447115dc:
> 1. Sort History by Earliest
> 
>     2. Open new Window
> 
>     3. Go to History page
> 
>     4. See that its sorted by Latest
> 
> 
> It should be also sorted by Earliest

Its better but not quite the same UX as how other dropdowns behave

https://github.com/user-attachments/assets/ddc11fea-27f5-4a02-b969-e8dd9b1fb6c2

anurag2787:
I have checked carefully but i am still having trouble in spotting exactly what behavior is different between my dropdown and the expected one could you please elaborate a bit more

efb4f5ff-1298-471a-8973-3d47447115dc:
@anurag2787 so in the first few seconds of the video the following happens. When sort order changes it automatically also changes in the other window.

When i navigate towards the History page this doesnt happen

anurag2787:
Thanks for clarifying I see the issue now and have fixed it so that the sort order syncs correctly

efb4f5ff-1298-471a-8973-3d47447115dc:
Apologies for the wait. We will get back to reviewing this as soon as possible

PikachuEXE:
Need all the test cases in testing section...
- Icon should change
- Earliest first should not show latest when new entry added
- Sorting doesn't change when user uses keywords to filter (I don't even know what this means
- Sort order change affects all other windows and new windows and future app instances

I definitely miss something I dunno what...

anurag2787:
I have added the testing section with the required test cases as mentioned please review whenever you have time. Thanks!

PikachuEXE (dismissed):

src/renderer/views/History/History.js
```diff
@@ -185,5 +217,10 @@ export default defineComponent({
     keyboardShortcutHandler: function (event) {
       ctrlFHandler(event, this.$refs.searchBar)
     },
+
+    getIconForSortPreference: (s) => getIconForSortPreference(s),
+    ...mapMutations([
+      'setsortBy'
+    ]),
```
absidue: This seems to be unused so can probably be removed.

anurag2787:
I have accidentally pushed yarn.lock file in my commit and am now facing some difficulties reverting it back to the original state could you please help me to resolve this

PikachuEXE:
Reversed

PR #7184, opened 2025-04-10
Use `maxVideoPlaybackRate` for the max value of Default Playback Rate

efb4f5ff-1298-471a-8973-3d47447115dc:
# Use `maxVideoPlaybackRate` for the max value of Default Playback Rate

## Pull Request Type

- [x] Bugfix

## Description

Default Playback Rate slider doesn't exceed max value set by Max Video Playback Rate slider

## Testing 

1. Set Max Video Playback Rate slider to a value, e.g. 2x
2. Use Default Playback Rate slider to check that you cant go beyond previously set value

## Desktop

- **OS: Ubuntu**
- **OS Version: 24.04 LTS**
- **FreeTube version: latest**

## Additional context

I noticed that the Default Playback Rate slider was set to max value of 8x but the Max Video Playback Rate slider goes up to 10x so i initially wanted to adjust that to 10x just to fix that inconsistency. Then i noticed that when I set the Max Video Playback Rate slider to 2x and i was able to set the Default Playback Rate slider to values >2x and the video player happily played at those speeds but didn't show them in the video player. So i guess a kind of 2 birds with one stone kind of situation here.

Also related but already existing behavior:

Sliders values don't automatically auto adjust. It requires user interaction before it uses the new values. If user doesnt interact with the slider than the player also plays the value. Any ideas how we could address that?

https://github.com/user-attachments/assets/7119745e-8078-448f-81c5-740a20c46045

### DIFF
--- a/src/renderer/components/player-settings/player-settings.vue
+++ b/src/renderer/components/player-settings/player-settings.vue
@@ -151,7 +151,7 @@
         :label="$t('Settings.Player Settings.Default Playback Rate')"
         :default-value="defaultPlayback"
         :min-value="parseFloat(videoPlaybackRateInterval)"
-        :max-value="8"
+        :max-value="maxVideoPlaybackRate"
         :step="parseFloat(videoPlaybackRateInterval)"
         value-extension="x"
         @change="updateDefaultPlayback"


kommunarr (approved):
> Sliders values don't automatically auto adjust. It requires user interaction before it uses the new values. If user doesnt interact with the slider than the player also plays the value. Any ideas how we could address that?

I'm not sure if there's a cleaner way to do it, but what comes to mind is applying a custom rounding function on either an intermediary function call in `player-settings.js` or as a side effect handler through our system we have set up for that.

absidue (approved):

PikachuEXE (approved):

PR #7186, opened 2025-04-11
Support LockupView videos in the up next feed

absidue:
# Support LockupView videos in the up next feed

## Pull Request Type

- [x] Bugfix

## Related issue

Surprisingly nobody has reported this yet

## Description

This pull request fixes the up next feed occasionally being empty, by adding support for `VIDEO` `LockupView`s and also fixes movies not being parsed correctly (their video ID was being set to undefined, which meant the thumbnail didn't appear and clicking on it produced an error). This doesn't currently support upcoming videos as I wasn't able to find an example of one appearing in the up next feed.

## Screenshots

Before with LockupViews
before-lockup-views

After with LockupViews
after-lockup-views

Before with CompactMovies
before-movies

After with CompactMovies
after-movies

## Testing

LockupViews appearing in the up next feed still seems to be somewhat random, so that isn't the easiest thing to test but testing the movie parsing changes is doable as you just need to find a YouTube Movie (they usually only have YouTube movies in their up next lists).

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 9b372e8dc335de6d860f28ee620a1d0e042759c1

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -1220,6 +1220,52 @@ function parseLockupView(lockupView, channelId = undefined, channelName = undefi
         videoCount: extractNumberFromString(thumbnailOverlayBadgeView.badges[0].text)
       }
     }
+    case 'VIDEO': {
+      let publishedText
+      let lengthSeconds = ''
+      let liveNow = false
+
+      /** @type {YTNodes.ThumbnailOverlayBadgeView | undefined} */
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
All other trailers work fine bit `Moana 2` does throw an error but i think its unrelated to this PR

VirtualBoxVM_g4yEcMXKyi

ChunkyProgrammer (approved):

PikachuEXE (approved):

PR #7213, opened 2025-04-16
Fix user-agent and referer for Innertube

absidue:
# Fix user-agent and referer for Innertube

## Pull Request Type

- [x] Bugfix

## Description

This pull request fixes the user-agent in the request payload not matching the user-agent header, as YouTube.js was chosing a random one for the request payload and also fixes the referer header for the session data request.

Thanks to @AudricV for discovering these inconsistencies.

## Testing

Test a few videos and check that they play correctly. While I wasn't able to confirm if this fixes the IP block issue that users have been reporting, as I couldn't reproduce that issue myself, this should hopefully help the situation. If someone can confirm that this does solve the issue we can link it with a closes line.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 765452d9daa44d9779f9d17a169249e815c6692d


PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

Eiji7:
@absidue "If someone can confirm that this does solve the issue we can link it with a closes line."

After downloading the `0.23.4` Beta for some time I was unable to fetch video list from subscriptions (it was just empty without any error). I have searched for desired video and watched it and then I was able to fetch all videos, weird but in previous version I had `IP block` error and I needed to open the video in `mpv`. Right now I'm playing the 2nd video and it seems to be fixed. Will give more details if I would have any problems related to this PR. Thanks for your fix! :heart:

Issue #6756, opened 2025-02-05
[Feature Request]: A toggle button for sponsorblock in the player.


shellheim:
### Guidelines

- [x] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

When you encounter an interesting sponsor segment that you want to watch, you have to go to the settings to disable sponsorblock and play the video from the beginning of the segment.

A great example of such a segment would be sponsorblock skipping segments in Louis Rosmann's video with Clinton the Cat. It would be nice to have a toggle in the player itself that toggles sponsorblock. LibreTube already does this pretty well.
 
#1380 is sort of similar to this, but that requires the user to press enter on every segment, which is a bit inconvenient. 

### Proposed Solution

A simple toggle button with the shield icon of sponsorblock. When sponsorblock is on we'll display this image on the button (telling the user that they can turn it off):

![Image](https://github.com/user-attachments/assets/f4bcf62c-504a-4659-ad45-87da169e32ea)

And when it's on we'll display a normal shield icon (telling the user they can turn it back on): 

![Image](https://github.com/user-attachments/assets/237e53c2-2b26-4bd1-93fa-552799a852ac)

### Alternatives Considered

Nil

### Issue Labels

new feature, ease of use improvement, improvement to existing feature

### Additional Information

N/A

hunkyburrito:
There also used to be a way to seek back into the skipped segment to watch it (#3116) which would be cool to see return if nothing else.

efb4f5ff-1298-471a-8973-3d47447115dc:
closing for the reasons stated https://github.com/FreeTubeApp/FreeTube/pull/7492#pullrequestreview-2884613161

PR #4978, opened 2024-04-18
DRAFT: Migrate video player from video.js to shaka-player

absidue:
# DRAFT: Migrate video player from video.js to shaka-player

## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation

## Related issue
* closes #624
* closes #661
* closes #959
* closes #1348
* closes #2013
* closes #2098
* closes #2365
* closes #4137
* closes #4039
* closes #3920
* closes #3560
* closes #3534
* closes #3163
* closes #2931
* closes #2455
* closes #845
* closes #1344

## Description
After almost a years worth of work, here is the shaka-player migration.

### Significant improvements and new features
* vp9 and opus during DASH playback. vp9 only works on the local API due to Invidious limitations
* The video player will now match the users display language when possible (41/45 locales) and custom player elements added by FreeTube are now translatable. shaka-player doesn't support Estonian `et` (eesti keel), Basque `eu` (Euskera), Galician `gl` (galego) and Icelandic `is` (Íslenska), for those languages it will use English (US) instead.
* Stats are now supported for all 3 Media Formats, not just DASH.
* Switching formats with keep your playback rate, selected subtitle, audio language and video quality settings and the player will no longer flash while switching.
* The player should be a lot more usuable on mobile screens, as player elements get moved into an overflow menu when they don't fit.
* Seeking on live streams when the uploader hasn't disabled it (4 hours back for normal ones, 2 hours on low latency live streams). Not supported when video proxying is enabled, as Invidious doesn't support proxying DASH live streams.
* You can now watch just finished live streams (Post-Live-DVR) from the start with the local API and you'll get thumbnails on the seek bar when they are available (the thumbnails seem to be missing on about 50% of Post-Live-DVR streams).
* The video player will pick up most settings changes without having to leave the watch page, such as changing the display language or the playback rate interval.
* Subtitles will now work on live streams that have them (very rare but they do exist), local API only.

### Significant bug fixes
* The subtitle translation feature now works on all videos, not just ones that are in English.
* Starting at a certain timestamp from the video URL or resuming a partially watched video, should now work all the time, instead of just randomly.
* Long videos start straigh away and no longer freeze FreeTube for almost a minute.

### Missing/Removed features
* shaka-player doesn't have any subtitle settings. Maybe in the future we could implement our own ones, but it's not-trivial to implement, so creating a replacement is out-of-scope for this pull request.
[...]

PR #7217, opened 2025-04-17
Allow unskipping of auto-skipped SponsorBlock segments

0je:
# Allow unskipping of auto-skipped SponsorBlock segments

via return key while toast is shown.
Another return hit reskips the segment.

Similar to how it works on youtube.com or Invidious with SponsorBlock.

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Relevant issues are:
#6756 #1380 same problem, different suggestion
#3116 PR that allowed seeking into otherwise auto-skipped segment, seems like it got lost with the move to the shaka video player.

## Description
When a sponsor segment was auto-skipped, I sometimes want to unskip it for at least 2 reasons.
The SponsorBlock segments are community provided and sometimes wrong/vandalism.
Another reason are funny transitions that make me wonder what the ad is about.
For now one would have to temporarily disable auto-skip / SponsorBlock and reload the video to watch the segment.
This PR adds the option to unskip an auto-skipped segment by hitting the return key while the "segment skipped" toast is shown.
Another return hit re-skips to the end of the segment.


absidue:
> #3116 PR that allowed seeking into otherwise auto-skipped segment, seems like it got lost with the move to the shaka video player.

It wasn't lost, it was intentionally removed. Quoting the migration PR #4978:

> I removed the ability to seek into sponsor block segments that are set to skip, as it was causing the skipping to not work properly, due to the existing implementation being fundamentally flawed. If someone would like to add something like that back in the future, it should require a specific user-interaction to unskip the segment, e.g. a button in the skip message, instead of allowing users to manually seek into the segment.

efb4f5ff-1298-471a-8973-3d47447115dc:
Just to be clear there are 2 ways with SP to get into a segment:
- By letting the user literally click into the sponsor segment and nothing would happen
- By skipping the segment but giving the user the option to "unskip" it by going back to the beginning of the skipped segment and letting it play normally.

This PR implements the latter one.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
How it is now:


https://github.com/user-attachments/assets/b4edd182-f114-4694-b618-bb899d4174cc



Make toasts the same as SP so that the toasts define that you have to press enter to unskip/reskip

https://github.com/user-attachments/assets/69c5f95c-6eea-4b4b-8c1d-79e6105443a4

PR #7281, opened 2025-04-26
Remove broken restart window keyboard shortcut

absidue:
# Remove broken restart window keyboard shortcut

## Pull Request Type

- [x] Bugfix

## Description

Despite advising users on many occasions that they should not use these keyboard shortcuts or only use them when there is no other option such as when the window crashes, users still use them and are then surprised that it breaks things. If only someone had warned them _deep sigh_. As the only valid use cases such as the window crashing, can also be fixed by closing the window and opening a new one, I see no reason to keep something that is known to be broken around.

If you disagree about removing it, then we need to find another solution such as displaying a blocking dialog each time a user tries to use those keyboard shortcuts, about the problems with it, also recommending that they do not report ANY issues they may experience in FreeTube if they use those keyboard shortcuts and require them to explicitly say yes every single time. That way if users still abuse it, they can no longer pretend they didn't know it would break things, so any problems that are caused by it are then their own fault.

## Testing

Check that CTRL+R does nothing.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** d5f6f7b61ac78e65c1f8d484c3bd0422111128f1

## Additional context

You may notice that various translation files have other changed strings, that is because translators are unfortunately still translating the placeholder values, in some cases even omitting them from the translation entirely and those are the ones that got flagged by the linter while I was committing the changes.


efb4f5ff-1298-471a-8973-3d47447115dc:
Unfortunately a much needed change as users clearly dont understand the harm they can do even with us repeating the implications of it multiple times. Sometimes the mental model of a browser really works against us and this is sadly one of them. 

Code looks good, will verify by testing later

PikachuEXE (approved):

Ein-Tim:
I’d recommend adding an explanation to the FAQs („Why did you remove the refresh shortcut?“) and link to this entry from the release notes to avoid people opening GitHub issues or complaining on other channels that „the feature has stopped working“.

Alternatively, if you don‘t want things like this in the docs/FAQ, I‘d definitely mention it in the release notes.

absidue:
@Ein-Tim Refreshing and restarting windows are two separate things, this only removes restarting windows. Refreshing, which is what you do on the subscription, trending and popular pages still exists, as well as the associated keyboard shortcuts. The misconception that restarting a window is just a refresh is likely one of the reasons why we are in this situation.

Ein-Tim:
@absidue Ahh, thanks for the explanation! So IIUC this PR will remove the ability to restart a window, but refreshing via the button in the top left coroner as well as via the corresponding shortcut will stay. 

But still: CTRL+R is now empty application wide, right? If yes: This will probably get noticed by users which will then complain that they can‘t use CTRL+R anymore and lead to unnecessary issues.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Tested and LGTM but im not ready to approve yet as my user minded view think we should resolve most playback errors before this should go in. That way users wont miss it that much and we'll receive less backlash

k0zyrev:
Sometimes I use Ctrl+R and it doesn't cause issues for me, I guess it's something with Linux users.

Maybe simply add an experimental option disabled by default that would allow us to use this shortcut? You could add any warning to the description of the function so that anyone who enables it knows they're on their own. Honestly, I've seed the ctrl+R being added in the change notes, but never seen "advising users on many occasions that they should not use these keyboard shortcuts"

absidue:
It's nothing to do with Linux, it's to do with the fact that this keyboard shortcut should not exist because FreeTube can't handle it cleanly. If you want to refresh specific pages, then implement proper refreshing on that page (yes it needs to be done on a page per page basis, because the restarting of windows, tear everything down and start it all up again approach that CTRL+R takes is way too messy). Disabling it by default and then having a "I think I know more than the devs, so let me do it anyway" setting is not going to happen, either it gets removed or we make it really inconvenient and annoying to use so that people really only ever use it when the window crashes.

Issue #7119, opened 2025-03-31
[Bug]: Intermittent - "Audio formats are not available for this video"


CoffeeMate2:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.

### Describe the bug

When opening a video from search results, or subscription page, rarely (perhaps 1/30 times) a message "Audio formats are not available for this video" is briefly displayed at the bottom of the program and the video will not play.
Going back and then re-opening the video seems to always resolve the issue, and the video then plays with no problem. 

### Expected Behavior

The video should play first time, as audio formats are available, as is evident of retrying the same video a second time and it working. 

### Issue Labels

inconsistent behavior

### FreeTube Version

v0.23.3 Beta

### Operating System Version

Microsoft Windows Version 10 22H2 (OS Build 19045:5608)

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

absidue:
Quoting the changelog, it literally says don't open bug reports for this:
> IMPORTANT
> Please note that you may occasionally encounter Legacy Formats (available video quality, 360p) while playing a video. This is expected behavior, **SO DO NOT SUBMIT BUG REPORTS FOR THIS**. In these cases, FreeTube's video playback wouldn't work at all, and an error message (No valid URLs to decipher error) would be shown. This workaround allows users to still watch videos in those cases.

Infinality:
In my experience with this bug, the error will be displayed, and the video won't play at all.  If I leave the video page and come back into it a few times, it will eventually work.  So, something isn't working properly, and in my opinion it's a legitimate bug, if only because there is inconsistent behavior when viewing the video page multiple times.

CoffeeMate2:
[...]

PR #7327, opened 2025-04-30
Use the streaming data from MWEB to avoid SABR-only responses

absidue:
# Use the streaming data from MWEB to avoid SABR-only responses

## Pull Request Type

- [x] Bugfix

## Related issue
- #7119

## Description

This pull request temporarily switches to using the streaming data from the MWEB client to avoid SABR-only responses, as the SABR implemenation still has various issues and this allows us to do a quick hotfix. The only known downside is that we lose support for multiple audio tracks (you'll only get the original one now) but Stable Volume is still available.

## Testing

Open a few videos and check that they work.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 818290941946e65703cc01c2f71cc9d1306ff365

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -226,6 +226,14 @@ export async function getLocalVideoInfo(id) {
 
   const info = await webInnertube.getInfo(id)
 
+  // temporary workaround for SABR-only responses
+  const mwebInfo = await webInnertube.getBasicInfo(id, 'MWEB')
+
+  if (mwebInfo.playability_status.status === 'OK' && mwebInfo.streaming_data) {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PikachuEXE (approved):

MarmadileManteater:
> Can't you make switching the client into an option in the settings so when youtube is tweaking something again users can try this?

That would increase maintenance requirements and not be as simple as switching between clients with a flag.

It also wouldn't have the benefit you are imagining it would have. You can not predict what clients will break at any given time. It makes more sense to just wait, see and hotfix.

Issue #1864, opened 2021-10-30
VIM like keybindings


starget24:
**Is your feature request related to a usage problem (not a bug)? Please describe.**
Basically to reduce at the minimum the usage of the hand killing mouse
**Describe the solution you'd like to see implemented**
Maybe something like Vimium for Firefox or Chrome

**Describe alternatives you've considered**
I tried using a browser with VIM like keybindings but it will be nice to have it in freetube so i don't need to enter to a sketchy ass site like youtube in my browser.

efb4f5ff-1298-471a-8973-3d47447115dc:
Related to https://github.com/FreeTubeApp/FreeTube/issues/251

UncleOilybag:
I would also like to see a VIM like approach to using FreeTube - it is so much easier than using a mouse or touchpad

RiQuY:
Would be nice to have this, can anyone from the dev team mention if this request is planned or out of scope?

efb4f5ff-1298-471a-8973-3d47447115dc:
@RiQuY it isnt planned but also not out of scope. There are just too many other things on our list that have a higher priority than this.

xeon826:
Thirded

fuguesoft:
plz plz plzplzplzplzplzplz
This is one thing I'll go back to the librewolf/tridactyl setup for

If there was vim navigation and customizable keybinds, I can't say I'd ever touch youtube.com proper again

not that I think this is your goal mind you

PR #7336, opened 2025-05-01
Vim waypoints

xeon826:
# VIM Navigation

## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
https://github.com/FreeTubeApp/FreeTube/issues/251#issuecomment-2164926447
https://github.com/FreeTubeApp/FreeTube/issues/1864
Closes https://github.com/FreeTubeApp/FreeTube/issues/1864

## Description
Implements VIM-like key-bindings, similar to Vimium.

## Screenshots 
image
image

## Testing 
I tested this with basic usage. Navigating to both general and video player keybinding contexts. 

## Desktop
- **OS:** Arch
- **OS Version:** Rolling
- **FreeTube version:** v0.23.1 Beta

## Additional context
Any conflicting native keybindings will change when vim navigation is enabled. The change will reflect in the keyboard shortcuts modal (Shift + ?). I've yet to implement back and forth page navigation but it should be pretty easy.


src/renderer/store/modules/utils.js
```diff
@@ -888,6 +901,159 @@ const mutations = {
     state.isKeyboardShortcutPromptShown = payload
   },
 
+  setAreVimWaypointsShown(state, { key }) {
+    const selectedLetterColor = '#D4AC3A'
+    if (key === 'f' && !document.querySelector('.vimHint')) {
+      const elements = getClickableElements()
+      displayHints(elements)
+      state.areVimWaypointsShown.selector.push(key)
+    } else if (['Esc', 'Escape'].includes(key) | (key === 'Backspace' && state.areVimWaypointsShown.selector.length === 1)) {
```
absidue, line 910:
```suggestion
    } else if (['Esc', 'Escape'].includes(key) || (key === 'Backspace' && state.areVimWaypointsShown.selector.length === 1)) {
```

src/renderer/store/modules/utils.js
```diff
@@ -34,6 +34,7 @@ const state = {
   showAddToPlaylistPrompt: false,
   showCreatePlaylistPrompt: false,
   isKeyboardShortcutPromptShown: false,
+  areVimWaypointsShown: { selector: [] },
```
absidue, line 37: From reading the code it looks like you are misusing this single variable for both an object and booleans.

efb4f5ff-1298-471a-8973-3d47447115dc:
I’m a bit concerned that the new PR might end up getting abandoned just like the previous one, especially since it was abandoned after one day. I’m worried that reviewing it might end up being a waste of time if it gets discarded again.

src/renderer/store/modules/utils.js
```diff
@@ -888,6 +901,159 @@ const mutations = {
...
+          break
+        }
+      }
+
+      // Pad the hint string to ensure its length is at least 'digits'.
+      while (hintStr.length < digits) {
+        hintStr.unshift(chars[0])
+      }
+
+      return hintStr.join('')
```
absidue, line 967: Please use the string padding functions that already exist in the JavaScript language instead of inventing your own.

src/renderer/store/modules/utils.js
```diff
@@ -888,6 +901,159 @@ const mutations = {
...
+          el.children[i].style = ''
+        }
+      }
+    })
+
+    function close() {
+      document.querySelectorAll('.vimHint').forEach((el) => el.remove())
+      state.areVimWaypointsShown.selector = []
+    }
+    console.warn(selector)
```
absidue, line 940: Please don't leave debugging code lying around, the whole point of the linter blocking `console.log` is to avoid this happening.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2065,6 +2067,15 @@ export default defineComponent({
      * @param {KeyboardEvent} event
      */
     function keyboardShortcutHandler(event) {
+      const vimEnabled = store.getters.getEnableVimNavigation
+      const passToVim =
+        // If vim keys are enabled, ctrl is not held down, and the key is not f, j, or k.
+        (vimEnabled && !event.ctrlKey && ['f', 'j', 'k'].includes(event.key)) || // OR
+        // The waypoints are already on the screen
+        (store.getters.getAreVimWaypointsShown.length && store.getters.getAreVimWaypointsShown.selector[0] === 'f')
```
absidue, line 2075: Please use computeds for these store values, like is done in the rest of the file.

absidue (commented):
At first glace there are a few issues with this pull request:
- Please fix the invalid code (referencing variables that you didn't actually create), missing translation strings and wrong indentation pointed out by the linter.
- The title of this pull request is misleading, it sounds like you are just implementing VIM waypoints but this actually changes various keyboard shortcuts, which personally I don't think is a good idea unless it is made extremely obvious to the user that that is what is happening.
- Please choose a unique keyboard shortcuts to show and hide the waypoints as well as for the other VIM interactions such as scrolling, they should not conflict with the existing keyboard shortcuts and hotkeys. Alternatively you could ensure that these VIM hotkeys are disabled on the watch page or make it very obvious to the user that the behaviour of various well-known keyboard shortcuts and hot keys will change when the VIM setting is enabled.
- Please don't unnecessarily place functions inside of functions, as that will result in the inner functions getting treated like variables and recreated every time the outer function is called.
- While a prompt is open only things inside that prompt can be triggered but this will show waypoints even on things that are not accessible while prompts are open.

src/renderer/store/modules/utils.js
xeon826: It's redundant to have a boolean to denote if they're shown or not which can be deduced by checking if the selector is empty or not.

absidue: That is correct but the comment was about the fact that in some places you are explicitly setting that variable to `true` and `false`.

src/renderer/store/modules/utils.js
xeon826: https://github.com/FreeTubeApp/FreeTube/pull/7336/commits/e4b381d7d9f601441c5e106af0071091c061a2c1

src/renderer/store/modules/utils.js
xeon826: https://github.com/FreeTubeApp/FreeTube/pull/7336/commits/e4b381d7d9f601441c5e106af0071091c061a2c1

src/renderer/store/modules/utils.js
xeon826: https://github.com/FreeTubeApp/FreeTube/pull/7336/commits/e010f52198e74573578558a739bc82858c4e85e7

absidue:
Instead of pushing and commenting after every single commit, please just do one push and comment at the end, so that we don't get spammed with notifications.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
xeon826: https://github.com/FreeTubeApp/FreeTube/pull/7336/commits/32a250e4920cd101d1f3257de80d2435c728eb79

xeon826:
I changed the hotkey to toggle the hints to "t" as it was previously assigned to "toggle play/pause" which is already handled by the space bar so it was a waste. I also changed `LARGE_REWIND` to be handled by h instead of j if vim is enabled because j and k are the only hotkeys that make sense for this functionality. I've denoted this in the hotkey popup and can add further info so the user knows if we want. There were some functions inside of `setAreVimWaypointsShown` that were only used once and didn't need to be functions so I replaced their references with their respective code. 

As for missing translations, is there a tool to do this or is it manual?

efb4f5ff-1298-471a-8973-3d47447115dc:
> As for missing translations, is there a tool to do this or is it manual?

Manual.

static/locales/en-US.yaml
```diff
@@ -1040,6 +1041,7 @@ Description:
 #Tooltips
 Tooltips:
   General Settings:
+    Enable VIM key-bindings: Allow navigation using VIM hotkeys.
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 1044: You're not saying anything new here as the toggle says the same https://github.com/FreeTubeApp/FreeTube/pull/7336/files#diff-4eaadcc25bfbd7ab1ba912e36d5e1bcbab346cab59546a970df7546d5fb7a145R277

Remove all code related to the tooltip

absidue:
Please revert all changes to translation files other than en-US. Manual means the translations will be done through weblate that can speak the language fluently, we don't want questionable machine translations. We are now back at the point that we were with the other pull request where there are a massive number of unnecessary changes in unrelated files.

absidue:
The t hotkey was assigned to theatre mode, which now doesn't have a hot key. Please chose a multi-key combination for the VIM hotkeys. Additionally if you want to break existing keyboard shortcuts when this is enabled there needs to be some very clear messaging for the user e.g. a banner or a large message next to the setting when it is enabled like with the proxy settings.

src/renderer/App.js
```diff
@@ -344,20 +351,73 @@ export default defineComponent({
...
+          break
+        case 'Tab':
+          this.showOutlines()
+          break
+        case 'L':
+        case 'l':
+          if ((process.platform !== 'darwin' && event.ctrlKey) || (process.platform === 'darwin' && event.metaKey)) {
+            this.$refs.topNav.focusSearch()
+          }
+          break
```
absidue, line 420: Alt+D and CTRL/CMD+L are handled in the TopNav component so they do not need to be duplicated here.

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc: you removed the tooltip but not the translation

xeon826:
> The t hotkey was assigned to theatre mode, which now doesn't have a hot key. Please chose a multi-key combination for the VIM hotkeys. Additionally if you want to break existing keyboard shortcuts when this is enabled there needs to be some very clear messaging for the user e.g. a banner or a large message next to the setting when it is enabled like with the proxy settings.

When I press t in the development branch, it toggles the playback of the video, are you sure it's being used for theater mode?

absidue:
Yes, I am, because this is where the `KeyboardShortcuts.VIDEO_PLAYER.GENERAL.THEATRE_MODE` property that you deleted is being used:
https://github.com/FreeTubeApp/FreeTube/blob/e2a53d0c86a74d0b4ed28c33d6c98f770862e618/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L2260-L2269

xeon826:
> Yes, I am, because this is where the `KeyboardShortcuts.VIDEO_PLAYER.GENERAL.THEATRE_MODE` property that you deleted is being used:
> 
> https://github.com/FreeTubeApp/FreeTube/blob/e2a53d0c86a74d0b4ed28c33d6c98f770862e618/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L2260-L2269
I'm saying when you open the video player and press t, it toggles the playback, not theater mode.

absidue:
For me it only does theatre mode, it doesn't pause or play the video.

efb4f5ff-1298-471a-8973-3d47447115dc:
There is no bug in the dev branch. You can verify that by looking at the code yourself

efb4f5ff-1298-471a-8973-3d47447115dc:
I see that you commuted something to the dev branch. That info in available to me through GH. This tends me to believe that you have made more commits locally to your dev branch.

xeon826:
- I added a small banner above the vim toggle to inform the user that the large rewind key will be remapped to h if enabled.
- Changed the way-point toggle key to w so we can keep the theater mode toggle as it is. It's important the the vim toggle key be easily accessible as most users will attest it's a very commonly used key particularly when working.
- Re-added the theater mode toggle.
- Created a subsection for vim-related hotkeys in the shortcut modal.

src/renderer/store/modules/utils.js
```diff
@@ -888,6 +901,146 @@ const mutations = {
...
+          fontSize: '13px',
+          fontFamily: '-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Arial, sans-serif',
+          fontWeight: '600',
+          borderRadius: '3px',
+          color: 'black',
+          zIndex: '10000', // Ensure it is above other content
+          padding: '1px 2px',
+          letterSpacing: '0.5px',
+          display: 'flex' // Use flexbox for letter alignment
+        }
```
absidue, line 935: As you are already assigning a class to this element, it would make a lot more sense to put the CSS in the App.css file. You'll need to use the `:deep()` selector as you are creating these elements outside of Vue.

xeon826:
Will have a look.

PR #7366, opened 2025-05-05
Hiding Watched Videos on Channel View

### LINKED ISSUE #4497: [Feature Request]: option to hide watched videos in channel view
### Guidelines

### Problem Description

in the channel view i would like to have an option to hide the already watched videos of the channel

### Proposed Solution

add a checkbox in the channel view to hide watched videos

### Alternatives Considered

maybe this could also be a global filter, if set, no already watched video is shown

### Issue Labels

new feature, new optional setting

### Additional Information

_No response_

palharesf:
# Issue 4497 - Hiding Watched Videos on Channel View

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
Closes #4497

## Description
The 'Hide Videos on Watch' setting has been moved from the 'Subscription' section to the 'Distraction Free' section. Now, it not only hides watched videos (and shorts and lives) from the subscription page, but also from the channel page (from the videos, shorts and lives tab, but not from the home tab).

## Screenshots
Current implementation:
image
image
image

Proposed change (settings page):
image
image

Channel pages (untoggled):
image
image
image

Channel pages (toggled):
image
image
image

Tooltip:
image

## Testing
(A)
- Navigate to a random channel
- Navigate to Videos tab
- Watch a video or manually mark it as watched -- it should only gray out
- Go to Settings --> Distraction Free --> General --> toggle 'Hide Videos on Watch'
- Return to the previously channel page
- Watched videos should not load anymore
- Untoggling the 'Hide Videos on Watch' and returning to the page should reload the videos

(B)
- Repeat all steps from the test (A), but choose a channel with few videos
- Leave two unwatched videos
- Ensure the toggle is on
- Return to the channel page
- Upon marking one of the last two videos as watched, the 'Sort By' dropdown will disappear (since there's no need to sort a single video, shorts or live)
- Screenshots:
image
image

## Desktop
Windows 10 Pro
OS Build:	19045.5796
Freetube v0.23.4 Beta

## Additional context

### DIFF
--- a/src/renderer/views/Channel/Channel.vue
+++ b/src/renderer/views/Channel/Channel.vue
@@ -41,7 +41,7 @@
       <div class="select-container">
         <FtSelect
           v-if="showVideoSortBy"
-          v-show="currentTab === 'videos' && latestVideos.length > 1"
+          v-show="currentTab === 'videos' && (showFetchMoreButton || filteredVideos.length > 1)"
           :value="videoSortBy"
           :select-names="videoLiveShortSelectNames"
           :select-values="videoLiveShortSelectValues"
@@ -51,7 +51,7 @@
         />
         <FtSelect
           v-if="!hideChannelShorts && showShortSortBy"
-          v-show="currentTab === 'shorts' && latestShorts.length > 1"
+          v-show="currentTab === 'shorts' && (showFetchMoreButton || filteredShorts.length > 1)"
           :value="shortSortBy"
           :select-names="videoLiveShortSelectNames"
           :select-values="videoLiveShortSelectValues"
@@ -61,7 +61,7 @@
         />
         <FtSelect
           v-if="!hideLiveStreams && showLiveSortBy"
-          v-show="currentTab === 'live' && latestLive.length > 1"
+          v-show="currentTab === 'live' && (showFetchMoreButton || filteredLive.length > 1)"
           :value="liveSortBy"
           :select-names="videoLiveShortSelectNames"
           :select-values="videoLiveShortSelectValues"


efb4f5ff-1298-471a-8973-3d47447115dc:
hi @palharesf in the testing section you mainly are mentioning how you tested it. It would be nice if you could provide some testcases (ofcourse we will be testing outside of those cases too but would be nice to have some kind of baseline)

palharesf:
Oh, of course @efb4f5ff-1298-471a-8973-3d47447115dc - I thought the request was for information on how I tested it hahaha

For a specific Channels page, the feature is differentiating between four kinds of states:
- No videos watched, toggle off: all videos should be visible
- No videos watched, toggle on: all videos should be visible
- One or more videos watched, toggle: all videos should be visible
- One or more videos watched, toggle on: one or few videos should not be visible

If I were to write an automated test for this, I would implement these four test cases.

Is that what you were alluding to, @efb4f5ff-1298-471a-8973-3d47447115dc ?

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
The Shorts and the Live tab also contains videos that can marked as watched so it would be a more consistent user experience to also include the toggle on these tabs

palharesf:
I didn't originally touch Shorts and Live as I felt those were outside the scope of the feature, but I'm happy to incorporate those changes tomorrow to harmonize the experience. I'll work on that tomorrow unless told otherwise

efb4f5ff-1298-471a-8973-3d47447115dc:
> Oh, of course @efb4f5ff-1298-471a-8973-3d47447115dc - I thought the request was for information on how I tested it hahaha
 
Maybe that section has to be revised so its more clear that those need to be provided, ill look into it.

> For a specific Channels page, the feature is differentiating between four kinds of states:
> - No videos watched, toggle off: all videos should be visible
> - No videos watched, toggle on: all videos should be visible
> - One or more videos watched, toggle: all videos should be visible
> - One or more videos watched, toggle on: one or few videos should not be visible
> If I were to write an automated test for this, I would implement these four test cases.
> 
> Is that what you were alluding to, @efb4f5ff-1298-471a-8973-3d47447115dc ?

Sorry i had to provide examples. From looking at your cases they look good. Everyone in our team does it a bit different depending on what type of PR they're sending in but these examples should give you a good view with what i meant with testcases. 

- https://github.com/FreeTubeApp/FreeTube/pull/5743
- https://github.com/FreeTubeApp/FreeTube/pull/7349
- https://github.com/FreeTubeApp/FreeTube/pull/5125
- https://github.com/FreeTubeApp/FreeTube/pull/7338

efb4f5ff-1298-471a-8973-3d47447115dc:
Also just a tip make sure your branch is up to date with the latest changes from the development branch before opening your PR.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
1. Sort by `Most Popular`
2. Mark a few videos as Watched
3. Enable Hide Watched
4. See that it doesnt hide the videos
5. Disable Hide Watched
6. See that it does hide the videos


https://github.com/user-attachments/assets/822e8229-6403-4419-86f8-d1d35f008774

palharesf:
Ok, I updated the 'Testing' section of the PR to reflect the examples provided, thank you for those.

I did run `git fetch origin` and `git merge origin/development` before running `git push origin feature/4497-hide-watched-videos` -- all the console messages I got indicated my branch was up-to-date with the dev branch, so I'm sorry if that was not the case.

And the case you mentioned (i.e. marking videos as watched and having them only hidden in the next refresh) was mentioned in my PR as something intentional - I decided against refreshing the video Panel every time one video was marked as watched because I considered it excessive. That said, I can work on an event watcher that tracks videos being marked as 'Watched' manually and forcing a refresh of the Video Panel if that's the preferred way moving forward. I'll include that in my next batch of changes along with the Live and Shorts pages. Thank you!

efb4f5ff-1298-471a-8973-3d47447115dc:
> And the case you mentioned (i.e. marking videos as watched and having them only hidden in the next refresh) was mentioned in my PR as something intentional - I decided against refreshing the video Panel every time one video was marked as watched because I considered it excessive. That said, I can work on an event watcher that tracks videos being marked as 'Watched' manually and forcing a refresh of the Video Panel if that's the preferred way moving forward. 

I think you interpreted my comment wrong. I pointed out a bug. Look at the screen recording again and try the steps to reproduce for yourself. It hides the videos when the toggle is OFF and doesnt hide them when the toggle is ON

palharesf:
> > And the case you mentioned (i.e. marking videos as watched and having them only hidden in the next refresh) was mentioned in my PR as something intentional - I decided against refreshing the video Panel every time one video was marked as watched because I considered it excessive. That said, I can work on an event watcher that tracks videos being marked as 'Watched' manually and forcing a refresh of the Video Panel if that's the preferred way moving forward.
> 
> I think you interpreted my comment wrong. I pointed out a bug. Look at the screen recording again and try the steps to reproduce for yourself. It hides the videos when the toggle is OFF and doesnt hide them when the toggle is ON

You're right, I had misunderstood your request. I apologize. I believe forcing the video panel to refresh once a video is marked as watched would fix this issue.

I've been trying to implement that, but the code for marking as watched is in 'ft-list-video.js' and the code for refreshing the video panel on the Channel page is, naturally, on Channel.vue. I don't want to tightly couple the two pages, so what's the preferred approach here? Using the Vuex store to manage state?

I don't have a lot of familiarity with Vuex, but I know how to create an event emitter and listener more generally and I assume that's the preferred route here. Any pointers would be welcome.

I'll implement the toggle on the Live and Shorts pages when this is working

absidue:
There are a few issues here but they are stem from the fact that you are overwriting the latestVideos array, which is why the setting is not working properly and also why you have to do the anti-pattern of fetching the data again from YouTube each time the switch is toggled.

So instead of overwriting the array and causing various problems, could instead create a computed which takes that array, checks the current value of the switch and if it is enabled returns a filtered copy of it, otherwise the original array. You can then reference that computed in the template instead of the latestVideos array. That way you don't have to do the unnecessary extra API requests and it will also automatically update when videos have been watched instead of only when you toggle the switch, it will also work properly with loading more data.

palharesf:
@absidue this is gold. I was not super familiar with the computed fields, but I implemented your suggestion and besides being simpler and cleaner, it works perfectly and is very fast.

I was also able to remove the code that forces a refresh of the videoPanel when the toggle is pressed, so everything looks much cleaner and responsive. Thanks a lot for this massive tip.

@efb4f5ff-1298-471a-8973-3d47447115dc I believe this also fixes the bug you encountered.

I'll implement the toggle for Live and Shorts before updating the PR, thank you all.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
These are some nice improvements.

I would remove the toggle when it changes sort order just like how it removes the dropdown. This gives it a bit of a cleaner look and feel

https://github.com/user-attachments/assets/27974a87-1e55-40f3-8bad-f7411bad8606

efb4f5ff-1298-471a-8973-3d47447115dc:
I tried to hide every video to see what would happen. The user must click on load more videos when all are currently loaded videos are marked as watched. Maybe its best to recognize when this is the case and auto load more videos. So if current window contains no videos because all are watched automatically load more. 


https://github.com/user-attachments/assets/5d217c62-bc1a-447e-90d8-e3fbbbcd6d8e

palharesf:
Really good tests, I appreciate it @efb4f5ff-1298-471a-8973-3d47447115dc . I have a decent idea on how to fix both, working on them now

palharesf:
Ok, I think my current changes fixes both of those issues. I recorded a screen cap that should show it:

electron_enqcPX6ttY

The one thing I noticed is that, when I mark all the videos on a page as 'Watched', I get the following error as a toast:

`TypeError: Cannot read properties of null (reading 'getContinuation')`

It looks like the error is being caught in the `getChannelVideosLocalMore()` function but I can't seem to follow what's happening. I'll push the current changes but would appreciate some guidance on what's happening in this case and how to prevent it.

absidue:
You are likely calling the get more functions when there are no more continuations/you got to the end. The get more functions were not designed to handle being called when there is no continuation left, as previously they were only called in response to a user clicking the button and the button was hidden when there was nothing left.

Personally I would have just left it up to the user to manually press load more, but then again I'm not the target audience for this, so it won't be spamming requests to YouTube on my computer...

src/renderer/views/Channel/Channel.vue
```diff
@@ -39,6 +39,14 @@
         :related-channels="relatedChannels"
       />
       <div class="select-container">
+        <FtToggleSwitch
+          v-if="(currentTab === 'videos') || (currentTab === 'shorts') || (currentTab === 'live')"
+          v-show="!isElementListLoading"
+          v-model="hideWatchedToggle"
```
absidue, line 45: FtToggleSwitch doesn't support v-model, you've not noticed it until now because you aren't actually tracking the value of the switch.
```suggestion
          :default-value="hideWatchedToggle"
```

src/renderer/views/Channel/Channel.vue
```diff
@@ -341,6 +350,12 @@ const isElementListLoading = ref(false)
 const isSearchTabLoading = ref(false)
 const currentTab = ref('videos')
 
+const hideWatchedToggle = ref(false)
+
+const updateHideWatchedToggle = () => {
+  hideWatchedToggle.value = !hideWatchedToggle.value
+}
```
absidue, line 357:
```suggestion
/**
 * @param {boolean} value
 */
function updateHideWatchedToggle(value) {
  hideWatchedToggle.value = value
}
```

src/renderer/views/Channel/Channel.vue
```diff
@@ -1048,6 +1063,34 @@ const videoContinuationData = shallowRef(null)
...
+const filteredVideos = computed(() => {
+  const filteredArray = hideWatchedToggle.value ? filterWatchedArray(latestVideos.value) : latestVideos.value
+  if (hideWatchedToggle.value && filteredArray.length === 0) {
+    if (process.env.SUPPORTS_LOCAL_API && apiUsed === 'local') {
+      getChannelVideosLocalMore()
+    } else {
+      channelInvidiousVideos()
+    }
+  }
+  return filteredArray
```
absidue, line 1075: Computeds should not have side-effects, please remove the fetch more stuff out if the computed.

From the Vue 3 docs but it also applies to Vue 2: https://vuejs.org/guide/essentials/computed.html#getters-should-be-side-effect-free

src/renderer/views/Channel/Channel.vue
```diff
@@ -2315,6 +2358,15 @@ function handleSubscription() {
...
 }
+
+function filterWatchedArray(videos) {
+  const historyCache = store.getters.getHistoryCacheById
+  return videos.filter(video => !historyEntryExists(video.videoId, historyCache))
+}
+
+function historyEntryExists(videoId, historyCache) {
+  return typeof historyCache[videoId] !== 'undefined'
+}
```
absidue, line 2369:
```suggestion
  return videos.filter(video => !Object.hasOwn(historyCache, video.videoId))
}
```

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
two things i just want to point out. 

1. If only one video is marked as unwatched you should hide the dropdown because there is no point in sorting one video.

https://github.com/user-attachments/assets/1a953d98-e99f-402c-aa94-fd844085e2aa

2. Here you see me marking an entire page as watched which prompted it to load more videos. It loaded allot of videos but there is also no page after this one, this is the last page it can load. If I would mark the entire page as Watched it shouldnt try to load more videos because there isnt anything to load (It could be that you already accounted for this but i wanted to point it out just in case)


https://github.com/user-attachments/assets/32d91473-8e42-4eff-b785-3343ed79086d

palharesf:
I updated the PR, and now fetching more videos won't happen automatically. I think this is a good compromise and avoids unnecessary issues that were being brought up by the automatic fetching of additional videos.

I'll take a look later at hiding the 'Sort By' selector once there's only one video available in the video panel. I think I know how to implement it. Here hoping that after this, we're ready to put a lid on it. I'm sure there are several additional edge cases we could account for, but I'm starting to worry we're over-complicating something that should be simpler. We can always improve in the future but what's already coded more than resolve the original issue I came to support with.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
Thanks @palharesf for creating this PR and making the necessary changes so this feature will have the same standards as other features within the application which in turn creates a coherent UX. I hope we'll see more of your contributions in the future!

src/renderer/views/Channel/Channel.vue
```diff
@@ -39,9 +39,17 @@
...
+          v-show="!isElementListLoading"
+          :default-value="hideWatchedToggle"
+          :label="$t('Channel.Hide Watched')"
+          compact
+          @change="updateHideWatchedToggle"
+        />
         <FtSelect
           v-if="showVideoSortBy"
-          v-show="currentTab === 'videos' && latestVideos.length > 0"
+          v-show="(currentTab === 'videos' && filteredVideos.length > 0 && !hideWatchedToggle) || (currentTab === 'videos' && filteredVideos.length > 1 && hideWatchedToggle)"
```
PikachuEXE, line 52: A bit too long
```suggestion
          v-show="currentTab === 'videos' && (hideWatchedToggle ? filteredVideos.length > 1 : filteredVideos.length > 0)"
```

src/renderer/views/Channel/Channel.vue
```diff
@@ -51,7 +59,7 @@
         />
         <FtSelect
           v-if="!hideChannelShorts && showShortSortBy"
-          v-show="currentTab === 'shorts' && latestShorts.length > 0"
+          v-show="(currentTab === 'shorts' && filteredShorts.length > 0 && !hideWatchedToggle) || (currentTab === 'shorts' && filteredShorts.length > 1 && hideWatchedToggle)"
```
PikachuEXE, line 62: similar to above

src/renderer/views/Channel/Channel.vue
```diff
@@ -61,7 +69,7 @@
         />
         <FtSelect
           v-if="!hideLiveStreams && showLiveSortBy"
-          v-show="currentTab === 'live' && latestLive.length > 0"
+          v-show="(currentTab === 'live' && filteredLive.length > 0 && !hideWatchedToggle) || (currentTab === 'live' && filteredLive.length > 1 && hideWatchedToggle)"
```
PikachuEXE, line 72: similar to above

src/renderer/views/Channel/Channel.vue
```diff
@@ -341,6 +350,15 @@ const isElementListLoading = ref(false)
 const isSearchTabLoading = ref(false)
 const currentTab = ref('videos')
 
+const hideWatchedToggle = ref(false)
```
PikachuEXE, line 353: It's not storing an element but the value
```suggestion
const hideWatchedEntries = ref(false)
```

PikachuEXE (changes requested):

src/renderer/views/Channel/Channel.vue
palharesf: Done - commit https://github.com/FreeTubeApp/FreeTube/pull/7366/commits/6ad9f5b1c65e4d47d952c9ff4982c9224bd30af4

src/renderer/views/Channel/Channel.vue
palharesf: Done - commit https://github.com/FreeTubeApp/FreeTube/pull/7366/commits/76aaf13b973894da42dac7354bd5eaf429f03469

src/renderer/views/Channel/Channel.vue
palharesf: Done - commit https://github.com/FreeTubeApp/FreeTube/pull/7366/commits/59e6b1a8205ae0179873b206350f665e6f0292ae

src/renderer/views/Channel/Channel.vue
```diff
@@ -341,6 +350,15 @@ const isElementListLoading = ref(false)
 const isSearchTabLoading = ref(false)
 const currentTab = ref('videos')
 
+const hideWatchedEntries = ref(false)
+
+/**
+ * @param {boolean} value
+ */
+function updatehideWatchedEntries(value) {
```
PikachuEXE, line 358: Find and replace mistake made with case insensitive mode I can see :)
```suggestion
function updateHideWatchedEntries(value) {
```

palharesf: Good eye, I even replaced entries one by one to try to avoid stuff like that and it still escaped me. Just fixed it, thank you very much

PikachuEXE (approved):

palharesf:
> Thanks @palharesf for creating this PR and making the necessary changes so this feature will have the same standards as other features within the application which in turn creates a coherent UX. I hope we'll see more of your contributions in the future!

My pleasure! I'd be happy to keep an eye out for features involving the Channel page, as this is the page I've got somewhat familiar with. If there are any issues any of you think I could get involved, feel free to @ me :)

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @palharesf 

After reviewing the PR, I think we made the wrong call in approving the direction for the "Hide Watched Videos" toggle on the Channel page. I’ve had some time to think it over, and I feel it would be better to avoid adding new settings outside of the Settings page.

Settings like this should really be centralized in the Settings page to prevent any confusion or the potential of other settings being added to random pages. This is something we've been consistent about, so I believe it makes more sense to maintain that approach.

What we should do instead is reuse the existing setting that’s listed in the Subscription settings and move it into the General section on the Distraction Free page. This will help keep the interface clean and consistent with our usual approach.

Let me know your thoughts on this and if you need help with the change.

palharesf:
> Hi @palharesf
> 
> After reviewing the PR, I think we made the wrong call in approving the direction for the "Hide Watched Videos" toggle on the Channel page. I’ve had some time to think it over, and I feel it would be better to avoid adding new settings outside of the Settings page.
> 
> Settings like this should really be centralized in the Settings page to prevent any confusion or the potential of other settings being added to random pages. This is something we've been consistent about, so I believe it makes more sense to maintain that approach.
> 
> What we should do instead is reuse the existing setting that’s listed in the Subscription settings and move it into the General section on the Distraction Free page. This will help keep the interface clean and consistent with our usual approach.
> 
> Let me know your thoughts on this and if you need help with the change.

Hi @efb4f5ff-1298-471a-8973-3d47447115dc,
Conceptually, I'm aligned with the direction of centralizing all settings on the Settings page instead of having custom settings applied into / managed by different pages.

I just want to re-hash what I see as the proposed implementation in your vision then (assuming we're starting from my edited branch):

1. Remove my toggle from the Channels Page
2. Add a toggle in the Settings page (under "Distraction Free", "General", titled 'Hide Watched Videos')
3. Update localization settings (to reflect the changes above)
4. Link all other functions and logic from the current Channels page toggle to the new Settings toggle

In summary, is that it?
I think the hardest part for me will be familiarizing myself with Vuex, but it's a good time as any to become familiar with it. I just want to make sure the direction is good before moving ahead.

Also, should I open a new PR or simply implement the changes over this PR / branch? I can also change the main post to reflect the updated approach. Let me know

efb4f5ff-1298-471a-8973-3d47447115dc:
>     1. Remove my toggle from the Channels Page
> 
>     2. Add a toggle in the Settings page (under "Distraction Free", "General", titled 'Hide Watched Videos')
> 
>     3. Update localization settings (to reflect the changes above)
> 
>     4. Link all other functions and logic from the current Channels page toggle to the new Settings toggle
> 
> 
> In summary, is that it? I think the hardest part for me will be familiarizing myself with Vuex, but it's a good time as any to become familiar with it. I just want to make sure the direction is good before moving ahead.

1. Correct
2. Correct but we can use the title of the already existing toggle that resides withing the Subscription Settings `Hide Videos on Watch`
3. See 2, so that would mean you can remove your localization and use the existing one. The placement does need to be changed so from https://github.com/FreeTubeApp/FreeTube/blob/development/static/locales/en-US.yaml#L531 to reside within the distraction free settings keys https://github.com/FreeTubeApp/FreeTube/blob/development/static/locales/en-US.yaml#L537
4. Correct, in addition to that remove the Subscription Settings `Hide Videos on Watch` toggle and add that functionality to the toggle you will add within the Distraction Free Settings. So that toggle will Hide Videos from the subscription page and the channel page.
5. Now that it will reside within the General section of the Distraction Free Settings I think it would also be good to add a tooltip with the description that this only applies to those two pages.


> 
> Also, should I open a new PR or simply implement the changes over this PR / branch? I can also change the main post to reflect the updated approach. Let me know

Feel free to take the approach you feel comfortable with

absidue:
Just to clarify you do not need to add a new setting, you would just need to move the existing distraction free setting around in the same file and then change your channel page logic to use that existing setting instead of its own switch.

palharesf:
Ok, good clarifications. Thank you, I'll get started on it now

palharesf:
OK a few questions to ensure I'm on the right track:

1. It looks like settings.js is where all components get assembled together, but Distraction Free Settings and Subscription Settings are actually two different components. So I would move the 'Hide Videos on Watch' toggle from subscription-settings.vue to distraction-settings.vue according to what we previously discussed right?
2. I assume I don't need to make any changes to the store if I just move the toggle from one settings component to the other, correct?
3. From there I can just access the settings value from the store within the Channels page and update all the logic to reference it right?

I'll look into the tooltip as soon as I got the basic logic working properly. Thank you all!

palharesf:
I implemented the base functionality (still missing the tooltip, working on it now) -- but it's not evident to me where are the conflicts that the bot is warning me of. I'd appreciate some guidance on that while I get the tooltip up and running

static/locales/en-US.yaml
```diff
@@ -1086,6 +1086,7 @@ Tooltips:
        The channel ID entered must be a complete match and is case sensitive.
     Hide Subscriptions Live: 'This setting is overridden by the app-wide "{appWideSetting}" setting, in the "{subsection}" section of the "{settingsSection}"'
     Hide Videos and Playlists Containing Text: Enter a word, word fragment, or phrase (case insensitive) to hide all videos & playlists whose original titles contain it throughout all of FreeTube, excluding only History, Your Playlists, and videos inside of playlists.
+    Hide Videos on Watch: 'This option hides watched videos, lives and shorts from the Subscriptions and Channels Pages. It does not affect the Home tab on the Channels page"'
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 1089:
```suggestion
    Hide Videos on Watch: This option hides Watched Videos, Live streams, and Shorts from the Subscriptions and Channels pages. It does not affect the Home tab on a Channel’s page.
```

palharesf:
Ok I think we're in a good spot. Tooltip implemented (and wording reviewed), functionality seems to be working for all the cases we expected (in the subscriptions page it's unchanged, for the Channels page it works for videos, lives and shorts), and the functionality to hide the Sorting button on the Channels page when there's only one video being displayed is still working. I still didn't understand the conflicts generated by GitHub Actions, but happy to resolve them with some guidance.

Please let me know if there's any additional work anyone notices that I missed. Otherwise, thank you all for the help and I'll start looking at related issues to tackle once we close this :)

efb4f5ff-1298-471a-8973-3d47447115dc:
We're all a bit busy at the moment. We will get back to you as soon as possible. I cant put my finger on why you are receiving the conflicts. What you can try is updating your branch so your changes are being made on top of the newest version of the dev branch

PikachuEXE:
The conflict is caused by migration to composition API https://github.com/FreeTubeApp/FreeTube/pull/7335
Some files edited in this PR (before merge) no longer exist
I did the merge for you since it's rare

PikachuEXE (commented):
The screenshots might need to be updated for later reviewers...

palharesf:
Thank you for addressing the conflicts @PikachuEXE. 

> The screenshots might need to be updated for later reviewers... `reload video list once a video is mark as 'Watched' manually` is done already too right?

Since we started using the computed property `filteredVideos` in the videoPanel, we don't need to reload the panel every time a video is watched. This is done automatically. I forgot to update the main PR post, I'll go back to it tomorrow and update the screenshots and tests. Thank you

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc, line 1089: I reread this again and thought it wasnt very clear

```suggestion
    Hide Videos on Watch: 'Hides watched videos from the Videos, Shorts, and Live tabs on the Subscription and Channel pages. This does not affect the Home tab on Channel pages'
```

palharesf:
Ok, I updated the main post with the current implementation, current screenshots and explanations

While doing some testing I noticed the logic for hiding the 'Sort By' dropdown when there was only one live left in the `videoPanel` was referencing to the `filteredShorts` variable instead of the `filteredLive` variable, so it was not working. I fixed that, but when I pushed my changes I think I undid the fix for conflicts that PikachuEXE had done -- apologies for that

palharesf:
Hi @absidue,

Sounds good, and sorry again. I had force-pushed yesterday without apparent negative consequences, and wrongly assumed I could do the same again without realizing it would undo the fixes, but I recognize my mistake. Are we able to revert my last push, so we return to the previous state (without conflicts)? There is only one line changed between that state and the current and might be an easier fix than addressing the conflicts again.

Since we're already at it, was there anything I could have done besides what I outlined here (https://github.com/FreeTubeApp/FreeTube/pull/7366#issuecomment-2909670290) to fix my branch?

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc: This one needs to be commited again

palharesf:
One change that is still required (but that I've been unable to push) is on 'Channels.vue', line 64. As far as I can see it still has `hideWatchedSubs ? filteredShorts.length > 1 : filteredShorts.length > 0`, where it should be `hideWatchedSubs ? filteredLive.length > 1 : filteredLive.length > 0`, but I am unable to push those changes from my side.

This was the fix that I had force-pushed previously and that broke the fix for the conflicts

src/renderer/views/Channel/Channel.vue
```diff
@@ -41,7 +41,7 @@
       <div class="select-container">
         <FtSelect
           v-if="showVideoSortBy"
-          v-show="currentTab === 'videos' && latestVideos.length > 0"
+          v-show="currentTab === 'videos' && (hideWatchedSubs ? filteredVideos.length > 1 : filteredVideos.length > 0)"
```
PikachuEXE, line 44: No idea why we need different logic based on `Hide Videos on Watch` enabled or not
Probably just need `filteredVideos.length > 1` here

Or do you mean the opposite?
When watched entries are hidden, allow changing the sorting due to having 1 entry does not mean 1 entry only in other orders (e.g. all except 1 watched, but all oldest still not watched)
However if this is the case the criteria should be on NOT filtered entries = `latestVideos.length > 1` (i.e. more than 1 entries from remote = possibility to have unwatched entries in other sorting orders

I might get some of this wrong let me know

src/renderer/views/Channel/Channel.vue
```diff
@@ -51,7 +51,7 @@
         />
         <FtSelect
           v-if="!hideChannelShorts && showShortSortBy"
-          v-show="currentTab === 'shorts' && latestShorts.length > 0"
+          v-show="currentTab === 'shorts' && (hideWatchedSubs ? filteredShorts.length > 1 : filteredShorts.length > 0)"
```
PikachuEXE, line 54: ditto

src/renderer/views/Channel/Channel.vue
```diff
@@ -61,7 +61,7 @@
         />
         <FtSelect
           v-if="!hideLiveStreams && showLiveSortBy"
-          v-show="currentTab === 'live' && latestLive.length > 0"
+          v-show="currentTab === 'live' && (hideWatchedSubs ? filteredLive.length > 1 : filteredLive.length > 0)"
```
PikachuEXE, line 64: ditto

src/renderer/views/Channel/Channel.vue
absidue: The original code checks YouTube's response for the dropdown, which only hides it when there is only 1 or 0 videos and there are no continuations. You need to ensure that you do not break this behaviour when hide watched is disabled.

So the important thing for this pull requests is that you cannot hide it just based on filteredVideos.length < 2, because if there is a continuation available, you don't actually know that all videos are hidden, because there may be some visible ones that were not fetched yet.

palharesf: My proposal was to only show the dropdown if we are in one of the tabs it would appear (`currentTab === 'videos'` and the same for shorts and lives) and if the array containing the videos to be displayed, after computing for hiding watched videos, has more than one video (`filteredVideos.length >1)

So basically just switching 
` v-show="currentTab === 'videos' && (hideWatchedSubs ? filteredVideos.length > 1 : filteredVideos.length > 0)"`

for
` v-show="currentTab === 'videos' && filteredVideos.length > 1"

It makes sense to me unless I'm missing something, and I just tested and it works. Let me know if you see any issues with that approach

palharesf: I also realized why I was not able to push my previous changes cleanly, my forked repo was behind the actual origin repo so that's why simply pulling the changes was not working. I have just synced my repo with the origin one, and when I try to pull the changes into my repo I finally get the conflicts. 

I assume that I can accept the remote repository changes by using `git checkout --theirs src/renderer/components/subscription-settings/subscription-settings.js` and `git checkout --theirs src/renderer/components/subscription-settings/subscription-settings.vue`, but I'm afraid of doing that without checking with you all and breaking things. If its easier to just suggest the changes I outlined in my previous comment directly into the code, I'm also happy with that approach

efb4f5ff-1298-471a-8973-3d47447115dc: Hi @palharesf just checking in. What is the status on this? Do you need any additional help?

palharesf: Hi, I still haven't been able to properly merge the changes from the remote branch onto my local copy of the repo, as outlined [here](https://github.com/FreeTubeApp/FreeTube/pull/7366#discussion_r2113912398).

It's unclear what I can do without breaking things, and as I have already broken stuff here in the past I decided to wait for guidance.

PikachuEXE: Just rename your existing branch (delete it later)
Check out this pull request (you can use `gh pr checkout 7366` if you got `gh` installed) and work from there
If you got something in the old branch committed but not pushed yet you can see try to apply those on new branch once you have done the above

palharesf: Ohhh interesting. Thank you for the guidance, I was not sure if I could bring a new branch 'up-to-speed' with this PR. I'll try it later today. I don't think I have `gh` installed but I'll check it out. Thank you very much

Ein-Tim: I don't know your needs, but if you prefer a GUI over the command line, I recommend using GitHub desktop: https://github.com/apps/desktop

palharesf: > The original code checks YouTube's response for the dropdown, which only hides it when there is only 1 or 0 videos and there are no continuations. You need to ensure that the original behaviour still exists when the hide watched setting is disabled.
> 
> So the important thing for this pull requests is that you cannot hide it just based on filteredVideos.length < 2, because if there is a continuation available, you don't actually know that all videos are hidden, because there may be some visible ones that were not fetched yet.

@absidue I am reviewing the code now, and since `filteredVideos` is computed from `latestVideos`, I would assume it preserves its properties i.e. if `latestVideos` had no continuations, `filteredVideos` will have no continuations. Is that reasonable?

As a way of testing that, I logged the length of both `filteredVideos` and `latestVideos` for a channel where I have hidden two videos from the first page. You can see that the offset is always two, i.e. even initially latestVideos does not hold data from the whole playlist, just from the first page. While I can try to incorporate data from the full playlist (including continuations) in the current check, it doesn't look like this is how it works currently with `latestVideos` unless I'm missing something here.

image

TL;DR - if the current implementation was taking continuations into account, I believe `latestVideos` should be initialized with 117 instead of 30. Let me know if I'm not understanding something.

palharesf: Assuming we agree on our point regarding continuations (i.e. my changes do not modify the base behavior), I wanted to push the last remaining commit but I'm getting an eslint error.

Changes I'm trying to commit:

image

Error message I'm getting (even after creating a new branch and checking out this PR):

image

I believe these changes would resolve the outstanding comments on this PR while addressing the concerns raised on the discussion here, but I'm unsure on how to push them with this new error. If it's easier, someone could suggest those changes and I could simple approve them on this PR, otherwise help on how to solve this issue locally would be welcome. Thanks

efb4f5ff-1298-471a-8973-3d47447115dc: You can try do request changes yourself on the lines you want to change. After the change is submitted a button should appear to commit the change.

https://github.com/user-attachments/assets/6124c64b-b598-4308-8c0e-88474ac938f2

PikachuEXE:
Solved merge conflict again and made changes for https://github.com/FreeTubeApp/FreeTube/pull/7366#pullrequestreview-2876788257 since it's not worth waiting weeks just a for a few changes to review this fully
I will review this tomorrow

PikachuEXE:
Tested a bit and realized that we should show the sort select when there are > 1 fetched entries (instead of shown=unwatched entries)
Otherwise the ability to see 1st page of other sorting orders is removed when most/all entries with default sort are watched and hidden

efb4f5ff-1298-471a-8973-3d47447115dc:
@palharesf No continuation and dropdown still showing if only one video in the tab present

VirtualBoxVM_kjTt0Wrevr

palharesf:
@efb4f5ff-1298-471a-8973-3d47447115dc I still don't understand the issue with continuations. As per my comment here (https://github.com/FreeTubeApp/FreeTube/pull/7366#discussion_r2132754590), it seems to me the original array didn't check for continuation for deciding when to show the "Fetch More" button, so I kept the same logic for the computed array. If we want to change that logic I think it's fine, but it seems to me it's a different issue from hiding watched videos altogether. If I'm wrong, I'd appreciate understanding how, since from my current point of view that doesn't seem like an issue. I'm less stuck on code and more stuck on comprehension here.

Regarding the dropdown, it's still visible when there's only one video being displayed on the channel because of the last change made by PikachuEXE (https://github.com/FreeTubeApp/FreeTube/pull/7366/commits/b232adf1ceec138c1c69eb44d01eceacc2a98c11). I tested this just now and this is happening because the array responsible for displaying videos on the screen is filteredVideos, whereas the array responsible for displaying the Sort element is latestVideos. They don't match when we have the 'Hide Videos on Watch' setting enabled, so naturally you'll have situations like this one, where the size of the arrays differ and hence the logic doesn't work as intended.

image

My proposed solution here is to revert and let filteredVideos command whether the Sort element would be displayed or not, but naturally I'd like @PikachuEXE two cents here since he's the one that changed this bit of logic and can show me the shortcomings of my suggestion.



Apologies if it seemed I was MIA, but I honestly did not know you all were waiting on me for the pending changes and I already forgot most of what I worked on in this PR since I dove into it a while ago.

PikachuEXE:
Updated display logic for sort element again
Now it's hidden when there is no next page and no. visible elements <= 1
i.e. if there is a next page or visible element > 1 = shown

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@palharesf thank you so much for you contribution and bearing with us. Normally merging a PR doesn't take this long. Let me know if you want to work on another issue and i can help you find one that fits your criteria

palharesf:
@efb4f5ff-1298-471a-8973-3d47447115dc thank you all for the patience and bearing with someone doing their first ever contribution to an open source project, hopefully the first of many. I will accept your offer, so feel free to @ me in any other non-critical or time sensitive issues you feel I could contribute. Happy to continue contributing with the project!

PR #7370, opened 2025-05-07
Add Support For Disabling Hardware Media Keys

### LINKED ISSUE #1465: Option to disable media keys support
Ubuntu often fails to get media keys to control the right application when there is more than one capable of playing audio open at the same time. Since I always mean for the media keys to control Spotify, I turn off media key support to other applications, like Firefox. I can't do this to Freetube.

I would like to be able to turn off media key support in Freetube, similarly to how it is possible to do this in Firefox.

**Additional context**
I only mean for the media keys to control Spotify, and having Freetube and Spotify open messes that behavior up. For example, I often would like to pause the music I am listening to so that I can watch a video on Freetube... Instead video on Fretube plays and I have to go to Spotify, pause manually then go back to Freetube. Having an option to turn off this feature in Freetube would solve my problem.

BentBoxCutter:
# Add Support For Disabling Hardware Media Keys

## Pull Request Type

- [ ] Bugfix
- [ x ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #1465 

## Description
Currently be default FreeTube has hardware media keys enabled to control playback (play/pause/next/previous button(s) on keyboards).  This is a behavior which comes from Electron.  This can be disabled by manually adding `-disable-features=HardwareMediaKeyHandling` command line argument however this is annoying to do and not all users are tech savvy.  Some users (myself included) prefer to use the hardware media keys to control music playback and would like the option to disable this feature.

This PR fixes the issue by adding a new toggle to the Experimental Settings section which appends the command line argument.

## Screenshots 
image

## Testing 

1. Launch FreeTube with no command line arguments & open any video.  Press the Play/Pause hardware media key (if your keyboard doesn't have one you will need to remap a key or just trust me).
    a.  You should see the video playback play/pause when you this the media keys 
2. Click on Settings & scroll down to Experimental Settings
3. Click on the 'Disable Hardware Media Keys' toggle and hit 'Yes Restart'
4. Navigate to any video and press the Play/Pause hardware media key
    a.  The playback should no play/pause

## Desktop

- **OS:** Windows 10 Pro
- **OS Version:** 10.0.19045 Build 19045
- **FreeTube version:** v0.23.4 Beta

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -39,6 +39,9 @@ const IpcChannels = {
   GET_REPLACE_HTTP_CACHE: 'get-replace-http-cache',
   TOGGLE_REPLACE_HTTP_CACHE: 'toggle-replace-http-cache',
 
+  GET_DISABLE_HARDWARE_MEDIA_KEYS: 'get-disable-hardware-media-keys',
+  TOGGLE_DISABLE_HARDWARE_MEDIA_KEYS: 'toggle-disable-hardware-media-keys',
+
   PLAYER_CACHE_GET: 'player-cache-get',
   PLAYER_CACHE_SET: 'player-cache-set',
 
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -259,6 +259,13 @@ function runApp() {
     app.commandLine.appendSwitch('disable-http-cache')
   }
 
+  const DISABLE_HARDWARE_MEDIA_KEYS_PATH = `${userDataPath}/experiment-disable-hardware-media-keys`
+  const disableHardwareMediaKeys = existsSync(DISABLE_HARDWARE_MEDIA_KEYS_PATH)
+
+  if (disableHardwareMediaKeys) {
+    app.commandLine.appendSwitch('disable-features', 'HardwareMediaKeyHandling')
+  }
+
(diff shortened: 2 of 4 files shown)


absidue:
As far as I can tell this isn't actually experimental, you just placed it there so you could abuse the workaround used for the replace HTTP cache setting, which only has that workaround because the plan is to eventually remove the setting and enable it by default so it is okay that it is not in the normal settings.

As this is not an experimental setting and you presumably want this to stay instead of being removed at some point in the future it has to go into the normal settings, not some temporary workaround like this. As that command line flag presumably has to be added before the ready event fires, that means you'll need to wait until we migrate the main process JavaScript to ESM, which also requires forking and migrating various dependencies to ESM first, so that we can use top-level await to wait for the settings to be loaded while blocking the rest of the app launch.

As that is not going to happen any time soon, I would recommend sticking to adding the command line flag manually, you can edit the shortcuts to FreeTube in the start menu and on the task bar with the command line flag, so you don't have to manually add it each time. Alternatively you can maintain a custom build for yourself that hardcodes that flag into the code, just note that we won't provide support for any issues that you experience with your custom build unless you can also reproduce them with the official, unmodified ones.

PR #7234, opened 2025-04-19
Create CODE_OF_CONDUCT.md

efb4f5ff-1298-471a-8973-3d47447115dc:
# Create CODE_OF_CONDUCT.md

## Pull Request Type

- [x] Documentation

## Description

CoC has been created after an internal discussion I had a while ago with some team members. The CoC is more for the community than for us because we get harassed very often

PR #7376, opened 2025-05-08
Add Code of Conduct

Ein-Tim:
## Pull Request Type
- [ ] Bugfix
- [ ] Feature Implementation
- [x] Documentation
- [ ] Other

## Related PR
This pull request is a follow up to #7234, which was closed due to a conflict of interest.

## Description
This PR takes the work done in #7234 and provides it again here. As I am providing this PR, there should be no conflict of interest. 
I would have loved to simply cherry pick the commits out of #7234 and provide them here, but since I‘m travelling at the moment and only have access to an iPad this is not possible. Should you be happy with this Code of Conduct but not want the commit be from me but rather have the original commits cherry picked, I can provide a PR with cherry picks in about two weeks, when I‘m back home. 

## Credits
All work that I am providing here was done by @efb4f5ff-1298-471a-8973-3d47447115dc, @ChunkyProgrammer, @kommunarr & @PikachuEXE.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,45 @@
+# FreeTube Code of Conduct
+
+Like the technical community as a whole, the FreeTube team and community is made up of a mixture of professionals and volunteers from all over the world, working on every aspect of the mission - including mentorship, teaching, and connecting people.
+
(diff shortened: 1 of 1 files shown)


Ein-Tim:
I'm back at my machine now and could also provide this PR with the commits cherry picked from the old PR. I will not do that unless there is feedback on this PR whether you consider merging it.

Issue #6000, opened 2024-10-28
[Bug]: no icon in taskbar


TotalCaesar659:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Open FreeTube
2. See there's no icon in taskbar and the upper-left window corner (screenshot of comparison with Firefox is attached)
![0](https://github.com/user-attachments/assets/eafb2361-d2d7-466a-af81-029fb7cd2600)

### Expected Behavior

FreeTube has to have the icon in taskbar and the upper-left window corner

### Issue Labels

visual bug

### FreeTube Version

v0.22.0

### Operating System Version

Lubuntu 24.04.1

### Installation Method

.deb

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

TotalCaesar659:
The bug is not fixed yet, so please remove "Stale" label.

efb4f5ff-1298-471a-8973-3d47447115dc:
@mariakeating any idea if something in this repo also needs to be changed. I saw that you opened up a PR for it in our flathub repo https://github.com/flathub/io.freetubeapp.FreeTube/pull/119

mariakeating:
Changes aren't needed here for the flathub version, at least when launching with wayland flags. As an aside, if they did use the flathub version then it's possibly related since I noticed the same bug when launching in X11 mode even with my PR but I didn't look into it.

TotalCaesar659:
> @TotalCaesar659 just as a sanity check did you install FreeTube through a .deb file or through flathub

I installed 0.22.0 DEB package and launched Appimage after that, they both have this bug. I didn't test Flathub so can't say about it.

[...]

PR #7392, opened 2025-05-09
Fix n/no-path-concat issues

absidue:
## Pull Request Type

- [x] Bugfix
- [x] Other

## Description

Currently we turn off the `n/no-path-concat` linter errors, this pull request fixes them instead. It also fixes the invalid path that we were using for the icon in the production builds, this was likely not noticed before because at least on Windows Electron just uses the icon from the exe as a fallback. While I'm not 100% sure, this may solve the issue were some Linux users were saying that the windows had no icons in the title bar.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** a560499ce11701697aa11e33b4bb1b9f8fb253f1


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
> While I'm not 100% sure, this may solve the issue where some Linux users were saying that the windows had no icons in the title bar.

Confirmed that it fixes https://github.com/FreeTubeApp/FreeTube/issues/6000

PikachuEXE (approved):

kommunarr (approved):

PR #7421, opened 2025-05-16
Downgrade electron 36 > 35, Fix video DASH playback

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Caused by upgrade in https://github.com/FreeTubeApp/FreeTube/pull/7403

## Description

See errors in screenshot (and other webm related errors in https://github.com/shaka-project/shaka-player/blob/v4.14.11/lib/dash/webm_segment_index_parser.js)

I've tried using shaka-player 4.14.9-12 in electron 36.2.1, not working

## Screenshots 

Screenshot 2025-05-16 at 17 41 57

## Testing

- Play https://youtu.be/9MjeHlWipWo?t=1598
- Ensure no error, no fallback to legacy format

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


efb4f5ff-1298-471a-8973-3d47447115dc:
I dont think we should downgrade intermediately. Its best to check what the root cause is and fix it otherwise we will be stuck on v35.x.x forever

PikachuEXE:
Reported issue in https://github.com/shaka-project/shaka-player/issues/8616

PikachuEXE:
As I said in [linked issue](https://github.com/shaka-project/shaka-player/issues/8616#issuecomment-2900576757) the issue seems fixed in Chrome 137
According to [Electron release timeline](https://www.electronjs.org/docs/latest/tutorial/electron-timelines) Chrome 137/138 = 37.x would only be released in 2025-Jun-24 (~1 month from now
I will keep this open until then (and any major release should be made after that

PR #7424, opened 2025-05-17
Add preload script and disable node integration in renderer processes

absidue:
## Pull Request Type

- [x] Security improvement

## Description

This pull request introduces a preload script, disables node integration and enables context isolation and sandboxing in the renderer processes/the user facing windows. A preload script is a small script that runs in a small Node.js like environment before any other code in the window runs, allowing us to expose some priviledged functions on the window object, through the contextBridge, to grant the page extra powers that aren't possible in a normal web browser, without having to give it a full blown Node.js environment. In our case we are mostly using it as a wrapper around our IPC calls.

I changed a few other things:
- `REQUEST_FULLSCREEN` and `REQUEST_PIP` are no longer IPC calls, we instead handle them directly in the preload script (`requestFullscreen` and `requestPip` respectively).
- The NATIVE_THEME_UPDATED main to renderer IPC call is now handled directly in the preload script which uses `evaluateJavaScript` to write the property to the body element.

I have also added typings/auto-complete for the `window.ftElectron` object through the `src/preload/preload-interface.d.ts` file, which is picked up automatically by the TypeScript language server in VS Code, I don't know how well other editors fare with it though.

## Testing

Please test this quite thoroughly, as this pull request contains a lot of small changes to many places in FreeTube so I may have missed something or have a typo.

Here's a build of it, so you can also test it without dev mode: https://github.com/absidue/FreeTube/actions/runs/15086330548

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** 617396972bc88858bcac79b8e03a6e2472fc8f48


PikachuEXE (commented):
Did not test system theme and external player and power save blocker (just lazy for last one
Otherwise tested with custom build

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Tested:
- [x] power save blocker
- [x] system theme

Untested:
- [ ] external player

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #7505, opened 2025-06-03
Make sure description isn't null when extracting chapters from description

ChunkyProgrammer:
## Pull Request Type
- [x] Bugfix

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/issues/7503

## Description
result.basic_info.short_description and result.secondary_info.description.text are both nullable strings and for whatever reason both returned null from Rune's testing

## Testing

- Go to https://youtu.be/A1IlubXQpL8
- Don't receive error, TypeError: Cannot read properties of undefined (reading 'matchAll') (error is inconsistent so we must try multiple times)

## Desktop
- **OS:** Fedora Linux
- **OS Version:** 42
- **FreeTube version:** latest git commit


absidue (changes requested):
Before we do a fix to just make the error go away, we should check that the description is actually completely missing in those responses or whether it has been moved to somewhere else in the response structure, in which case we should extract it from the new location. My suspicion is that it is the latter (yt-dlp has had reports recently about the translated title not being extracted, which sounds related) and an A/B test which would explain why it can't be reproduced consistently.

Unfortunately @efb4f5ff-1298-471a-8973-3d47447115dc wasn't able to reproduce it with the devtools attached, so we currently only have the error message (in electron/chromium logs are captured all the time, requests and responses are only captured while devtools are open).

I'm not blocking this pull request entirely, I just want to be sure that this is the right fix. If we don't have an answer right before we do the next release, I am happy to approve it as-is so that users don't see the error.

efb4f5ff-1298-471a-8973-3d47447115dc:
I've caught the error again and shared the responses

ghost:
Looks like it has a placeholder (descriptionPlaceholder) saying that the description is empty now:

```json
              {
                "expandableVideoDescriptionBodyRenderer": {
                  "descriptionPlaceholder": {
                    "content": "No description has been added to this video.",
                    "styleRuns": [
                      {
                        "startIndex": 0,
                        "fontColor": 4284506208,
                        "italic": true
                      }
                    ]
                  }
                }
              },
```

absidue:
As it looks like the description is actually missing and it seems like YouTube is just switching how they handle videos without a description, I'll remove my block.

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
Got this message

VirtualBoxVM_yVQPBCMe6F

absidue (approved):

Issue #746, opened 2020-10-30
Store data and settings in portable directory with portable version


zahroc:
It seems that the portable version store its settings and data within the AppData folder structure in Windows.
For a portable application, the data and settings should reside within the application's directory.

rddim:
@zahroc 

You can use [FreeTubePortable](https://github.com/rddim/FreeTubePortable/releases) as a workaround. Its deal also with the registry strings.

zahroc:
Correct, that is why I think the portable version should NOT use the registry settings. It should keep the data within the portable directory.

snakyjake1:
I'm using the portable version.  It is saving to:
%AppData%\Roaming\FreeTube

One of the bigger problems I have is I run the portable version from a network share.  The portable version isn't using the network share.

rddim:
@snakyjake1 

I make some tests and:

1) If you run it from `\\server\share` the GUI will not appear on the screen but the `%AppData\Roaming\FreeTube` will be created, also some `FreeTube.exe` processes are started (check Task Manager).

2) But if you map your network share as drive with letter (for example `Z:\`) FreeTube will work, of course will store the settings in the `%AppData%\Roaming\FreeTube` on your device. In my case it starts slowly from the mapped network share.

In the next release I will change the `SupportsUNC=yes` to `SupportsUNC=no` for the portable launcher that I maintain. This will not running the FreeTubePortable from `\\server\share` and will inform the user why this happen.

zahroc:
Is it possible to add a config variable to place the configuration in a user specified location and not %AppData%\..... not matter where it is run from? This was my original post subject.

GilgusMaximus:
At the moment, the data path is hardcoded as the path Electron uses by default

rddim:
I found this code https://github.com/getferdi/ferdi/blob/develop/src/index.js#L13:L23

And modified the `index.js` with the code below and it's working.
```js
...
import Datastore from 'nedb'

const path = require('path')

// Set app directory before loading user modules
if (process.env.FREETUBE_APPDATA_DIR != null) {
  app.setPath('appData', process.env.FREETUBE_APPDATA_DIR)
  app.setPath('userData', path.join(app.getPath('appData')))
} else if (process.env.PORTABLE_EXECUTABLE_DIR != null) {
  app.setPath('appData', process.env.PORTABLE_EXECUTABLE_DIR, `${app.name}AppData`)
[...]

PR #7513, opened 2025-06-04
Add new setting for storing user data inside app folder for windows

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/746

## Description

It's common for a portable app to store user data (settings and other data files) 

https://portableapps.com/about/what_is_a_portable_app#definition

> A portable app is a computer program that doesn't need to be installed into Windows like traditional apps so you can carry it with you on a portable device or synced cloud folder and use on any Windows computer. When your USB flash drive, portable hard drive, or other portable device is plugged in or your cloud drive is synced, you have access to your software and personal data just as you would on your own PC. And when you unplug the device, none of your personal data is left behind.

## Screenshots 

image

## Testing
** Backup data before testing **
Suggest to test with non dev version to be accurate (I have never tested this with dev on windows)
Custom build: https://github.com/PikachuEXE/FreeTube/actions/runs/15431198924

A. Windows
- Open app
- Ensure all data like subscribed channels/profiles, sub cache, playlist, settings, etc. are still present
- Ensure new setting DOES appear and disabled
- Turn OFF `Replace HTTP Cache` (for later special setting sync test)
- Close app
- Create folder `userData` under the app folder (if you want to test auto folder creation you can do that later)
  e.g. if executable is at `C:\FreeTube\Freetube.exe`, create `C:\FreeTube\userData`
- Create empty file `experiment-replace-http-cache` inside that folder (which should be absent when `Replace HTTP Cache` disabled)
- Open app, turn ON new setting
- Ensure all data like subscribed channels/profiles, sub cache, playlist, settings, etc. are still present
- Ensure  `Replace HTTP Cache` still OFF (and empty file `experiment-replace-http-cache` is gone)
- Ensure data files (`*.db`) now present in new `userData` folder in app folder
- Turn ON `Replace HTTP Cache`
- Ensure no empty file `experiment-replace-http-cache` present in default user data folder
- Optionally do some updates and ensure only files in app folder updated not those in default folder
- DELETE data files in default user data folder
- Turn OFF new setting
- Ensure all data like subscribed channels/profiles, sub cache, playlist, settings, etc. are still present
-  Ensure  `Replace HTTP Cache` still ON
- Ensure data files now present in default `userData` folder (`%APPDATA%\FreeTube`, see https://www.electronjs.org/docs/latest/api/app#appgetpathname)

B. Non windows (macOS/Linux)
- Open app
- Ensure all data like subscribed channels/profiles, sub cache, playlist, settings, etc. are still present
- Ensure new setting does NOT appear

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -45,6 +45,9 @@ const IpcChannels = {
   GET_SCREENSHOT_FALLBACK_FOLDER: 'get-screenshot-fallback-folder',
   CHOOSE_DEFAULT_FOLDER: 'choose-default-folder',
   WRITE_TO_DEFAULT_FOLDER: 'write-to-default-folder',
+
+  GET_STORE_USER_DATA_IN_APP_FOLDER: 'get-store-user-data-in-app-folder',
+  TOGGLE_STORE_USER_DATA_IN_APP_FOLDER: 'toggle-store-user-data-in-app-folder',
 }
 
 const DBActions = {
@@ -222,6 +225,9 @@ const MIXED_SEARCH_HISTORY_ENTRIES_DISPLAY_LIMIT = 4
 // Displayed on the about page and used in the main.js file to only allow bitcoin URLs with this wallet address to be opened
 const ABOUT_BITCOIN_ADDRESS = '1Lih7Ho5gnxb1CwPD4o59ss78pwo2T91eS'
 
+// Filename for enabling user data storage in app folder (Windows)
+const STORE_USER_DATA_IN_APP_FOLDER_SWITCH_FILENAME = 'store-user-data-in-app-folder'
+
 export {
   IpcChannels,
   DBActions,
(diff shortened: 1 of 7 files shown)


PikachuEXE:
Ya also this doesn't work with portable version
I admit I have never thought about those 2 versions coz I have never used/tried them (as a user)
Marking as WIP

PikachuEXE:
Updated for portable version and NSIS installer version, custom build updated

absidue:
Not sure patching the installer is the right approach. I think if we add this, it should only be available in the portable version/zip files, as it doesn't make much sense to show and allow it with the installer version, as the installer is inherently not portable and the entire point of this is to make FreeTube more portable.

PikachuEXE:
Updated installer version (by detecting the existence of uninstaller executable) to have new setting hidden and disabled (even with flag file created)

efb4f5ff-1298-471a-8973-3d47447115dc:
I share the same sentiment tbh

FoxMcloud5655:
Then how about not advertising setup-free files (what you call portable) as portable when it's not truly portable?  This made me unreasonably upset when I figured out my settings weren't being stored on my external drive where the program actually resides when it's clearly marked as a portable application.

PR #7519, opened 2025-06-05
Fix some template strings in locale files, add script to find problematic values

ChunkyProgrammer:
## Pull Request Type
- [x] Bugfix

## Description
I ended up updating some code that I wrote for Invidious to find locales that use incorrect template strings/are missing template strings. https://github.com/iv-org/invidious/issues/1951#issuecomment-1947659377

This should make finding issues with other locales a lot easier

Ideally we would upstream some of this functionality into vue-i18n-eslint

## Screenshots 

image

## Testing
- `yarn run checkforbadtemplates` should create a `locale-errors.json` file at the root of the project as well as log the errors to the console. I didn't fix every issue as some of them aren't that straightforward without knowing the language that is being translated.

## Desktop

- **OS:** Fedora Linux
- **OS Version:** 42

### DIFF
--- a/_scripts/findMissingTemplates.mjs
+++ b/_scripts/findMissingTemplates.mjs
@@ -64,8 +64,8 @@ function isMissingInterpolation(defaultValue, otherValue, filename) {
     // not translated yet, we don't care
     return false
   }
-  const defaultMatches = defaultValue.match(/{[^}]*}/g)
-  const otherMatches = otherValue.match(/{[^}]*}/g)
+  const defaultMatches = Array.from(new Set(defaultValue.match(/{[^}]*}/g)))
+  const otherMatches = Array.from(new Set(otherValue.match(/{[^}]*}/g)))
 
   if (defaultMatches) {
     if (!otherMatches) {


PikachuEXE (approved):

_scripts/findMissingTemplates.mjs
```diff
@@ -0,0 +1,76 @@
...
+
+const filesInLocaleDir = await readdir(localesPath)
+
+for (const file of filesInLocaleDir) {
+  if (file !== defaultLocale) {
+    const fileData = loadYaml(await readFile(`${localesPath}/${file}`, { encoding: 'utf-8' }))
+    const fileDataKeys = Object.keys(fileData)
+    addErrors(defaultData, fileData, defaultKeys, fileDataKeys, file)
+  }
+}
```
absidue: Do we need to filter out the `activeLocales.json` file here as that isn't YAML?

_scripts/findMissingTemplates.mjs
```diff
@@ -0,0 +1,76 @@
+import { readdir, readFile, writeFile } from 'node:fs/promises'
```
absidue: I'm curious about the choice of using the async filesystem APIs as nothing seems to be done in parallel in this file?

_scripts/findMissingTemplates.mjs
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+writeFile('locale-errors.json', JSON.stringify(errors, null, 2))
+
+function addErrors(originalData, newData, originalKeys, newKeys, file) {
+  newKeys.forEach(fdk => {
+    if (!originalKeys.includes(fdk)) {
+      // errors.push({ fileName: file, error: 'extra key found', key: fdk })
+    } else {
```
absidue: Was this forgotten or intentionally left commented out?

ghost: I intentionally left it commented out since it adds a lot more "errors" to the file but it's something we should go through eventually

_scripts/findMissingTemplates.mjs
ghost: I think originally I was going to make it async but it started to get a bit more complicated with recursion so I decided against it to keep the code simpler

_scripts/findMissingTemplates.mjs
ghost: I didn't notice any issues but it's probably best to filter all non yaml files

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

absidue (dismissed):

PikachuEXE:
Wondering if duplicate variables can/should be ignored
e.g.
- `This setting is overridden by the app-wide "{appWideSetting}" setting, in the "{subsection}" section of the "{settingsSection}"`
- `此设置被应用级的 "{appWideSetting}" 设置所覆盖，"{appWideSetting}" 位于 "{subsection}" 部分，该部分在 "{settingsSection}" 中`
`{appWideSetting}` repeated twice but valid (I can read this locale
image

absidue (approved):

PikachuEXE (approved):

Issue #6194, opened 2024-11-18
[Feature Request]: Add apply button to search filters modal


digitalcraftsman:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Applying search filters is a bit unconvienent, as I have to:

1. click on the funnel icon
2. set the search filters
3. close the search filters modal
4. click on the magnifier icon to trigger a search

To speed up things, step three can be eliminated.

### Proposed Solution

How? add a new "Search"-button next to the "Close"-button in the search filters modal that immediately triggers a search with the search query and the selected filters:

![grafik](https://github.com/user-attachments/assets/6126917c-5d35-4c61-bc13-004daeb3a0a2)

### Alternatives Considered

None

### Issue Labels

ease of use improvement

efb4f5ff-1298-471a-8973-3d47447115dc:
see review comments on https://github.com/FreeTubeApp/FreeTube/pull/6277 https://github.com/FreeTubeApp/FreeTube/pull/6627 to improve your PR

PR #7531, opened 2025-06-07
add search filter reset button

### LINKED ISSUE #7396: [Feature Request]: Search Filters Reset Button
### Guidelines

### Problem Description

The search filter could benefit from a reset button, so you don't have to click so many times to return to the default search settings when using several filters.

### Proposed Solution

Simply add a reset button :)

### Alternatives Considered

-

### Issue Labels

ease of use improvement

### Additional Information

_No response_

ozrendev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #7396 

## Description
Add button that clears active search filters

## Screenshots 
Before

After

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/FtSearchFilters/FtSearchFilters.vue
+++ b/src/renderer/components/FtSearchFilters/FtSearchFilters.vue
@@ -51,6 +51,12 @@
     </FtFlexBox>
     <div class="searchFilterCloseButtonContainer">
       <FtButton
+        :label="$t('Clear filters')"
+        background-color="var(--accent-color)"
+        text-color="var(--text-with-main-color)"
+        @click="clearFilters"
+      />
+      <FtButton
         :label="$t('Close')"
         background-color="var(--primary-color)"
         text-color="var(--text-with-main-color)"
@@ -255,6 +261,15 @@ function hideSearchFilters() {
 function isVideoOrMovieOrAll(type) {
   return type === 'video' || type === 'movie' || type === 'all'
 }
+
+function clearFilters() {
+  sortByValue.value = SORT_BY_VALUES[0]
+  timeValue.value = TIME_VALUES[0]
+  typeValue.value = TYPE_VALUES[0]
+  durationValue.value = DURATION_VALUES[0]
+  featuresValue.value = []
+}
+
 </script>
 
 <style scoped src="./FtSearchFilters.css" />


src/renderer/components/FtSearchFilters/FtSearchFilters.vue
```diff
@@ -50,6 +50,12 @@
       />
     </FtFlexBox>
     <div class="searchFilterCloseButtonContainer">
+      <FtButton
+        :label="$t('Clear filters')"
```
PikachuEXE: Also need an entry in `static/locales/en-US.yaml` so people can translate
```suggestion
        :label="$t('Search Filters.Clear Filters')"
```

PikachuEXE (changes requested):

src/renderer/components/FtSearchFilters/FtSearchFilters.vue
ozrendev: Thanks for the quick response, have committed those changes

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
- I looks like that the close and clear filters button use different kinds of red
- Change secondary theme to Dracula green and see that words in clear filters label is green. See that the words within the buttons of the Data settings are black

ozrendev:
> * I looks like that the close and clear filters button use different kinds of red
> 
>     * Change secondary theme to Dracula green and see that words in clear filters label is green. See that the words within the buttons of the Data settings are black

The clear filters button is using the secondary color var as I thought this would be less confusing. I can change it to the primary color though or whichever is suitable?

efb4f5ff-1298-471a-8973-3d47447115dc:
Please read my comment again. I do not have anything against the secondary color theme. I have an issue with the text within it

ozrendev:
Ah okay I got it now, should be fixed

PikachuEXE:
I think you misunderstood
Expected: (See button at bottom right with black text
image

New button:
image

static/locales/en-US.yaml
```diff
@@ -118,6 +118,7 @@ Search Filters:
   Fetching results. Please wait: Fetching results. Please wait
   Fetch more results: Fetch more results
   There are no more results for this search: There are no more results for this search
+  Clear Filters: Clear filters
```
PikachuEXE: I think most of the places use capitalized words
e.g. settings page
image

```suggestion
  Clear Filters: Clear Filters
```

ozrendev: Definitely misunderstood, this should follow expected behavior. I left the close button as-is from before

efb4f5ff-1298-471a-8973-3d47447115dc:
Probably very out of scope: I need someone else to verify because i dont know if its my panel put if my secondary color theme is set to Red the Close button got a different shade of Red then the Clear Filters button

PikachuEXE:
Dunno what you are talking about
image

Using IPS panel

efb4f5ff-1298-471a-8973-3d47447115dc:
Ok disregard what i said. Ty for testing :)

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
FreeTube_qDlZNi1r5o

Summoning @kommunarr @ozrendev @pkrasicki for this discussion. The image above shows how this currently looks like in dev branch. Only real change compared to this PR is that the cancel button is now always white which is consistent within the whole application.

From my daily use of this feature something feels a bit off and i would like to share that with you to see if its just a me problem or something bigger (UI/UX issue). 

Everytime i apply a filter i somehow automatically gravitate towards the Clear Filters button as if my brain thinks that is has search now functionality applied to it. First i thought it must be because it is a different color compared to the cancel button but now i think this is happening because it is a similar style button as the close button and the positioning of this button. 

So in my mind this could be solved if the functionally is visually moved somewhere else in the filters modal and maybe a different style of button. idk something that is used often to clear stuff.

Does this make sense at all or am i just used to this placement from other applications/websites and try to apply it here?

PikachuEXE:
I agree Clear Filters shouldn't be styled as main action (the one that should be would be a new button for "Search, though I tried once and implementing it seems difficult...

Edit: issue related to adding search https://github.com/FreeTubeApp/FreeTube/issues/6194 also see attempted PR's linked to that issue

pkrasicki:
Here:
freetube-search-filters

But you might want to add a text label to this button that says "Reset".

efb4f5ff-1298-471a-8973-3d47447115dc:
I found this icon. Would this icon portrait the action better?
firefox_9mPYricr4v

Edit: @pkrasicki i like the positioning of the icon in your screenshot!

pkrasicki:
Yeah, that one is better.

efb4f5ff-1298-471a-8973-3d47447115dc:
@pkrasicki Would you like to PR it if others approve of this placement and icon?

@PikachuEXE @kommunarr @ozrendev thoughts on discussion?

pkrasicki:
Not really, it's just a quick mockup I did in Gimp. _You might also want to consider repositioning the refresh button on subscriptions page: https://github.com/FreeTubeApp/FreeTube/discussions/6293#discussioncomment-11643324._

PikachuEXE:
If it's shown always and it's shown as disabled by default then I guess it's easy enough to learn that it's a reset button

efb4f5ff-1298-471a-8973-3d47447115dc:
> If it's shown always and it's shown as disabled by default then I guess it's easy enough to learn that it's a reset button

i agree

pkrasicki:
I would display it only when it's needed. And if the icon doesn't make it clear what the button does on its own, add a text label, so that it looks like one of those buttons:

freetube-video-share-buttons

Fine, here is another mockup:

freetube-search-filters2

freetube-search-filters3

pkrasicki:
Now that I think about it, some things don't make sense in this modal. The "Sort By" column shouldn't be here, it's not a filter. The radio buttons are not a great solution either. Generally a filter is something you add. So it's strange that the user here can select type "All Types", which just means no filter. Same with time and duration.

YouTube's solution looks like this, but it's not great either:

youtube-filters-modal

Edit: I guess it's not that bad when compared to YouTube, I just think it could be better.

efb4f5ff-1298-471a-8973-3d47447115dc:
Every time an user applies a filter on YT it performs a search. If we would copy this behavior over than FT users will be ratelimited way faster because it will do multiple requests. We have implemented something that applies to our usecase.

Regarding how it should look like. I think we need to be consistent here. The filter icon in the top nav is just an icon and when you hover on it you will see the description of it. IMO we should do the same with the Clear filters icon.

We could make it disabled by default or make it appear once a filter has been chosen.

ozrendev:
> We could make it disabled by default or make it appear once a filter has been chosen.

Yeah I agree having the filter icon appear when a filter is selected should give the user enough feedback to know it's a clear icon (with tooltip)

I know it's just a mockup, but also having the icon a bit bigger (same size as the current filter icon) would help. The position looks good!

efb4f5ff-1298-471a-8973-3d47447115dc:
@ozrendev would you like to create a pr for it?

ozrendev:
> @ozrendev would you like to create a pr for it?

Yeah, I'll do this soon!

PR #5885, opened 2024-10-17
Undo Feature For Remove From Playlist.

Soham456:
## Pull Request Type

- [ X ] Feature Implementation

## Related issue
"closes #5421"

## Additional context
resolved yarn.lock conflict

PR #7544, opened 2025-06-10
Update single user playlist view to remove videos immediately on exit

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Undo introduced in https://github.com/FreeTubeApp/FreeTube/pull/5885

## Description

1. Error raised if out of playlist view when scheduled task for actual removing videos from playlist run
2. IMO annoying to need to stay in the view for 5s when I am just trying to remove 1 finished video and play another one
  Undo still useful for removing multiple videos

Screenshot 2025-06-10 at 09 19 35

This PR makes single playlist view remove videos queued for removal after 5s to be removed immediately once 

## Testing

- Create some really long user playlist (easier for testing, e.g. copy one from LTT long playlist
- Create 2+ windows to ensure change sync (optional, nothing changed on that side
- Remove 2+ videos, undo, ensure all removal undone (no change on other windows
- Remove 2+ videos, wait 5s, repeat N times, ensure videos are really removed (can go to another view and back to ensure
- Remove 2+ videos, exit view immediately (go back to playlists view or about/settings view or start playing one of the items)
  Ensure videos removed and toast closed on view changed (no longer able to undo)

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -213,7 +213,7 @@ User Playlists:
       This video cannot be moved up.: This video cannot be moved up.
       This video cannot be moved down.: This video cannot be moved down.
       Video has been removed: Video has been removed
-      Video has been removed. Click here to undo.: Video has been removed. Click here to undo.
+      Video has been removed. Click here to undo.: Video has been removed. Click here to undo. Exiting this view will remove them immediately (cannot undo).
       There was a problem with removing this video: There was a problem with removing this video
 
       This playlist is already being used for quick bookmark.: This playlist is already being used for quick bookmark.
--- a/src/renderer/components/FtToast/FtToast.vue
+++ b/src/renderer/components/FtToast/FtToast.vue
@@ -37,9 +37,9 @@ let idCounter = 0
 const toasts = shallowReactive([])
 
 /**
- * @param {CustomEvent<{ message: string, time: number | null, action: Function | null }>} event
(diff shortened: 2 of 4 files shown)


static/locales/en-US.yaml
```diff
@@ -213,7 +213,7 @@ User Playlists:
       This video cannot be moved up.: This video cannot be moved up.
       This video cannot be moved down.: This video cannot be moved down.
       Video has been removed: Video has been removed
-      Video has been removed. Click here to undo.: Video has been removed. Click here to undo.
+      Video has been removed. Click here to undo.: Video has been removed. Click here to undo. Exiting this view will remove them immediately (cannot undo).
```
efb4f5ff-1298-471a-8973-3d47447115dc: Tbh i dont think this change is needed as it was a bug in the first place

PikachuEXE:
> So the total watch time gets updated immediately but video count not. Best to also update the video count.

Addressed

> It says "Video" has been removed but i removed multiple.

Not related to the issue. Maybe someone can create a followup PR? :)

> Video count doesn't get updated on playlist page

More unrelated to this issue than the above one. That's asking for instant update of local playlist attributes and I dunno if performance would be worse (and I don't see a need for instant update

absidue (approved):

PikachuEXE:
@ChunkyProgrammer ~

ChunkyProgrammer (approved):

Issue #7563, opened 2025-06-15
[Feature Request]: New comment type - Voice Reply comment


ArthurKun21:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Problem Description

New type of comment not shown on the freetube

### Proposed Solution

Be able to parse this new type of comment

### Alternatives Considered

None

### Issue Labels

new feature

### Additional Information

https://www.youtube.com/watch?v=ZJ1N9PLNzT4

Youtube

![Image](https://github.com/user-attachments/assets/28f8d39a-b101-43a7-901d-d1afeb6799f3)

Freetube

![Image](https://github.com/user-attachments/assets/8c81e385-9461-4112-a86f-22a0be95bf58)

[bundle-1749990089295.log](https://github.com/user-attachments/files/20745502/bundle-1749990089295.log)

PR #7573, opened 2025-06-17
Local API: Add support for transcript voice replies

ChunkyProgrammer:
## Pull Request Type
- [x] Feature Implementation

## Related issue
closes https://github.com/FreeTubeApp/FreeTube/issues/7563

## Description
Adds support for displaying voice reply transcript. We may want to add additional styling in the future (ex: an icon)
Relies on: https://github.com/LuanRT/YouTube.js/pull/981

## Screenshots 
reply is now shown instead of empty

## Testing
- go to https://www.youtube.com/watch?v=ZJ1N9PLNzT4
- view replies to netherwolf100's comment

## Desktop
- **OS:** Fedora Linux
- **OS Version:** 42

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -1649,6 +1649,8 @@ export function parseLocalComment(comment, commentThread = undefined) {
     hasReplyToken = true
   }
 
+  const commentTextRuns = comment.voice_reply_container?.transcript_text ? comment.voice_reply_container.transcript_text.runs : comment.content.runs
+
   return {
     id: comment.comment_id,
     dataType: 'local',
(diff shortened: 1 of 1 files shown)


absidue (commented):
To reduce the duplicate code, I wonder if it makes sense to figure out which runs to use in a variable before the return. E.g.

```js
const textRuns = comment.voice_reply_transcript ? comment.voice_reply_transcript.runs : comment.content.runs

// ...

text: Autolinker.link(parseLocalTextRuns(textRuns, 16, { looseChannelNameDetection: true })),
```

efb4f5ff-1298-471a-8973-3d47447115dc:
Im not sure if this fully closes the issue. If voice reply isn't supported in the country you're viewing it in you will see the transcript. 

So we show the transcript but not the actual voice reply if it is supported in that country

absidue:
This always shows the transcript. My understanding was that the website only has the transcript and the playback of the voice replies is only possible in the mobile apps or do you know a country that also supports playback on the website?

The actual playback would be a lot more complicated because if I am remembering it correctly from the last time I looked into it, the voice replies are backed by an unlisted, audio-only YouTube video, so not just some audio file we can easily play.

efb4f5ff-1298-471a-8973-3d47447115dc:
relies on https://github.com/FreeTubeApp/FreeTube/pull/7701

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Hmm im not seeing it

https://github.com/user-attachments/assets/59d6e592-0d39-431e-a862-40ec94c7695f

ghost:
Oops, i forgot to push my rebase @efb4f5ff-1298-471a-8973-3d47447115dc try pulling and doing a yarn install now

absidue (approved):

PikachuEXE (approved):

PR #7607, opened 2025-06-28
Enable Breton Language

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type
- [x] Other - enable translated locale

## Related issue
User requested this in General room and is translated to 100%

## Description
Enable Breton to be used in the app

## Testing
See Breton in the drop down of language setting


efb4f5ff-1298-471a-8973-3d47447115dc:
uhhh not sure what happened here

PR #7650, opened 2025-07-08
Deduplicate nested development dependencies in the yarn.lock files

absidue:
## Pull Request Type

- [x] Other

## Description

This pull request deduplicates some of the nested development dependencies in the `yarn.lock` file. I wasn't able to fully deduplicate them, due to differences in major versions or some packages requiring a specific version. On my machine the size of the `node_modules` folder went from 818 MB (858236064 bytes) to 759 MB (796660800 bytes), which works out to a ~61 MB reduction. That means less disk usage on developer and GitHub Actions machines as well as faster installs of the dependencies.

## Testing

- Successful production build: https://github.com/absidue/FreeTube/actions/runs/16150577673
- The linter run on this pull request.
- `yarn dev` (if it builds and electron launches it worked).

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -66,21 +66,7 @@
     "@jridgewell/trace-mapping" "^0.3.28"
     jsesc "^3.0.2"
 
-"@babel/helper-annotate-as-pure@^7.18.6":
-  version "7.18.6"
-  resolved "https://registry.yarnpkg.com/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.18.6.tgz#eaa49f6f80d5a33f9a5dd2276e6d6e451be0a6bb"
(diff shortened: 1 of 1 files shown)


ghost:
Did you manually edit the lock file or clear it and then run yarn install (or another option)?

absidue:
Manually edited and validated by invoking `yarn install` which ignores invalid entries, logs them and replaces them with corrected ones. I also locally ran, `yarn lint`, `yarn dev` and `yarn build` to make sure they worked correctly and then the linting and building was also done again here in GitHub Actions.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #7719, opened 2025-07-21
Move General section within the Distraction Free Settings to top

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Visual change

## Description

General being at the bottom of the the Distraction Free Settings seems a bit silly so i moved it to the top to make to make the order of the sections more logical

## Screenshots 

VirtualBoxVM_VJwMLTsGyR

## Testing

1. Go to the Distraction Free Settings
2. Check if General section is the first one listed

### DIFF
--- a/src/renderer/components/distraction-settings/distraction-settings.css
+++ b/src/renderer/components/distraction-settings/distraction-settings.css
@@ -3,3 +3,7 @@
     display: none;
   }
 }
+
+.containingTextFlexBox {
+  margin-block-end: 2em;
+}
--- a/src/renderer/components/distraction-settings/distraction-settings.vue
+++ b/src/renderer/components/distraction-settings/distraction-settings.vue
@@ -76,7 +76,7 @@
         @toggle-show-tags="handleAddedChannelsHidden"
       />
     </ft-flex-box>
-    <ft-flex-box>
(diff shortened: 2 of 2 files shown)


PikachuEXE:
I think the spacing between the box and the group title need some adjustment...
Existing (0.5em
image

2em
image

efb4f5ff-1298-471a-8973-3d47447115dc:
> I think the spacing between the box and the group title need some adjustment.

Good catch. What would be best here. Create a separate class for this case?

So it would be the same as this but just with 2.0em https://github.com/FreeTubeApp/FreeTube/blob/df428b1b8f0e62bedc774499554b6feb96409cbc/src/renderer/components/FtSettingsSection/FtSettingsSection.scss#L43-L46

PikachuEXE:
Add class name for `Settings.Distraction Free Settings.Hide Videos and Playlists Containing Text` (or the flex box
And add `margin-block-end 2.0em` (as in keep your distance from my butt

efb4f5ff-1298-471a-8973-3d47447115dc:
Srry i tested something very unrelated and noticed i pushed it towards this branch. Reverted those commits and have added the proper code now

PikachuEXE (approved):

PR #7722, opened 2025-07-21
Change Settings order

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Visual change

## Related issue

#7719 

## Description

This PR tries to implement logical changes to the order of the Settings.

|Old Order|New Order|
|-----------|------------|
|1. General|1. General |
|2. Theme |2. Theme   |
|3. Player|3. Player|
|4. External Player|4. External Player|
|5. Subscription|5. Subscription|
|6. Distraction Free|6. Distraction Free  |
|7. Privacy|**7. Parental control**|
|8. Data|8. Privacy|
|9. Proxy|9. Data|
|10. Download|10. Proxy|
|11. Parental control|11. Download|
|12. SponsorBlock|12. SponsorBlock|
|13. Experimental|**13. Password**|
|14. Password|**14. Experimental**|

- 7 = Parental controls, because it makes more sense to group that with 6
- 13 = Password, see reasoning 14
- 14 = Experimental, last because settings that are not part of "Normal" settings belong at the bottom

## Testing

1. Go to Settings
2. Observe Setting order changes

### DIFF
--- a/src/renderer/views/Settings/Settings.js
+++ b/src/renderer/views/Settings/Settings.js
@@ -128,15 +128,15 @@ export default defineComponent({
             ]
           : []),
         {
-          type: 'password-settings',
-          title: this.$t('Settings.Password Settings.Password Settings'),
-          icon: 'key'
-        },
-        {
           type: 'sponsor-block-settings',
           title: this.$t('Settings.SponsorBlock Settings.SponsorBlock Settings'),
           // TODO: replace with SponsorBlock icon
           icon: 'shield'
+        },
+        {
+          type: 'password-settings',
+          title: this.$t('Settings.Password Settings.Password Settings'),
+          icon: 'key'
         },
         ...(process.env.IS_ELECTRON
           ? [{


PikachuEXE:
I think password should still be at the end coz it doesn't make sense to lock settings before I read all the settings section before (unless you jump to password section)
Parental control is for other users not myself so I am not so sure if that should even be moved...

I would at most swap 11 & 12 (from old) or move 11 to 13

efb4f5ff-1298-471a-8973-3d47447115dc:
> I think password should still be at the end coz it doesn't make sense to lock settings before I read all the settings section before 

An user reads all the settings the first time they install an application and maybe after an update. Not all the time. What you're saying resonates more to us doing that allot because of testing and what not

> Parental control is for other users not myself so I am not so sure if that should even be moved...

Distraction Free Settings and Parental Control do the same they hide elements from the user thats why i grouped them. The only real difference is the target audience

> I would at most swap 11 & 12 (from old) or move 11 to 13

umm im not following... what would you like to be moved around from the new list?

PikachuEXE:
I would still say `SponsorBlock` IS part of the app (just disabled by default like IV API, proxy, external player...
Swap 12 & 13 in new list then it's fine

efb4f5ff-1298-471a-8973-3d47447115dc:
> I would still say `SponsorBlock` IS part of the app (just disabled by default like IV API, proxy, external player... Swap 12 & 13 in new list then it's fine

Sure that reasoning makes sense

efb4f5ff-1298-471a-8973-3d47447115dc:
Done

PikachuEXE (approved):

absidue (approved):

ChunkyProgrammer (approved):

PR #7757, opened 2025-07-30
Feat: Display player overlay controls on keyboard interaction

### LINKED ISSUE #7458: [Feature Request]: Display video progress indicator after invoking rewind/fast-forward commands, or changing playback speed
### Guidelines

### Problem Description

The video progress indicator (how much time is elapsed) does not get summoned and display after issuing a rewind (<kbd>←</kbd>) or fast-forward (<kbd>→</kbd>) command, or playback speed adjustment.

### Proposed Solution

The video progress indicator should display on any timeline movement, or playback speed adjustment, just like it does in most other video-playing applications, including YouTube web and VLC.

### Alternatives Considered

Moving the mouse is the only way I could find to display the video progress indicator.

### Issue Labels

### Additional Information

_No response_

caetano-dev:
## Pull Request Type

- [x] Feature Implementation

## Related issue

Closes #7458 

## Description

Shows the video controls when the users interacts with the video using the keyboard, similar to YouTube.

## Screenshots

https://github.com/user-attachments/assets/190a0880-1e7c-448c-8cb9-f8b329f6bad8

## Testing

1. Open any video
2. Use J, L, <, >, M and arrow keys to interact with the video
3. See the controls show up

## Desktop

- **OS:** Linux
- **OS Version:** Ubuntu 24
- **FreeTube version:** v0.23.5 Beta


src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -3035,6 +3042,7 @@ export default defineComponent({
       valueChangeIcon.value = icon
       showValueChangePopup.value = true
       invertValueChangeContentOrder.value = invertContentOrder
+      showOverlayControls()
```
PikachuEXE: This makes `showOverlayControls` called twice in `seekBySeconds`
Maybe just call `showOverlayControls` under some of the `showValueChange()` calls?

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -3024,6 +3025,12 @@ export default defineComponent({
     const invertValueChangeContentOrder = ref(false)
     let valueChangeTimeout = null
 
+    function showOverlayControls() {
+      if (container.value) {
+        container.value.dispatchEvent(new MouseEvent('mousemove'))
+      }
```
PikachuEXE: Fake event is not needed, method available to make UI shown
https://shaka-player-demo.appspot.com/docs/api/shaka.ui.Controls.html
```suggestion
      ui.getControls().showUI()
```

caetano-dev:
Changes applied!

kommunarr (commented):
**suggestion:** put `showOverlayControls()` at the end of the `showValueChange` implementation since it appears to be being done after each use of it

caetano-dev:
Changes applied. Thanks for the suggestion!

efb4f5ff-1298-471a-8973-3d47447115dc:
so i dug a little more into what YT does with their other keyboard shortcuts and this is what i found out:

Shortcuts YT shows the player UI but FT does not (yet)

Addressed by this PR:
* J
* L
* left arrow
* right arrow
* O | <
* P | >
* arrow up
* arrow down
* m

Not addressed:
* home
* end
* 0..9
* ctrl + left (it does not here but i think that might be an oversight because they do it for ctrl + right and other playback related controls)
* ctrl + right 
* f (it does this sometimes within FT)
* m
* c

Works on YT but not useful to implement within FT
* shift + P 
* shift + N

i suggest following them

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2577,6 +2584,8 @@ export default defineComponent({
 
       document.removeEventListener('keydown', keyboardShortcutHandler)
       document.addEventListener('keydown', keyboardShortcutHandler)
+      document.addEventListener('fullscreenchange', showOverlayControls)
+      document.addEventListener('webkitfullscreenchange', showOverlayControls)
```
caetano-dev: For full screen, I am using event listeners because just calling `showOverlayControls` after the user presses F was not working when they leave the mode.

And on macOS, using `fullscreenchange` alone was not enough. That is why I included `webkitfullscreenchange`.

caetano-dev:
It is now showing the overlay for all the shortcuts you mentioned, excluding Shift P and Shift N.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Functionally wise LGTM but i prefer @absidue to sign off code wise first

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
absidue: As that extra event is only needed on macOS please wrap it in a `if (process.platform === 'darwin) {}` check (obviously also the removeEventListener part below).

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2237,6 +2239,7 @@ export default defineComponent({
             // Rewind by the time-skip interval (in seconds)
             seekBySeconds(-defaultSkipInterval.value * player.getPlaybackRate(), false, true)
           }
+          if (canChapterJump(event, 'previous')) showOverlayControls()
```
absidue: Same issue as with your playlist duration pull request, rather than integrating it into the existing code, you decided to copy-paste and duplicate code. Please move the `showOverlayControls()` call into the `if` block above.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2247,6 +2250,7 @@ export default defineComponent({
             // Fast-Forward by the time-skip interval (in seconds)
             seekBySeconds(defaultSkipInterval.value * player.getPlaybackRate(), false, true)
           }
+          if (canChapterJump(event, 'next')) showOverlayControls()
```
absidue: Ditto

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -3043,6 +3058,7 @@ export default defineComponent({
       valueChangeTimeout = setTimeout(() => {
         showValueChangePopup.value = false
       }, 2000)
+      showOverlayControls()
```
absidue, line 3073:
```suggestion

      showOverlayControls()
```

caetano-dev:
Changes applied!

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2577,6 +2584,10 @@ export default defineComponent({
 
       document.removeEventListener('keydown', keyboardShortcutHandler)
       document.addEventListener('keydown', keyboardShortcutHandler)
+      document.addEventListener('fullscreenchange', showOverlayControls)
+      if (process.platform === 'darwin') {
+        document.addEventListener('webkitfullscreenchange', showOverlayControls)
```
PikachuEXE: If this is for browsers (not electron), no point hiding behind `process.platform`
If it's for macOS, electron just use recent Chromium so no need for this (searched electron code base

https://caniuse.com/mdn-api_document_fullscreenchange_event

Unless it doesn't work in specific OS/browser I don't understand the purpose of this

caetano-dev:
@PikachuEXE After more testing, I've found that the video controls only appear when a user presses 'F' if their mouse is already over the video.

Maybe this is what @efb4f5ff-1298-471a-8973-3d47447115dc meant in https://github.com/FreeTubeApp/FreeTube/pull/7757#issuecomment-3180528879 by saying "f (it does this sometimes within FT)".

The `eventListeners` I added were actually not working. My tests were probably a false positive because of my mouse position, so I deleted them.

I haven't found a good way to make the controls appear when _exiting_ fullscreen mode. Simulating a `mousemove` event also did't work.

Should we consider this an edge case we can ignore for now? Entering fullscreen mode will always show the controls, unless the user has more than 1 monitor. Leaving it will depend on their mouse position.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
PikachuEXE: A working one (plz test yourself) seems to be
(need to import `nextTick` from vue
```js
      document.addEventListener('fullscreenchange', () => {
        nextTick(showOverlayControls)
      })
```

caetano-dev: It does work. Thanks!

caetano-dev:
Control overlays now show with F.

Thanks @PikachuEXE

PikachuEXE:
Waiting for others to review consistency with YT (which I know nothing

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -1,4 +1,4 @@
-import { computed, defineComponent, onBeforeUnmount, onMounted, reactive, ref, shallowRef, watch } from 'vue'
+import { computed, defineComponent, onBeforeUnmount, onMounted, reactive, ref, shallowRef, watch, nextTick } from 'vue'
```
absidue:
```suggestion
import { computed, defineComponent, nextTick, onBeforeUnmount, onMounted, reactive, ref, shallowRef, watch } from 'vue'
```

efb4f5ff-1298-471a-8973-3d47447115dc:
@caetano-dev apologies for the delay. Could you merge dev branch into this so videos are playable again

caetano-dev:
Rebased.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
im receiving errors when pressing `f`

VirtualBoxVM_FcmLP1dZq1

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

kommunarr (dismissed):

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2925,6 +2936,9 @@ export default defineComponent({
       document.body.classList.remove('playerFullWindow')
 
       document.removeEventListener('keydown', keyboardShortcutHandler)
+      document.removeEventListener('fullscreenchange', () => {
+        nextTick(showOverlayControls)
+      })
```
PikachuEXE: Need a named function (same function object) to be put into `addEventListener` & `removeEventListener` (like `keyboardShortcutHandler` above
If you declare unnamed function twice they would be 2 function objects not the same function

caetano-dev:
Changes applied.

PikachuEXE (approved):

kommunarr (approved):

absidue (approved):

PR #7797, opened 2025-08-08
Feature: Silence skip

### LINKED ISSUE #1766: [Feature Request] Skip the silence in media playback
No, this feature request is not related to any bugs or usage problems.

There is a unique feature in Google ExoPlayer (used by NewPipe) that allows the viewer to fast forward and skip the silence in media playback. I would like to have a similar feature in FreeTube that is toggled on and off somewhere.

I struggled to find any media player that has a similar feature.

**Screenshots**
N/A

**Additional context**
I have provided a link below for the relevant code in ExoPlayer that accomplishes this task.

http://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/audio/SilenceSkippingAudioProcessor.java

some-git-user:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #1766

## Description

This PR implements a simple silence skip logic, to dynamically determine silent video parts, related to the individual max volume of the audio source.

## Screenshots 

https://github.com/user-attachments/assets/f236ff1e-e788-4c45-83d1-bf425f2127d1

## Testing

- Open any video (prefer a podcast with silent or mute parts between conversations).
- Toggle the function via the overflow menu Button.
- Watch/hear the player increasing playback speed to 3x between silent/mute audio parts and decreasing to the previous speed when audio is detected.
- Toggle the function again to deactivate and fallback to the previous speed.
- When the feature is enabled and you change the speed higher then 3x, the feature will deactivate itself.

## Desktop

- **OS:** debian
- **OS Version:** 13
- **FreeTube version:** v0.23.13 Beta


efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Remove all the locale files. Only keep en-us

efb4f5ff-1298-471a-8973-3d47447115dc:
Im a bit confused by this skip rate. Please elaborate why you chose for 2.5x instead of e.g. 10x

Lets say the ending of a certain video is 60sec it will take 24sec to skip the silence. Thats way too long imo.

Why not detect when there is no audio and detect when audio is playing again and instantly skip towards that timestamp instead of relying on playback rate to handle this

some-git-user:
> Im a bit confused by this skip rate. Please elaborate why you chose for 2.5x instead of e.g. 10x
> 
> Lets say the ending of a certain video is 60sec it will take 24sec to skip the silence. Thats way too long imo.
> 
> Why not detect when there is no audio and detect when audio is playing again and instantly skip towards that timestamp instead of relying on playback rate to handle this
the

The main goal of this feature is to skip (accelerate) silence between conversations, or other silent parts, not to fast forward to the end of a video.
I tried different speed steps with multiple videos and 2.5x was the best one in my tests, since faster steps could also "cut off" the beginning of the next non silent part.

I'm not sure how to "detect when there is no audio and detect when audio is playing again and instantly skip towards that timestamp" .
Maybe I'm wrong here, but to know this you must buffer and analyze the whole video before the playback start, so the code knows the parts with no audio.
I also tried to create a little UI, to let the user make individual changes for every video, to match the needed audio threshold, but I'm not familiar with Vue and the lack of Typescript made it even harder for me to get into this project.

efb4f5ff-1298-471a-8973-3d47447115dc:
This implementation works really well, especially considering that our player is the limiting factor in this case.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
I would like to be a button for this inside the player instead of outside the player

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
- Please move the icon to the left of the autoplay icon
- It should remember its state in the watch session. Examples -> https://github.com/FreeTubeApp/FreeTube/pull/5903 https://github.com/FreeTubeApp/FreeTube/pull/6449
- It probably also should receive a toggle in the Player settings called Enable [name of the feature] by Default

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
- If i use my left/right arrow keys to skip forward/backward it sometimes skips 2,5x because the skip entered a silence bit. I think thats undesired behavior as it skipped 12,5seconds. The creator wasnt silent for 12,5sec. Context: Arrow Left/Right, Fast-Forward X seconds based on the Fast-Forward Interval and current Video Playback Rate (Ex: 5 seconds Fast-Forward Interval and Video Playback Rate 2x, fast-forwards video 10 seconds)

https://github.com/user-attachments/assets/914cbacd-684f-488c-adc4-b17629a0a666

- Put the Silence Skip within the overflow menu beneath loop for both desktop and mobile
- If i enable this feature within the player settings it turns is on everytime even if i turned it off in the watch session. What the user does within the Watch session should be retained. If user navigates to lets say the subscriptions page and clicks on a video that will be the start of a new Watch session and the setting should be enabled by default.
- You introduced a new icon. See Before and after pics
firefox_EddjIAP0uw
VirtualBoxVM_WLR4rMjz2V

- Edit the PR body to reflect latest changes including tests
- Remove notification popup
- Move toggle to the left side
VirtualBoxVM_bsj99BU9Xp

- Creator talks continuously but it still get activated. idk if thats a real concern or not.

https://github.com/user-attachments/assets/1130db5e-08fe-449b-b152-79aec8d87a99

- Unsure if related but i receive allot of the following error
VirtualBoxVM_jscEjcoDVR

- I wonder if there are icons that could better represent the functionality of this feature, as I feel the current ones don’t quite capture it.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @some-git-user could you rebase into dev branch so that the playback issues are resolved and this PR can be tested

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc could you please point me to an code example to address following issue?
"_If i enable this feature within the player settings it turns is on everytime even if i turned it off in the watch session. What the user does within the Watch session should be retained. If user navigates to lets say the subscriptions page and clicks on a video that will be the start of a new Watch session and the setting should be enabled by default_."

I can't find where to persist the watch session and ignore the Vuex.Store settings.

efb4f5ff-1298-471a-8973-3d47447115dc:
> It should remember its state in the watch session. Examples -> #5903 #6449

Hi @some-git-user i pointed you to a few examples some time ago. Did you look at these? 

Also could you rebase again so that playback is functions properly

src/constants.js
```diff
@@ -224,7 +224,9 @@ const PlayerIcons = {
...
+  RECTANGLE_DEFAULT: 'M160-160q-33 0-56.5-23.5T80-240v-480q0-33 23.5-56.5T160-800h640q33 0 56.5 23.5T880-720v480q0 33-23.5 56.5T800-160H160Zm0-80h640v-480H160v480Zm0 0v-480 480Z',
+  SKIP_SILENCE_DEFAULT: 'M360-840v-80h240v80H360Zm80 440h80v-240h-80v240Zm40 320q-74 0-139.5-28.5T226-186q-49-49-77.5-114.5T120-440q0-74 28.5-139.5T226-694q49-49 114.5-77.5T480-800q62 0 119 20t107 58l56-56 56 56-56 56q38 50 58 107t20 119q0 74-28.5 139.5T734-186q-49 49-114.5 77.5T480-80Zm0-80q116 0 198-82t82-198q0-116-82-198t-198-82q-116 0-198 82t-82 198q0 116 82 198t198 82Zm0-280Z',
+  SKIP_SILENCE_FILLED: 'M360-840v-80h240v80H360ZM480-80q-75 0-140.5-28.5T225-186q-49-49-77-114.5T120-440q0-74 28.5-139.5T226-694q49-49 114.5-77.5T480-800q63 0 120 21t104 59l58-58 56 56-56 58q36 47 57 104t21 120q0 74-28 139.5T735-186q-49 49-114.5 77.5T480-80Zm0-360Zm0-80h268q-18-62-61.5-109T584-700L480-520Zm-70 40 134-232q-59-15-121.5-2.5T306-660l104 180Zm-206 80h206L276-632q-42 47-62.5 106.5T204-400Zm172 220 104-180H212q18 62 61.5 109T376-180Zm40 12q66 17 128 1.5T654-220L550-400 416-168Zm268-80q44-48 63.5-107.5T756-480H550l134 232Z',
```
absidue: The keys should be the name of the Material Symbol with a suffix saying whether it was the default or filled variant, as [searching for skip silence on the Material Symbols page](https://fonts.google.com/icons?icon.query=skip+silence) doesn't produce any results, it looks like you've invented your own name here.

src/renderer/store/modules/settings.js
```diff
@@ -313,6 +313,7 @@ const state = {
   uiScale: 100,
   userPlaylistsSortBy: 'latest_played_first',
   userHistorySortBy: 'latest_played_first',
+  skipSilenceEnabled: false,
```
absidue: Please move this new setting up with the other player settings, as it doesn't have any side effects it doesn't belong in the "The settings below have side effects" section.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -3026,14 +3080,118 @@ export default defineComponent({
...
+     *
+     * The detection logic calculates the maximum and average amplitude of the audio signal. If a silent segment
+     * is detected and persists for a defined minimum duration, the video is fast-forwarded and the output is smoothly
+     * muted using a GainNode to avoid click artifacts. When non-silent audio is detected and persists for a minimum duration,
+     * the output is smoothly unmuted and playback speed returns to normal, with a additional delay to further reduce audio clicks.
+     *
+     * The function continuously analyzes the audio stream using requestAnimationFrame, adapting playback and muting in real time.
+     * All transitions for muting and unmuting use smooth ramping via setTargetAtTime for click-free audio..
+     */
+    function skipSilence() {
```
absidue: This function is in the wrong place too, it's called by the button in the player or by keyboard shortcuts not by the watch page component, so it doesn't belong in the `functions used by the watch page` region.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -811,13 +819,15 @@ export default defineComponent({
...
+          'ft_screenshot',
+          'ft_skip_silence_toggle',
           'ft_autoplay_toggle',
           props.format === 'legacy' ? 'ft_legacy_quality' : 'quality',
           'playback_rate',
           'captions',
           'ft_audio_tracks',
           'loop',
-          'ft_screenshot',
+          'ft_skip_silence_toggle',
```
absidue: Why is `ft_skip_silence_toggle` listed twice and why was the screenshot button moved?

static/locales/en-US.yaml
```diff
@@ -1215,3 +1216,6 @@ KeyboardShortcutPrompt:
   End: Seek to the end of the video
   Skip to Next Video: Skip to the next video in a playlist or next recommended video
   Skip to Previous Video: Skip to the previous video in a playlist
+
+SilenceSkip:
+  Skip Silence: Skip Silence
```
absidue: Please place this with the translation strings for other player buttons.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -3026,14 +3080,118 @@ export default defineComponent({
...
+     *
+     * The function continuously analyzes the audio stream using requestAnimationFrame, adapting playback and muting in real time.
+     * All transitions for muting and unmuting use smooth ramping via setTargetAtTime for click-free audio..
+     */
+    function skipSilence() {
+      if (!isSilenceSkipEnabled.value) {
+        isSilenceSkipEnabled.value = true
+      } else {
+        isSilenceSkipEnabled.value = false
+      }
```
absidue:
```suggestion
      isSilenceSkipEnabled.value = !isSilenceSkipEnabled.value
```

static/locales/en-US.yaml
```diff
@@ -944,6 +945,8 @@ Video:
     You appear to be offline: You appear to be offline.
     Playback will resume automatically when your connection comes back: Playback will resume automatically when your connection comes back.
     Skipped segment: 'Skipped {segmentCategory} segment'
+    SilenceSkip:
+      Skip Silence: Skip Silence
```
efb4f5ff-1298-471a-8973-3d47447115dc: I think this isnt needed because it only has one key below it

```suggestion
    Skip Silence: Skip Silence
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
efb4f5ff-1298-471a-8973-3d47447115dc: @some-git-user you forgot to address this

src/renderer/components/ft-shaka-video-player/player-components/SkipSilenceButton.js
```diff
@@ -0,0 +1,72 @@
...
+    this.eventManager.listen(events, 'localeChanged', () => {
+      this.updateLocalisedStrings_()
+    })
+
+    this.updateLocalisedStrings_()
+  }
+
+  /** @private */
+  updateLocalisedStrings_() {
+    this.nameSpan_.textContent = this.button_.ariaLabel = i18n.t('Video.Player.SilenceSkip.Skip Silence')
```
efb4f5ff-1298-471a-8973-3d47447115dc:
```suggestion
    this.nameSpan_.textContent = this.button_.ariaLabel = i18n.t('Video.Player.Skip Silence')
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -811,13 +819,14 @@ export default defineComponent({
 
       if (onlyUseOverFlowMenu.value) {
         uiConfig.overflowMenuButtons = [
+          'ft_screenshot',
```
efb4f5ff-1298-471a-8973-3d47447115dc:
```suggestion
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -811,13 +819,14 @@ export default defineComponent({
...
           'ft_autoplay_toggle',
           props.format === 'legacy' ? 'ft_legacy_quality' : 'quality',
           'playback_rate',
           'captions',
           'ft_audio_tracks',
           'loop',
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 833:
```suggestion
          'loop',
          'ft_skip_silence_toggle',
          'ft_screenshot',
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -811,13 +819,14 @@ export default defineComponent({
...
-          'ft_screenshot',
+          'ft_skip_silence_toggle',
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 834:
```suggestion
```

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
- If i enable this feature my playback rate defaults to 1x and i cant change it

https://github.com/user-attachments/assets/1fcdba5f-5c71-448d-8492-5612dc131e97


https://github.com/user-attachments/assets/5926d695-f9d2-443e-81fb-508c96e8f4ad

- O and P keyboard shortcuts are used to increase or decrease playback speed but because of the lock to 1x its behaving in weird ways

https://github.com/user-attachments/assets/1d9dfc56-7ffe-4b17-95ff-35650b263f28

- Discussion point: What should we do about people that are using a playback speed above 3x? I mean if i use 4x to watch a video and silent parts kicks in it probably shouldnt touch the playback speed right?

> If i use my left/right arrow keys to skip forward/backward it sometimes skips 2,5x because the skip entered a silence bit. I think thats undesired behavior as it skipped 12,5seconds. The creator wasnt silent for 12,5sec. Context: Arrow Left/Right, Fast-Forward X seconds based on the Fast-Forward Interval and current Video Playback Rate (Ex: 5 seconds Fast-Forward Interval and Video Playback Rate 2x, fast-forwards video 10 seconds)

- Other shortcuts are also affected by this: J and L

efb4f5ff-1298-471a-8973-3d47447115dc:
Umm i think this isnt supposed to happen. It default to 10x now

https://github.com/user-attachments/assets/0c62341e-d865-4eae-b97a-a43112d8cb28

> Discussion point: What should we do about people that are using a playback speed above 3x? I mean if i use 4x to watch a video and silent parts kicks in it probably shouldnt touch the playback speed right?

I think we probably should implement not touching the playback speed if its above 3x

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc I used the user setting values for the fast and normal playback speed. I tested multiple values with various videos and it worked fine even with 10x forward speed. So in your case it's at 10x speed because this should also be your "Max Video Playback Rate"?
image

Do you have problems with speeds above 3x ? I can't unmute in your provided video examples. Are they recorded with sound?
image

efb4f5ff-1298-471a-8973-3d47447115dc:
@some-git-user 

> Do you have problems with speeds above 3x ? I can't unmute in your provided video examples. Are they recorded with sound?

The arent recorded with sound but i think this case doesnt apply anymore because of the change you explained

> I used the user setting values for the fast and normal playback speed. I tested multiple values with various videos and it worked fine even with 10x forward speed. So in your case it's at 10x speed because this should also be your "Max Video Playback Rate"?

Yes it is 10x. I will reevaluate!

efb4f5ff-1298-471a-8973-3d47447115dc:
My Max video playback rate is set to 10x in the settings. Within the first 10sec of this video https://youtu.be/gT0rd69e5fo it will skip some silence but it also cuts of 2 words that have been said

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc I think the 10x speed is too aggressive in some cases. The default 3x should be enough in most cases?

efb4f5ff-1298-471a-8973-3d47447115dc:
i think 3x should be enough but i would like to hear from @PikachuEXE and @absidue before we decide on the right speed

PikachuEXE:
I found it sound too rushy with it enabled, and I am not sure if it's better or worse to add N seconds before/after the speed up
e.g. https://youtu.be/l8h4QgtNL0M
And I am not sure if it also skip through something (but I only tested it briefly so could be wrong)
On the speed I think 3x is fine (the gap/skip too much thing is bigger and once solved the speed while actual silence is not that important
Need longer term testing

src/constants.js
```diff
@@ -246,13 +248,20 @@ const MIXED_SEARCH_HISTORY_ENTRIES_DISPLAY_LIMIT = 4
 // Displayed on the about page and used in the main.js file to only allow bitcoin URLs with this wallet address to be opened
 const ABOUT_BITCOIN_ADDRESS = '1Lih7Ho5gnxb1CwPD4o59ss78pwo2T91eS'
 
+const SilenceSkip = {
+  SILENCE_DETECTION_MULTIPLIER: 4, // Multiplier for silence detection. Higher = more sensitive
+  MIN_SILENCE_DURATION: 150, // Min silence duration in ms. Higher for longer silence before skipping. Lower for faster reaction.
+  MIN_SOUND_DURATION: 5, // Min sound duration in ms. Higher to avoid false positives for short sounds.
```
PikachuEXE: FT got values in ms and s, no way to tell which one
```suggestion
  MIN_SILENCE_DURATION_MS: 150, // Min silence duration in ms. Higher for longer silence before skipping. Lower for faster reaction.
  MIN_SOUND_DURATION_MS: 5, // Min sound duration in ms. Higher to avoid false positives for short sounds.
```

efb4f5ff-1298-471a-8973-3d47447115dc:
3x seemed fine to me before it changed so maybe we should go back to that? Or are there other more logical speeds for this?

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc would it be a problem,to let it be configurable?

efb4f5ff-1298-471a-8973-3d47447115dc:
@some-git-user i prefer to not make this feature configurable. I prefer to go with good defaults instead of creating yet another slider/toggle in the settings

efb4f5ff-1298-471a-8973-3d47447115dc:
@some-git-user any issues you're facing we can help with?

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc I hadn't much time this month. I rechecked your comment about the "good defaults". Currently the playback speeds are related to the "Default Playback Rate" and "Max Video Playback Rate" settings. Removing this relation and set the skip silence to e.g. 3x will break the logic when the "Default Playback Rate" is set to like 4x speed. This would lead again to the issue, that the skip silence speed could be slower than the regular "Default Playback Rate". Maybe I'm wrong, but letting the user set the speeds, like it's in the current implementation, would be a good default for most users?

efb4f5ff-1298-471a-8973-3d47447115dc:
@some-git-user sorry for the delay.

After internal discussion we decided that its best to go for a good default instead of making it configurable. So what it should do is apply that number when there is a silent part and also only apply it if that speed is above the current speed.

Reasons for this decision are:
* We want to minimize overskipping as much as possible and prevent any performance issues, as higher playback rates demand more bandwidth and processing power.
* I also think that if your current speed is set to something extreme, like 8x, you're clearly prioritizing speed over skipping a few silent parts. In such "mad man" cases, I'd argue that pushing it to 10x results in diminishing returns.

3x felt good but maybe we can fiddle around with it some more to see if a slightly higher speed is possible

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc I updated the test instructions to align with the new behavior

efb4f5ff-1298-471a-8973-3d47447115dc:
PR LGTM but i need some feedback from @absidue and @PikachuEXE to be sure about the following points

- What do we think about the default skip speed? See previous reasoning for this speed https://github.com/FreeTubeApp/FreeTube/pull/7797#issuecomment-3718168682
- The Skip Silence button will be set to off when the user goes above 3x. Would it make sense to hide it in this case and show when speed is below 3x?
- Should Skip by Default toggle in settings be locked/visually greyed out when default playback speed is above 3x?

PikachuEXE:
1. 3x sounds fine unless proven otherwise
2. Just hide it
3. Yes and not sure if we can show users why it's locked

efb4f5ff-1298-471a-8973-3d47447115dc:
> 1. 3x sounds fine unless proven otherwise
> 
>     2. Just hide it
> 
>     3. Yes and not sure if we can show users why it's locked

@some-git-user lets go ahead and at least implement the first two suggestions. What are your thoughts on 3?

some-git-user:
@efb4f5ff-1298-471a-8973-3d47447115dc can't we keep the current button logic and use the already implemented `showValueChange` function to explain the behavior?
I think it could be confusing if the button visibility changes based on the current playback speed?
I attached a short demonstration of the `showValueChange`  implementation:

https://github.com/user-attachments/assets/7fa9f5f5-ee01-4e25-878f-002da87583ed

efb4f5ff-1298-471a-8973-3d47447115dc:
@some-git-user 
We would like to thank you for the PR and for the effort put into working within the current limitations of our player.
As you know the ideal behavior for this feature would be detecting silent sections and skipping directly to the point where audio resumes.
While this is the best implementation given the player limitations it is not the behavior we ultimately want.
For that reason we do not think it makes sense to merge this. If the player gains the capability to support real autoskip in the future we would be happy to revisit the idea then.
We would again like to emphasize that we appreciate the work that went into this and the attempt to push the feature forward.

PR #6915, opened 2025-02-24
Added hide to tray on minimize option

Devenor:
# Added hide to tray on minimize option

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #59
## Description

Adds optional hide to system tray behaviour on  minimizing the main window. The settings have an option to enable or disable it in the general section. The tray icon has a small menu for closing the app or opening the window again, and it can be opened by clicking on the icon too.

## Screenshots 

**Screenshot is outdated, only hide to tray on minimize exists now.**
image

image

## Testing 

- Go to Settings -> General and enable Minimize to system tray.
- When minimizing a window, the window will hide and a tray icon will appear.
- [Only on Windows] If there's only one window in the tray, left clicking the tray will open the window and the tray will disappear.
- Right clicking the tray will show a menu with an option to quit the app, and one submenu for each minimized window.
- Clicking the submenu will show one option to open the window and another to close it. When clicking an option, the submenu will disappear.
- If there are no more windows minimized, on Windows the tray will disappear, and on Linux it will remain with just the quit option.
- When a window is shown from the tray, it will retain its original size and position.
- When disabling the setting, if there are minimized windows they'll appear minimized in the taskbar.
- This setting is not avaliable for MacOS.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** v0.23.2 Beta

PR #7856, opened 2025-08-16
Refactor/cleanup tray related code after PR merged

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/pull/6915

## Description

Cleanup per https://github.com/FreeTubeApp/FreeTube/pull/6915#pullrequestreview-3125848598
Addressed stuff:
- Commented code
- Using `.length` as boolean -> `.length === 0 / > 0`
- If (!something) else -> if (something) else
- Move tray related code in main process to be behind `process.platform !== 'darwin'` so they should be removed on build time

## Screenshots 

Nope

## Testing

Same as https://github.com/FreeTubeApp/FreeTube/pull/6915

- Update `_scripts/webpack.main.config.js` from `'process.platform': `'${process.platform}'`,` to `'process.platform': '\'darwin\'',` (replace `darwin with other values later
- `yarn run pack && npx prettier@2.8.8 --write --no-config dist/main.js`
- open `dist/main.js`
- Check `setIgnoreDoubleClickEvents` absent with `darwin`, present with other values

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -274,7 +274,7 @@ function runApp() {
   let mainWindow
   let startupUrl
   let tray = null
-  let trayOnMinimize
+  let trayOnMinimize = false
   let trayWindows = []
   const trayMaximizedWindows = {}
 
@@ -333,7 +333,7 @@ function runApp() {
         if (!openDeepLinksInNewWindow) {
           // Just focus the main window (instead of starting a new instance)
           if (mainWindow.isMinimized()) {
-            if (trayOnMinimize) {
+            if (process.platform !== 'darwin' && trayOnMinimize) {
               trayClick(mainWindow)
             } else {
               mainWindow.restore()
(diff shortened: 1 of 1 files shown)


absidue:
> Move tray related code in main process to be behind process.platform !== 'darwin' so they should be removed on build time (not idea how to test

The webpack config for the main file sets the value of `process.platform` to `process.platform`, so of your current machine/build machine, but you can also hardcode it to `'darwin'`, `'win32'` or `'linux'` to see what the build output looks like for other OS'. To make the build output more readable I use prettier 2 (3 and later don't seem to reformat files for some reason) as it's aggressive reformatting works quite well on mimified files e.g. `npx prettier@2.8.8 --write --no-config dist/main.js`.
As for navigating around the file faster, picking some things that won't change between input and output such as strings or methods provided by the runtime (electron, node, browser) e.g. `isMinimized` and searching for them in the output works quite well.

Those are just the steps I take when I want to look at the output, but if you find some better ways to do it, I am all ears :).

PikachuEXE:
Updated testing section and locally tested (from previous macOS only testing

kommunarr (approved):

Issue #6863, opened 2025-02-21
[Feature Request]: AppImageUpdate support


TheAssassin:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

An AppImage is provided on the release page, but updates need to be downloaded manually. With [AppImageUpdate](https://github.com/AppImageCommunity/AppImageUpdate), the AppImages could be updated conveniently and efficiently.

### Proposed Solution

To support AppImageUpdate, you would need to pass `-u <updateinformation>` to appimagetool during build time. Please see https://docs.appimage.org/packaging-guide/optional/updates.html#using-appimagetool for more information. appimagetool then also generates the auxiliary `.zsync` file which you need to upload along with the AppImage. 

The following update information should work if you stop marking your releases as prereleases (or make an actual stable release at some point):

```
gh-releases-zsync|FreeTubeApp|FreeTube|latest|freetube-*x86_64.AppImage.zsync
```

### Alternatives Considered

You could embed AppImageUpdate to support self-updates. But that also 

### Issue Labels

ease of use improvement

absidue:
From what you are saying this would only be possible if we stopped doing pre-releases, so for the foreseeable future this feature request is not plausible.

TheAssassin:
No, not necessarily. We could perhaps support prereleases in the update information in the AppImage spec if you intend to keep the prereleases. Please don't hesitate to open an issue.

absidue:
As it sounds like you are one of the people that work on AppImageUpdate and as you are the one asking for AppImageUpdate support, may I suggest that you should be the one to open an issue on your side for the necessary changes. You clearly know more about it and are more invested in it, than we are, so it seems like the more logical option.

Once it would actually be technically possible for us to add AppImageUpdate support, without having to claim that pre-releases are not actually pre-releases, you are welcome to revive this issue thread.

TheAssassin:
Opened PR #8153.

eMPee584:
Why did you delete my comment though? @absidue 
That is really unusual as I did not post anything offensive, I was just posting a serious question.

absidue:
[...]

PR #7951, opened 2025-08-27
Make appimage updateable if project sets up `latest` tag [UNTESTED]

eMPee584:
Seems this is all it needs, according to some quick research..
https://github.com/search?q=appimage+updateable&type=pullrequests

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
This adds a line to the build and release gh workflows for supporting appimage updates.

## Testing
It is referring to a `latest` tag that doesn't exist yet, so that would have to be created.


absidue:
As mentioned here, by the maintainer of AppImageUpdate, this won't work as AppImageUpdate only supports stable GitHub releases and FreeTube only has prereleases: https://github.com/FreeTubeApp/FreeTube/issues/6863

PR #7962, opened 2025-08-30
Skip unnecessary shorts subscription cache writes

absidue:
## Pull Request Type

- [x] Performance improvement

## Description

On the channel page we don't have published dates for shorts, we can only update the metadata of shorts that are already in the cache (e.g. title changes or higher view counts). At the moment we write to the database every time, as long as you are subscribed to the channel, even if you don't have a shorts cache for that channel (we write an empty array) or none of the shorts in the cache were in the update payload (we write the existing cache entry again). This pull request corrects that so we only write to the database when there is already a cache entry and at least one shorts in the update payload matches one of the shorts in the cache. This will make the biggest difference for people that don't use the shorts subscription tab but have the shorts tab visible on channel pages.

## Desktop

- **OS:** Windows
- **OS Version:** 10

### DIFF
--- a/src/datastores/handlers/base.js
+++ b/src/datastores/handlers/base.js
@@ -308,16 +308,17 @@ class SubscriptionCache {
     )
   }
 
-  static updateShortsWithChannelPageShortsByChannelId(channelId, entries) {
-    return db.subscriptionCache.findOneAsync({ _id: channelId }, { shorts: 1 }).then((doc) => {
-      if (doc == null) { return }
+  static async updateShortsWithChannelPageShortsByChannelId(channelId, entries) {
+    const doc = await db.subscriptionCache.findOneAsync({ _id: channelId }, { shorts: 1 })
 
-      const shorts = doc.shorts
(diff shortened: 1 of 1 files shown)


src/datastores/handlers/base.js
```diff
@@ -308,17 +308,18 @@ class SubscriptionCache {
...
 
-  static updateShortsWithChannelPageShortsByChannelId(channelId, entries) {
-    return db.subscriptionCache.findOneAsync({ _id: channelId }, { shorts: 1 }).then((doc) => {
-      if (doc == null) { return }
+  static async updateShortsWithChannelPageShortsByChannelId(channelId, entries) {
+    const doc = await db.subscriptionCache.findOneAsync({ _id: channelId }, { shorts: 1 })
 
-      const shorts = doc.shorts
-      const cacheShorts = Array.isArray(shorts) ? shorts : []
+    if (Array.isArray(doc?.shorts)) {
```
PikachuEXE: Have early return here instead of a big nested if?

PikachuEXE (approved):

PR #7970, opened 2025-08-31
Transform in-app URLs to YouTube URLs when dragging them

absidue:
## Pull Request Type

- [x] Feature Implementation

## Description

Web browsers, including Electron make links draggable by default, to allow users to easily drag them into other applications or into text boxes within the same application or website. In FreeTube this works great for external URLs but if you tried doing that for an in-app URL, you would still copy the in-app URL. This pull request fixes that by intercepting the drag operations on in-app URLs and transforming them to YouTube ones.

Please note that this does not add support for dragging things that were not already draggable, it also only changes the URLs in the dragged content, it does not add or remove any other information. Additionally the link preview during the drag operation will still mention the in-app URL, as the browser generates that preview before we swap the URL but it will output the correct URL when dropped.

## Testing

Try dragging in-app URLs such as video and channel links into FreeTube's search bar or other applications such as your web browser. Please keep the things mentioned in paragraph two of the description in mind while testing.

Please test both dev mode and production builds, as the in-app URLs differ (http://localhost:9080/ and app://bundle/index.html respectively).

Test build: https://github.com/absidue/FreeTube/actions/runs/17360383850

## Desktop

- **OS:** Windows
- **OS Version:** 10


PikachuEXE (approved):
Both dev & test build tested

kommunarr (approved):

PR #7995, opened 2025-09-07
Added New Window option to Tray

Devenor:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Adds #7049 functionality to the Tray

## Description

Tray menu option enabling to open a new Freetube instance.

## Screenshots 

imagen

## Testing

When clicking the New Window option, a new window opens. Other windows hidden in the tray must remain there.

## Desktop

- **OS: Windows, Linux**
- **OS Version: Windows 10, Ubuntu 22.04.3 LTS**
- **FreeTube version: 0.23.8 Beta**

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -717,6 +717,8 @@ function runApp() {
           window.show()
         }
       }
+
+      if (trayWindows.length === BrowserWindow.getAllWindows().length) { mainWindow = window }
     } else if (trayWindows.length > 0) {
       window.close()
     }
@@ -1008,6 +1010,14 @@ function runApp() {
         if (trayOnMinimize) {
           newWindow.hide()
           manageTray(newWindow)
+
+          if (newWindow === mainWindow) {
+            // A timer is needed because getFocusedWindow doesn't update until the minimize event ends
+            setTimeout(() => {
+              const newMainWindow = BrowserWindow.getFocusedWindow() || BrowserWindow.getAllWindows().find(window => window.isVisible())
+              if (newMainWindow) { mainWindow = newMainWindow }
+            }, 100)
+          }
         }
       })
 


src/main/index.js
```diff
@@ -752,28 +752,37 @@ function runApp() {
...
     tray.setContextMenu(menu)
   }
 
+  function defaultTrayMenu() {
+    return [
+      {
+        label: 'New Window',
+        click: () => createWindow({
+          showWindowNow: true,
+          replaceMainWindow: true
```
absidue: This should either be `false` all the time or only `true` if the window that is currently set as the main window is minimised to the tray.

Devenor:
Done, but I was thinking that when the main window is hidden to the tray and there are other visible windows, none of them become the main window. Should I change that too?

absidue:
If it is easy to do, sure, if not then it would be better to leave it for a separate pull request.

Devenor:
I've added the following:
- When all windows are hidden, the first one to be shown becomes the new main window.
- If the main window is hidden, the one that has the focus becomes the new main window.
- If no window has the focus (for example if they are behind another app), the last opened window becomes the new main window.

Seems to work fine. My only concern is about the timer in Linux, it works in my distro, but we might need to increase it a bit if it's not enough in others.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @Devenor apologies for the delay. Past weeks were a bit hectic because we needed to adapt to YouTube's changes to get playback to unbreak.

Anyway i tested the PR and it looks good.

> I've added the following:
> - When all windows are hidden, the first one to be shown becomes the new main window.
> - If the main window is hidden, the one that has the focus becomes the new main window.
> - If no window has the focus (for example if they are behind another app), the last opened window becomes the new main window.
> 
> Seems to work fine. My only concern is about the timer in Linux, it works in my distro, but we might need to increase it a bit if it's not enough in others.

Could you maybe elaborate how i should test these changes

Devenor:
Hey @efb4f5ff-1298-471a-8973-3d47447115dc , no worries, I imagined you'd be busy!

If you've tested the PR in all your distros and it works, it should be fine as it is. I just wanted to make sure because I had to add a small delay to check if another window has the focus after minimizing the main one, and we had some issues when using a timer on Linux in my first PR (the occasional flicker, if you still remember)

In any case, if the timer were too short, the only thing that could happen is that the new main window would become the first visible one found, which may or may not be the one that has the focus.

absidue (approved):

PR #8033, opened 2025-09-14
Remove n argument in sed

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Bugfix

## Related issue

https://github.com/FreeTubeApp/FreeTubeApp.io/commit/fc8429129da74069e8132b0641679270bd049440#diff-51d6bfb8792aec023c93824fd8db464a816a4971f89acc4360e5d5cdbe33d95f

## Description

The `-in` part of the sed command is likely incorrect but i couldn't test this to verify it. The `-i` option is used for in-place editing, and it should be followed by the file extension you want to create (e.g., -i.bak to create a .bak file as a backup).
`-in` could be interpreted as trying to create a backup with the .n extension, which might be why we're seeing a file like index.phpn.

### DIFF
--- a/.github/workflows/updateSite.yml
+++ b/.github/workflows/updateSite.yml
@@ -52,7 +52,7 @@ jobs:
         git checkout master
     - name: Update index.php
       run: |
-        sed -in 's/${{ steps.previous.outputs.result  }}/${{ steps.current.outputs.result }}/' src/index.php
+        sed -i 's/${{ steps.previous.outputs.result  }}/${{ steps.current.outputs.result }}/' src/index.php
     - name: Commit Files
       uses: stefanzweifel/git-auto-commit-action@v6
       with:


.github/workflows/updateSite.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         git checkout master
     - name: Update index.php
       run: |
-        sed -in 's/${{ steps.previous.outputs.result  }}/${{ steps.current.outputs.result }}/' src/index.php
+        sed -i 's/${{ steps.previous.outputs.result  }}/${{ steps.current.outputs.result }}/' src/index.php
```
PikachuEXE: Need to replace all
2025-09-14 20_45_01-prox1 - Proxmox Console - Vivaldi

```suggestion
        sed -i 's/${{ steps.previous.outputs.result  }}/${{ steps.current.outputs.result }}/g' src/index.php
```

PikachuEXE:
To learn more I found https://linuxize.com/post/how-to-use-sed-to-find-and-replace-string-in-files/

ChunkyProgrammer (approved):

PR #8056, opened 2025-09-19
Downloads feature with muxing audio and video to single file

bardcrab:
## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
I've implemented downloads manager feature with muxing audio and video to single file. It uses the Mediabunny library for muxing. The video download button dropdown is extended to allow multiple selection of available download items. The "Download" item at the end of the dropdown open file saving dialog. As for output format there are MP4 and MKV available. Then download get placed to the "Downloads" list where downloading progress is displayed. When only one item is selected it will be handled in usual way without remuxing.


absidue (changes requested):
Thank you for your pull request, unfortunately there are currently quite a few issues with it that need to be resolved:
- This currently depends on a branch in your fork of mediabunny. Relying on a dynamic target such as a branch in some unknown person's GitHub repo (an exact commit would have been better), is a security issue. Additionally based on the discussions in the main mediabunny repository, it looks like they are planning to take a different approach for merging multiple files together than you did, so this code will need refactoring at a later point in time to fit what upstream is doing and based our experience with external contributors not coming back to clean up their own code, that burden would likely land on us. Therefore it is likely better to wait until the approach is finalised in the main mediabunny repository and use that directly.
- You have included a lot of commented out code in your pull request, while yes commenting out code is fine during local development, it should be removed once you are ready to submit your changes as a pull request as it bloats the changes and makes it more difficult to review
- This completely ignores all of the existing download settings.
- It also does Electron only stuff but doesn't guard is properly with the IS_ELECTRON build flag, so it will cause errors outside of Electron such as in FreeTubeAndroid.
- New Vue components need to be written in the composition API style, notice how most of the components are already in that style. 
- In various places you've invented your own code style rather than following what the surrounding code is doing e.g. `is_variant` should be `isVariant`.

bardcrab:
Thanks for review!

- As for the MediaBunny library I plan to update the code in FreeTube once the library will get ready. I just want this feature become available in FreeTube quicker.
- About the commented out code - I just tried to keep some code for later in such lame way. I will clean-up the code and update the PR.
- The download settings in FreeTube is specifying how the file download will be handled. But in this case I performing downloading of several streams and I already need them in memory, and there are no need to store them separately on disk. I don't see how I can apply the app download settings here.
- May I ask you to point me what Electron specific things I need to handle accordingly?
- I will try to re-shape my custom components to make them in composition style.
- I was hoping that linter is enough. I will refactor the code.

absidue:
We have actually decided that we will be removing the download functionality entirely, it will break with SABR so would require a complete rewrite which we don't want to do (the download functionality has been unmaintained for a long time now, so has issues such as getting itself throttled), additionally dedicated downloaders such as yt-dlp are much more powerful and we have been telling users to use those instead of FreeTube's one for a while now.

I'll close this pull request as it is no longer required.

bardcrab:
Got it.
Actually good point about the SABR. I haven't considered it before.
But I think it is worth to have a download feature in FreeTube despite that other downloaders exists.
Maybe I or someone else will implement SABR downloading and then open new PR - who knows :)

absidue:
For the moment it's out of scope for FreeTube so even if you work on a PR we would probably not merge it. If you want to download things use a dedicated downloader like yt-dlp.

PR #8109, opened 2025-10-11
Update Sort by strings

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Refining strings

## Description

Since the introduction of these particular strings i had the feeling that some of them felt off and could be refined but i never really knew how. I recently had access to a YT account and saw how YT handled the wording of those strings which triggered me to adapt to that style of wording. We already use that on a few places which made it an easier choice to change these strings.

Also the word "first" is removed because that is redundant

## Additional context

There are more strings that can be clarified better. More string refining PR's are coming.

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -200,14 +200,14 @@ User Playlists:
     NameAscending: 'A-Z'
     NameDescending: 'Z-A'
 
-    LatestCreatedFirst: 'Recently Created'
-    EarliestCreatedFirst: 'Earliest Created'
+    LatestCreatedFirst: 'Date Created (Newest)'
+    EarliestCreatedFirst: 'Date Created (Oldest)'
 
-    LatestUpdatedFirst: 'Recently Updated'
-    EarliestUpdatedFirst: 'Earliest Updated'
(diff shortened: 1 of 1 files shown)


static/locales/en-US.yaml
```diff
@@ -951,16 +951,16 @@ Playlist:
...
   Last Updated On: Last Updated On
   Sort By:
-    DateAddedNewest: Latest added first
-    DateAddedOldest: Earliest added first
-    PublishedNewest: Latest published first
-    PublishedOldest: Earliest published first
+    DateAddedNewest: Date added (Newest)
+    DateAddedOldest: Date added (Oldest)
+    PublishedNewest: Date published (Newest)
+    PublishedOldest: Date published (Oldest)
```
PikachuEXE, line 957: Wondering why only these are not capitalized

efb4f5ff-1298-471a-8973-3d47447115dc: I will also do a followup PR for https://github.com/FreeTubeApp/FreeTube/issues/5000

so this is doing that already but i forgot to adjust the others. I guess it doesnt really matter because im going to adjust all strings pretty soon

PR #8119, opened 2025-10-12
Hotfix release - v0.23.10

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Release

## Testing

- Test if all functionality added works correctly

### DIFF
--- a/_scripts/sigFrameConfig.js
+++ b/_scripts/sigFrameConfig.js
@@ -9,5 +9,5 @@ const script = rawScript
 
 module.exports.sigFrameTemplateParameters = {
   sigFrameSrc: `data:text/html,${encodeURIComponent(`<!doctype html><script>${script}</script>`)}`,
-  sigFrameCspHash: `sha512-${hash('sha512', script, 'base64')}`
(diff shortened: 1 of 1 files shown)


_scripts/sigFrameConfig.js
```diff
@@ -0,0 +1,15 @@
+const { hash } = require('crypto')
+const { join } = require('path')
+const { readFileSync } = require('fs')
+
+const path = join(__dirname, '../src/renderer/sigFrameScript.js')
+const rawScript = readFileSync(path, 'utf8')
+
+const script = process.env.NODE_ENV === 'development'
+  ? rawScript
+  : require('terser').minify_sync({ [path]: rawScript }).code
```
absidue: Looks like `minify_sync` was only added in later versions of terser, so for the hotfix lets just use the unminified source, it's not very big anyway.
```suggestion
const script = rawScript
```

efb4f5ff-1298-471a-8973-3d47447115dc:
Received upon startup, videos wont load

FreeTube_uK5mTh5OdE

PR #8131, opened 2025-10-14
Fix application icon on Wayland

RevengeRip:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
On Wayland application id must match installed `.desktop` entry, otherwise compositor won't show application icon. 
## Screenshots 
**Before:** 
freetube-before
**After:**
freetube-after

## Testing

1. Run FreeTube with next command `freetube --ozone-platform=wayland` on KDE Wayland session
2. Look at the application icon

## Desktop
- **OS:** Arch Linux
- **OS Version:** Rolling Release
- **FreeTube version:** 0.23.10-beta

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "freetube",
   "productName": "FreeTube",
+  "desktopName": "freetube",
   "description": "A private YouTube client",
   "version": "0.23.11",
   "license": "AGPL-3.0-or-later",


absidue:
> In order for this to work desktop entry need to have corresponding name io.freetubeapp.freetube.desktop

As the `.desktop` file in the official FreeTube builds is called `FreeTube.desktop` that means this pull request won't actually work.

absidue:
I think you may have misunderstood my comment, you need to change the `desktopName` field to match the current desktop file name, renaming the desktop file to match the name you invented for the `desktopName` is the wrong approach.

absidue:
Closing as requested by the author, as they can't close themselves while the repo is locked.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Doenst seem to work


https://github.com/user-attachments/assets/42c7ebaf-d8db-4909-8c7e-41aa59c63929

RevengeRip:
I'm not sure if it should work with AppImages since they're not installing their .desktop entry in the system
I think that was one of the reason for creating xdg-toplevel-icon Wayland protocol: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/269

This PR should've solved the issue for system-wide installed FreeTube instance, where desktop entry is installed. For the AppImage support xdg-toplevel-icon support should be merged in Electron first: https://github.com/electron/electron/issues/45638

efb4f5ff-1298-471a-8973-3d47447115dc:
Used a nightly build and showed FT icon. Used a test build which included the change from this PR and it showed the wayland icon.

RevengeRip:
I've noticed that in nightly build desktop entry name changed from `freetube.desktop` to `FreeTube.desktop` compared to the latest release, which makes things work as intended ootb and there's no need for this PR (it will actually break things now). But I heard concerns that changing desktop entry name may break things for some users. Was this change intentional?
изображение

absidue:
We didn't change the desktop name on our end, the change is likely due to the master branch being far behind the development branch, so is still using much older versions of Electron and electron-builder, as we haven't done a feature release since January (only bug fix ones).

When you make a pull request to FreeTube always compare with the development branch or the latest nightly build, never compare against the master branch/a release version as those will always be behind.

absidue:
As you are not the first person that has done a pull request to fix a bug that they saw in a release version without bothering to test the development branch first to see if their change is even necessary, is there something you would suggest we could do from our side to prevent other potential contributors from making the same mistake?

RevengeRip:
Nah, that was my mistake, I should've read contribution guide properly. In my case most Arch Linux users are getting their software from AUR from either https://aur.archlinux.org/packages/freetube-git or https://aur.archlinux.org/packages/freetube so it's very unusual to have a package for Arch Linux, especially when it has wrong extension (pacman packages are `.pkg.tar.zst` archives, `.pacman` extension was clearly made up by electron devs). Even If I would use `freetube-git` I would still experience this issue since unlike upstream `.pacman` package AUR packages don't ship Electron themselves but build against system electron which is currently outdated compared to development branch.

PR #8103, opened 2025-10-07
Update next video countdown toast to be refreshed per second instead of using workaround

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Extracted from https://github.com/FreeTubeApp/FreeTube/pull/8047 (except the update of `Playing Next Video Interval` part
If this is merged before https://github.com/FreeTubeApp/FreeTube/pull/8094, that PR requires a minor fix, if reversed, this PR should be updated
Closes https://github.com/FreeTubeApp/FreeTube/issues/2340

## Description

Count down message (e.g. Playing next video in N seconds) currently implement as series of toasts shown 700ms (per 1s)
This PR just make the toast show for N seconds and update message every second (hardcoded to be 1s as no use case requires otherwise)

## Screenshots 

Lazy, test yourself :)

## Testing

- Use magic to spawn a playlist with >= 2 videos
- Play non-last one, enable autoplay if necessary, skip to near the end
- Watch countdown works and clicking on it still cancels

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

PR #8168, opened 2025-10-21
Fix toast with timeout 0 will be displayed for default 3s

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Reported in https://github.com/FreeTubeApp/FreeTube/pull/8103#issuecomment-3422567662

## Description

Toast shown with default timeout 3s when 0 is passed in
This PR just make `showToast` with `0` timeout passed in to do nothing

## Testing

- Update player autoplay countdown to 0s
- Play any playlist
- Ensure no countdown seen

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/FtToast/FtToast.vue
+++ b/src/renderer/components/FtToast/FtToast.vue
@@ -41,9 +41,6 @@ const toasts = shallowReactive([])
  * @param {CustomEvent<{ message: string | (({elapsedMs: number, remainingMs: number}) => string), time: number | null, action: Function | null, abortSignal: AbortSignal | null }>} event
  */
 function open({ detail: { message, time, action, abortSignal } }) {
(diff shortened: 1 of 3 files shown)


absidue (changes requested):
Could you please fix the actual issue, which is the unnecessary showToast call in Watch.js, instead of adding a workaround.

PikachuEXE:
Updated, added a warning in `showToast` in case there are new calls with `0` time during development

PikachuEXE:
The warning hopefully will help make developers/reviewers aware of the 0 value passed in during testing
So the comment is wrong

ChunkyProgrammer (approved):

Issue #8157, opened 2025-10-19
[Feature Request]: Remove input limit of 3 on title blacklist feature


An-anonymous-coder:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Problem Description

There are thousands (maybe millions) of videos that spam hashtags in the title. Quite simply, it's annoying and distracting. There's already a setting to "Display Titles Without Excessive Capitali[z]ation And Punctuation", but no option to fix excessive hashtags. The blacklist feature cannot be used to fix this because of an arbitrary minimum of 3 characters. 

### Proposed Solution

Create a setting to "Display Titles Without Hashtags" which will automatically remove all hashtags from titles.

### Alternatives Considered

1. Remove the arbitrary limit on the blacklist feature to include single character filters like the hashtag symbol.
2. Channels that spam hashtags in the title often post low quality content, even on videos without hashtags. Because of this, it would be nice to have a setting to automatically blacklist any channels that have hashtags in a video title. This check can be done every time a page is loaded.

### Issue Labels

ease of use improvement, new optional setting, improvement to existing feature, new feature

PR #8174, opened 2025-10-22
Change containing text input limit

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Feature Implementation?

## Related issue

Closes https://github.com/FreeTubeApp/FreeTube/issues/8157
https://github.com/FreeTubeApp/FreeTube/pull/4202

## Description

Input limit of 3 was set in the PR that introduced this feature. It was an arbitrary limit and now someone raised an issue because their usecase is to block one specific character. This PR changes the arbitrary limit to 1.

## Screenshots 

Before:
FreeTube_IyG1SyfkYZ

After:
VirtualBoxVM_w3Yc9Z2UOD

## Testing

1. Go to Settings -> Distraction Free -> General -> Hide Videos and Playlists containing text
2. Input anything below the below/above the removed limit
3. Check if content is still being blocked

## Desktop

- **OS:  Windows 11**
- **OS Version: 24H2**

## Additional context

This limit was introduced in these specific commits:
- https://github.com/FreeTubeApp/FreeTube/pull/4202/commits/1b4f4aac4081b199644fbdaef462d6f4f16fe24b
- https://github.com/FreeTubeApp/FreeTube/pull/4202/commits/8c978ef80a720b8c2054836b8aaddecf63fceaa5

### DIFF
--- a/src/renderer/components/FtInputTags/FtInputTags.vue
+++ b/src/renderer/components/FtInputTags/FtInputTags.vue
@@ -155,7 +155,7 @@ async function updateTags(text) {
   const trimmedText = text.trim()
 
   if (props.minInputLength > trimmedText.length) {
-    showToast(t('Trimmed input must be at least N character long', { length: props.minInputLength }, props.minInputLength))
+    showToast(t('Trimmed input must be at least N characters long', { length: props.minInputLength }, props.minInputLength))
     return
   }
 


absidue (commented):
I haven't tested it yet, but based on the code, you could type in multiple spaces and click save/press <kbd>enter</kbd>, it would get saved as an empty string and then all videos would be hidden. So we should probably still check that that trimmed input is at least 1 character.

absidue:
Just to be clear, I still think we should trim the input as we do now, but we shouldn't allow the user to add empty filters.

src/renderer/components/FtInputTags/FtInputTags.vue
PikachuEXE, line 1: Maybe code in this file shouldn't be removed (and locale entries) since the only limit to be removed is `3` not `1`?

src/renderer/components/FtInputTags/FtInputTags.vue
```diff
@@ -155,7 +155,7 @@ async function updateTags(text) {
   const trimmedText = text.trim()
 
   if (props.minInputLength > trimmedText.length) {
-    showToast(t('Trimmed input must be at least N characters long', { length: props.minInputLength }, props.minInputLength))
+    showToast(t('Trimmed input must be at least 1 character long', { length: props.minInputLength }, props.minInputLength))
```
PikachuEXE: I know value > 1 isn't used right now but is it meaningful to remove translation for value > 1...?
```suggestion
    showToast(t('Trimmed input must be at least N character long', { length: props.minInputLength }, props.minInputLength))
```

static/locales/en-US.yaml
```diff
@@ -1130,7 +1130,7 @@ Screenshot Success: Saved screenshot
 Screenshot Error: Screenshot failed. {error}
 Channel Hidden: '{channel} added to channel filter'
 Channel Unhidden: '{channel} removed from channel filter'
-Trimmed input must be at least N characters long: Trimmed input must be at least 1 character long | Trimmed input must be at least {length} characters long
+Trimmed input must be at least 1 character long: Trimmed input must be at least 1 character long
```
PikachuEXE:
```suggestion
Trimmed input must be at least N characters long: Trimmed input must be at least 1 character long | Trimmed input must be at least {length} characters long
```

src/renderer/components/FtInputTags/FtInputTags.vue
efb4f5ff-1298-471a-8973-3d47447115dc: I dont see why we should keep something that we wont use anymore so thats why i removed it

PikachuEXE: This is not removing the entry but replacing the entry (and limit the possible translatable value to 1)
I don't see a point in changing (not removing) an entry which is **already translated**

efb4f5ff-1298-471a-8973-3d47447115dc: im keeping this part which is already translated:
`Trimmed input must be at least 1 character long`

The only thing that will be changed is:
`| Trimmed input must be at least {length} characters long`

Which i feel like we wont need as i dont see us increasing the limit again

PikachuEXE: That's not what I am talking about
The key is already translated
https://github.com/search?q=repo%3AFreeTubeApp%2FFreeTube+%22Trimmed+input+must+be+at+least+N+characters+long%22&type=code
Changing the key means it has to be translated again? For what?

efb4f5ff-1298-471a-8973-3d47447115dc: ahhhh its about the key sorry i will make the changes

src/renderer/components/FtInputTags/FtInputTags.vue
```diff
@@ -155,7 +155,7 @@ async function updateTags(text) {
   const trimmedText = text.trim()
 
   if (props.minInputLength > trimmedText.length) {
-    showToast(t('Trimmed input must be at least N characters long', { length: props.minInputLength }, props.minInputLength))
+    showToast(t('Trimmed input must be at least N character long', { length: props.minInputLength }, props.minInputLength))
     return
```
efb4f5ff-1298-471a-8973-3d47447115dc:
```suggestion
    showToast(t('Trimmed input must be at least N characters long', { length: props.minInputLength }, props.minInputLength))
    return
```

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer @kommunarr

kommunarr (approved):

PR #8175, opened 2025-10-22
Remove Hide Channel from the kebab menu on the Subscriptions page

### LINKED ISSUE #7868: [Feature Request]: Remove `Hide Channel` from the kebab menu on the Subscriptions page
### Guidelines

### Problem Description

Mentioned in the Hide Videos from Channels tooltip in the General section of the Distractions Free Settings

> Enter a channel ID to hide all videos, playlists and the channel itself from appearing in search, trending, most popular and recommended.

The Hide channel button on the Subscription page serves no use to the user as it doesnt do anything there.

### Proposed Solution

Remove that option from the kebab menu

### Alternatives Considered

none

### Issue Labels

ease of use improvement

### Additional Information

Created this because user being confused in #7660

ozrendev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #7868

## Description

Removes the "Hide/Show Channel" option from menu on FtListVideo component when user is on the Subscriptions page

## Screenshots 

Before

After

## Testing

Navigate to Subscriptions page (including Video, Shorts, Live, Community tabs) and check for absence of "Hide/Show Channels" option
Navigate to any other place where a video/thumbnail is and see that the option is visible

### DIFF
--- a/src/renderer/components/ft-list-video/ft-list-video.js
+++ b/src/renderer/components/ft-list-video/ft-list-video.js
@@ -174,7 +174,7 @@ export default defineComponent({
     },
 
     inSubscriptions: function () {
-      return this.$route.name === 'subscriptions'
+      return this.$route.name === 'subscriptions' || this.$route.name === 'default'
     },
 
     inUserPlaylist: function () {


src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -327,22 +327,25 @@ export default defineComponent({
       if (this.channelId !== null) {
         const hiddenChannels = JSON.parse(this.$store.getters.getChannelsHidden)
         const channelShouldBeHidden = hiddenChannels.some(c => c.name === this.channelId)
+        const isSubscriptionsPage = window.location.href.endsWith('subscriptions')
```
absidue: Please check the current page the proper way. This file already has code for checking the history page, please copy that.

When an application uses a routing library you should always go through that that, but even if it doesn't checking a URL as a string with endswith is still wrong, as it doesn't actually check for the URL for that specific page, just that it contains the word subscriptions. I'm surprised this didn't get flagged by CodeQL, which has flagged all other bad URL validation in the past.

ozrendev: Sorry about that, I tried to look for the proper way to do this but couldn't figure it out. Thanks for your help, it made it much easier.

For these types of questions, what's the best way to ask before making a PR? Through the matrix room or to message you (or others) directly?

absidue: Reading the existing code is the best way to answer these kinds of questions, alternatively search on the Internet, Vue is quite popular and vue-router is the official routing library for Vue, so you will definitely find information about it.

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -328,21 +332,23 @@ export default defineComponent({
         const hiddenChannels = JSON.parse(this.$store.getters.getChannelsHidden)
         const channelShouldBeHidden = hiddenChannels.some(c => c.name === this.channelId)
 
-        options.push(
-          {
-            type: 'divider'
-          },
+        if (!this.inSubscriptions) {
```
absidue: Please move the check to the same if that checks that the channel ID is not null.

src/renderer/components/ft-list-video/ft-list-video.js
```diff
@@ -173,6 +173,10 @@ export default defineComponent({
       return this.$route.name === 'history'
     },
 
+    inSubscriptions: function () {
+      return this.$route.name === 'subscriptions'
```
absidue: Please also check for 'default', see:
https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/router/index.js#L23-L38

```suggestion
      return this.$route.name === 'subscriptions' || this.$route.name === 'default'
```

ozrendev:
Wouldn't this make sense to also remove the option inside each channel too? (for videos, shorts, live, etc tabs)

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm you could say the same thing for all other pages

Idk tbh. We could do if subscribed removed hide channel option everywhere

absidue:
> Wouldn't this make sense to also remove the option inside each channel too? (for videos, shorts, live, etc tabs) 
> 
> 
> 
> 

No, not unless we add the option somewhere else on the channel page. Visiting a specific channel to block it is a valid use case.

absidue (approved):

PR #8238, opened 2025-11-06
Specify window position in constructor for better wayland compatibility

absidue:
## Pull Request Type

- [x] Bugfix

## Description

According to the Electron documentation wayland doesn't allow apps to resize or move their own windows after they have been created, as we currently do just that, this pull request switches to passing the size and position in the BrowserWindow constructor instead. The recommendation of running on X11 isn't relevant for us, as we only need to set the initial size and position of the window, we don't need to move or resize it programatically after that.

https://www.electronjs.org/docs/latest/api/browser-window#platform-notices

## Testing

I haven't tested this on Linux but I did check that it still works as expected on Windows 11.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -930,10 +930,27 @@ function runApp() {
       return nativeTheme.shouldUseDarkColors ? '#212121' : '#f1f1f1'
     })
 
-    /**
-     * Initial window options
-     */
-    const commonBrowserWindowOptions = {
+    let savedBounds, savedMaximized, savedFullScreen
+
+    const boundsDoc = await baseHandlers.settings._findOne('bounds')
+    if (typeof boundsDoc?.value === 'object') {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@ChunkyProgrammer @kommunarr 👋

ghost:
I believe @MarmadileManteater is using a tiling window manager so, I think it'd be a good idea for them to approve/test this PR to make sure it works correctly.

absidue:
It is possible that any kind of positioning even through the constructor won't work on wayland, if that's the case there is nothing we can do about it on our side (you would need to get the wayland spec changed and the change implemented in enough places that Chromium/Electron would adopt it).

PR #8248, opened 2025-11-09
Validate sender in all IPC event handlers

absidue:
## Pull Request Type

- [x] Security improvement

## Description

At the moment we only validate that the sender of the IPC event was actually the FreeTube page in a few event handlers, this pull request adds the checks to the remaining ones.

## Testing

Test a few things that use IPC calls e.g. turning on and off the experimental HTTP cache or having multiple windows open and changing some data in one window and check that it is synced to the other window.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -351,9 +351,19 @@ function runApp() {
           replaceMainWindow: false,
           showWindowNow: true,
         })
-        ipcMain.once(IpcChannels.APP_READY, () => {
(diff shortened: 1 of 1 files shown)


absidue:
Yay, more Vue 3 migration issues, not caused by this PR.

absidue:
Updated the test case to something that actually works.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #8267, opened 2025-11-11
Remove upperCase on various places

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Sentence case prep

## Related issue

#5000 

## Description

In preparation for #5000 i looked around the code for stuff that wouldnt make sense anymore after the transition to sentence case. As that issue pointed out, transforming text is one of them.

- Since we introduced tabs across all our pages we've drawn inspiration from the Channel page on YouTube. For a while now YT is using lowercase for their Channel page tabs so this update reflects that change on the Subscription, Channel and Trending pages.
- The "Subscribe" button on YouTube has not used all caps for the past three years so i've updated it to match their current style.
- Video format selection labels no longer use all caps
- The "Upcoming" label has been updated to lowercase to match YouTube’s style, and the same change has been applied to the "Watched" label for consistency.

## Screenshots 

YT:
firefox_aA0rjyDXtD
firefox_Zyhwb1WMFy
firefox_mfpaX2gAXQ

FT after changes:
VirtualBoxVM_1W1Y7pGA98
VirtualBoxVM_DgnuRkrJvB
VirtualBoxVM_DJcZA9TRGn
VirtualBoxVM_cNuB7fWEap

## Testing

- Go to the Subscription, Trending and Channel pages and check that all the tabs arent using uppercase anymore
- Go to your Channels list and check that the Subscribe button isnt using uppercase anymore
- Find an Upcoming video and check that the upcoming label isnt using uppercase anymore
- Mark a video as Watched to see that the Watched label is the same as what i just wrote
- Go to a video and select the Video formats button. See that it uses sentence casing.

## Desktop

- **OS: Windows 11**
- **OS Version: 24H2**

### DIFF
--- a/src/renderer/scss-partials/_ft-list-item.scss
+++ b/src/renderer/scss-partials/_ft-list-item.scss
@@ -388,9 +388,14 @@ $watched-transition-duration: 0.5s;
   }
 }
 
+.videoWatched,
+.live,
+.upcoming {
+  user-select: none;
+}
+
 .live {
   text-transform: uppercase;
-  user-select: none;
 }
 
 // we use h3 for semantic reasons but don't want to keep the h3 style


src/renderer/scss-partials/_ft-list-item.scss
```diff
@@ -388,9 +388,7 @@ $watched-transition-duration: 0.5s;
   }
 }
 
-.videoWatched,
-.live,
-.upcoming {
+.live {
```
ghost, line 397: I think we want to keep the `user-select: none;` for videoWatched and upcoming?

efb4f5ff-1298-471a-8973-3d47447115dc: Good catch. Fixed now.

PikachuEXE (approved):

PR #8273, opened 2025-11-12
Sentence casing part 1

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Chore

## Related issue

#5000 

## Description

**This PR is drafted because i'd like to reach a consensus on the changes first before i change all the related files**

Notable changes:
https://github.com/efb4f5ff-1298-471a-8973-3d47447115dc/FreeTube/blob/5c7be48e038b60e32366ca591bcf8c1f65e351b9/static/locales/en-US.yaml#L4-L26 is made the same as browser and other Electron application menu bars

General rules of thumb im adhering to in all my PR's:
* Apply proper punctuation when using multiple sentences to both the key and value
* Remove punctuation when using one sentence (95% already does this so i just make it consistent)
* Apply proper casing to abbreviations if its being used already to both the key and value
* Apply proper casing when a specific setting or button is being named in a message (e.g. `If you want to import your subscriptions you can go to **Data settings** and select **Import subscriptions** or you can search for a channel and subscribe to them.`)

## Testing

Look around in the UI to check that sentence casing has been applied properly

## Desktop

- **OS: Windows 11**
- **OS Version: 24H2**

### DIFF
--- a/src/renderer/components/FtInput/FtInput.vue
+++ b/src/renderer/components/FtInput/FtInput.vue
@@ -32,8 +32,8 @@
       :class="{
         visible: inputDataPresent || showOptions
       }"
-      :aria-label="t('Search Bar.Clear Input')"
-      :title="t('Search Bar.Clear Input')"
+      :aria-label="t('Search bar.Clear input')"
+      :title="t('Search bar.Clear input')"
       @click="handleClearTextClick"
     >
       <FontAwesomeIcon
--- a/src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
+++ b/src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
@@ -296,7 +296,7 @@ const parsedViewCount = computed(() => {
     return null
   }
 
-  return t('Global.Counts.View Count', { count: formatNumber(props.viewCount) }, props.viewCount)
+  return t('Global.Counts.View count', { count: formatNumber(props.viewCount) }, props.viewCount)
 })
 
 const dateString = computed(() => {


static/locales/en-US.yaml
```diff
@@ -1,5 +1,5 @@
 # Put the name of your locale in the same language
-Locale Name: English (US)
+Locale name: English (US)
```
PikachuEXE, line 2: https://github.com/search?q=repo%3AFreeTubeApp%2FFreeTube+%22Locale+Name%22+language%3AJavaScript&type=code

This key is used but I see no point updating it

Reviewing other changes later

efb4f5ff-1298-471a-8973-3d47447115dc: Thats correct. I changed it pure for consistency. Lmk if you prefer me to revert that

PikachuEXE: https://github.com/FreeTubeApp/FreeTube/issues/5000 is about changing labels, keys can be changed if value updated but in this case there is no value change so it seems pointless

efb4f5ff-1298-471a-8973-3d47447115dc: Makes sense. There are more changes in here like that. I would like to hear from the others first before i start changing stuff though

static/locales/en-US.yaml
```diff
@@ -43,59 +43,59 @@ Global:
...
+    Like count: 1 like | {count} likes
+    Comment count: 1 comment | {count} comments
+    Watching count: 1 watching | {count} watching
 
-# Search Bar
-Search / Go to URL: Search / Go to URL
-Search Bar:
-  Clear Input: Clear Input
+# Search bar
+Search / go to URL: Search / go to URL
```
absidue, line 57: I would revert this change as the word go is still at the start because of the slash.

static/locales/en-US.yaml
absidue: I think renaming the key is okay too, just make sure you replace all references too it (also in the ProcessLocalesPlugin).

static/locales/en-US.yaml
```diff
@@ -13,27 +13,27 @@ Cut: Cut
...
-Version {versionNumber} is now available!  Click for more details: Version {versionNumber} is now available!  Click
-  for more details
-Download From Site: Download From Site
-A new blog is now available, {blogTitle}. Click to view more: A new blog is now available, {blogTitle}.
-  Click to view more
+Version {versionNumber} is now available!  Click for more details.: Version {versionNumber} is now available!  Click
+  for more details.
+Download from site: Download from site
+A new blog is now available, {blogTitle}. Click to view more.: A new blog is now available, {blogTitle}.
+  Click to view more.
```
absidue: I would prefer to keep periods out keys if possible, as it makes using the translation strings more annoying `t('parent.key without a period')` vs `t('parent["key with . a period"]')`, I don't mind adding them to the values/the actual strings.

Off-topic: I would prefer to use shorter keys in general, even if that means that the key doesn't match the text.

Issue #8246, opened 2025-11-09
[Feature Request]: Asymetric fwd/rev skipping


AlkalineJaunt:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Problem Description

I'd like to be able to jump forward in 30 sec intervals, and rewind in 10 second ones.

The [docco](https://docs.freetubeapp.io/usage/keyboard-shortcuts/) seems to suggest that this feature existed once...

> Left Arrow 	Rewind X seconds based on the Rewind Interval and current Video Playback Rate (Ex: 5 seconds Rewind Interval and Video Playback Rate 1x, rewinds video 5 seconds)

>Right Arrow 	Fast-Forward X seconds based on the Fast-Forward Interval and current Video Playback Rate (Ex: 5 seconds Fast-Forward Interval and Video Playback Rate 2x, fast-forwards video 10 seconds)

This is a great feature of antennapod and smplayer and I use it a lot.

### Proposed Solution

Add some extra setup and logic when skipping.

### Alternatives Considered

Can't think of any.

### Issue Labels

improvement to existing feature

efb4f5ff-1298-471a-8973-3d47447115dc:
We are doing the same as what YT does here so we arent going to change that. Also your interpretation of our documentation is wrong. We state nowhere that we offer individual control over the playback speed per shortcut

AlkalineJaunt:
Would a PR be considered?

PR #8280, opened 2025-11-14
Allow asymmetric forward/backward skipping

AlkalineJaunt:
Added different skip values for skipping by left/right arrow (small skips) and my mouse wheel.

Changed the layout of a couple of settings to imply their relatedness.

## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Addresses my query here: #8246
That issue was given a terse "no" and I had no further response when I followed up re a PR.

## Description

I like to skip forward further then I skip back. Jumping past boring stuff is fast, and it's easy to get back to where you want if you accidentally jumped a bit too far.

Once I started I found a the mouse skip setting was related, so I unconnected that and made the setting finer. It makes sense to me that keyboard and mouse skipping would be (able to be) different. Personally I don't have a clicky mouse wheel so it's very easy to scroll right to the end of the vid if you aren't careful :).

After this I moved a couple of the player settings to put like with like. I have no idea whose baby the settings are, but I'm happy to work on the arrangement if anybody hates what I've done.

This is my first attempt at a contribution, it's only small. I hope nobody is offended by me jumping in out of the blue.

## Screenshots 

Before:
image
After:
image

## Testing

When Jumping fwd and back with the arrow keys it's pretty obvious if the settings are being obeyed. When skipping with the mouse wheel it's less obvious, but adjusting the skip interval can be seen to adjust the wheel seek speed.

## Desktop

- Kbuntu
- 24.04
- 0.23.12


efb4f5ff-1298-471a-8973-3d47447115dc:
The issue was closed so the answer is no

PR #8310, opened 2025-11-18
Remove unnecessary new Promise call in ProcessLocalesPlugin

absidue:
## Pull Request Type

- [x] Other

## Description

This pull request removes the unecessary `new Promise()` calls around the async functions in our ProcessLocalesPlugin webpack plugin.

## Testing

1. The linter workflow on this pull request being successful indicates that these changes work correctly for release production.
2. Run `yarn dev`, if the webpack build doesn't fail these changes work correctly for dev mode.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/_scripts/ProcessLocalesPlugin.js
+++ b/_scripts/ProcessLocalesPlugin.js
@@ -45,7 +45,6 @@ class ProcessLocalesPlugin {
 
   /** @param {import('webpack').Compiler} compiler  */
   apply(compiler) {
-    const { CachedSource, RawSource } = compiler.webpack.sources
     const { Compilation, DefinePlugin } = compiler.webpack
 
     new DefinePlugin({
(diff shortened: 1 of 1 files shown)


PikachuEXE:
Not caused by this PR but anyway to make it gives warning about en-US instead of en (dev
image

absidue:
That's unfortunately a vue-i18n "feature": https://vue-i18n.intlify.dev/guide/essentials/fallback#implicit-fallback-using-locales

While we can stop it happening by adding a exclamation mark to the end of the locale code when setting the locale, that presumably also means that everywhere we read the locale, we would need to remove it before passing the locale code to other APIs e.g. the Intl ones. So I haven't tried doing that yet.

efb4f5ff-1298-471a-8973-3d47447115dc:
String look like this now.  Also unrelated?

VirtualBoxVM_JF4pXuCvwA

absidue:
That's very odd, I presume you just switched to this branch and then ran `yarn dev`?

efb4f5ff-1298-471a-8973-3d47447115dc:
Correct also ran yarn before running yarn dev

absidue:
@efb4f5ff-1298-471a-8973-3d47447115dc I have taken a different approach now, could you please try again?

kommunarr (approved):

PR #8417, opened 2025-12-14
Migrate the App component to the composition API

absidue:
## Pull Request Type

## Description

This pull request migrates the `App` component to the composition API. I also fixed some of the issues with event handlers not being unregistered when the `App` component is hot-reloaded in development, which caused external links to open multiple times and rearranged the elements in this component, so we can hopefully remove some of the CSS `z-index` mess that we have in a follow up. Additionally we start loading the `en-US` locale as soon as the App component is created, as it is the fallback locale so is always required, which should cut down on the messages about not finding strings when FreeTube starts, as it can be loaded in parallel with user data.

## Testing

Test the functionality of the App component, such as changing theme settings, the updates banner (modify the package.json to use an older version), click on external links, etc.

## Desktop

- **OS:** Windows 
- **OS Version:** 11

### DIFF
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -252,13 +252,19 @@ function updateTheme() {
 updateTheme()
 
 const showUpdatesBanner = ref(false)
-const updateBannerMessage = ref('')
+const latestVersionNumber = ref('')
 const showReleaseNotes = ref(false)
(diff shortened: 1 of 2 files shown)


absidue:
Fixed both issues, please test again.

PR #8419, opened 2025-12-14
SPAM

sickelap:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Use `window.location.reload()` to reload the view on button press.

## Screenshots 

Before:
before

After:
after

## Testing

- Navigate to all views and pressed reload

## Desktop

- **OS:** Linux x86_64
- **OS Version:** Pop_OS 22.04 LTS (Linux 6.17.4-76061704-generic)
- **FreeTube version:** development

## Additional context

Currently in channel view switching to any other tab and pressing refresh reloaded page goes back to initial first one. (Do we need to persist tab?)

### DIFF
--- a/src/renderer/components/TopNav/TopNav.vue
+++ b/src/renderer/components/TopNav/TopNav.vue
@@ -43,6 +43,15 @@
         :title="forwardText"
         @click="historyForward"
       />
+      <FtIconButton
+        class="navIconButton"
+        :icon="['fas', 'refresh']"
+        :theme="null"
+        :size="20"
+        :use-shadow="false"
+        :title="$t('Reload')"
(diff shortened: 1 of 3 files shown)


absidue:
This is  the completely wrong solution and has all the same problems that the CTRL+R shortcut had before it was removed.

Restarting the entire window, nuking everything in memory, clearing caches and loading all settings and user data from the disk again, just because you want to refresh a single page, is clearly completely wrong approach. If you want to refresh a specific page, it has to be implemented on a page by page basis, see the existing refresh buttons on the subscriptions, trending and popular pages for correctly implemented examples.

PR #8481, opened 2026-01-04
adding number of views filter

andrewarrow:
## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
Added maxViews to search settings state and setSearchMaxViews mutation

## Screenshots
https://i.imgur.com/AdHWcIK.png

## Testing
The filter works client-side by filtering results after they're fetched, since YouTube's API doesn't support filtering by view count directly. This means you may see fewer results than a typical page if many videos exceed the view threshold.

## Desktop
- **OS: macOS
- **OS Version: 26.1
- **FreeTube version: 0.23.12

## Additional context
Useful for finding "gems" of videos not yet popular.

### DIFF
--- a/src/renderer/components/FtSearchFilters/FtSearchFilters.vue
+++ b/src/renderer/components/FtSearchFilters/FtSearchFilters.vue
@@ -54,6 +54,13 @@
         :values="DURATION_VALUES"
         class="searchRadio"
       />
+      <FtRadioButton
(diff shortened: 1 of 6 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
Sorry but we only allow filters that YT has in their list

Issue #8359, opened 2025-11-27
[Bug]: Application quits itself on Ubuntu after clicking on Show window in Tray menu


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

1. Minimize application
2. Go to tray
3. Click on Show to Show the minimized window
4. App quits

This is 100% wayland related because the issue disappears after running `unset XDG_SESSION_TYPE`

https://github.com/user-attachments/assets/77114a2c-eedc-4b96-9bd4-132cd4cab0f2

### Expected Behavior

App shouldnt quit and window shows

### Issue Labels

feature stopped working

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/b124ed2efeccdef548a3eb1f448284e330ba3778

### Operating System Version

Ubuntu 24.04.3 LTS

### Installation Method

other

### Primary API used

Local API

### Nightly Build

- [x] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
@Devenor would it be possible for you to look into this one

Devenor:
It's an Electron issue that has already been reported [here](https://github.com/electron/electron/issues/49253). It exits with the same error, and I can confirm that the 38.1.2 version still works, as they mention in the bug report.

efb4f5ff-1298-471a-8973-3d47447115dc:
@Denvenor would you be interested to open a PR with [the workaround](https://github.com/electron/electron/issues/49253#issuecomment-3960109030)

Devenor:
Sure, I'll take a look.

Devenor:
[...]

PR #8494, opened 2026-01-06
Add a show all windows button to the tray menu

### LINKED ISSUE #8479: [Feature Request]: Add "Show All" to the tray icon menu
### Guidelines

### Problem Description

When "minimise to system tray" is active and you have minimised a bunch of windows to the tray, getting them to show all at once again must be done for each window manually.

For example Windows has the feature to show the desktop (clicking the lower left corner when this feature is enabled) which minimises all open windows at once. Usually to the taskbar, but for applications that implemented a mimimise-to-tray feature like FreeTube, windows get hidden behind the tray icon. As everything is supposed to work and is expected by the Windows user.

If you've had lots of windows open, minimising them takes a second, unminising takes much longer and is tedious. A quick way to do one thing should also have a quick way to reverse it, imho. Therefore some apps have a "show all hidden windows" option in their tray menu. I think FreeTube needs that, too. Maybe even a "hide all windows" for when the OS doesn't support something like show desktop or if it works differently or if the user has disabled it.

### Proposed Solution

Add a "Show all" option in the tray icon menu over or under the "New window" option.

### Alternatives Considered

Not really an alternative but the workaround I currently use: Go to settings and temporarily disable "Minimise to tray" which does the same thing as a "Show all" option would do, then immediately enable the "Minimise to tray" again.

### Issue Labels

new feature, improvement to existing feature

### Additional Information

_No response_

absidue:
## Pull Request Type

- [x] Feature Implementation

## Related issue

- closes #8479

## Description

This pull request adds a show all windows button to the tray menu, which does what it says on the tin: shows all hidden windows.

## Testing

The tray menu is only implemented on Windows and Linux

1. Enable the minimize to tray setting
2. Open multiple windows
3. Minimize all or most of them
4. Click on "Show all windows" in the tray menu
5. All windows should appear and the tray icon should disappear.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -797,14 +797,14 @@ function runApp() {
   function defaultTrayMenu() {
     return [
       {
-        label: 'New window',
+        label: 'New Window',
         click: () => createWindow({
           showWindowNow: true,
           replaceMainWindow: trayWindows.some(item => item.id === mainWindow.id)
         })
       },
       {
-        label: 'Show all windows',
+        label: 'Show All Windows',
         click: () => {
           // Use while loop instead of for loop as trayClick modifies the trayWindows array
           while (trayWindows.length > 0) {


src/main/index.js
```diff
@@ -797,12 +797,21 @@ function runApp() {
   function defaultTrayMenu() {
     return [
       {
-        label: 'New Window',
+        label: 'New window',
```
PikachuEXE: I think title case should still be used like browser/electron menu/macOS app context menu

Image

Image

efb4f5ff-1298-471a-8973-3d47447115dc: i agree. Thats why i didnt touch it in https://github.com/FreeTubeApp/FreeTube/pull/8273. So this and https://github.com/FreeTubeApp/FreeTube/pull/8494/changes#diff-f4ca67281c6be819c2166841d7631ffd5316d4c3a3b40e01a512c4113a537b2cR807 should be updated to title case

efb4f5ff-1298-471a-8973-3d47447115dc:
rip i cant properly test this in my Ubuntu VM due to #8359 :/ 
Ill try to bisect when it broke

PikachuEXE:
I tested this on Windows via https://github.com/PikachuEXE/FreeTube/actions/runs/20767925766

efb4f5ff-1298-471a-8973-3d47447115dc:
Tested it too with your build and seems to be working perfectly. 

I have one maybe nitpick point of discussion. My first minimized window is the last to be unminimized and my last minimized window is the first one that will show. Should we maybe change it to the other way around so the first window that is minimized is the first to show?

The current approach is very logical from the point that it was the user's last opened window to show first but i thought to raise the point anyway. 

I think the current approach is good as is

PikachuEXE:
I have no preference here coz I don't use it :P

absidue:
I don't mind either way, I van reverse it if you would prefer that?

efb4f5ff-1298-471a-8973-3d47447115dc:
No i think its good but i only raised it just in case both of you wanted to see it changed

ChunkyProgrammer (approved):

PR #8495, opened 2026-01-07
Create Build workflows for PRs

ArthurKun21:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Creating an accessible build for testing PR-related builds. The code was copy-paste from https://github.com/FreeTubeApp/FreeTube/blob/development/.github/workflows/build.yml while updating the name of uploads to have `PR`-`<pr-number>`

Since we already have nightly builds for the `development` branch, I figured it won't be too bad to also have a nightly build for PRs for test

Also made it a separate workflow file for separate tracking in the `actions` tab

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

I made a sample here https://github.com/ArthurKun21/FreeTube/pull/6 fixed the naming problem that I didn't noticed.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -77,22 +77,10 @@ jobs:
       id: versionNumber
       uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd #v8.0.0
       env:
-        IS_DEV: ${{ contains(github.ref, 'development') }}
-        IS_RC: ${{ contains(github.ref, 'RC') }}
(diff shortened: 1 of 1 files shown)


absidue (commented):
If you want to test pull requests you should use dev mode, on the rare occasion that a custom build including pull request changes is actually needed one will be provided and as a user you really should not use changes from pull requests before they have been reviewed and tested. So doing full builds for every pull request is a waste of CI resources and puts users at risk as they'll be using completely untested and unvalidated builds, test build workflows only make sense for projects where they are actually needed (e.g. NewPipe because you need to install it on your phone).

PR #7514, opened 2025-06-04
feat: add option to hide the channel link

DontBlameMe99:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #3193

## Description

This adds a option to the settings (parental controls) which allows the Channel Link to be hidden.

## Screenshots 

image
image
image

## Testing

Open various videos, enable and disable the option

## Desktop

- **OS:** Arch Linux
- **FreeTube version:** Newest Development branch ([this commit](https://github.com/FreeTubeApp/FreeTube/commit/15bedb96a2a39ac2187cfc8ebfe33e27bb5262dd))

## Additional context

This was a requested feature (marked with good first issue) and I implemented it.
I am new to contributing actual code to FreeTube. I have previously only contributed themes, so please let me know if anything should be done differently.

PR #8502, opened 2026-01-09
Add Parental Control toggle to disable channel link from videos

### LINKED ISSUE #8375: [Feature Request]: Remove/limit access to channels.
### Guidelines

### Problem Description

I want to limit my child's access to the back catalogue of the channels I am subscribed to. You can limit the number of videos in the subscription list but clicking on the channel name or going to the channels tab allows you unrestricted access to all of the videos

### Proposed Solution

It would be helpful to have a toggle to remove the channel category and remove the ability to click the channel names in the subscription list.

### Alternatives Considered

Another option would to apply the subscription limit of showing x amount of videos to also apply to the channels page. Being able to only show the most recent 5 videos or whatever amount would limit the access to the back catalogue and stop overconsumption.

### Issue Labels

ease of use improvement

### Additional Information

_No response_

ozrendev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #8375

## Description

Adds a toggle under Parental Control that allows the user to disable the channel link from video thumbnails.

## Testing

Go to Settings -> Parental Control -> Enabled "Disable Channel Link on Videos"
View any video thumbnail and see that channel link can't be clicked and no longer appears as a link

## Additional context
The original issue called for also hiding the "Channels" button in the side bar, but I feel like this should be a separate PR since I think it makes the most sense to be another toggle under "Distraction Free -> Side Bar"
I'm not 100% sure on this, so wanted to add this here for feedback?


efb4f5ff-1298-471a-8973-3d47447115dc:
> The original issue called for also hiding the "Channels" button in the side bar, but I feel like this should be a separate PR since I think it makes the most sense to be another toggle under "Distraction Free -> Side Bar"
I'm not 100% sure on this, so wanted to add this here for feedback?

indeed best to kept separate and add it to the Side bar

src/renderer/store/modules/settings.js
```diff
@@ -169,6 +169,7 @@ const state = {
   defaultViewingMode: 'default',
   defaultVideoFormat: 'dash',
   disableSmoothScrolling: false,
+  disableChannelLink: false,
```
PikachuEXE: I think this name is quite ambiguous coz I can clearly see channel links in watch page...
Need something more specific

static/locales/en-US.yaml
```diff
@@ -673,6 +673,14 @@ Settings:
     Hide Uploader on Watch page: Hide Uploader on Watch page
     Show Family Friendly Only: Show Family Friendly Only
     Hide Search Bar: Hide Search Bar
+    Disable Channel Link on Videos : Disable Channel Link on Vidoes
```
PikachuEXE: This is not copy and paste!? (key > value)
Also got the same issue on where does this apply (basically the same as above)
```suggestion
    Disable Channel Link on Videos : Disable Channel Link on Videos
```

PikachuEXE (changes requested):

efb4f5ff-1298-471a-8973-3d47447115dc:
So im not really sure if this even needs to be implemented. Isnt it possible to limit this kind of access with our existing settings?

Just and FYI for the team below is the usecase listed in the issue

> I want to limit my child's access to the back catalogue of the channels I am subscribed to. You can limit the number of videos in the subscription list but clicking on the channel name or going to the channels tab allows you unrestricted access to all of the videos

src/renderer/store/modules/settings.js
ozrendev: How about something like `disableChannelLinkOnVideoThumbnail`?

static/locales/en-US.yaml
ozrendev: Oops, my bad! Same as above too, Disable Channel Link on Video Thumbnails?

ozrendev:
> So im not really sure if this even needs to be implemented. Isnt it possible to limit this kind of access with our existing settings?
> 
> Just and FYI for the team below is the usecase listed in the issue
> 
> > I want to limit my child's access to the back catalogue of the channels I am subscribed to. You can limit the number of videos in the subscription list but clicking on the channel name or going to the channels tab allows you unrestricted access to all of the videos

How would you limit it with existing settings? I had a look but couldn't find a way to do it so I'm curious to test it out.

I think overall the toggle can be useful, ie for kids controls, extra control for stopping binge watching a whole channel's content? But having an option to also hide the channel in the sidebar would be necessary I think, am happy to open another PR for that.

PikachuEXE:
I think `Hide Uploader on Watch page` is the closes we have for toggling that
But it hides not just disable the link on watch page (and also it said watch page only...)
But I can't think of a better option to group these 2 together (or even should they be grouped

Is there another option this "link hiding" can fall under? Or do we need the new option to be used for more behaviour (which might be implemented yet? Discuss~

ozrendev:
> I think `Hide Uploader on Watch page` is the closes we have for toggling that But it hides not just disable the link on watch page (and also it said watch page only...) But I can't think of a better option to group these 2 together (or even should they be grouped
> 
> Is there another option this "link hiding" can fall under? Or do we need the new option to be used for more behaviour (which might be implemented yet? Discuss~

Can you please clarify which page you are describing when you say Watch Page? This PR should be active when you search and see a list of videos, or when viewing a video and seeing the "recommended" videos in the side bar. I am happy with whichever wording, either Hide or Disable is fine - I would say "disable" because you can still see the channel name but not click the link, but either is fine in my opinion.

I think the option for toggling hide/show link here should stay in Parental Controls, since it's a restrictive option that limits function, like most other toggles in that category. The option for hiding the "Channel" button in the Side Bar would make more sense to me to be grouped with Distraction Free -> Side Bar (with the rest of the side bar button toggles)

PikachuEXE:
`Hide Uploader on Watch page` is an existing option, not a suggestion for naming this new option, just to be clear

Watch Page should mean watch video page (coz what else can be watched)

If this new option applies to all "video boxes" then I dunno what to name it yet.
`Disable Channel Link on Videos` sounds fine but I am not sure if the link in the watch page should be disabled too or not (are there other places with channel links tied to video(s)? dunno)

ozrendev:
Ah okay, thanks for making it clear!

Yeah I see what you mean. The component name is `ft-list-video`, but I'd call those thumbnails so maybe something like `Disable Channel Link on Video Thumbnails` or `Disable Link on Thumbnails` to keep it short?

PikachuEXE:
Actually I think this option can (not sure should) be used to disable channel link in watch page

(bottom left
image

From https://github.com/FreeTubeApp/FreeTube/issues/8375
> I want to limit my child's access to the back catalogue of the channels I am subscribed to. You can limit the number of videos in the subscription list but clicking on the channel name or going to the channels tab allows you unrestricted access to all of the videos

I know we can hide channel name on watch page, but maybe disabling the link and still showing the channel link with this option is good too?

I know we have https://github.com/FreeTubeApp/FreeTube/pull/7514 but disable might be a better implementation

src/renderer/store/modules/settings.js
PikachuEXE: This would be a good name if we really use it to disable all channel links not just on video thumbnails :)
```suggestion
  disableChannelLinks: false,
```

ozrendev:
> I know we can hide channel name on watch page, but maybe disabling the link and still showing the channel link with this option is good too?
> 
> I know we have #7514 but disable might be a better implementation

Yeah, I feel like since there is already an option under Parental Controls, the user could just toggle both in order to achieve this. I like the granular approach to locking down little bits at a time, so you can very finely tune the level of control.

But disabling the channel link also makes sense, especially for this option and it's more clear that ALL channel links will be disabled by the toggle. My preference would go to the first option using both toggles, but I think both are good approaches.

What do some of the other reviewers think about these approaches?

src/renderer/store/modules/settings.js
PikachuEXE: Just update this first~~ (kind of waiting for sorting bug to be fixed for 0.24.0, this one probably won't be reviewed until that's released

ozrendev: All good, will wait until then

efb4f5ff-1298-471a-8973-3d47447115dc:
Sorry for the late response. Reading everything back and maybe we should combine this feature with the hide watch page one? Meaning that the watch page one will be visible but can't be clicked and no longer appears as link?

PikachuEXE:
Agreed and the old setting should be migrated (if too difficult I can do the setting migration in a separate PR
Though in this PR the channel link on watch page should be disabled
Also the translation typo not fixed yet~

ozrendev:
The setting should be combined now so that it removes link from the channel name and thumbnail under the Watch page. I feel it's a little verbose, but not sure of another method since those elements are nested inside `RouterLink` components. Hope I understood the above correctly

Also typos/store variables have been re-named and fixed

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -503,6 +517,8 @@ function removeFromQuickBookmarkPlaylist() {
   // TODO: Maybe show playlist name
   showToast(t('Video.Video has been removed from your saved list'))
 }
+
+const disableChannelLinks = computed(() => store.getters.getDisableChannelLinks)
```
PikachuEXE:
```suggestion
const enableChannelLinks = computed(() => !store.getters.getDisableChannelLinks)
```

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -42,6 +42,7 @@
           v-if="!hideUploader"
         >
           <RouterLink
+            v-if="!disableChannelLinks"
```
PikachuEXE: I am not a fan of double negative
```suggestion
            v-if="enableChannelLinks"
```

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -50,18 +51,31 @@
               alt=""
             >
           </RouterLink>
+          <img
+            v-if="disableChannelLinks"
```
PikachuEXE:
```suggestion
            v-else
```

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -50,18 +51,31 @@
...
+            class="channelThumbnail revertCursor"
+            alt=""
+          >
         </div>
         <div>
           <div
             v-if="!hideUploader"
           >
             <RouterLink
+              v-if="!disableChannelLinks"
```
PikachuEXE: ditto

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -50,18 +51,31 @@
...
               :to="`/channel/${channelId}`"
               class="channelName"
               dir="auto"
             >
               {{ channelName }}
             </RouterLink>
+            <div
+              v-if="disableChannelLinks"
```
PikachuEXE: ditto

src/renderer/components/WatchVideoInfo/WatchVideoInfo.css
```diff
@@ -97,6 +97,10 @@
   gap: 4px;
 }
 
+.revertCursor {
```
PikachuEXE: I haven't tested this yet so not sure if this is really needed

static/locales/en-US.yaml
```diff
@@ -673,6 +673,14 @@ Settings:
...
     Show Family Friendly Only: Show Family Friendly Only
     Hide Search Bar: Hide Search Bar
+    Disable Channel Link on Videos : Disable Channel Link on Videos
+  Download Settings:
+    Download Settings: Download
+    Ask Download Path: Ask for download path
+    Choose Path: Choose Path
+    Download Behavior: Download Behavior
+    Download in app: Download in app
+    Open in web browser: Open in web browser
```
PikachuEXE: I am gonna guess this is from merging dev...?

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Found link to channel:

https://github.com/user-attachments/assets/4e93ce5c-79a2-4928-a4d2-04c6273e643b

I think this one is out of scope but i can also go to a channel like this, thoughts?:

https://github.com/user-attachments/assets/efb65ae1-281a-49b0-b0bf-ba363fe59416


Some weird behavior:

https://github.com/user-attachments/assets/35cef41f-f2f8-4caf-a8f6-8b551c7bacf5

https://github.com/user-attachments/assets/3f8b2dd8-3861-425d-a49c-a7e6fcc65b4e

https://github.com/user-attachments/assets/d946fe68-bc1f-43e4-83de-92c1763c4469

https://github.com/user-attachments/assets/00fe7574-080c-45e4-b03f-deb3c3857037

PikachuEXE:
Since https://github.com/FreeTubeApp/FreeTube/issues/8375 is asking for blocking access to channel, not just via videos
I think channel links anywhere should be blocked

static/locales/en-US.yaml
ozrendev: Yeah, I believe so

src/renderer/components/WatchVideoInfo/WatchVideoInfo.css
ozrendev: I changed this to use `initial` instead - it just stops the cursor staying as `pointer`

ozrendev:
Hopefully the new commit should solve that weird behavior. Also made it clearer by removing the double negatives.

Will have a look at disabling channel links from those other places. Let me know if you find anywhere else it can be accessed

ozrendev:
I've pushed changes for both Playlists and Posts. 

Since the original issue was relating to subscribed channels, I feel that Posts are in scope. The user might want to limit access, but still have all Subscriptions options enabled.

I think that covers most video related channelLinks. The question is where to stop, as you could potentially still access the uploader's channel if you have comments enabled and the uploader has commented on the video. But I feel like this is a stretch and out of scope, since at that point you would probably just disable comments anyway. What do you think?

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm i see your point but i also can see a parent giving their child educational content so and sometimes with that there are helpful comment from the author itself or clarifications. So i would tend to say its in scope.

Also i found the last one

https://github.com/user-attachments/assets/9c65d752-f439-453a-83cf-33ef794390c7

This is definitely out of scope 😄 

https://github.com/user-attachments/assets/ad30c8ef-444a-4ba6-b5ad-bad226b921aa

static/locales/en-US.yaml
```diff
@@ -671,6 +671,14 @@ Settings:
     Hide Uploader on Watch page: Hide Uploader on Watch page
     Show Family Friendly Only: Show Family Friendly Only
     Hide Search Bar: Hide Search Bar
+    Disable Channel Link on Videos : Disable Channel Link on Videos
```
efb4f5ff-1298-471a-8973-3d47447115dc: Probably should be renamed now because it doesnt apply only on videos

ozrendev:
As I was making the above changes, the manually copying/pasting depending on `enableChannelLinks` seemed a bit tedious and prone to error/future edits. I went looking for a solution and [found this workaround for `<router-view>`](https://github.com/vuejs/vue-router/pull/2098#issuecomment-652283739)

I just made commit [c46623f](https://github.com/FreeTubeApp/FreeTube/pull/8502/commits/c46623f3a39452bd952a534a6af44d1fd915d4b6) so that it's easier to show the difference in context. What does everyone think? Keep it or revert back to previous method?

PikachuEXE (dismissed):
It works but I have no idea if others are fine with this style of conditional component

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
LGTM, but i would like @absidue to check it out first

static/locales/en-US.yaml
absidue: Even if this came from a merge, the fact that it is still here means the merge wasn't clean and as those download strings no longer exist on the development branch, merging this pull request as is would add those strings back.

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -50,18 +51,31 @@
...
             <RouterLink
+              v-if="enableChannelLinks"
               :to="`/channel/${channelId}`"
               class="channelName"
               dir="auto"
             >
               {{ channelName }}
             </RouterLink>
+            <div
+              v-if="!enableChannelLinks"
```
absidue:
```suggestion
              v-else
```

src/renderer/components/PlaylistInfo/PlaylistInfo.vue
```diff
@@ -129,6 +129,11 @@
         v-else
         class="playlistChannel"
       >
+        <img
+          class="channelThumbnail"
```
absidue: The channel thumbnail wasn't here before, because when YouTube doesn't return a channel link here it also doesn't return a thumbnail, so you need to wrap the channel thumbnail in a if statement.
```suggestion
        <img
          v-if="channelThumbnail"
          class="channelThumbnail"
```

src/renderer/components/FtListChannel/FtListChannel.vue
```diff
@@ -34,15 +46,29 @@
             {{ name }}
           </h3>
         </router-link>
+        <div
+          v-if="!enableChannelLinks"
```
absidue:
```suggestion
          v-else
```

src/renderer/components/FtListChannel/FtListChannel.vue
```diff
@@ -20,10 +21,21 @@
           alt=""
         >
       </router-link>
+      <div
+        v-if="!enableChannelLinks"
```
absidue:
```suggestion
        v-else
```

src/renderer/components/FtListChannel/FtListChannel.vue
```diff
@@ -34,15 +46,29 @@
...
             :to="`/channel/${id}`"
           >
             {{ handle }}
           </router-link>
+          <bdi
+            v-if="handle !== null && !enableChannelLinks"
+            class="handle"
+          >
+            {{ handle }}
+          </bdi>
```
absidue: It would probably be better to wrap both of these tags in a `<template>` tag with a `v-if` that checks the `handle` variable, then inside of it just use a `v-if` and `v-else` for the enableChannelTags variable.

efb4f5ff-1298-471a-8973-3d47447115dc:
@ozrendev anything else left todo?

ozrendev:
> @ozrendev anything else left todo?

Yeah, that should be it. I tested it all again and don't think I missed anything

efb4f5ff-1298-471a-8973-3d47447115dc:
Somehow the playlist controls moved

VirtualBoxVM_QtPDhXPDeU

ozrendev:
That should be fixed now! Thanks for always testing it so thoroughly, didn't expect that one

efb4f5ff-1298-471a-8973-3d47447115dc:
only thing i can find is this. It makes me think i can click on this but i clearly cant

https://github.com/user-attachments/assets/79c60067-a02d-4bb4-893a-d19ab57e1491

PikachuEXE:
Can you fix? should be just style @ozrendev

ozrendev:
@PikachuEXE @efb4f5ff-1298-471a-8973-3d47447115dc I've pushed the fix now

PikachuEXE (dismissed):
I tested with dev merged (due to pnpm PR

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

src/renderer/components/ft-list-video/ft-list-video.vue
```diff
@@ -123,17 +123,15 @@
...
           class="channelName"
           dir="auto"
           :to="`/channel/${channelId}`"
         >
           {{ channelName }}
-        </router-link>
-        <bdi v-else-if="channelName !== null">
-          {{ channelName }}
-        </bdi>
+        </component>
```
absidue, line 134: With the else block remove it means that if there is a channel name but no ID it won't show anything anymore.

absidue (commented):
Sorry for the late review. Unfortunately I still can't approve it yet because this would introduce a regression.

src/renderer/components/ft-list-video/ft-list-video.vue
ozrendev: I've re-added the `bdi` element back in now. I haven't tested it as I wasn't sure where to find a case where there's a channel name but without ID

absidue: I can't give you an example off the top of my head, however if FreeTube has some handing for something it means that it exists or at least existed at some point in time. As a general rule you can assume that every field in YouTube's innertube responses are optional, even for the same node type it can be missing different fields on different pages, but as treating everything as optional is a massive amount of work and very inconvenient FreeTube only has handling for situations that someone actually encountered at some point.

ozrendev: Thanks for explaining, I misread the `v-else-if` when I removed it so it was a mistake. It makes sense that leaving these edge cases are much better than having a situation where something that was previously handled is now not handled. The above was saying that commit [93b63b2](https://github.com/FreeTubeApp/FreeTube/pull/8502/commits/93b63b268f2d8f2bd2cc302a9dd0671d9095acb4) was untested, but checking it looks correct now?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

PR #8510, opened 2026-01-11
save the tab selection and sort on channel page

andrewarrow:
## Pull Request Type
- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
When on a channel page, if I select the video tab, change the sort, it doesn't remember this selection. After selecting a video and clicking back I'm back on home tab and my sort selection for video is gone. This makes the tab selection and sort sticky.

## Testing
Goto a channel page, select video tab, change sort, select a video, then select back button.

## Desktop

- **OS: macOS
- **OS Version: 26.1
- **FreeTube version: 0.23.12

### DIFF
--- a/src/renderer/views/Channel/Channel.vue
+++ b/src/renderer/views/Channel/Channel.vue
@@ -336,6 +336,43 @@ let channelInstance = null
 let apiUsed = ''
 let mayContainContentFromOtherChannels = false
 
+/**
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
We have issued a temporarily block to interact with our repo as you have now opened multiple pull requests that either dont make sense at all or contains very bad code

PR #8582, opened 2026-01-24
Decipher the live HLS manifest URL too

absidue:
## Pull Request Type

- [x] Bugfix

## Related issue

- https://github.com/FreeTubeApp/FreeTube/pull/8530

## Description

I realised that in the rush to get the deciphering fix for live DASH into a hotfix release that I forgot to apply the same treatment to the HLS URL which we use as a fallback on the watch page here:

https://github.com/FreeTubeApp/FreeTube/blob/07ce7215c598717935d2015de8f1a0d48a34ac79/src/renderer/views/Watch/Watch.js#L675-L681

## Testing

1. Add `info.streaming_data.dash_manifest_url = undefined` on line 547 in `src/helpers/api/local.js` to remove the DASH manifest URL from the response
2. Open https://youtube.com/@live
3. Pick a live stream that is currently live
4. It should still play
5. Trying to switch to the audio formats should show a toast that audio formats are not available for this video.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -546,10 +546,20 @@ export async function getLocalVideoInfo(id) {
     }
 
     if (info.streaming_data.dash_manifest_url) {
-      info.streaming_data.dash_manifest_url = await decipherDashManifestUrl(
+      info.streaming_data.dash_manifest_url = await decipherManifestUrl(
         info.streaming_data.dash_manifest_url,
         webInnertube.session.player,
-        contentPoToken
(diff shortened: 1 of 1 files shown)


absidue:
That's expected, seeking is only possible with the DASH manifest, which is why we treat the HLS one as the backup. That's why seeking on live streams in FreeTube only became a thing when we switched from video.js to shaka-player which supports the non-standard independent segments that YouTube uses in their live DASH manifests.

PR #8607, opened 2026-01-30
Show a message instead of hiding the posts tab when RSS is enabled

### LINKED ISSUE #5716: [Feature Request]: Show a message instead of hiding the community tab on the subscriptions page when it is unavailable because of RSS
### Guidelines

### Problem Description

When the user has 125 or more subscriptions in the currently active profile FreeTube switches to RSS (as indicated by a toast message displayed at the time) and as the community posts are not available over RSS FreeTube currently hides that tab. To users it is not directly obivious on the subscriptions page itself why the community tab was hidden.

### Proposed Solution

Instead of hiding the community tab when using RSS, FreeTube should instead display a message inside of the tab explaining why community posts cannot be shown.

### Issue Labels

### LINKED ISSUE #8104: [Feature Request]: Improve force RSS behavior when subscription limit is reached
### Guidelines

### Problem Description

In the Settings under Subscriptions, there is an option called "Fetch Feeds from RSS" which can be set to disabled, even when RSS fetching is enforced due to the user being subscribed to 125 channels. This can be confusing, as the user might not understand why this setting still applies, even though RSS is technically disabled.

Additionally, the user is informed that they have a "large number of subscriptions," but they're not given clear instructions on what to do to stop the enforcement of RSS fetching. This creates an unclear user experience, as it's not immediately obvious what steps need to be taken to resolve or adjust the situation.

### Proposed Solution

Visually grey out the toggle after the user first receives the RSS is forced message. This way, the message will only be displayed once, rather than every time. In addition to that, adjust the message to inform the user of the option they have to prevent RSS from being forced in the future.

### Alternatives Considered

None.

### Issue Labels

visual improvement, ease of use improvement, display more information to user

### Additional Information

_No response_

absidue:
## Pull Request Type

- [x] Feature Implementation

## Related issue

- closes #5716

## Description

Show the posts tab on the subscriptions with an explanation when "Fetch Feeds from RSS" is enabled instead of hiding it.

## Screenshots

Screenshot of the FreeTube subscriptions page with the &quot;Posts&quot; tab selected, showing the new message explaining why posts are unavailable when RSS is enabled.

## Testing

1. Enable the "Fetch Feeds from RSS" setting and check that the message shows up on the subscription posts tab.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -135,6 +135,9 @@ Subscriptions:
   Load More Posts: Load More Posts
   Subscriptions Tabs: Subscriptions Tabs
   All Subscription Tabs Hidden: 'All subscription tabs are hidden. To see content here, please unhide some tabs in the "{subsection}" section in "{settingsSection}".'
+  Posts:
+    RSS Message: The "{setting}" setting is enabled and YouTube does not provide RSS feeds for posts.
+    Forced RSS Message: This profile has a large number of subscriptions. Forcing RSS to avoid rate limiting. YouTube does not provide RSS feeds for posts.
 More: More
 Channels:
   Channels: Channels
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -125,8 +125,7 @@ Subscriptions:
   Subscriptions: Subscriptions
   # channels that were likely deleted
   Error Channels: Channels with Errors
-  This profile has a large number of subscriptions. Forcing RSS to avoid rate limiting: This
-    profile has a large number of subscriptions.  Forcing RSS to avoid rate limiting
+  RSS is forced for profiles with 125 or more subscriptions to avoid rate limiting: RSS is forced for profiles with 125 or more subscriptions to avoid rate limiting
   'Your Subscription list is currently empty. Start adding subscriptions to see them here.': Your Subscription list is currently empty. If you want to import your subscriptions you can go to Data Settings and select Import Subscriptions or you can search for a channel and subscribe to them.
   Disabled Automatic Fetching: You have disabled automatic subscription fetching. Refresh subscriptions to see them here.
   Empty Channels: Your subscribed channels currently does not have any videos.
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -131,6 +131,8 @@ Subscriptions:
   Load More Posts: Load More Posts
   Subscriptions Tabs: Subscriptions Tabs
   All Subscription Tabs Hidden: 'All subscription tabs are hidden. To see content here, please unhide some tabs in the "{subsection}" section in "{settingsSection}".'
+  Posts:
+    RSS Message: YouTube does not provide RSS feeds for posts. Disable the "{rssSetting}" setting to view posts. Enable the "{hideSetting}" setting to hide this tab.
 More: More
 Channels:
   Channels: Channels
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -132,7 +132,7 @@ Subscriptions:
   Subscriptions Tabs: Subscriptions Tabs
   All Subscription Tabs Hidden: 'All subscription tabs are hidden. To see content here, please unhide some tabs in the "{subsection}" section in "{settingsSection}".'
   Posts:
-    RSS Message: YouTube does not provide RSS feeds for posts. Disable the "{rssSetting}" setting to view posts. Enable the "{hideSetting}" setting to hide this tab.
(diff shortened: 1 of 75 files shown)


static/locales/en-US.yaml
```diff
@@ -135,6 +135,9 @@ Subscriptions:
   Load More Posts: Load More Posts
   Subscriptions Tabs: Subscriptions Tabs
   All Subscription Tabs Hidden: 'All subscription tabs are hidden. To see content here, please unhide some tabs in the "{subsection}" section in "{settingsSection}".'
+  Posts:
+    RSS Message: The "{setting}" setting is enabled and YouTube does not provide RSS feeds for posts.
```
efb4f5ff-1298-471a-8973-3d47447115dc:
```suggestion
    RSS Message: The "{setting}" setting is enabled, but YouTube does not provide RSS feeds for posts. To view posts, disable this setting. To hide this tab, go to the "{subsection}" section in "{settingsSection}".
```

static/locales/en-US.yaml
```diff
@@ -135,6 +135,9 @@ Subscriptions:
...
+    Forced RSS Message: This profile has a large number of subscriptions. Forcing RSS to avoid rate limiting. YouTube does not provide RSS feeds for posts.
```
efb4f5ff-1298-471a-8973-3d47447115dc:
```suggestion
    Forced RSS Message: RSS is forced for profiles with 125 or more subscriptions to avoid rate limiting. YouTube does not provide RSS feeds for posts. To disable forced RSS, reduce your subscription count or split your subscriptions across multiple profiles. To hide this tab, go to the "{subsection}" section in "{settingsSection}".
```

PikachuEXE:
Not sure if that dim out effect on specific profiles would confuse users
I don't think we have any profile specific setting so far
If we want to tell user that FT is forcing RSS for current profile (even when force RSS disabled) I think it probably should be done in another way

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Taken from issue

> Visually grey out the toggle after the user first receives the RSS is forced message. This way, the message will only be displayed once, rather than every time.

Show toast once per tab per session


~Also like discussed update the message at
https://github.com/FreeTubeApp/FreeTube/blob/development/static/locales/en-US.yaml#L128~

~To: “RSS is forced for profiles with 125 or more subscriptions to avoid rate limiting.”~

efb4f5ff-1298-471a-8973-3d47447115dc:
> Not sure if that dim out effect on specific profiles would confuse users
I don't think we have any profile specific setting so far
If we want to tell user that FT is forcing RSS for current profile (even when force RSS disabled) I think it probably should be done in another way

@PikachuEXE what about this: If users doesnt explicitly toggles RSS in the settings then it will be dimmed out and forced if limit is reached in profile A but when user switches to profile B that sits below the limit then the setting will be turned off

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc: > To disable forced RSS, reduce your subscription count or split your subscriptions across multiple profiles.

I would also like to link profiles to https://docs.freetubeapp.io/usage/profiles/

efb4f5ff-1298-471a-8973-3d47447115dc:
RSS forcing doenst exist anymore so this PR only needs to show the Posts tab and display a message.

PikachuEXE (dismissed):

static/locales/en-US.yaml
```diff
@@ -131,6 +131,8 @@ Subscriptions:
   Load More Posts: Load More Posts
   Subscriptions Tabs: Subscriptions Tabs
   All Subscription Tabs Hidden: 'All subscription tabs are hidden. To see content here, please unhide some tabs in the "{subsection}" section in "{settingsSection}".'
+  Posts:
+    RSS Message: YouTube does not provide RSS feeds for posts. Disable the "{rssSetting}" setting to view posts. Enable the "{hideSetting}" setting to hide this tab.
```
efb4f5ff-1298-471a-8973-3d47447115dc: I added an 'or' to the sentence to emphasize that the user has multiple options for resolving the current situation

```suggestion
    RSS Message: YouTube does not provide RSS feeds for posts. Disable the "{rssSetting}" setting to view posts or enable the "{hideSetting}" setting to hide this tab.
```

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@radmorecameron 👋

Shadorc (approved):

PR #8621, opened 2026-02-04
Feature: Drag and Drop

### LINKED ISSUE #4872: [Feature Request]: Drag and drop videos in edit mode

Android789515:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #4872 

## Description

Adds drag and drop ordering when custom sort is enabled for playlist videos both in list and grid view.

## Grab bar on hover (list view only)
ListHover

### List drag and drop
ListDnD

### Grid drag and drop
GridDnD

#### Testing instructions
1. Clone `https://github.com/Android789515/FreeTube.git`
2. `cd FreeTube`
3. `yarn install`
4. `yarn run dev`
5. Create a playlist and fill it with videos
6. Play around with the drag and drop feature
7. Switch to list or grid view and play around

## Desktop

- **OS: EndeavourOS Linux x86_64**
- **OS Version: Ganymede Neo | kernel -> 6.12.68-1-lts**
- **FreeTube version: v0.23.13 Beta**

### DIFF
--- a/static/locales/zh-CN.yaml
+++ b/static/locales/zh-CN.yaml
@@ -1116,5 +1116,5 @@ KeyboardShortcutPrompt:
   Skip to Next Video: 跳到播放列表中下一个视频或下一个推荐的视频
   Skip to Previous Video: 跳到播放列表中前一个视频
 shortcutLabelSeparator: ｜
-Compact side navigation: 压缩侧边栏导航
+Compact side navigation: 紧凑侧边栏导航
 Expand side navigation: 扩展侧边栏导航


Android789515:
I fixed the issue with draggable channel playlist videos. I went ahead and added something that I think will help indicate that grid playlist videos can be dragged without getting in the way or screwing up the layout. Let me know what you think. It only shows on hover.
unwatched-hover
watched-hover

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm its an interesting approach, idk tbh. I would need to test it to get a feel of it. 

@PikachuEXE @absidue opinions?

Android789515:
You want me to push my changes, so you can test it out?

PikachuEXE:
I tested a bit and the visual effect is confusing

https://github.com/user-attachments/assets/3fdefed3-9642-4aef-ab42-6f016e11a166

Android789515:
> I tested a bit and the visual effect is confusing
> Screen.Recording.2026-02-09.at.09.50.18.mp4

That's strange. I don't experience that behavior on my system.

https://github.com/user-attachments/assets/e7d42dea-eb15-4061-8cf8-f0b960d1b4f3

Could you please share your system details so I can try to replicate it?

PikachuEXE:
I am simply running `yarn dev` on macOS
Would it look like the same if I build the app? Yes
Is it macOS specific? I dunno (no access to windows yet

Custom build: https://github.com/PikachuEXE/FreeTube/actions/runs/21813666795

Android789515:
> I am simply running `yarn dev` on macOS Would it look like the same if I build the app? Yes Is it macOS specific? I dunno (no access to windows yet
> 
> Custom build: https://github.com/PikachuEXE/FreeTube/actions/runs/21813666795

I will spin up a MacOS VM and investigate this.

Android789515:
I found the culprit and fixed it. Here is a video of the fix applied. I will go ahead and push all my changes to this PR now. I fixed channel playlists being draggable (they weren't supposed to be), I added the grab bar icon to the grid view videos on hover, and I fixed this rubber banding @PikachuEXE was experiencing on MacOS.

https://github.com/user-attachments/assets/9bcbc626-bfd3-475a-b839-4a2520803b5f

PikachuEXE:
Seems fixed on macOS which is great
Wonder how it works on touch screen... (no idea how many use touch screen for this)

One small improvement I can think of is to change cursor to `grab` when hover above the grab indicator at least (whether it should be `grab` for the whole box... for now I think not)
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/cursor
image

Tested list view
Tested online playlist
Tested other sort orders..
What else

Custom build for daily testing: https://github.com/PikachuEXE/FreeTube/actions/runs/21847160650

PikachuEXE:
For some reason I can no longer right click on images and copy the URL (which I use quite often
I need to fallback to old version until that is fixed

Update 1: The original context menu, image items (middle 3) missing in this PR
2026-02-10 10_44_54-Watch Later - FreeTube

Android789515:
> For some reason I can no longer right click on images and copy the URL (which I use quite often I need to fallback to old version until that is fixed
> 
> Update 1: The original context menu, image items (middle 3) missing in this PR 2026-02-10 10_44_54-Watch Later - FreeTube

Went ahead and fixed the issue. Context menu items show up, and I added the grab cursor to the grab bar.

Android789515:
Forgot you mentioned touch screen users... as far as I know, the HTML drag and drop API only works with mouse.

image


There is a polyfill that adds touch screen support.

https://github.com/drag-drop-touch-js/dragdroptouch

Otherwise I will have to look at a library to handle this or make my own API.

Android789515:
I just checked on YouTube and they don't even have drag and drop playlist ordering for touch screen users. Granted, this is their web version. I don't want to install their app on my phone.

absidue:
The drag and drop API should work on touch screens without requiring polyfills.

Please follow the pure CSS, hover, focus and always show on touchscreens approach that we use for the existing video buttons for the drag icon too.

This is an incomplete review, I will come back and do a full code review and test the pull request later (the current event handling setup looks like it will unnecessarily recreate functions on every render).

Android789515:
> The drag and drop API should work on touch screens without requiring polyfills.
> 
> Please follow the pure CSS, hover, focus and always show on touchscreens approach that we use for the existing video buttons for the drag icon too.
> 
> This is an incomplete review, I will come back and do a full code review and test the pull request later (the current event handling setup looks like it will unnecessarily recreate functions on every render).

I will go ahead and make the grab icon use the approach for the other ones.

Could you clarify what you mean by pure CSS hover focus? I'm not sure what you mean. I tried to match whatever CSS style was being used in the stylesheets. Some are in SCSS, some are in Vanilla CSS.

Drag and Drop API does not support touch events. Touching and holding on the screen is ambiguous, it could mean select text, open context menu.. gestures come into play, multiple fingers. It can be done, but I will have to make a custom drag and drop instead of using the HTML API.

Android789515:
Here is the icon added to the stack. Which do you like? Left, middle, right, other?

ft-left
ft-middle
ft-right

absidue:
Please leave the icon where it was before. I am asking you to show and hide it with CSS like we do with the other icons, because you are currently doing it through JavaScript with mouseenter and mouseleave events.

> Drag and Drop API does not support touch events. Touching and holding on the screen is ambiguous, it could mean select text, open context menu.. gestures come into play, multiple fingers. It can be done, but I will have to make a custom drag and drop instead of using the HTML API.

Not sure where you are getting your information from (some LLM hallucination?), because demos using only the drag and drop API, no polyfills, work just fine on my mobile phone. E.g. https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Kanban_board

We know for a fact that people use FreeTube on their Linux phones and FreeTubeAndroid exists, so touch support **using the standard APIs** would be very good to have in this pull request.

efb4f5ff-1298-471a-8973-3d47447115dc:
Opinions needed on the following:

I would accept the placement of the icon to be above the thumbnail in grid view but only for custom sorting. 

Like shown in screenshot here (yes its using numbers but you get the point) https://github.com/FreeTubeApp/FreeTube/pull/4949#issuecomment-2078512221

This feels less complicated from an user perspective and easier to understand

Android789515:
> Please leave the icon where it was before. I am asking you to show and hide it with CSS like we do with the other icons, because you are currently doing it through JavaScript with mouseenter and mouseleave events.
> 
> > Drag and Drop API does not support touch events. Touching and holding on the screen is ambiguous, it could mean select text, open context menu.. gestures come into play, multiple fingers. It can be done, but I will have to make a custom drag and drop instead of using the HTML API.
> 
> Not sure where you are getting your information from (some LLM hallucination?), because demos using only the drag and drop API, no polyfills, work just fine on my mobile phone. E.g. https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Kanban_board
> 
> We know for a fact that people use FreeTube on their Linux phones and FreeTubeAndroid exists, so touch support **using the standard APIs** would be very good to have in this pull request.

It's not from AI. Touch event support for DnD is spotty and not supported on most mobile devices' browsers because the API is based on mouse events. I will still try to implement it but I don't have a linux phone or an android to test it out and see if it works. Someone else will have to do it. The thing with Electron on desktop is touch events simply won't trigger the drag API nor drag draggable elements (natively). So I think I'll ditch the vanilla DnD API.

Android789515:
For clarification this what happens when I try the example https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Kanban_board on my iPhone (Safari and Orion browsers)

https://github.com/user-attachments/assets/35ff414e-8e2a-4fa8-b964-a05e233f0aa7

Android789515:
After trying to implement touch functionality for drag and drop I found it much much easier to use this package: `@dragdroptouch/drag-drop-touch` it's two lines of code to make it work

```js
import { enableDragDropTouch } from '@dragdroptouch/drag-drop-touch'

...

enableDragDropTouch()
```

Here is a video of me testing out the DnD with touch enabled in Electron dev tools:

https://github.com/user-attachments/assets/dd5fd8cb-af6d-4554-bd24-bad256371f34

What do you think? Is it alright to use this package to add touch support?

Android789515:
I added smooth auto scrolling when dragging videos up and down.

https://github.com/user-attachments/assets/03c5cba2-82f0-48b7-b26b-d144ffae996f

Android789515:
> Opinions needed on the following:
> 
> I would accept the placement of the icon to be above the thumbnail in grid view but only for custom sorting.
> 
> Like shown in screenshot here (yes its using numbers but you get the point) [#4949 (comment)](https://github.com/FreeTubeApp/FreeTube/pull/4949#issuecomment-2078512221)
> 
> This feels less complicated from an user perspective and easier to understand

Do you want both numbers and the grab bar?

efb4f5ff-1298-471a-8973-3d47447115dc:
@Android789515 no i want the grab bar in Grid mode with the placement of the numbers shown in that picture.

Android789515:
> @Android789515 no i want the grab bar in Grid mode with the placement of the numbers shown in that picture.

Since the list view shows numbers, I figure it makes sense to reuse the component that provides the number (`FtListVideoNumbered`). The grid view renders the videos with a completely different structure, and I'm not sure why. (`FtElementList` -> `FtListLazyWrapper`)

It will help me if I understand the difference between these approaches:

`TransitionGroup` -> `FtListVideoNumbered` -> `ft-list-video`

`FtElementList` -> `FtListLazyWrapper` -> `ft-list-video`

And the difference between `FtElementList` and `FtListLazyWrapper`.

absidue:
> > @Android789515 no i want the grab bar in Grid mode with the placement of the numbers shown in that picture.
> 
> Since the list view shows numbers, I figure it makes sense to reuse the component that provides the number (`FtListVideoNumbered`). The grid view renders the videos with a completely different structure, and I'm not sure why. (`FtElementList` -> `FtListLazyWrapper`)
> 
> It will help me if I understand the difference between these approaches:
> 
> `TransitionGroup` -> `FtListVideoNumbered` -> `ft-list-video`
> 
> `FtElementList` -> `FtListLazyWrapper` -> `ft-list-video`
> 
> And the difference between `FtElementList` and `FtListLazyWrapper`.

The grid view for playlists was a later addition and as it didn't require any playlist specific functionality it was implemented using the generic FtElementList. FtListVideoNumbered is a playlist specific component, not intended to be used anywhere else in the app, so it is intentionally left out of the generic list/grid rendering components and is also hardcoded to only render the list view.

FtElementList takes a list of items (videos, playlists, channels, hashtags, posts) and creates a list of FtListLazyWrapper components. FtListLazyWrapper is a wrapper around the more specific FtListVideo, FtListChannel etc components that delays creating and rendering the more specific component until it is scrolled into view (or immediately if it is one of the first 16 items), it solves some of the performance problems that occur if you render all of the more specific components in one go without the complexity of virtual scrolling (not impossible but very hard to do with FreeTube supporting so many different screen and window sizes and the components all being dynamically sized).

If you want to render something inside the FtListVideo component you should do it inside that component, please don't try and hack it on top from another component in the chain.

absidue:
Just tested this pull request for the first time and as suspected this breaks dragging the video title and channel links, but not just on the playlist page but throughout the entire app.

Android789515:
> Just tested this pull request for the first time and as suspected this breaks dragging the video title and channel links, but not just on the playlist page but throughout the entire app.

Alright, thanks for the feedback... I wasn't aware those needed to be dragged anywhere.

absidue:
While it is not implemented in the release versions yet it is already in the development branch and the nightlies. The preview incorrectly shows the in-app link (it unfortunately doesn't update when we replace the link inside the datatransfer object) but when you drop it, it drops the equivalent youtube.com link. Dragging a link has the same uses as copying it e.g. dragging it into your web browser, a chat app, a notetaking app or even from one FreeTube window into another one if you want to open the video/playlist/channel in that window.

absidue:
As for touch, while yes Electron itself doesn't seem to have touch support on the drag-and-drop API, iOS Safari definitely does so I presume that the native Android webview that FreeTubeAndroid uses does too.

I did have to fix the drag and drop implementation in this pull request to be more complete (actually setting some drag data rather than starting an empty drag and passing the data around behind the scenes, cancelling the default drop behaviour so that JavaScript can handle the drop rather than the web view) to get it to work on iOS Safari as mentioned here: https://stackoverflow.com/a/69109382. Once I did those changes it worked fine without requiring polyfills.

Running FreeTube outside of Electron or FreeTube Android is unsupported and therefore mostly broken but for the sake of testing the drag and drop inside local playlists that doesn't require any actual internet connectivity it works alright.

1. Start FreeTube with `yarn run dev`
2. Open your test playlist
3. Click the export icon and select FreeTube (.db)
4. Close FreeTube
5. Transfer the playlists database to your phone (e.g. with `qrcp send --keep-alive path/to/playlists.db`)
6. Run `yarn run dev:web`
7. Open `http://{your computer's IP address}:9080` in the browser on your phone
8. Click on settings
9. In the general section set the thumbnail preference to hidden (they won't load anyway so this fixes the UI a bit).
10. In the data section click on import playlists and select your playlist.db file
11. Go to the user playlist page
12. Click on the playlist you just imported
13. Set the sort option to custom
14. Hold down on a video to drag it (you need to hold on an empty area next to the title for example, so that you don't drag the title text or image instead).
15. You can stop the webpack dev server with CTRL+C

Android789515:
> As for touch, while yes Electron itself doesn't seem to have touch support on the drag-and-drop API, iOS Safari definitely does so I presume that the native Android webview that FreeTubeAndroid uses does too.
> 
> I did have to fix the drag and drop implementation in this pull request to be more complete (actually setting some drag data rather than starting an empty drag and passing the data around behind the scenes, cancelling the default drop behaviour so that JavaScript can handle the drop rather than the web view) to get it to work on iOS Safari as mentioned here: https://stackoverflow.com/a/69109382. Once I did those changes it worked fine without requiring polyfills.
> 
> Running FreeTube outside of Electron or FreeTube Android is unsupported and therefore mostly broken but for the sake of testing the drag and drop inside local playlists that doesn't require any actual internet connectivity it works alright.
> 
> 1. Start FreeTube with `yarn run dev`
> 2. Open your test playlist
> 3. Click the export icon and select FreeTube (.db)
> 4. Close FreeTube
> 5. Transfer the playlists database to your phone (e.g. with `qrcp send --keep-alive path/to/playlists.db`)
> 6. Run `yarn run dev:web`
> 7. Open `http://{your computer's IP address}:9080` in the browser on your phone
> 8. Click on settings
> 9. In the general section set the thumbnail preference to hidden (they won't load anyway so this fixes the UI a bit).
> 10. In the data section click on import playlists and select your playlist.db file
> 11. Go to the user playlist page
> 12. Click on the playlist you just imported
> 13. Set the sort option to custom
> 14. Hold down on a video to drag it (you need to hold on an empty area next to the title for example, so that you don't drag the title text or image instead).
> 15. You can stop the webpack dev server with CTRL+C

That's good news. I'll remove the package I added, then. Sounds touch works fine after the changes you made.

Android789515:
> @Android789515 no i want the grab bar in Grid mode with the placement of the numbers shown in that picture.

How's this?

image

efb4f5ff-1298-471a-8973-3d47447115dc:
@Android789515 i dont want numbering on grid view...

i want the grab bar on the position of the numbering

Android789515:
Please be patient. I am trying my best to interperet what you are both requesting. I moved the icon to the spot you asked, now.

I believe I have now applied all the changes requested, except for optimizing the drag handlers:

There was a concern that this code would recreate the event handlers each render.
```js
const { dragVideo, moveDraggedVideo, afterDrag } = handleDragAndDrop(emit)
```

I would like some advice on how to optimize this. I don't know Vue very well, or what the best practices regarding this topic would be. My thoughts would be to put it in a call to computed and reference the drag handlers that way.

Edit: for video order to update in real time when dragging (eg. on drag enter), I can't use the event data transfer to track the dragged video, because it's only accessible in the drop event. That's why I chose to have a ref track the dragged video, instead of using the event data transfer. Though, I still added something to the data transfer for drag and drop to work correctly for touch devices.

From MDN:
image

efb4f5ff-1298-471a-8973-3d47447115dc:
> Please be patient. I am trying my best to interperet what you are both requesting. I moved the icon to the spot you asked, now.

Sorry if i came of rude to you, that wasnt my intention. I was being annoyed at myself for not explaining it well

Android789515:
> Please be patient. I am trying my best to interperet what you are both requesting. I moved the icon to the spot you asked, now.
> 
> I believe I have now applied all the changes requested, except for optimizing the drag handlers:
> 
> There was a concern that this code would recreate the event handlers each render.
> 
> ```js
> const { dragVideo, moveDraggedVideo, afterDrag } = handleDragAndDrop(emit)
> ```
> 
> I would like some advice on how to optimize this. I don't know Vue very well, or what the best practices regarding this topic would be. My thoughts would be to put it in a call to computed and reference the drag handlers that way.
> 
> Edit: for video order to update in real time when dragging (eg. on drag enter), I can't use the event data transfer to track the dragged video, because it's only accessible in the drop event. That's why I chose to have a ref track the dragged video, instead of using the event data transfer. Though, I still added something to the data transfer for drag and drop to work correctly for touch devices.
> 
> From MDN: image

I cached my drag handlers inside `computed`. Now that that is done, I submitted all my changes for review.

src/renderer/components/AutoScrollWrapper/AutoScrollWrapper.scss
absidue, line 1: In the rest of the code base we have been moving away from SCSS, please inline the height variable, rename the file to be a CSS file and then remove the stylelint ignore comment (please also remember to remove the `lang="scss"` attribute in the `.vue` file).

src/renderer/components/AutoScrollWrapper/AutoScrollWrapper.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+  <div>
+    <div
+      v-if="hotZoneEnabled"
+      class="hotZone topZone"
+      v-on="{
+        dragenter: () => setScrollDirection('up'),
+        dragleave: clearScrollDirection,
+      }"
```
absidue, line 9: The rest of the codebase uses the less verbose `@eventhandler` event handler style, please use that here too.

src/renderer/components/AutoScrollWrapper/AutoScrollWrapper.vue
```diff
@@ -0,0 +1,88 @@
...
+
+    <slot />
+
+    <div
+      v-if="hotZoneEnabled"
+      class="hotZone bottomZone"
+      v-on="{
+        dragenter: () => setScrollDirection('down'),
+        dragleave: clearScrollDirection,
+      }"
```
absidue, line 20: The rest of the codebase uses the less verbose `@eventhandler` event handler style, please use that here too.

src/renderer/components/AutoScrollWrapper/AutoScrollWrapper.vue
```diff
@@ -0,0 +1,88 @@
...
+    clearInterval(scrollInterval)
+  })
+})
+</script>
+
+<style
+  scoped
+  src="./AutoScrollWrapper.scss"
+  lang="scss"
+/>
```
absidue, line 88: Please write this on a single line, like in the rest of the code base.
```suggestion
<style scoped src="./AutoScrollWrapper.css" />
```

src/renderer/components/FtListLazyWrapper/FtListLazyWrapper.vue
```diff
@@ -294,6 +329,17 @@ function moveVideoDown(videoId, playlistItemId) {
 function removeFromPlaylist(videoId, playlistItemId) {
   emit('remove-from-playlist', videoId, playlistItemId)
 }
+
+const { value: { dragVideo, moveDraggedVideo, afterDrag } } = computed(() => handleDragAndDrop(emit))
```
absidue, line 333: Please undo these unnecessary computed changes (in both files), that section only runs once per component instance anyway. My original comment was actually about the event handlers in the template. Vue uses the virtual DOM rendering approach, do the `<template>` section gets compiled into a function that is run every time the component renders and the output is diffed against what was returned last time and the changes are applied to the real DOM. So in this case because all 5 of the newly added event handlers are inline, instead of being written in the `<script>` section, the get recreated and as JavaScript functions are objects, two identical function definitions are considered two different objects, so Vue will remove and re-add the event handlers on every render because it thinks they have changed.

So to summarise please undo the computed changes and move the object containing those 5 event handlers into a variable in the `<script>` section and reference that in the template.

src/renderer/components/ft-list-video/ft-list-video.vue
absidue, line 1: I haven't tested it yet but based on the fact that you are still cancelling link drags in this file I strongly suspect that link dragging still won't work properly. The link dragging handlers are in the `App` component, so the drag events need to get all the way up there without being blocked or intercepted in-between.

src/renderer/views/Playlist/Playlist.vue
```diff
@@ -85,6 +87,15 @@
...
             :playlist-items-length="shownPlaylistItems.length"
             :can-remove-from-playlist="true"
+            :dragged-video="draggedVideo"
+            :is-sort-order-custom="isSortOrderCustom"
+            :is-video-dragging="isVideoDragging()"
+            @drag-video="setDraggedVideo"
+            @drag-video-end="() => setDraggedVideo({
+              videoId: null,
+              playlistItemId: null,
+            })"
```
absidue, line 97: Please define this function once in the `<script>` section and reference it in both places.

src/renderer/views/Playlist/Playlist.vue
```diff
@@ -750,6 +776,50 @@ function moveVideoDown(videoId, playlistItemId) {
...
+
+  const procedure = moveDown ? moveVideoDown : moveVideoUp
+
+  const difference = moveDown ? draggedOverIndex - droppedIndex : droppedIndex - draggedOverIndex
+
+  // Need to let the store dispatch each call
+  // to make grid layout work
+  range(0, difference).forEach(() => {
+    procedure(droppedVideoId, droppedPlaylistItemId)
+  })
```
absidue, line 820: Please refactor this to do a single call to the store and database, duplicate code from move up or move down if you have to, but please don't spam writes to the Vuex store and database/disk, apart it being an inefficient way to achieve what can be done in a single operation, each write will also trigger watchers and recomputation in everything that watches the playlists in the store (so every FtListVideo component on screen, as it needs to check if the video is in the quick bookmark playlist).

Additionally for the future if you want to run a loop a certain number of times, a standard for loop is a lot more efficient than creating a holey array, filling it with numbers and then looping over it with Array#forEach.

absidue (changes requested):

Android789515:
You can apply the changes yourself, or I can close this pull request. I don't appreciate the way you have been treating me this whole time. There are much kinder ways to say these things. Though, I appreciate the few kind things you have said.

You didn't read my comment on why I chose to dispatch each call. **Consider that and then tell me what I can do to fix it because I did refactor move video up and move video down but grid videos were not moving unless I let the store dispatch. And I don't know why**

Creating an array of n items is not that inefficient it just creates an array with a length. I definitely didn't need to use range populated with numbers for the approach, but it's much more terse and is negligable in performance. The numbers will also be useful in other cases if it's used elsewhere.**You could have simply asked me to use a loop and that would have been fine. No extra commentary.**

I'd appreciate more communcation regarding the link dragging issue as whatever I have tried has been blind as this is not a repo I have worked on. And every change I make is met with passive aggressiveness.

Android789515:
This PR is all yours. I don't want to contribute anymore. I hope you guys enjoy continuing work on this. I found drag and drop very helpful for custom sorting.

absidue:
I am sorry that our feedback asking for maintainable, performant code and decent UI/UX were not to your liking. If you change your mind in the future, may I suggest choosing a smaller task as that will receive less feedback, so you might be more motivated to stick with it until it has been merged.

You always need to dispatch to the store and database when you change data in FreeTube, otherwise the changes won't be saved. I am unable to find a comment from you that explains why you were unable to do a single store dispatch for the "entire" move instead of repeatedly dispatching a move offset by one (your comment just says that you need to dispatch to the store). It is definitely possible to do it in one go I would not have asked for the change otherwise.

Android789515:
1. Maintainable: this is why I extracted event handlers into a helper. This is why I made my auto scroll a component. This is why I JSDoc-ed everything.
2. Performant: I tried to address your concern of recreating functions each render (which I don't see why that's an issue anyway unless you have a bottleneck). My range approach for moving videos has a time complexity of 2N which is still O(n). Same as a for loop.
3. Decent UI/UX: I have no problem doing this.

My issue is your behavior, and how you chose to communicate by condescending.

Android789515:
I went ahead and checked the box to allow edits by maintainers. That way if you all want to continue this feature, you can.

Issue #6661, opened 2025-01-27
[Feature Request]: Automated color contrast accessibility testing


kommunarr:
### Guidelines

- [x] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

The addition of new themes can introduce many different types of accessibility issues with regards to the color contrast. This can be time-intensive for contributors to manually test and identify.

### Proposed Solution

Implement automated digital accessibility testing for themes. This means:
- For base themes with related primary & secondary theme colors (e.g., Dracula & Dracula Cyan), [require that there is a 3+:1 color contrast](https://www.w3.org/TR/WCAG20-TECHS/G183.html) between them.
- Ensure that each base theme's text colors [have a 4.5+:1 (WCAG AA) or 7+:1 (WCAG AAA) color contrast](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) with both the `card-bg-color`, `bg-color`, and `secondary-card-bg-color`.
- Require that each primary & secondary theme's text-with color [has a 4.5+:1 (WCAG AA) or 7+:1 (WCAG AAA) color contrast](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html) with them.
- (arguable / post-MVP because it is not followed by many of our themes currently) [Require that there is a 3+:1 color contrast](https://www.w3.org/TR/WCAG20-TECHS/G183.html) with each base theme's `search-bar-color` and `secondary-card-bg-color` against the `card-bg-color`.

Automated enforcement of this as a part of our build process can be done as part of the same effort or split into a separate PR if deemed appropriate.

### Alternatives Considered

Don't implement this feature.

### Issue Labels

new feature

### Additional Information

I suggest that this be done before merging any future theme PRs. Otherwise, any new theme PRs need to have exhaustive link & screenshot evidence of manual accessibility testing for all relevant color combinations before being considered ready to review.

PR #8664, opened 2026-02-14
Setup playwright tests for accessibility testing

radmorecameron:
## Pull Request Type
- [x] Other

## Related issue
related to: #6661 (a static analyzer of some sort might be better for this use case though?)

## Description
This PR sets up initial accessibility testing through playwright and axe.

The tests are currently failing due to accessibility issues, I fixed one of the issues but there's still 2 accessibility issues for mobile/tablet and 4 for desktop.

## Testing

To run the tests, you can use `yarn run test:e2e`, you may need to run `npx playwright install ` first

## Future

This is just an initial implementation but in the future it'd probably be worth it to:
- run these tests in CI (only if UI code is touched? ie: not dependabot)
- add more tests (ex: tests that change theme)
- mock API requests so more can be tested

## Desktop

- **OS:** Fedora Linux
- **OS Version:** 43


radmorecameron:
Oh yeah for added context these are the failures:

For mobile/tablet it doesn't like:
- no h1 on page
- multiple nav elements (/ role = navigation) but no aria-label 

```json
[
  {
    "description": "Ensure landmarks are unique",
    "help": "Landmarks should have a unique role or role/label/title (i.e. accessible name) combination",
    "helpUrl": "https://dequeuniversity.com/rules/axe/4.11/landmark-unique?application=playwright",
    "id": "landmark-unique",
    "impact": "moderate",
    "nodes": [
      {
        "all": [],
        "any": [
          {
            "data": { "accessibleText": null, "role": "navigation" },
            "id": "landmark-is-unique",
            "impact": "moderate",
            "message": "The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable",
            "relatedNodes": [
              {
                "html": "<div data-v-2ac94927=\"\" data-v-4afaba5a=\"\" data-v-2fa9d48e=\"\" class=\"ft-flex-box sideNav closed\" role=\"navigation\">",
                "target": [".sideNav"]
              }
            ]
          }
        ],
        "failureSummary": "Fix any of the following:The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable",
        "html": "<nav data-v-50bc178e=\"\" data-v-2fa9d48e=\"\" class=\"topNav\">",
        "impact": "moderate",
        "none": [],
        "target": ["nav"]
      }
    ],
    "tags": ["cat.semantics", "best-practice"]
  },
  {
    "description": "Ensure that the page, or at least one of its frames contains a level-one heading",
    "help": "Page should contain a level-one heading",
    "helpUrl": "https://dequeuniversity.com/rules/axe/4.11/page-has-heading-one?application=playwright",
    "id": "page-has-heading-one",
    "impact": "moderate",
    "nodes": [
      {
        "all": [
          {
            "data": null,
            "id": "page-has-heading-one",
            "impact": "moderate",
            "message": "Page must have a level-one heading",
            "relatedNodes": []
          }
        ],
        "any": [],
        "failureSummary": "Fix all of the following:Page must have a level-one heading",
        "html": "<html lang=\"en-US\">",
        "impact": "moderate",
        "none": [],
        "target": ["html"]
      }
    ],
    "tags": ["cat.semantics", "best-practice"]
  }
]
```

For desktop, it's a lot of buttons without text/titles/labels and colour contrast issues as well

absidue (changes requested):
The biggest issue I see with this pull request at the moment is that instead of running inside Electron like FreeTube does, it uses the unsupported, mostly broken web build. Unless it is possible to do it inside Electron, I don't think it is worth merging this, as running tests in an environment that doesn't match how FreeTube is actually used and where most of the functionality is unsupported and broken, is not a proper test.

Additionally could you please rename the pull request and code references to say accessibility testing, so it actually matches what it does (calling accessibility tests, in the wrong environment on only 3 pages "end to end tests" is quite misleading)

P.S. We only have the web build in-place in CI because it serves as a smokescreen for downstream FreeTubeAndroid, to check that Electron only APIs are properly guarded by build flags. FeeeTubeAndroid has reimplemented a lot of the Electron functionality with Android APIs instead so it doesn't use the web build either.

radmorecameron:
> Additionally could you please rename the pull request and code references to say accessibility testing, so it actually matches what it does (calling accessibility tests, in the wrong environment on only 3 pages "end to end tests" is quite misleading)

Thanks for looking at this @absidue :smile: 

1) It should now be running through electron instead (the only drawback to this is that the tests can't be ran in parallel)
2) For now it's only accessibility tests but it can definitely expand to other types of tests so I don't think it'd be a good idea to make the code references "accessibility tests" but I think that makes sense for the PR title

package.json
```diff
@@ -51,7 +51,9 @@
     "pack:botGuardScript": "webpack --config _scripts/webpack.botGuardScript.config.js",
     "postinstall": "yarn run --silent patch-shaka",
     "checkforbadtemplates": "node _scripts/findMissingTemplates.mjs",
-    "ci": "yarn install --silent --frozen-lockfile --network-concurrency 1"
+    "ci": "yarn install --silent --frozen-lockfile --network-concurrency 1",
+    "test:playwright": "run-s pack test:playwright-nopack",
+    "test:playwright-nopack": "playwright test"
```
radmorecameron, line 54: I found the pack command to be a bit slow on my machine so i added an option for "no-pack" so pack doesn't run every time you want to run the test (ex: if you're only working on tests and not updating code). Let me know if this should be removed.

eslint.config.mjs
```diff
@@ -14,10 +14,12 @@ import eslintPluginYml from 'eslint-plugin-yml'
 import { neostandard } from 'neostandard/lib/main.js'
 import jsdoc from 'eslint-plugin-jsdoc'
 import freetube from './_scripts/eslint-rules/plugin.mjs'
+import eslintPluginPlaywright from 'eslint-plugin-playwright'
 
 import activeLocales from './static/locales/activeLocales.json' with { type: 'json' }
+import { defineConfig } from "eslint/config";
 
-export default [
+export default defineConfig([
```
radmorecameron, line 23: `defineConfig` allows `extends` to work for the playwright section

package.json
absidue: Rather than creating a release build and using that to run the tests, a different approach could be to use the existing dev runner script with the `--remote-debug` arg in combination with `playwright.chromium.connectOverCDP`, although you then have to worry about closing the dev runner script afterwards (you would want to run the dev runner script once, connect once and reuse the same connection and instance for all tests, they don't run in parallel anyway)...

tests/playwright/helpers.mjs
```diff
@@ -0,0 +1,59 @@
...
+    await electronApp.close()
+  })
+}
+
+/**
+ * @param {string} name
+ * @param {(function({ app: import('@playwright/test').ElectronApplication, page: import('@playwright/test').Page }) : Promise<void>)} func
+*/
+export const electronDesktopTest = async (name, func) => {
+  await electronTest(name, devices['Desktop Firefox'].viewport, func)
```
absidue: Lets use `Desktop Chrome` here as that is closer to reality (Electron is based on Chromium).

eslint.config.mjs
```diff
@@ -14,10 +14,12 @@ import eslintPluginYml from 'eslint-plugin-yml'
 import { neostandard } from 'neostandard/lib/main.js'
 import jsdoc from 'eslint-plugin-jsdoc'
 import freetube from './_scripts/eslint-rules/plugin.mjs'
+import eslintPluginPlaywright from 'eslint-plugin-playwright'
 
 import activeLocales from './static/locales/activeLocales.json' with { type: 'json' }
+import { defineConfig } from "eslint/config";
```
absidue: Please move the `eslint/config` import to the top of the imports list and `eslint-plugin-playwright` one above the FreeTube one, as external imports being above the local imports is a pattern that we have for most of the JavaScript files in this repo.

package.json
absidue, line 1: Please include the `test-results` directory that you added to the `.gitignore` file in the clean script.

absidue, line 1: Please rebase on the development branch to pull in the latest changes to the `package.json` scripts.

absidue (commented):
In the future it would probably be good to add a test build to FreeTube, that uses in-memory nedb databases rather than a production build that shares the development data stores. That's not a major issue for this pull request as it doesn't modify any significant user data but it would definitely need to be done before any tests that do actual data changes are implemented.

package.json
radmorecameron: I'll try to look into this a bit more sometime next week (and if I run into too many issues, I'll look into a separate build for tests instead) :smile:

radmorecameron:
I went for the test build (almost exactly the same as the production build, just uses an inmemory database). I think reusing the same dev runner might cause some headaches in the future if tests are added that modify state (we'd need to reset the state of the db after a test is ran)

_scripts/webpack.main.config.js
```diff
@@ -5,74 +5,77 @@ const JsonMinimizerPlugin = require('json-minimizer-webpack-plugin')
 
 const isDevMode = process.env.NODE_ENV === 'development'
 
-/** @type {import('webpack').Configuration} */
```
radmorecameron, line 8: I'd recommend hiding whitespaces when looking at the changes to this file.
Config was just wrapped in a method that is exported so the IS_TESTS env variable could be set properly: https://webpack.js.org/guides/environment-variables/ (I followed what was in this guide)

efb4f5ff-1298-471a-8973-3d47447115dc:
@radmorecameron Thank you for putting so much work into this. 

That being said this reminds me of another situation we've experienced in the past where we setup tests for something but didnt address the issues pointed out by the tests before merging the tests pr itself. This resulted in tests failing on every PR which wasnt very helpful as the tests failed on stuff that were unrelated to the changes and ultimately made us turn off the tests

Ill propose the following to prevent getting into the same situation again:
We shelve introducing the tests for now and focus on addressing the issues pointed out by the tests. After those are addressed we can come back to this.

> The tests are currently failing due to accessibility issues, I fixed one of the issues but there's still 2 accessibility issues for mobile/tablet and 4 for desktop.

Feel free to create a PR for the one you fixed already

PR #8673, opened 2026-02-15
Add SponsorBlock unskip and reskip feature

### LINKED ISSUE #1380: SponsorBlock - Auto-skip as an option

Pillowytuba:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #1380

## Description

Adds unskipping and reskipping segments from SponsorBlock. When a segment is skipped it gives the option to press enter in unskip it. When watching the segment you can also choose to press enter again to reskip the segment. This feature is present in the SponsorBlock extension.

## Testing

1. Enable SponsorBlock in settings
2. Go to any video with a segment to auto skip.
3. When the segment skips you will see the option to unskip
4. Once you press enter you will also see the option to reskip

## Desktop

- **OS: Arch Linux**
- **OS Version: N/A**
- **FreeTube version: v0.23.13 Beta**

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -479,7 +479,7 @@ export default defineComponent({
      * @type {import('vue').Ref<{uuid: string, translatedCategory: string, timeoutId: number, startTime: number, endTime: number, unskipped: boolean}[]>}
      */
     const skippedSponsorBlockSegments = ref([])
-    const ignoredSponsorBlockSegments = ref(new Set())
+    const ignoredSponsorBlockSegments = reactive(new Set())
 
     async function setupSponsorBlock() {
       let segments, averageDuration
@@ -511,7 +511,7 @@ export default defineComponent({
     function skipSponsorBlockSegments(currentTime) {
       skippedSponsorBlockSegments.value = skippedSponsorBlockSegments.value.filter(segment => {
         if (segment.unskipped && (currentTime >= segment.endTime || currentTime < segment.startTime)) {
-          ignoredSponsorBlockSegments.value.delete(segment.uuid)
+          ignoredSponsorBlockSegments.delete(segment.uuid)
           return false
         }
         return true
@@ -529,7 +529,7 @@ export default defineComponent({
       const skippedSegments = []
 
       sponsorBlockSegments.forEach(segment => {
-        if (!ignoredSponsorBlockSegments.value.has(segment.uuid) && autoSkip.has(segment.category) && currentTime < segment.endTime &&
+        if (!ignoredSponsorBlockSegments.has(segment.uuid) && autoSkip.has(segment.category) && currentTime < segment.endTime &&
           (segment.startTime <= currentTime ||
             // if we already have a segment to skip, check if there are any that are less than 150ms later,
             // so that we can skip them all in one go (especially useful on slow connections)
@@ -565,7 +565,7 @@ export default defineComponent({
               clearInterval(segment.intervalId)
               skippedSponsorBlockSegments.value.splice(index, 1)
             }
-            ignoredSponsorBlockSegments.value.delete(uuid)
+            ignoredSponsorBlockSegments.delete(uuid)
           }
 
           const existingSkip = skippedSponsorBlockSegments.value.find(skipped => skipped.uuid === uuid)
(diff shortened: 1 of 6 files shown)


src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -476,9 +476,10 @@ export default defineComponent({
     /**
      * Yes a map would be much more suitable for this (unlike objects they retain the order that items were inserted),
      * but Vue 2 doesn't support reactivity on Maps, so we have to use an array instead
-     * @type {import('vue').Ref<{uuid: string, translatedCategory: string, timeoutId: number}[]>}
+     * @type {import('vue').Ref<{uuid: string, translatedCategory: string, timeoutId: number, startTime: number, endTime: number, unskipped: boolean}[]>}
      */
     const skippedSponsorBlockSegments = ref([])
+    const ignoredSponsorBlockSegments = ref(new Set())
```
absidue: When you are making a collection like a `Map` or `Set` deeply reactive and you never reassign the variable it is better to use `reactive` (saves needing to write `.value` everytime and has slightly less overhead because Vue doesn't need to track the `.value` access). Remember to remove `.value` from all of the accesses below.
```suggestion
    const ignoredSponsorBlockSegments = reactive(new Set())
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -547,26 +556,46 @@ export default defineComponent({
...
-              skippedSponsorBlockSegments.value.splice(index, 1)
-            }, 2000)
+            existingSkip.timeLeft = 4
+            existingSkip.intervalId = setInterval(() => {
+              existingSkip.timeLeft--
+            }, 1000)
+            existingSkip.timeoutId = setTimeout(removeSegment, 4000)
           } else {
-            skippedSponsorBlockSegments.value.push({
+            const segmentObj = reactive({
```
absidue: `reactive()`, `ref()`, `computed()` etc should only ever be used at the top level (directly inside the setup function) so that they get registered to the correct component and cleaned up when the component is unmounted/destroyed. Dynamically creating a reactive object like this means that it will either never get cleaned up (so a memory leak) or registered to another component creating completely unpredicatable behaviour because it will track that components lifecycle rather than this one.

In this case a better approach would be to add the object to the reactive array first, so it becomes reactive by the deep-reactivity applied to the array and then get the reactive proxy from the array to use in the `setInterval()` callback.

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2275,6 +2304,28 @@ export default defineComponent({
       }
 
       switch (event.key.toLowerCase()) {
+        case 'enter': {
```
absidue: Please register the keyboard shortcut correctly, so that it shows up in the keyboard shortcut list in the settings (you can see how it is done for all the other ones). Please also open a pull request to the FreeTubeApp/FreeTube-Docs repository to mention this new keyboard shortcut.

efb4f5ff-1298-471a-8973-3d47447115dc:
requires rebase to make playback work

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Hi @Pillowytuba apologies for the delay there were allot of things going on behind the scenes. 

The reskip notification in your implementation doesnt have a timer. I would like to see that timer implemented there also because now the reskip notification stays there until the segment is done but the user doesnt really knows that

https://github.com/user-attachments/assets/154e3e32-86fa-49e0-b5c4-6d8dafb691c1

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
If i skip or reskip sometimes the message and counter will disappear. When using the extension in the browser the message+counter will never disappear except for when user dismisses it manually

efb4f5ff-1298-471a-8973-3d47447115dc:
> If i skip or reskip sometimes the message and counter will disappear. When using the extension in the browser the message+counter will never disappear except for when user dismisses it manually

Sorry for the late review. Could you also add the manual dismiss

Issue #6699, opened 2025-01-31
[Feature Request]: Bookmark Exisiting Playlists


Privacy-Now:
### Guidelines

- [x] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

Channels have existing playlists that may continue to have videos added to them upon release by their content creators.

In order to save/bookmark the playlist to come back to easily I have to recreate the playlist and add the videos manually to the playlist that I created locally. If new videos were to be added to the original channel playlist from the content creator then I must periodically return to the channel, check for a new video, and manually add it to my custom playlist.

### Proposed Solution

Let people save an existing channel playlist to our playlist tab and then have it update with new videos as the content creator adds them

### Alternatives Considered

Display a notification and/or badge counter that a new video was added

### Issue Labels

ease of use improvement

### Additional Information

Example: MangoJelly Solutions for FreeCAD channel has a playlist called "[Basic Beginners FreeCAD v1 (Public)](https://youtube.com/playlist?list=PLWuyJLVUNtc3UYXXfSglVpfWdX31F-e5S)". 

I am learning how to use the software so I wanted to save the playlist so I can easily access it to practice the next lesson. 

I had to create a new playlist for myself, name it the same thing, and manually add each video to it and add the playlist's description as well. At the time I created it the content creator had 24 lessons in the original. I recently checked back on his original playlist and found that I was not aware of lesson 25 that released 5 days prior, I then manually added lesson 25 to my personal recreation of said playlist.

efb4f5ff-1298-471a-8973-3d47447115dc:
duplicate of https://github.com/FreeTubeApp/FreeTube/issues/312

ArthurKun21:
I think the playlist could make use of another key-value with "url" to represent the youtube url of the playlist

```json
{
    "playlistName":"Playlist name",
    "description":"",
    "videos":[],
    "url": ""
}
```

so that it would check for the url first and if it is present and has valid youtube url link, it would load it otherwise it would fallback to the videos which is what the current behavior is.

This should also make it easier to be backup and restore

PR #8680, opened 2026-02-17
 Feature: Subscribe to playlists #312

### LINKED ISSUE #312: Subscribe to playlists

Finebouche:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #312

## Description
Implemented a button to subscribe to playlist. The playlist now appear in the sidebar with the other channels. 
The main feed of the app now update the playlist content as well as the channel contents. 

I tried to stay as close as possible to the original channel subscription logic. 

It still needs some tweaks, but I wanted to publish this first to be sure I was going to the right direction. 

## Testing
You can subscribe to any playlist and check if the feed is updated. 

To be decided : 
- Remove the subscribe button if the channel is already subscrided ? 
- Fix duplicates directly in DB managment ? 

To do :
- Cosmetic points for the sidebar
- same update on shorts and lives ?

## Desktop

- **OS:** macos 
- **OS Version:** 26.2
- **FreeTube version:** 0.23.13


absidue (changes requested):
While I know that the original feature request was to subscribe to playlists on the subscription page, a lot has changed in FreeTube in the last 7 years and rather than pretending that playlists are channels and hacking them into the subscription page it would be much better if they were treated like user playlists/playlists created inside FreeTube are (they would be a special kind of user playlist that is read-only and would essentially just act as a bookmark for the playlist on YouTube), that is also a lot closer to how it is handled by YouTube themselves.

Finebouche:
> While I know that the original feature request was to subscribe to playlists on the subscription page, a lot has changed in FreeTube in the last 7 years and rather than pretending that playlists are channels and hacking them into the subscription page it would be much better if they were treated like user playlists/playlists created inside FreeTube are (they would be a special kind of user playlist that is read-only and would essentially just act as a bookmark for the playlist on YouTube), that is also a lot closer to how it is handled by YouTube themselves.

Not sure what this means, as a personnal preference I would want to have the video agregated in the main subcription feed to follow when my favorite shows have a new video on huge channels that have many. 

Do you suggest to just have them in the list of playlists ? And that's it ?

absidue:
> Not sure what this means, as a personnal preference I would want to have the video agregated in the main subcription feed to follow when my favorite shows have a new video on huge channels that have many.
>
> Do you suggest to just have them in the list of playlists ? And that's it ?

Correct, yes. Cramming playlists into the channel subscriptions page is something you are welcome to do and maintain in a custom personal build for yourself but doesn't belong in upstream FreeTube. The goal is to match YouTube's behaviour in this regard (so just remembering playlists), not to turn the subscriptions page into a generic RSS reader.

k0zyrev:
> The goal is to match YouTube's behaviour in this regard (so just remembering playlists), not to turn the subscriptions page into a generic RSS reader.

When _subscribing_ to a playlist, it would make more sense if new content appeared somewhere on the "_Subscriptions_" page. What if add another tab called "Playlists" and it would be obvious that these playlists are dynamic ones which you have subscribed to?

image

Finebouche:
Looks like previous comment was deleted but indeed, would it work to have the behavior described by @absidue  AND as well to have a separate place were channels videos are aggregated ? 

(it could also be toggled in the settings ?) 

What do you think ?

absidue:
@Finebouche What I said above still applies, suggesting to hide it behind an option doesn't change that.

absidue:
I have closed the feature request to avoid future misunderstandings and re-opened one that is more inline with what I said above.

I will leave this pull request open for a while, in case you want to implement the requested changes, if not feel free to close this pull request.

efb4f5ff-1298-471a-8973-3d47447115dc:
#6699

Finebouche:
Alright, so I added the subscribed playlist in the User playlist as a seperate card.

Capture d’écran 2026-03-13 à 01 14 20

I also added a item to the subscription parameters, when "on" Freetube add the playlist vids to subscription page, (by default it's off so only the channel are added).

Capture d’écran 2026-03-13 à 01 16 00

edit: last commit display the correct number of videos in the playlists

absidue (changes requested):
I'm going to give you the benefit of the doubt and assume you just misread what I said rather than actively choosing to ignore it.

Please get rid of the setting and all connections to the subscriptions (as I already previously mentioned). It should be stored in the playlist database, fed through the playlist datastore handlers, the playlist IPC channels and stored in the playlists store and then only shown on the playlists page. Additionally it should have a dedicated button on the playlist page (near where the copy playlist button is), not be using the subscription button.

Once the idea behind the pull request is working in the right direction then we can start doing code reviews and test the pull request.

Finebouche:
Hey, I actually completly skip your comment yesterday when I started working again on the PR and then got the email of notificaiton only this morning, so my previous comment and commits were not taking your reply into account unfortunatly.

Finebouche:
However, the feature i wanted to implement was "Playlists subscription", not "Playslist bookmark" so yes, this Pull Request can be trashed I guess

efb4f5ff-1298-471a-8973-3d47447115dc:
> Hey, I actually completly skip your comment yesterday when I started working again on the PR and then got the email of notificaiton only this morning, so my previous comment and commits were not taking your reply into account unfortunatly.

Ah that is very unfortunate indeed

> However, the feature i wanted to implement was "Playlists subscription", not "Playslist bookmark" so yes, this Pull Request can be trashed I guess

@Finebouche Just to be clear, are you saying that this can be closed because #6699 is not what you would like to implement?

Finebouche:
Yes, I am closing it

Finebouche:
also if Youtube ever allow to subscribe to Podcast (which are basically playlist) we can reuse this code I guess

PR #8682, opened 2026-02-17
Improve cache handling in GitHub Actions workflows

absidue:
## Pull Request Type

- [x] Security improvement

## Description

Currently we read and write the yarn cache in the build, linter and release workflows, however as anyone can submit a pull request, this enables a potential scenario where someone can get something written to the cache in a pull request through the linter workflow and restored from the cache in the build and release workflows, which we don't want (known as cache poisoning).

This pull request switches to the following caching behaviour:
- Linter workflow: can only read caches, which keeps linting "fast" in pull requests that don't change the dependencies
- Build workflow: can read and write caches as it runs only runs trusted branches or when manually triggered
- Release workflow: To be on the safe side, this workflow does not read or write caches

## Testing

- Successful run of the build workflow with no matching caches (save only): https://github.com/absidue/FreeTube/actions/runs/22118913662
- Successful run of the build workflow with matching caches (restore only): https://github.com/absidue/FreeTube/actions/runs/22119253812

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,20 +59,38 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
-      with:
-        persist-credentials: false
-
     - name: 'Use faster D: drive for yarn cache on Windows'
       if: startsWith(matrix.os, 'windows')
       shell: cmd
       run: yarn config set cache-folder D:\ft_yarn_cache
 
+    - name: Get yarn cache directory
+      id: cache_dir
+      shell: bash
+      run: |
+        {
+          echo 'cache_dir<<EOF'
+          yarn cache dir
+          echo EOF
+        } >> "$GITHUB_OUTPUT"
+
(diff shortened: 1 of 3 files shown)


PikachuEXE:
Got wrong cache key?
https://github.com/absidue/FreeTube/actions/runs/22118913662/job/63934026545

Update 1: Just checked windows & linux, all `runner.arch` sticked to one version only for some reason

absidue:
That is the same as it was before. Because we don't include anything arch specific from npm in FreeTube builds other than Electron, which electron-builder downloads itself, we only care about the OS and arch of the build machine when saving and restoring the cache (we need to be able to run the linters and build tools on that machine).

Windows x64 and arm64 are both built on the Windows x64 runner, macOS x64 and arm64 are both built on the macOS arm64 runner and Linux x64, arm64 and arm7l are all built on the Linux x64 runner. Which means there are only 3 different runner types and only 3 different cache keys, so if the Windows x64 finishes before Windows arm64, Windows x64 will save the cache and Windows arm64 will log a warning about the cache already existing.

The same warning used to appear for the old caching solution too, just only in the logs instead of as a big warning on the page: https://github.com/FreeTubeApp/FreeTube/actions/runs/21997506267/job/63561267885

ChunkyProgrammer (approved):

Issue #8527, opened 2026-01-16
[Bug]: API error on Mac


Libellule1971:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

When clicking on a video I have this message since few days (with all instances): 
Local API Error (Click to copy): Error: UNPLAYABLE Video unavailable

Get back to Invidious API

Invidious API Error (Click to copy): SyntaxError: Unexpected token 'A', "<API disabled"... is not valid JSON

### Expected Behavior

I usually can read any video with a short delay.

### Issue Labels

API issue

### FreeTube Version

v0.23.12 Bêta

### Operating System Version

Tahoe 26.2 (macOS)

### Installation Method

.dmg

### Primary API used

Local API

### Last Known Working FreeTube Version (If Any)

v0.23.11

### Additional Information

<img width="1367" height="856" alt="Image" src="https://github.com/user-attachments/assets/0f788409-cc67-4577-867e-76c1c10b6081" />

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Libellule1971:
I aslo have sometimes IP blocked issue

efb4f5ff-1298-471a-8973-3d47447115dc:
@Libellule1971 so to be clear what is your Primary API set to? Are you receiving the video unavailable error for every video?

bioked:
I get the same error, and on every video. I'm on Local API but invidious also doesn't work.

bioked:
- Local API Error: Error: [UNPLAYABLE] Video unavailable
-  Falling back to Invidious API
- Invidious API Error: TypeError: Failed to fetch
- Legacy formats are not available for this video

Libellule1971:
[...]

Issue #7883, opened 2025-08-22
[Bug]: Local API Error : Error This Video is unavailable.


Icraft343:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

 It also gave me the Invidious Error as well. I am using v0.23.7 Beta of Freetube. This happens for all videos I watch.
Local API Error: This video is unavailable
Falling back to Invidious API
Invidious Error: SyntaxError: Unexpected token 'A', "API disabled" is not valid JSON
Legacy formats are not available for this video

### Expected Behavior

I am a bit new to this, but I tried changing Invidious instances and using VPNs, but not much changes.

### Issue Labels

API issue

### FreeTube Version

v0.23.7 

### Operating System Version

Windows 10 22H2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Issue #5720, opened 2024-09-20
[Bug]: SyntaxError: Unexpected token '<', "<html> <h"... is not valid JSON (when wanting to use a working Invidious instance)


linlinxza:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. Add any Invidious instance.
2. Try playing a video

### Expected Behavior

Should just play video.

### Issue Labels

API issue, feature stopped working

### FreeTube Version

v0.21.3 Beta

### Operating System Version

KDE Neon (Ubuntu 22.04)

### Installation Method

Flathub

### Primary API used

Invidious API

### Additional Information

Presents this error:

SyntaxError: Unexpected token '<', "<html>
<h"... is not valid JSON

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

efb4f5ff-1298-471a-8973-3d47447115dc:
Duplicate of #5534

IV instances are down because YT changed something again

linlinxza:
> Duplicate of #5534
> 
> IV instances are down because YT changed something again
> 
> Use the Local API

Thankfully, with all the nonsense Google is doing, I really don't care all that much for YouTube. Thank goodness there is Odysee and Peertube.

Issue #5506, opened 2024-08-01
[Bug]: Invidious API Error: SyntaxError: Unexpected end of JSON input


gokuthehedgehog56:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

1. With Invidious set as API, select a video
2. Video will error, giving a syntax error

### Expected Behavior

Video plays as normal

### Issue Labels

API issue

### FreeTube Version

v0.21.2 beta

### Operating System Version

Windows 10 Home, version 22H2

### Installation Method

.exe

### Primary API used

Invidious API

### Nightly Build

- [ ] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

thingsiplay:
This depends on the Invidious instance and needs to be fixed there. I already found a working instance and currently watch videos with FreeTube using https://yt.artemislena.eu as the instance. So from FreeTube's side, this is working with Invidious API.

Prazgreen1:
not planned? This is a freetube issue and not an invidious issue as invidious instances I have tested to work have this issue in freetube.

efb4f5ff-1298-471a-8973-3d47447115dc:
@Prazgreen1 please refrain from commenting when you don't understand the inner workings of the application

Prazgreen1:
Nvm about my previous comment. It was not reasonable. Anyways, I checked every invidious instance currently online that I could find, and all had this issue. I misread the above comment thinking it said that freetube was the issue rather than the invidious instance. I am mostly familiar with web frontend, however considering that using the invidious api with the public instances currently online, the option to embed videos from invidious could be a decent alternative patch

PR #8756, opened 2026-03-06
Fix: handle non-JSON Invidious API responses gracefully

Houmgaor:
## Summary

- When an Invidious instance returns non-JSON responses (e.g. plain text `"API disabled"`), the app currently crashes with `SyntaxError: Unexpected token 'A', "API disabled" is not valid JSON`
- This change reads the response as text first, checks `response.ok`, and wraps `JSON.parse` in a try/catch, so the actual server message is surfaced as the error instead of a cryptic JSON parse failure
- This is the single most commonly reported error in the app, affecting users whose Invidious instances have disabled their APIs

relates to #7883, #8527, #5720, #5506

## Test plan

- [ ] Configure an Invidious instance that has its API disabled and verify the error message now shows the actual server response (e.g. "API disabled") instead of a JSON SyntaxError
- [ ] Verify normal Invidious API usage still works correctly with a working instance
- [ ] Verify that Invidious JSON error responses (e.g. `{"error": "..."}`) are still handled as before


absidue:
> Please refrain from using AI/LLMs to create issues, pull requests, comments, or any other content submitted to the FreeTube project.

https://docs.freetubeapp.io/community/rules/

PR #8762, opened 2026-03-06
Migrate the WatchVideoPlaylist component to the composition API

absidue:
## Pull Request Type

- [x] Performance improvement

## Description

This pull request migrates the `WatchVideoPlaylist` component to the composition API. While testing the changes I noticed that shuffling was slow so I also improved the performance of it by switching to the [modern version of the Fisher-Yates shuffle algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm), as pushing to an array in a loop requires resizing arrays and splicing an array requires moving items around in the array each time an item is removed, the new in-place shuffling algorithm is a lot faster.

## Screenshots

Measuremeants with my 16k item test playlist:

Before (notice how it is so slow that after clicking the shuffle button a few times in a row, V8 decides it needs to apply aggressive optimisations to the function, to avoid freezing the UI for 5.5 seconds):
before

After (the new approach is fast enough that V8 doesn't need to apply such aggressive optimisations):
after

While yes in the end the old approach ends up being faster because of the aggressive optimisations that V8 applies after clicking the shuffle button a few times, in real world situations users will probably only click the shuffle button once so a shuffle algorithm that is fast from the start is definitely the better choice.

## Testing

Test both user playlists and YouTube playlists, as well as the various functions of the watch page playlist component.

## Desktop

- **OS:** Windows
- **OS Version:** 11


src/renderer/components/WatchVideoPlaylist/WatchVideoPlaylist.vue
```diff
@@ -0,0 +1,812 @@
...
+    const errorMessage = t('Invidious API Error (Click to copy)')
+    showToast(`${errorMessage}: ${err}`, 10000, () => {
+      copyToClipboard(err)
+    })
+    if (process.env.SUPPORTS_LOCAL_API && backendPreference.value === 'invidious' && backendFallback.value) {
+      showToast(t('Falling back to Local API'))
+      getPlaylistInformationLocal()
+    } else {
+      isLoading.value = false
+      // TODO: Show toast with error message
```
ghost: Nit: We show the toast with an error message before the if statement. I don't think we need this todo (im guessing this was here before migration)

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
On User or Creator playlists when you enable shuffle and skip to next video you receive an error. This error wont occur when shuffle is disbaled

https://github.com/user-attachments/assets/b7fc00ec-72d7-4616-8261-4c7bf33f0a07

efb4f5ff-1298-471a-8973-3d47447115dc:
needs rebase into dev

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
I found that the autoplay button that controls autoplay when loop enabled is hidden on last video of playlist when loop enabled (but still affects the autoplay
Also autoplay control totally hidden when loop disabled

image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE just to understand correctly. The last video on the playlist does not show the autoplay button except for when loop is enabled?

PikachuEXE:
Last video = invisible whether loop enabled or not
image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
Ohhh okay yeah i see it now, good catch!

absidue:
Fixed

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #8809, opened 2026-03-15
Properly sanitize HTML

absidue:
## Pull Request Type

- [x] Security improvement

## Description

Currently we have a mix of different ways of handling raw HTML, none of them are bulletproof. This pull request removes all uses of `v-html` and replaces them with a custom `v-safer-html` directive that uses the brand new HTML Sanitizer API which we can use with the Electron 41 upgrade and falls back to the DOMPurify library outside of Electron when the HTML Sanitizer API is not available. I went with two different sanitizing configurations a default one that is very strict that only allows the `<br>`, `<b>`, `<i>`, `<s>`, `<a>`, `` elements, which is used in most places and a lenient one that uses the default Sanitizer/DOMpurify configurations that only strip dangerours things (e.g. `<script>` tags and inline event handlers) that is used for the changelog display.

## Testing
1. Check that the About page still looks correct
2. Check that video descriptions look correct and clicking on timestamps still works (as playback is broken you don't actually need to check that it seeked, just that it scrolled to the top of the page, as that happens in the same event handler).
3. Check that comments with formatting look correct
4. Hard-code `USE_NATIVE_SANITIZER` to `false` and remove the dompurify alias in the renderer webpack config, then repeat steps 1-3

If you are unsure about what videos to use for 2 and 3, the LinusTechTips WAN show ones usually have timestamps in both the description and comments, as well as formatting in the comments.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -270,7 +270,11 @@ export default [
     rules: {
       '@stylistic/space-before-function-paren': 'off',
       '@stylistic/comma-dangle': ['error', 'only-multiline'],
-      'vue/no-v-html': 'off',
+
+      // Ban v-html as it inserts HTML via innerHTML without sanitizing it
+      // if inserting raw HTML is unavoidable the custom v-safer-html directive should be used
+      // which sanitizes the HTML before inserting it into the DOM
+      'vue/no-v-html': 'error',
 
       'no-console': ['error', {
         allow: ['warn', 'error'],
@@ -339,6 +343,15 @@ export default [
     languageOptions: {
       globals: {
         __FREETUBE_ALLOWED_PATHS__: 'readable'
+      }
+    }
+  },
+  {
+    files: ['src/renderer/directives/vSaferHtml.js'],
+    languageOptions: {
+      globals: {
+        // Fix Sanitizer not being listed in `globals` yet, remove it it gets added in the future
+        Sanitizer: 'readable'
       }
     }
   },
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -350,7 +350,7 @@ export default [
     files: ['src/renderer/directives/vSaferHtml.js'],
     languageOptions: {
       globals: {
-        // Fix Sanitizer not being listed in `globals` yet, remove it it gets added in the future
+        // Fix Sanitizer not being listed in `globals` yet, remove it when it gets added in the future
         Sanitizer: 'readable'
       }
     }
--- a/src/renderer/directives/vSaferHtml.js
+++ b/src/renderer/directives/vSaferHtml.js
@@ -33,7 +33,7 @@ export const vSaferHtml = (element, { value, oldValue, modifiers }) => {
             // live chat emojis (see parseLocalTextRuns)
             {
               name: 'img',
-              attributes: ['alt', 'height', 'loading', 'src', 'style', 'width']
+              attributes: ['alt', 'height', 'loading', 'src', 'style', 'title', 'width']
             }
           ]
         })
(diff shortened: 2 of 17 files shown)


PikachuEXE (dismissed):

eslint.config.mjs
```diff
@@ -342,6 +346,15 @@ export default [
       }
     }
   },
+  {
+    files: ['src/renderer/directives/vSaferHtml.js'],
+    languageOptions: {
+      globals: {
+        // Fix Sanitizer not being listed in `globals` yet, remove it it gets added in the future
```
ghost:
```suggestion
        // Fix Sanitizer not being listed in `globals` yet, remove it when it gets added in the future
```

?

src/renderer/directives/vSaferHtml.js
```diff
@@ -0,0 +1,54 @@
...
+            'i',
+            's',
+            {
+              name: 'a',
+              attributes: ['data-time', 'dir', 'href', 'lang', 'tabindex']
+            },
+            // live chat emojis (see parseLocalTextRuns)
+            {
+              name: 'img',
+              attributes: ['alt', 'height', 'loading', 'src', 'style', 'width']
```
ghost: Should we have title attribute as well?
https://github.com/iv-org/invidious/blob/749791cdf1316bc89415d27d503042d3f6b3f398/src/invidious/comments/content.cr#L73

absidue: Probably yes, I didn't check the Invidious stuff in-depth.

absidue:
```suggestion
              attributes: ['alt', 'height', 'loading', 'src', 'style', 'title', 'width']
```

src/renderer/directives/vSaferHtml.js
```diff
@@ -0,0 +1,54 @@
...
+          ]
+        })
+      }
+
+      element.setHTML(value, { sanitizer })
+    } else {
+      if (domPurifyStrictConfig === undefined) {
+        domPurifyStrictConfig = {
+          ALLOWED_TAGS: ['br', 'b', 'i', 's', 'a', 'img'],
+          ALLOWED_ATTR: ['alt', 'data-time', 'dir', 'height', 'href', 'lang', 'loading', 'src', 'style', 'tabindex', 'width']
```
absidue:
```suggestion
          ALLOWED_ATTR: ['alt', 'data-time', 'dir', 'height', 'href', 'lang', 'loading', 'src', 'style', 'tabindex', 'title', 'width']
```

PikachuEXE (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
> Hard-code USE_NATIVE_SANITIZER to false

i did 

> 1. Check that the About page still looks correct

Doesnt seem to look correct 

https://github.com/user-attachments/assets/a2076e7c-11a0-48e7-ac07-0b8830fe845e

> 2. Check that video descriptions look correct and clicking on timestamps still works (as playback is broken you don't actually need to check that it seeked, just that it scrolled to the top of the page, as that happens in the same event handler).
> 3. Check that comments with formatting look correct

Same here

https://github.com/user-attachments/assets/21ff209d-42dd-4940-b9d8-de3ebbfc56d6

absidue:
Fixed the test steps, you need to remove the dompurify override in the renderer webpack config (see second file in the diff view of this pull request).

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #8831, opened 2026-03-18
Fix age-restricted videos breaking the watch page

absidue:
## Pull Request Type

- [x] Bugfix

## Related issue

Mentioned in the Matrix FreeTube General chat

## Description

YouTube made some changes to how the WEB_EMBEDDED_PLAYER client works, so it currently throws errors in FreeTube, which means that instead of showing the age-restricted message to the user, we show an empty page with an error message. As fixing the WEB_EMBEDDED_PLAYER support still requires some investigation and the workaround only works for a very small number of age-restricted videos anyway, this pull request currently only fixes the watch page breaking so that we show the user friendly message as originally intended but doesn't fix the WEB_EMBEDDED_PLAYER requests.

## Screenshots

Before:
Before the change. Broken watch page with &quot;This video is unavailable&quot; message.

After:
After the change. Properly displayed watch page with the player replaced with a user friendly error message about the video being age-restricted.

## Testing

1. Open an age-restricted video e.g. Linus Torvalds NVIDIA rant: https://youtu.be/iYWzMvlj2RQ
2. The player should be replaced with a message about the video being age-restricted, up next should be empty like on YouTube but all the other information should be shown (description, channel information, comments, etc).

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -470,28 +470,32 @@ export async function getLocalVideoInfo(id) {
       info.playability_status.reason === 'Sign in to confirm your age') ||
     (hasTrailer && trailerIsAgeRestricted)
   ) {
-    const webEmbeddedInnertube = await createInnertube({ clientType: ClientType.WEB_EMBEDDED })
-    webEmbeddedInnertube.session.context.client.visitorData = webInnertube.session.context.client.visitorData
+    try {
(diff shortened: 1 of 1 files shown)


PikachuEXE (approved):
TIL that video is age restricted on YT

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ChunkyProgrammer (approved):

PR #8855, opened 2026-03-21
Reduce workflow contention with changes to long running workflows

absidue:
## Pull Request Type

- [x] Other

## Description

As an open source project GitHub gives us unlimited (within reason) GitHub Actions minutes for free and doesn't charge us for storage use, we are however still limited by how many workflows we can run in parallel (I can't find the actual limit in the docs but from my observations the limit seems to be around 4-6), when we hit the limit any other workflows sit in the queue but the timeout already starts counting then. `ubuntu-slim` based workflows only have a 15 minute timeout. This is especially relevant for the weekly dependabot batch, as merging each pull request triggers at least 4 workflows, more if it causes conflicts with other one and dependabot has to resolve the conflicts, so unless we wait about 10 minutes between each merge, the majority of the workflows will sit in the queue for a while.

This pull request does not speed any of those workflows up but it should still help by skipping unnecessary runs or aborting duplicate runs early on long running workflows:
| Workflows | Changes |
| --- | --- |
| Build | Add `paths-ignore` to skip running on changes to other workflows that don't result in a different build e.g. Dependabot updates to the stale workflow. |
| Conflicts and Linter | Add cancel in progress for changes to the same branch/pull request, as we only care that the latest commit passes the linter and that there are no conflicts with the most recent commit. |
| No Response | Currently this runs on a daily schedule and on every single comment but we can skip most comments as they will be on issues without the "waiting for response from author" label or from other accounts that are not the original author. |

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,26 +1,23 @@
-# This is a basic workflow to help you get started with Actions
-
 name: Linter
 
-# Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the master branch
 on:
   pull_request:
     branches: [ master, development, '**-RC' ]
 
 permissions: {}
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  # This workflow contains a single job called "build"
   lint:
-    # The type of runner that the job will run on
+    name: Linter
     runs-on: ubuntu-latest
 
     permissions:
       contents: read
 
-    # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
     - name: Get yarn cache directory
       id: cache_dir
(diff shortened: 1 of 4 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (dismissed):

ChunkyProgrammer (dismissed):

absidue:
Ugh I think I broke the mandatory merge check by adding a name to the lint job in the Linter workflow, a bit dumb of GitHub that they tied it to the display name.

.github/workflows/linter.yml
```diff
@@ -1,26 +1,23 @@
...
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  # This workflow contains a single job called "build"
   lint:
-    # The type of runner that the job will run on
+    name: Linter
```
absidue:
```suggestion
```

ChunkyProgrammer (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #8857, opened 2026-03-21
Address profile channel list TODOs about using Sets with Vue 3

absidue:
## Pull Request Type

- [x] Refactor

## Description

This pull request resolves some open TODOs in the profile channel list components about using Sets instead of arrays on Vue 3. I also switched another array to shallow ref as we never modify the array itself, we only change it by setting the variable.

## Testing

Check that selecting, adding and removing channels works correctly on the profile edit page.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/components/FtProfileChannelList/FtProfileChannelList.vue
+++ b/src/renderer/components/FtProfileChannelList/FtProfileChannelList.vue
@@ -14,8 +14,8 @@
           :channel-id="channel.id"
           :channel-name="channel.name"
           :channel-thumbnail="channel.thumbnail"
-          selectable
-          :selected="selected.includes(channel.id)"
+          :selectable="true"
+          :selected="selected.has(channel.id)"
           @change="handleChannelToggle(channel.id)"
         />
       </FtFlexBox>
@@ -48,7 +48,7 @@
 </template>
 
 <script setup>
-import { computed, ref, watch } from 'vue'
+import { computed, reactive, ref, shallowRef, watch } from 'vue'
 import { useI18n } from '../../composables/use-i18n-polyfill'
 
 import FtCard from '../ft-card/ft-card.vue'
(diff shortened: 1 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

src/renderer/components/FtProfileChannelList/FtProfileChannelList.vue
```diff
@@ -14,8 +14,8 @@
           :channel-id="channel.id"
           :channel-name="channel.name"
           :channel-thumbnail="channel.thumbnail"
-          selectable
-          :selected="selected.includes(channel.id)"
+          :selectable="true"
```
PikachuEXE, line 17: What are advantages using this form

PikachuEXE (approved):

src/renderer/components/FtProfileChannelList/FtProfileChannelList.vue
absidue: We currently use a mix of both styles in the codebase, in most cases we explicitly pass in that value. Functionally it doesn't change anything in this specific case but it can in others.

`selectable` compiles to `selectable: ""`, `:selectable="true"` compiles to `selectable: true` (which terser transforms to `selectable: !0` to save two characters).

When the prop only accepts `Boolean`s everything works fine, but if it accepts both booleans and strings things get funky:
- prop with type `Boolean` and input value `""` -> `true`
- prop with type `[Boolean, String]` and input value `""` -> `true`
- prop with type `[String, Boolean]` and input value `""` -> `""`

Vue has implicit boolean casting behaviours when Boolean is listed first in the prop type, but in other cases it just passes through the original value which can be a bit confusing.

In the long term I do plan to switch the whole code base to the explicit name + value attribute format, as it ls less confusing, especially to people that might not know about that implicit behaviour in Vue and enable a lint rule to enforce it but in this pull request it was a drive-by change, that wasn't relevant to the core changes.

ChunkyProgrammer (approved):

PR #7729, opened 2025-07-23
Enhancement: Add fast forward and rewind pop up to video player

caetano-dev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

Adds a pop up when the user fast forwards or rewinds a video.

## Screenshots

a

b

## Testing

1. Open a video
2. Use the keyboard to fast forward or rewind
3. See a popup with an arrow and the seconds.

## Desktop

- **OS:** Linux
- **OS Version:** Ubuntu 24.04.2 LTS
- **FreeTube version:** v0.23.5 Beta

PR #8870, opened 2026-03-24
Ensure popup does not obstruct video title in full screen

caetano-dev:
## Pull Request Type

- [x] Other

## Related issue
Requested by https://github.com/FreeTubeApp/FreeTube/pull/7729#issuecomment-4119695838, since the previous implementation would appear on top of the titles in full screen mode.

## Description
Increases the top scss value for the popup, so that is does not obstruct the title when in full screen.

## Screenshots
Capture d’écran_2026-03-24_14-27-28 (copie 1)

Capture d’écran_2026-03-24_14-36-19

## Testing
Open video with a long title in full screen
Use left or right arrows
See the pop up below the title

## Desktop

- OS: Linux
- OS Version: Ubuntu 24.04.2 LTS
- FreeTube version: v0.23.15 Beta

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
@@ -60,7 +60,7 @@
 
 .valueChangePopup {
   position: absolute;
-  top: 20px;
+  top: 70px;
   left: 50%;
   transform: translateX(-50%);
   padding: 10px;


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
Lgtm, this moves all popups not just the fastforward/backward which is good. Also the placement is moved in all views but seems to be similar as what yt is doing so im ok with it

Bit off topic
I wonder if we should place left/right popups on the left/right middle side like yt
caetano-dev thoughts?

edit: removed offtopic section and will be moved to another PR

caetano-dev:
That makes sense. I implemented this locally to see what it would look like.
What do you think?

https://github.com/user-attachments/assets/35f01abd-8a36-4b18-a260-761a02912961

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm this feels a bit in the way.  maybe a bit more towards the edge? See yt for reference

caetano-dev:
Like this?

aaaa
bbbb
ccccc

efb4f5ff-1298-471a-8973-3d47447115dc:
@caetano-dev that placement lgtm but i meant it more as a followup PR to this one so this can stay focused on fixing the popup not being in the way of the title. Sorry for the confusion on that. Would it be possible for you to revert the changes regarding the placement of the left/right popups and create a separate PR for it?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Thanks @caetano-dev again sorry for the confusion

caetano-dev:
haha, no problem! I'm writing the new PR right now.

PikachuEXE (approved):

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@caetano-dev thanks for sending this one in. Random thing popedup (you see what i did there :D) again. I tried to play with the opacity of the popup because i feel like its blocking a bit much content but im not getting the desired result. Im either having it to be to solid or too opaque. I think YT has a good balance in this but i think they might be playing with other css stuff.

Would it be possible for you to look into this?

caetano-dev:
I'm on it.

PR #6655, opened 2025-01-27
Fixes #6330: Commented out overflow-x: hidden to resolve scrollbar issue

sullivanlng:
# Title

Fix unnecessary vertical scrollbar in settings page

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #6330 

## Description

This pull request fixes an issue where an unnecessary vertical scrollbar appeared in the settings page for certain languages (e.g., German, Turkish).

Changes made:
- Commented out the overflow-x: hidden property in .settingsSections and .switchRow within the Settings.css file.

## Screenshots 

Before :
20250121_CAP_Bug

After : 
20250121_CAP_BugFix

## Testing 

Approach:
- Tested across multiple languages that previously showed the issue (German, Polish, Turkish).
- Used the toggle device bar to simulate different screen sizes and validate responsive behavior.

Behavior Confirmed:
- The scrollbar no longer appears unnecessarily.
- No new horizontal scrollbars were introduced as a side effect.

## Desktop

- **OS:** Windows 11
- **OS Version:** 23H2
- **FreeTube version:** 0.22.1

## Additional context

- Left the commented code in place to make reactivation easier, if needed in the future.

PR #8873, opened 2026-03-25
Fix: Remove overflow-x: hidden to fix extra scrollbar in settings

cyphercodes:
## Pull Request Type
- [x] Bugfix

## Description
This pull request fixes an issue where an unnecessary vertical scrollbar appeared in the settings page for certain languages (e.g., German, Turkish).

Changes made:
- Removed the overflow-x: hidden property from .settingsSections and .switchRow within the Settings.css file.

## Testing
- Tested across multiple languages that previously showed the issue (German, Polish, Turkish).
- Validated responsive behavior at different screen sizes.

### DIFF
--- a/src/renderer/views/Settings/Settings.css
+++ b/src/renderer/views/Settings/Settings.css
@@ -5,7 +5,6 @@
 
 .settingsSections,
 .switchRow {
-  overflow-x: hidden;
   max-inline-size: 90%;
 }
 
@@ -41,6 +40,7 @@
   .settingsSections,
   .switchRow {
     max-inline-size: 100%;
+    overflow-x: hidden;
   }
 }
 


efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
https://github.com/FreeTubeApp/FreeTube/pull/6655#pullrequestreview-2576703172

cyphercodes:
@kommunarr Thanks for the review! I've updated the fix to address the horizontal scrollbar issue on compressed screens.

**Changes made:**
- Removed aoverflow-x: hiddena from the default a.settingsSectionsa and a.switchRowa rules (fixes the extra vertical scrollbar in German/Turkish)
- Added aoverflow-x: hiddena back only for screens <= 1200px via media query (prevents horizontal scrollbars on compressed screens)

This should now fix the original issue without introducing the horizontal scrollbar regression you identified in PR #6655.

cyphercodes:
Addressed the review feedback regarding horizontal scrollbar on compressed screens.

**Changes made:**
- Applied  to both  AND  in the  media query (previously it was only on )

This ensures both elements get the overflow protection on compressed screen sizes, preventing the horizontal scrollbar regression identified in PR #6655.

Thanks for the review!

PR #8877, opened 2026-03-26
Move skip popups to the edges of the screen

caetano-dev:
## Pull Request Type

- [x] Other

## Related issue
This is an improvement from PR https://github.com/FreeTubeApp/FreeTube/pull/8870, as suggested by this comment https://github.com/FreeTubeApp/FreeTube/pull/8870#pullrequestreview-4001873134

## Description

This PR moves the skip popups to the right or left side, making it easier for the user to understand what is happening.

## Screenshots
image

image

## Testing

1. Open a video
2. Press the left or right arrow
3. See the popups on the left or right
4. Press the up or down arrow or M key (or any other that triggers a popup)
5. See the popups on the original place

## Desktop

- **OS:** Linux
- **OS Version:** Ubuntu 24.04.2 LTS
- **FreeTube version:** v0.23.15 Beta


src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.css
```diff
@@ -60,9 +60,7 @@
 
 .valueChangePopup {
   position: absolute;
-  top: 20px;
-  left: 50%;
-  transform: translateX(-50%);
+  top: 70px;
```
efb4f5ff-1298-471a-8973-3d47447115dc: The `top: 70px` is from the other pr and doesnt belong here :D

caetano-dev: Fixed.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
I think having to look at different places for the notification is bad for UX

Note: especially on 4K fullscreen

efb4f5ff-1298-471a-8973-3d47447115dc:
Hmm tbh i didnt think of that. Also dont have an argument against so ill let the others chime in and be the deciding factor

efb4f5ff-1298-471a-8973-3d47447115dc:
I have given this some more thought and i agree with Pika. Thanks for opening the PR on my request and im sorry for not thinking a bit more about this before requesting this

caetano-dev:
No problem!

PR #8910, opened 2026-03-31
Show channels with errors sorted by name

jgresty:
Sorry about the drive by pull request but this was bothering me and I thought it would be less friction to just post a fix.

## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
n/a

## Description

When channels fail to get updated for whatever reason, they are inserted into an array of channels with errors in the order they finish processing. Since this is indeterministic it results in the channels being presented to the user in a seemingly random order.

This sorts the list of channels with errors when presenting them so they are always in a consistent order.

## Testing

- Subscribe to multiple channels
- Stub out `getLocalChannelVideos` in `src/renderer/helpers/api/local.js` to always return `null`
- Refresh video subscriptions a few times, they should always render in the same order

## Desktop

- **OS:** any
- **OS Version:** n/a
- **FreeTube version:** v0.23.15-beta

### DIFF
--- a/src/renderer/components/SubscriptionsTabUi/SubscriptionsTabUi.vue
+++ b/src/renderer/components/SubscriptionsTabUi/SubscriptionsTabUi.vue
@@ -9,7 +9,7 @@
       <h3> {{ $t("Subscriptions.Error Channels") }}</h3>
       <FtFlexBox>
         <FtChannelBubble
-          v-for="channel in errorChannels"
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
its doubled now

Image

PR #8915, opened 2026-03-31
Update electron-builder config so alpine can drop their patch

radmorecameron:
## Pull Request Type
- [x] Other

## Description
The way electron apps are being packaged in Alpine (through electron_tasje) requires a patch for FreeTube to build. This change should allow Alpine to drop that patch https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/freetube/tasje-dotdash.patch

## Testing
```bash
yarn run build
```

## Desktop
- **OS:** Fedora Linux
- **OS Version:** 43

### DIFF
--- a/_scripts/ebuilder.config.mjs
+++ b/_scripts/ebuilder.config.mjs
@@ -21,7 +21,7 @@ export default {
   files: [
     '_icons/iconColor.*',
     'icon.svg',
-    './dist/**/*',
+    'dist/**/*',
     '!dist/web/*',
     '!node_modules/**/*',
   ],


absidue (approved):
Confirmed that the included files are the same before and after this change in official builds with the "List files in app.asar" step in these test builds:
- Before: https://github.com/absidue/FreeTube/actions/runs/24036511173
- After: https://github.com/absidue/FreeTube/actions/runs/24036761472

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #8974, opened 2026-04-06
Add setting for shuffling subscription feed

### LINKED ISSUE #7907: [Feature Request]: Add shuffle mode / random sorting to subscriber view
### Guidelines

### Problem Description

Sorry if this issue already exists, I went through the ones mentioning "sort" but perhaps I missed one. **I would like to propose a shuffle mode or random sorting to subscriber view.**

**What is my problem:**

Basically, I'm subscribed to a ton of channels since I like diversity and variation, and I find youtube's own suggestion algorithm too harmful to rely on. I don't like watching many things from the same person all the time. So what I like is seeing a list of videos that is as random and from as many different people as possible, but from a curated list of people, hence my subscriber count is very high.

Sadly, the current sorting of the subscriber view gets in the way of variety. I've noticed it sort things by time, which in practice means that the youtubers uploading often are almost always at the top.

### Proposed Solution

I suggest a "Subscription" setting that offers a "random" sorting mode for the subscription view, rather than by default showing the newest uploads at the top.

**I feel like this is a natural extension of "Limit the number of videos displayed for each channel":** It's already possible to reduce the videos shown per subscriber, which boosts variety to have mass uploaders dominate less. Random sorting would boost it further.

### Alternatives Considered

I actually can't think of any good ones, I'm open to ideas.

### Issue Labels

new feature, new optional setting

### Additional Information

_No response_

TaranT1:
Added a setting which allows the option to randomize the order of videos in the subscription feed

## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #7907 

## Description
Adds a "shuffle" mode to the subscription video list
This takes the videos which would normally be shown in the subscription list and then displays them in a random order

## Testing
First go to subscriptions page and observe the videos are in order by upload date descending
Go to the settings page and apply the "Randomize Order" toggle
Go back to subscriptions page and observe the same videos are still displayed however in a random order

## Desktop
- **OS:MacOS**
- **OS Version:Tahoe 26.3**
- **FreeTube version:V0.24.0-beta**


absidue (commented):
Looks like the merge wasn't clean, please correct that. Additionally the shuffling should be done in the SubscriptionTab component, please also look at the more efficient shuffling implementation that the WatchVideoPlaylist component is using, which avoids creating unnecessary extra objects and arrays.

PR #5922, opened 2024-10-23
Handle unplayable content better with the local API

absidue:
# Handle unplayable content better with the local API

## Pull Request Type

- [x] Feature Implementation

## Related issue
closes #3238

## Description
This pull request introduces better handling for unplayable videos with the local API. As Invidious only returns the error message we cannot do anything like this for the Invidious backend, the same goes for private videos because YouTube only returns the please log-in message for those.

Currently when you encounter an unplayable video in FreeTube the watch page just breaks, this pull request changes that for members-only and age-restricted videos as well as when the user is IP blocked and has backend fallback disabled, by parsing as much metadata as possible so that the only missing information is the video player itself. For the IP blocked case, that required falling back to information from the `/next` response as the `/player` response is missing a lot of information when the IP block error is returned.

As age-restricted videos and member-only videos require a YouTube account, this pull request intentionally doesn't attempt to fallback to Invidious for those videos, as they will fail with Invidious too.

## Screenshots
<details><summary>Members-only video</summary>

members-only

</details>

<details><summary>Age-restricted video</summary>

age-restricted

</details>

<details><summary>Age-restricted movie trailer</summary>

age-restricted-trailer

</details>

<details><summary>IP blocked with backend fallback disabled</summary>

ip-block

</details>

## Testing

**Make sure you are using the local API.**
* Age-restricted video: https://youtu.be/ZQsK9WzDiuM
* Members-only video: https://youtu.be/eJ-AaE9GnzQ (taken from https://github.com/FreeTubeApp/FreeTube/issues/3238#issuecomment-1468301470)
* Age-restricted movie trailer: https://youtu.be/z-M8HU4BAG4 (searched for "MaXXXine", if that movie isn't available in your region, you can use your search engine to find 18+ movies and try searching for them on YouTube until you find one)

If you have access to a blocked IP address you can also test that the error message is shown instead of the player when backend fallback is disabled and that it still falls back to the Invidious API when backend fallback is enabled.

Regression test, as the trailer handling code changed:
* Movie trailer: https://youtu.be/3HMXp2_-_es (searched for "Blade Runner 2049", if that movie isn't available in your region try searching for other ones)

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:** b1839a25aeba0374024090dabbe8d0216bac069c

PR #8991, opened 2026-04-10
fix(watch): show persistent error for unavailable and removed videos

### LINKED ISSUE #8472: [Feature Request]: Removed video screen
### Guidelines

### Problem Description

If a video has been removed by its uploader, clicking on said video will merely give a blank video page (empty video, upload date 31/12/1969, etc).
There is a tiny message that says something to the effect of "Invidious API Error: Video removed by uploader", but if you're not actively looking for it, then you likely won't see it before it disappears.

### Proposed Solution

A page, not unlike the usual YouTube page, clearly stating "Video has been removed by uploader", and the video/channel name if it was saved (e.g if the video appeared in Trending).

### Alternatives Considered

Make error messages persist if video did not load.

### Issue Labels

### Additional Information

Image
Basically what I'm looking for.

aditya0155:
## Pull Request Type
- [x] Bugfix

## Related issue
closes #8472

## Description

Opening a removed/unavailable video currently shows a blank watch page with a error that disappears after a few seconds. The existing error overlay (used for members-only, age-restricted, DRM videos) never triggers because the catch blocks do not set `errorMessage` variable for unavailable videos.

I have made changes so that unavailable/removed video errors now set `errorMessage`, so the error stays visible inside the player instead of only appearing as a temporary toast.

Since unavailable videos may not have a usable thumbnail, the player area can shrink and hide the error. To avoid that, I added YouTube's official unavailable video fallback thumbnails:

- `https://www.youtube.com/img/desktop/unavailable/unavailable_video.png` for light themes
- `https://www.youtube.com/img/desktop/unavailable/unavailable_video_dark_theme.png` for dark themes

The thumbnail is chosen by checking the actual page background color with `calculateColorLuminance`, so it should keep working with current and future themes.

It also skips falling back to Invidious when the Local API error contains `unavailable`, since unavailable videos will fail there too.

Finally, I removed the redundant API error toast when the same message is already shown persistently in the player. The toast is now only shown when FreeTube is actually falling back to the other backend.

While testing the review feedback, I also found that the stale thumbnail/title issue existed before this PR. When FreeTube is playing a normal video and then switches to an unavailable video, the unavailable video request fails, but some data from the previous video, like `thumbnail`, `videoTitle`, `channelName`, and `channelThumbnail`, can still be kept around.

Before this PR, that was easier to miss because FreeTube mostly stayed visually on the previous video and showed a temporary unavailable-video toast. After this PR, the issue became more obvious because the unavailable message stays inside the player, so it could appear on top of the previous video’s thumbnail/title.

The fix is to clear the previous video’s state when the watch route changes, before loading the next video. Then, if the new video is unavailable, FreeTube shows the unavailable fallback thumbnail and persistent error instead of accidentally reusing details from the last normal video.

## Screenshots
**Before:** 

the error of video unavailable stays for only few seconds so if you're not actively looking for it, then you likely won't see it before it disappears.

image

**After:**

_Testing On Multiple Themes_

https://github.com/user-attachments/assets/1e89c33b-56cf-4da2-9ab6-4a12574e4553

_Testing On Normal Video_

https://github.com/user-attachments/assets/3cd5d4c0-db48-4f38-b002-234ba7b30035

---

_Before this PR during watch route change to unavailable video_

https://github.com/user-attachments/assets/075919d4-c170-4464-b7dc-846f8f70e78e

_After this PR during watch route change to unavailable video_

https://github.com/user-attachments/assets/74ee9038-9c02-4895-9cdc-1c8a5d75735c

## Testing

1. Open https://www.youtube.com/watch?v=Y0Aqc174FMM
2. Check that the "Video unavailable" error is shown inside the player and does not disappear.
3. Check that there is no duplicate API error toast when the player already shows the same error.
4. Open a normal video and check that it still plays.
5. Open other unplayable videos, like members-only or age-restricted videos, and check that their existing player error behavior still works.
6. Also retest the feature with different themes. 

## Desktop
- **OS:** Windows
- **OS Version:** 11
- **FreeTube version:** 0.24.0

### DIFF
--- a/src/renderer/helpers/colors.js
+++ b/src/renderer/helpers/colors.js
@@ -99,10 +99,27 @@ export function getRandomColor() {
 }
 
 export function calculateColorLuminance(colorValue) {
-  const cutHex = colorValue.substring(1, 7)
-  const colorValueR = parseInt(cutHex.substring(0, 2), 16)
-  const colorValueG = parseInt(cutHex.substring(2, 4), 16)
-  const colorValueB = parseInt(cutHex.substring(4, 6), 16)
+  let colorValues
+
+  if (colorValue.startsWith('#')) {
+    const cutHex = colorValue.length === 4
+      ? colorValue.slice(1).split('').map(value => value + value).join('')
+      : colorValue.substring(1, 7)
+
+    colorValues = [
+      parseInt(cutHex.substring(0, 2), 16),
+      parseInt(cutHex.substring(2, 4), 16),
+      parseInt(cutHex.substring(4, 6), 16)
+    ]
+  } else {
+    colorValues = colorValue.match(/\d+(\.\d+)?/g)?.slice(0, 3).map(Number)
+  }
+
+  if (!colorValues || colorValues.some(value => isNaN(value))) {
+    return '#FFFFFF'
+  }
+
+  const [colorValueR, colorValueG, colorValueB] = colorValues
 
   const luminance = (0.299 * colorValueR + 0.587 * colorValueG + 0.114 * colorValueB) / 255
 
--- a/src/renderer/views/Watch/Watch.js
+++ b/src/renderer/views/Watch/Watch.js
@@ -13,6 +13,7 @@ import WatchVideoLiveChat from '../../components/WatchVideoLiveChat/WatchVideoLi
 import WatchVideoPlaylist from '../../components/WatchVideoPlaylist/WatchVideoPlaylist.vue'
 import WatchVideoRecommendations from '../../components/WatchVideoRecommendations/WatchVideoRecommendations.vue'
 import FtAgeRestricted from '../../components/FtAgeRestricted/FtAgeRestricted.vue'
+import { calculateColorLuminance } from '../../helpers/colors'
 import {
   buildVTTFileLocally,
   copyToClipboard,
@@ -59,16 +60,6 @@ const UNAVAILABLE_VIDEO_THUMBNAILS = {
   light: 'https://www.youtube.com/img/desktop/unavailable/unavailable_video.png',
   dark: 'https://www.youtube.com/img/desktop/unavailable/unavailable_video_dark_theme.png'
 }
-const LIGHT_BASE_THEMES = new Set([
-  'light',
-  'pastelPink',
-  'catppuccinLatte',
-  'solarizedLight',
-  'gruvboxLight',
-  'everforestLightHard',
-  'everforestLightMedium',
-  'everforestLightLow'
-])
 
 export default defineComponent({
   name: 'Watch',
@@ -204,9 +195,6 @@ export default defineComponent({
     },
     backendFallback: function () {
       return this.$store.getters.getBackendFallback
-    },
-    baseTheme: function () {
-      return this.$store.getters.getBaseTheme
     },
     currentInvidiousInstanceUrl: function () {
       return this.$store.getters.getCurrentInvidiousInstanceUrl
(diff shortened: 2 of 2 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Hi @aditya0155 sorry for the delayed review. I had a random thought on this. What about using 

https://www.youtube.com/img/desktop/unavailable/unavailable_video.png for light themes and https://www.youtube.com/img/desktop/unavailable/unavailable_video_dark_theme.png for dark themes

aditya0155:
@efb4f5ff-1298-471a-8973-3d47447115dc hey I have tried to implement the required changes feel free to check whenever you get time.

I proposed 2 ways of solving this issue 

1st : 

1st commit tests: 

https://github.com/user-attachments/assets/d6cd28db-9b0c-43f8-bc89-0e1d881b1c19

https://github.com/user-attachments/assets/4bddd5ff-a9a2-42ef-aad3-351cb0b01d14


although there is 1 problem 

const LIGHT_BASE_THEMES = new Set([
  'light',
  'pastelPink',
  'catppuccinLatte',
  'solarizedLight',
  'gruvboxLight',
  'everforestLightHard',
  'everforestLightMedium',
  'everforestLightLow'
`])

`LIGHT_BASE_THEMES` is fragile because future light themes would need to be added here too but this is the most minimal edit and fixing the root problem is trickier because the app does not currently have a shared way to ask whether the active theme is light or dark.

aditya0155:
final commit (fixing the root issue: recommended): 

Instead of maintaining a hardcoded list of light themes, it now reads the page's actual background color at render time and uses the existing `calculateColorLuminance` helper to decide light vs dark. 

also I extended `calculateColorLuminance` to handle rgb()/rgba() strings (not just hex like previously) since `getComputedStyle` returns those formats. This way, new themes automatically get the correct thumbnail without anyone needing to update the previous `LIGHT_BASE_THEMES` list.


Testing videos:

_Testing On Multiple Themes_


https://github.com/user-attachments/assets/1e89c33b-56cf-4da2-9ab6-4a12574e4553


_Testing On Normal Video_



https://github.com/user-attachments/assets/3cd5d4c0-db48-4f38-b002-234ba7b30035

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Looked at https://github.com/FreeTubeApp/FreeTube/pull/5922 briefly and it looks like the toast can be removed as the message in the player is the same.

Please also update the details in the PR body :)

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):
LGTM! Thank you for implementing this :)

PikachuEXE:
Is it expected/desired to show last normal video thumbnail/title when switching from viewing normal video to unavailable video?
image

efb4f5ff-1298-471a-8973-3d47447115dc:
Doenst seem desired

aditya0155:
I think the root cause existed even before this PR. When FreeTube is playing a normal video and then switches to an unavailable video, the unavailable video request fails, but some watch page state (thumbnail, videoTitle, channelName, channelThumbnail etc etc) from the previous video is still kept around.

Before this PR, this was easier to miss because FreeTube would mostly just stay on the previous video visually and show a temporary “video unavailable” toast although some times the button next to search bar doesn't work previously which was a bit frustrating (you can see from the video). After this PR, the issue is more obvious because the unavailable message stays inside the player, so it ends up being shown over the previous video’s thumbnail/title.

https://github.com/user-attachments/assets/075919d4-c170-4464-b7dc-846f8f70e78e


I have implemented a fix is to clear the previous video’s state data when the watch route changes, "before" loading the next video. Then, if the new video is unavailable, FreeTube can show the unavailable fallback thumbnail and error instead of accidentally reusing details from the last normal video.

https://github.com/user-attachments/assets/74ee9038-9c02-4895-9cdc-1c8a5d75735c

Also, I have updated the PR body

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
@aditya0155 Thank you for addressing this! Hope to see more from you in the future :)

PikachuEXE (approved):

src/renderer/helpers/colors.js
```diff
@@ -99,10 +99,27 @@ export function getRandomColor() {
...
+    ]
+  } else {
+    colorValues = colorValue.match(/\d+(\.\d+)?/g)?.slice(0, 3).map(Number)
+  }
+
+  if (!colorValues || colorValues.some(value => isNaN(value))) {
+    return '#FFFFFF'
+  }
+
+  const [colorValueR, colorValueG, colorValueB] = colorValues
```
absidue, line 122: Not a big fan of this unnecessarily complicated bit of code, but as it works and this merge request has been open for a while, I'll leave it for now.

absidue (approved):

PR #9020, opened 2026-04-16
Fix SOCKS proxies not using authentication by using proxy-chain

Gabi-Zar:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description

The current implementation of proxy authentication does not work with SOCKS proxies because Electron/Chromium does not support it.
To fix this, I implemented a local proxy that redirects traffic to the SOCKS5 proxy using [proxy-chain ](https://www.npmjs.com/package/proxy-chain) to expose the SOCKS proxy through a local HTTP proxy.
The local proxy is launched via an Electron [utilityProcess](https://www.electronjs.org/docs/latest/api/utility-process) only when the user configures a SOCKS proxy with a username and password.
The first time the proxy is launched, a firewall popup may appear. This will only occur for users who configure a SOCKS proxy with authentication.
The local proxy only listens on localhost and runs as a child process of FreeTube, so it is killed when the app exits.

## Testing

1. Enter a SOCKS5 or SOCKS4 proxy with username and password
2. A local proxy is automatically launched on a free port chosen by the OS
3. Click on "test proxy"
4. Check the reported proxy IP address

If you do not have a SOCKS5 proxy with authentication, you can set up a local proxy using [3proxy](https://github.com/3proxy/3proxy). Here is an example configuration file:
```cfg
nserver 8.8.8.8
nserver 1.1.1.1

auth strong
users username:CL:password

allow username
socks -p1080
```

## Desktop

- **OS:** Windows 11
- **OS Version:** Version 25H2
- **FreeTube version:** v0.24.0

### DIFF
--- a/_scripts/ebuilder.config.mjs
+++ b/_scripts/ebuilder.config.mjs
@@ -4,6 +4,7 @@ import packageDetails from '../package.json' with { type: 'json' }
 export default {
   appId: `io.freetubeapp.${packageDetails.name}`,
   copyright: 'Copyleft © 2020-2026 freetubeapp@protonmail.com',
+  asarUnpack: ['dist/proxyProcess.js'],
   // asar: false,
   // compression: 'store',
   productName: packageDetails.productName,
--- a/_scripts/webpack.main.config.js
+++ b/_scripts/webpack.main.config.js
@@ -12,6 +12,7 @@ const config = {
   devtool: isDevMode ? 'eval-cheap-module-source-map' : false,
   entry: {
     main: path.join(__dirname, '../src/main/index.js'),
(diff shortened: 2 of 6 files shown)


absidue:
As it is not supported by Chromium, I would prefer if we just hid the username and password fields in the UI when a SOCKS option is selected, we should never having been showing those fields for something that doesn't work. Rather than pulling in 9 new dependencies into FreeTube, launching extra processes and bypassing the app.asar file.

I'm going to close this pull request, you are welcome to open a new one to hide those fields in the UI as mentioned above, if not someone else can do it. I'll leave a comment on the issue too.

PR #7684, opened 2025-07-16
CI: Use faster D: drive for yarn cache on Windows

absidue:
## Pull Request Type

- [x] Other

## Description

Yarn's default cache location is `C:\Users\{username}\AppData\Local\Yarn\Cache\v6`, which works fine on most computers where the main drive is generally pretty fast, but on GitHub Actions the `D:` drive seems to be a lot faster than the `C:` drive, so the default cache location is not ideal. This pull request changes yarn's config in the build and release workflows on the Windows runners to `D:\ft_yarn_cache`.

I've recorded the times from runs before and after the change both without a cache (lockfile changed due to dependency updates etc) and with a cache (dependencies didn't change since the previous run), as you can see the difference is pretty impressive.

| Run \\ Step | "Use Node.js 22.x"<br>no cache | "Run yarn run ci"<br>no cache | | "Use Node.js 22.x"<br>with cache | "Run yarn run ci"<br>with cache |
| --- | --- | --- | --- | --- | --- |
| "win-x64" before | 6s | 3m 51s | | 1m 29s | 26s |
| "win-x64" after | 6s | 1m 34s | | 16s | 20s |
| | | | | | |
| "win-arm64" before | 7s | 3m 49s | | 1m 34s | 28s |
| "win-arm64" after | 5s | 1m 51s | | 20s | 21s |

## Testing

Before without cache: https://github.com/FreeTubeApp/FreeTube/actions/runs/16276244757
Before with cache: https://github.com/FreeTubeApp/FreeTube/actions/runs/16263531496/job/45913993365
After without cache: https://github.com/absidue/FreeTube/actions/runs/16315682784
After with cache: https://github.com/absidue/FreeTube/actions/runs/16315902686/

## Additional context

I'm not sure why the simple step of calling `yarn config set cache-folder D:\ft_yarn_cache` is so slow in GitHub Actions (8-18s) but even with those extra seconds restoring the cache and running yarn install is still faster.

PR #1692, opened 2021-09-13
Switch from npm to yarn as package manager

PikachuEXE:
**Important note**
We may remove your pull request if you do not use this provided PR template correctly.

**Pull Request Type**
Please select what type of pull request this is:
- [x] Feature Implementation

**Related issue**
[Recent Build Failures](https://github.com/FreeTubeApp/FreeTube/actions/workflows/build.yml)
Due to `cb() never called!` (Example issue in npm: https://github.com/npm/cli/issues/417)

**Description**
Use `yarn` (1.x) instead of `npm`
Use `yarn.lock` instead of `package-lock.json`
Use `yarn install --frozen-lockfile` instead of `npm ci`

**Screenshots (if appropriate)**
Takes less time to install packages
image

**Testing (for code that is not small enough to be easily understandable)**
Has this pull request been tested?
Please describe shortly how you tested it and whether there are any ramifications remaining. 

**Additional context**

PR #9041, opened 2026-04-20
Switch from yarn to pnpm as package manager

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

Previous switch and related PRs:
- https://github.com/FreeTubeApp/FreeTube/pull/1692
- https://github.com/FreeTubeApp/FreeTube/pull/7075
- https://github.com/FreeTubeApp/FreeTube/pull/7684

## Description

PR opened for open discussion
This might improve performance (not measured yet) but should reduce disk usage (on dev machines)
Also the default settings like https://pnpm.io/next/settings#allowbuilds should improve security
Hidden dependency like `terser` will be caught (i.e. cannot use a package if not specified directly in `package.json`)

### Compared to yarn classic/yarn berry(v2+)
https://oleksiipopov.com/blog/npm-package-managers-comparison/
https://pnpm.io/feature-comparison (no classic

## Screenshots 

Run it in console yourself

### Local
- [Install pnpm somehow](https://pnpm.io/installation) (I use `npm i -g pnpm@latest-10`)
- `rm -rf ./node_modules` (MUST be run when switching between yarn & pnpm otherwise you are on your own
- Run pnpm install, various scripts including dev & lint & build (e.g. `pnpm i && pnpm run dev`

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

[`--network-concurrency`](https://pnpm.io/settings#networkconcurrency) is removed intentionally to see if it necessary on CI (no way to test this until we use it a long enough period of time)
Cache save/restore still using the same steps as when we use yarn
Cache location on Windows on GH CI is already on D drive, so removed steps added in https://github.com/FreeTubeApp/FreeTube/pull/7684

image

See example build https://github.com/PikachuEXE/FreeTube/actions/runs/24647743297/job/72063783295


absidue:
Absolutely massive YESSS to this from my side!! Still need to test it but if it works as well as it should I'll approve it straight away.

Doesn't have to be part of this pull request but it looks like `pnpm run` supports a `--parallel` flag and passing a regex instead of an exact script name to run multiple scripts at the same time, so it might be possible to drop the `npm-run-all2` dependency.

After this is merged the docs will need updating too and it is probably worth pinging some known maintainers of third-party builds (@efb4f5ff-1298-471a-8973-3d47447115dc might still have a list from when we changed the build artifact names months ago).

PikachuEXE:
Separate PR I guess
Example run (should succeed just started): https://github.com/PikachuEXE/FreeTube/actions/runs/24657690328/job/72095274825

pnpm-workspace.yaml
```diff
@@ -0,0 +1,6 @@
+# https://github.com/pnpm/pnpm/issues/11070
+dedupePeers: true
+
+# https://pnpm.io/next/settings#allowbuilds
+allowBuilds:
+  electron: true
```
absidue: Lefthook's postinstall script needs to run to install git hooks for the first time (e.g. first time contributors), it would also be a good idea to explicitly set `electron-winstaller` and `unrs-resolver` to false, as we don't need them and that gets rid of the ignored build scripts warning.

```suggestion
  electron: true # Downloading electron executables for development
  electron-winstaller: false # We build NSIS installers on Windows, not squirrel ones
  lefthook: true # Installing git hooks
  unrs-resolver: false # Works around an npm specific optional dependencies bug, so we don't need it with pnpm
```

absidue (commented):
Linting runs 2 seconds slower on my machine, likely due to the overhead of needing to resolve all of the hardlinks, but it's not that bad and we definitely need to move away from yarn classic as that is unmaintained. Another major advantage looking at the runs in your repo is that the GitHub Actions caches for pnpm are only 160 MB compared to the 400 MB ones that we get with yarn, which is good news for pull requests that stay open for a long time because they'll get to use the caches for longer before they get evicted.

absidue (approved):

ChunkyProgrammer (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
All commands passed but run dev isnt showing me a window :(

https://github.com/user-attachments/assets/a5c32707-0a3f-4925-b0aa-e7bcca7b9b8d

PikachuEXE:
I created a openSUSE Slowroll VM on proxmox with KDE and I can't reproduce your issue
2026-04-23 08_57_07-QEMU (pika-workstation-ubuntu) - noVNC - Vivaldi

Make sure `rm -rf ./node_modules` is run and also try to reproduce it with different machines/config (different distro/DE?)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

sorry had issues with sandbox permissions -.-

PR #9042, opened 2026-04-20
Feature: Screenshot copy to clipboard

### LINKED ISSUE #8641: [Feature Request]: Save screenshot to clipboard
### Guidelines

### Problem Description

When using videos to take notes and triying to attach screenshots it can be tedious to manually save the screenshot and then import on your notetaking app.

### Proposed Solution

Have a shortcut/button/change default setting that allows to save a screenshot directly to clipboard.

### Alternatives Considered

Thus far the alternative is using a normal browser with an extension that does that.

### Issue Labels

ease of use improvement

### Additional Information

_No response_

Shadorc:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #8641 

## Description

- Replace "Ask for Save Folder" toggle with a dropdown allowing to chose a new option: copy to clipboard

## Screenshots 

After:
image
image
image
image

## Testing

**Copy to Clipboard:**
1. Select "Screenshot Mode: Copy to Clipboard"
2. Verify that only all screenshot options are either invisible or removed 
3. Verify that the screenshot button in the player is present
4. Verify that clicking on the screenshot button copy the screenshot to clipboard and does not save on disk

## Desktop

- **OS:** Bazzite (Fedora based Linux distro)
- **OS Version:** 43
- **FreeTube version:** 0.24.0

## Additional context

- ~~I'm not really satisfied of the "both" option, not really future proof if more modes are added. It could be two toggles, but again not really sure what the UI would look likle with more modes. Let me know what you think.~~  
- ~~Electron Clipboard API only supports JPEG and PNG but from my tests it seems that ultimately the image ends up as PNG anyways~~  
- ~~This feature is breaking backward compatibility as the "enable screenshot" option is gone. Not sure if there are migrations code or if retrocoompatibility should be kept no matter what.~~ 

I'm new to this project so I still have a bit to learn about!


absidue:
Please use the standard web navigator.clipboard API, like we do for copying text, rather than going to the main process and using the Electron specific APIs. Copying to the clipboard should be possible in FreeTubeAndroid too just like prompting for a save location (only the save to default folder is electron only).

I agree with you that "both" is odd, please remove it. Additionally rather than replacing the enable screenshot switch it would be better if you replaced the prompt for save location one. That dropdown would still default to prompting for the save location, in Electron builds it should have 3 options (the two existing ones plus copying to the clipboard) outside of Electron e.g. in FreeTubeAndroid it should only have prompting for the save location or copying to the clipboard.

As you already discovered it only makes sense to write PNGs to the clipboard, that id because images sit on the clipboard in operating system specific bitmap formats, not PNGs or JPEGs, so writing a PNG avoids losing details that would be lost in the lossy conversion to JPEG. As copying to the clipboard will always use PNG, the format and quality options should be hidden or at least disabled when the clipboard option is selected.

Testing an actual FreeTubeAndroid build is a bit more involved, so you can check the settings UI outside of Electron with `pnpm run dev:web`, that is unusable for most things as FreeTube cannot fetch information from YouTube in a normal web browser but for "offline" stuff like the settings page it should work fine.

Shadorc:
Thanks for your review!

I have applied the requested changes and updated my initial messsage accordingly.
I was also able to test in browser to simulate `USING_ELECTRON`

static/locales/en-US.yaml
```diff
@@ -476,9 +476,13 @@ Settings:
       8k: 8k
     Screenshot:
       Enable: Enable Screenshot
+      Mode: Screenshot Mode
+      Modes:
+        Ask Path: Ask for Save Folder
+        Do Not Ask Path: Do not ask for Save Folder
```
efb4f5ff-1298-471a-8973-3d47447115dc: I feel like this doesnt really tell the user what it does. I mean yes it doesnt ask for a save folder but what does it do instead. The Save screenshot folder does appear but still could be confusing.

I suggest renaming it to something like
`Save screenshot to`

Shadorc: I've updated the localization file. I've left the enum 'do_not_ask' unchanged because I think it's more understandable for internal enum name but I could change it too.  
Also, I'm not really sure what are the rules for loca keys/values, some have all words starting with uppercase, some just have the first one and some are a mix of both

absidue: Do not ask is technically incorrect, because if you choose the default folder option but don't select a folder in the settings it will prompt you for one the first time you take a screenshot.

My suggestions would be:
- `prompt_folder`
- `default_folder`
- `clipboard`

My idea is that including folder in the values allows for theoretical future options e.g. some kind of ask each time setting that would prompt you each time whether you want to copy or save the screenshot.

static/locales/en-US.yaml
```diff
@@ -1117,7 +1121,9 @@ Age Restricted:
   This video is age restricted: This video is age restricted
 External link opening has been disabled in the general settings: 'External link opening has been disabled in the general settings'
 Screenshot Success: Saved screenshot
+Screenshot Clipboard Success: Saved screenshot to clipboard
 Screenshot Error: Screenshot failed. {error}
+Screenshot Clipboard Error: Screenshot copy to clipboard failed. {error}
```
absidue: As this placeholder never gets set in the code so it is better to remove it.

```suggestion
Screenshot Clipboard Error: Screenshot copy to clipboard failed.
```

src/renderer/helpers/utils.js
```diff
@@ -190,20 +190,33 @@ export function showToast(message, time = null, action = null, abortSignal = nul
...
+        await navigator.clipboard.writeText(content)
+      }
+
       if (messageOnSuccess !== null) {
         showToast(messageOnSuccess)
       }
     } catch (error) {
-      console.error(`Failed to copy ${content} to clipboard`, error)
+      if (content instanceof Blob) {
+        console.error('Failed to copy image to clipboard', error)
```
absidue: As the blob copying might be used for something else in the future and this isn't a user facing message it probably makes sense to log the mime type too.
```suggestion
        console.error(`Failed to data of type "${content.type}" to clipboard`, error)
```

absidue (commented):
In addition to the other two comments, could you please also add a migration from the old setting to the new one in `src/datastore/handlers/base.js`, you could take inspiration from the other setting migrations that are already there.

static/locales/en-US.yaml
efb4f5ff-1298-471a-8973-3d47447115dc: > Also, I'm not really sure what are the rules for loca keys/values, some have all words starting with uppercase, some just have the first one and some are a mix of both

To solve this we will switch to https://github.com/FreeTubeApp/FreeTube/pull/8273 sentence casing when i have time to do that so feel free to apply it but it isnt necessary

Shadorc:
Thanks for your feedbacks, I have updated the merge request accordingly

src/datastores/handlers/base.js
```diff
@@ -48,6 +48,20 @@ class Settings {
...
+    // In FreeTube 0.24.0, the "Screenshot Mode" setting only had two options that were represented as a toggle named "Ask path"
+    // This is a one time migration to preserve users' Screenshot prompt preference through this change.
+    const screenshotAskPath = await db.settings.findOneAsync({ _id: 'screenshotAskPath' })
+
+    if (screenshotAskPath) {
+      if (screenshotAskPath.value) {
+        await this.upsert('screenshotMode', 'prompt_folder')
+      } else {
+        await this.upsert('screenshotMode', 'default_folder')
+      }
```
absidue:
```suggestion
      await this.upsert('screenshotMode', screenshotAskPath.value ? 'prompt_folder' : 'default_folder')
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
absidue: Instead of creating two blobs when the clipboard mode is active, you could override the screenshot format up here instead.

```suggestion
      // Navigator Clipboard API only supports PNG
      const format = screenshotMode.value === 'clipboard' ? 'png' screenshotFormat.value
```

Shadorc: Indeed, now that we don't allow both mode to be selected at the same time, it's better to have only one blob
I will also move some vars inside the other if, like `filenameWithExtension` which is unused in case of clipboard mode

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Lets move the dropdown right below the `Enable Screenshot` toggle. This way the options based on what is selected will show below it

Image

Shadorc:
Done
I've also hidden screenshot options when clipboard is selected because it has no meaning
image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
> I've also hidden screenshot options when clipboard is selected because it has no meaning

You've somehow read my mind :O, thanks for doing this. PR LGTM, hope to see more from you in the future :)

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

static/locales/en-US.yaml
```diff
@@ -476,9 +476,13 @@ Settings:
       8k: 8k
     Screenshot:
       Enable: Enable Screenshot
+      Mode: Screenshot Mode
+      Modes:
+        Ask Path: Ask for Save Folder
+        Save Screenshot To: Save Screenshot to
```
PikachuEXE:
```suggestion
        Save Screenshot To Folder: Save Screenshot to Folder
```

src/renderer/components/PlayerSettings/PlayerSettings.vue
```diff
@@ -598,24 +602,35 @@ async function handleUpdateScreenshotFormat(format) {
   getScreenshotFilenameExample(screenshotFilenamePattern.value)
 }
 
-/** @type {import('vue').ComputedRef<number>} */
-const screenshotQuality = computed(() => store.getters.getScreenshotQuality)
+const screenshotModeNames = computed(() => [
+  t('Settings.Player Settings.Screenshot.Modes.Ask Path'),
+  ...process.env.IS_ELECTRON ? [t('Settings.Player Settings.Screenshot.Modes.Save Screenshot To')] : [],
```
PikachuEXE:
```suggestion
  ...process.env.IS_ELECTRON ? [t('Settings.Player Settings.Screenshot.Modes.Save Screenshot to Folder')] : [],
```

PikachuEXE (changes requested):
Only the middle one is missing the action target

Image

Local preview of proposed change
Image

If you want it to look better especially when select closed just add class name and add style to it

Image

Image

Image

Image

Shadorc:
@PikachuEXE Thanks for your suggestion. I've used `inline-size: fit-content;` instead, if that's ok for you, to support long localized texts

PikachuEXE:
The spacing doesn't look right
image

If padding removed, it looks wrong when longer options selected
image

image

Edit 1: It submitted before I can paste all images...

Also I forgot I need to check narrow screen view...
image

PikachuEXE:
Well maybe just update label text to `Save to Folder` and we don't have to mess with the style (no idea why I didn't think of this before

Edit: local preview
image
image
image

Shadorc:
I've scratched my head at this problem because I really wanted the content to fit the `select` for localization. I discovered that the correct solution would be 

```css
.screenshotModeSelect {
  min-inline-size: fit-content;
}

.screenshotModeSelect .select-text {
  margin-right: 15px;
}
``` 

but `.select-text` is scoped, I cannot add an attribute outside of the `FtSelect.css` file.
I don't think I can fix this easily without messing with `FtSelect`, so I will just apply your short text suggestion.

Thank you!

Shadorc:
Results:
image
image

PikachuEXE (dismissed):

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

PR #9093, opened 2026-05-02
Fix: Translated video titles do not depend on user locale

### LINKED ISSUE #2049: [Feature Request]: Disable forced title translations
### Guidelines

### Problem Description

Sometimes Youtube will show titles of videos translated into what it thinks is my language. I believe this can either be with Google Translate or the video uploader can [provide the translation themselves](https://support.google.com/youtube/answer/6289575?hl=en-GB). In either case, the translation can be undesirable if you understand the language and want to see the original title of the video, which will still be in the HTML page title but nowhere else. I believe it's impossible to disable this misfeature on Youtube itself. 

### Proposed Solution

It would be great to disable the translations and see all video titles with their original language. 

### Alternatives Considered

I've tried changing the Freetube language to the original language of the video title and this doesn't seem to be enough to see the original, it continues to show me the English translations instead.

### Issue Labels

### Additional Information

_No response_

Shadorc:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

~~https://github.com/FreeTubeApp/FreeTube/issues/2049~~
This is, in fact, two separate issues. I will create an issue to link to if you think that this change could be useful.

## Description

This merge request is a bit rough, but it's more a way to ask questions and show a bit of a proof-of-concept ~~because the concerned issue is locked https://github.com/FreeTubeApp/FreeTube/issues/2049~~

~~As I read issues on Youtube.js repo, I've found that absidue already knew that this 'lang' field existed, however, it didn't seem to be discussed as a solution for issue 2049. My main question is then: did I miss something? I've read all the thread, multiple times, but I may have missed something subtle indicating why it couldn't be fixed using the 'lang' field. The trending page is already returning the videos with their original titles, so it should be fine for subscription page too (also, the trending page is using 'location', which does not work for the subscription page and I can't really explain it).~~

~~The following questions will be irrelevant depending on the previous answer but I will go on:~~

- Do we want to have the innertube lang always depend on the current FreeTube locale? My implementation is doing so but it could be a parameter
- If YouTube does not support the specified 'lang', it will return error 400. This is currently not handled in the initial state of this MR as I wanted to keep things simple for a first draft, but we could imagine that either FreeTube will not support more langs than YouTube already does, or we could have a hardcoded list of supported langs by YouTube and check against it, defaulting to English.

## Testing

- Define FreeTube locale as english
- Go to this channel with FreeTube (https://www.youtube.com/channel/UCCMxHHciWRBBouzk-PGzmtQ)
- Check that the first video is in english (WHO KILLED THE RTS?)
- Define FreeTube locale as french
- Check that the first video is in french (QUI A TUE LES RTS ?)

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -11,6 +11,7 @@ import {
   getChannelPlaylistId,
   getRelativeTimeFromDate,
 } from '../utils'
+import i18n from '../../i18n'
 
 const TRACKING_PARAM_NAMES = [
   'utm_source',
@@ -95,6 +96,7 @@ async function createInnertube({ withPlayer = false, location = undefined, safet
 
     retrieve_player: !!withPlayer,
     location: location,
+    lang: i18n.global.locale,
     enable_safety_mode: !!safetyMode,
     client_type: clientType,
 


efb4f5ff-1298-471a-8973-3d47447115dc:
Idk what they do in https://github.com/FreeTubeApp/FreeTube/issues/2049#issuecomment-3461334875 but maybe useful?

Shadorc:
I just realized there are two separate issues. The bug #2049 refers to the fact that the original title of a video cannot be displayed, regardless of the user's language. My pull request fixes the issue where the default translation language is not the user's language. So these are different issues, and the extension you're talking about is indeed interesting.
I’ll update my pull request to specify that this addresses a different issue than #2049

absidue:
This pull request has two major issues:
1. FreeTube and YouTube support a different set of languages and even for the languages that they both support they use different language codes so you will need to create a mapping, you cannot just pass the FreeTube language code into YouTube.
2. As YouTube doesn't give us raw data, instead returning translated preformatted display ready strings in many places, that means a lot of FreeTube's information extraction relies on parsing those language specific strings which is currently written to only handle English strings, so if you want to pass the FreeTube language to YouTube that means you need to adapt all of those bits of code to handle every single language, which is not only a massive undertaking for the initial implementation but also a massive maintenance burden, as YouTube is moving more and more things to those translated preformatted strings.

So my stance on this, just like it is has been in all other threads that suggested this, is a big no.

Shadorc:
Thanks for your clarification!

I was confused about the trending tab being able to return the video info in the correct locale. After some testing, it seems that it's because this tab is using an API endpoint that returns the title and description based on 'location' locale but keep other fields in 'en' (default 'lang' value). 

Concerning `yt-anti-translate`, from what I understand, it seems to use a trick that consists of querying the same video info but with 'lo' as lang. This is a valid lang but not fully localized by YouTube. It always return the original title/description/etc. It's a pretty clever trick but that will probably not last long. I've tried with FreeTube, it works for title and description, but the parsing of other fields are breaking as absidue explained (for example the published date will be '1 ມື້ກ່ອນ' for '1 day'). 
https://github.com/artisticfox8/yt-anti-translate/blob/12da0aee9f267f0520f553ca0847d178d8f9ae9d/app/src/content_channelbranding.js#L135

There is still something that intrigates me, in ` Watch.js` line 454
```
// extract localised title first and fall back to the not localised one
this.videoTitle = result.primary_info?.title.text?.trim() ?? result.basic_info.title?.trim()
``` 
I'm not sure about this comment. If the title is always translated, how could we 'chose'? And if we are indeed able to chose, why do we chose the translated one and not `basic_info`?
Looking at the values, `basic_info` seems to be the original title, not translated

absidue:
> I'm not sure about this comment. If the title is always translated, how could we 'chose'? And if we are indeed able to chose, why do we chose the translated one and not basic_info?
Looking at the values, basic_info seems to be the original title, not translated

The watch page is the only place where YouTube returns both the original title and the translated one, FreeTube intentionally shows the translated title on the watch page so that it matches the rest of the app because many users complained about it being understandably annoying that it didn't match and before you come up with the dumb idea of making a watch page request for every video shown anywhere in FreeTube, FreeTube already has more than enough issues with playback issues and IP blocks, spamming even more watch page requests would just speed run the IP blocks and playback issues.

Shadorc:
Thank you for your explanations. I didn't have the full picture, and as we discussed in another MR, probably didn't spend enough time looking for all the information. 
Concerning my "dumb idea", it was not my proposition to make one request per video. It is indeed not exactly smart. I was just answering to efb4f5ff-1298-471a-8973-3d47447115dc about my understanding of the https://github.com/artisticfox8/yt-anti-translate extension. And, according to my tests, it is indeed making 2x more requests when the extension is installed. I wanted to test with FreeTube, just to see what it could do in order to fully answer the question, but it would never have been my idea to submit something like this.

PR #9119, opened 2026-05-10
Fix: permanent player tooltip 

### LINKED ISSUE #8046: [Feature Request]: Ensure player control labels do not overlap or interfere with each other
### Guidelines

### Problem Description

1. Click on a player button, eg play/pause
2. Now use the shortcut for it, k
3. Label is now permanently showing and obstructs other player labels

https://github.com/user-attachments/assets/90ad0660-92e0-43fc-b4fc-ff200be8d941

### Proposed Solution

Label shouldnt be permanently showing so it doesnt obstruct other player labels

### Alternatives Considered

None.

### Issue Labels

visual improvement

### Additional Information

_No response_

Shadorc:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/8046

## Description

When a button is clicked, it gains `:focus`, and then, when a shortcut is used (it can be any button in the player and the button clicked does not need to be hovered anymore), the `:focus-visible` state becomes `true` and stay like this until the focus is lost by clicking elsewhere. 
According to [Shaka Player's CSS rules](https://github.com/shaka-project/shaka-player/blob/main/ui/less/tooltip.less#L48)), the `:focus-visible` pseudo-class keeps the tooltip visible until focus is removed.

Calling [blur()](https://www.w3schools.com/jsref/met_html_blur.asp) after a keyboard shortcut is triggered removes the  focus by clearing the `:focus-visible` state, hiding the tooltip.

## Screenshots 

No visible change to show

## Testing

Scenario 1
1. Click on a player button, eg play/pause
2. Now use the shortcut for it, k
3. Label is now **not** permanently showing and **does not** obstruct other player labels

Scenario 2 (thanks aditya0155)
1. Click on a player button, eg play/pause
2. Now use the shortcut for another one, eg m
3. Label is now **not** permanently showing and **does not** obstruct other player labels

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -2238,6 +2238,23 @@ export default defineComponent({
       seekBySeconds(dist, true)
     }
 
+    // Blur player buttons to remove :focus-visible state, preventing tooltips from staying visible
+    const buttonWithTooltipClasses = [
+      'shaka-play-button',
+      'shaka-fullscreen-button',
+      'shaka-mute-button',
+      'shaka-pip-button',
+      'full-window-button',
+      'theatre-button',
+      'screenshot-button',
+    ]
+    function blurTooltipButtons() {
+      const element = document.activeElement
+      if (buttonWithTooltipClasses.some(className => element.classList.contains(className))) {
+        element.blur()
+      }
+    }
+
     /**
      * @param {KeyboardEvent} event
      */
@@ -2289,25 +2306,6 @@ export default defineComponent({
       if (event.shiftKey && event.key.toLowerCase() === 'p') {
         emit('skip-to-prev')
         return
-      }
-
-      // Blur player buttons to remove :focus-visible state, preventing tooltips from staying visible
-      const buttonWithTooltipClasses = [
-        'shaka-play-button',
-        'shaka-fullscreen-button',
-        'shaka-mute-button',
-        'shaka-pip-button',
-        'full-window-button',
-        'theatre-button',
-        'screenshot-button',
-      ]
-      function blurTooltipButtons() {
-        for (const buttonClass of buttonWithTooltipClasses) {
-          const button = document.querySelector(`.${buttonClass}`)
-          if (button) {
-            button.blur()
-          }
-        }
       }
 
       switch (event.key.toLowerCase()) {
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -2292,16 +2292,16 @@ export default defineComponent({
       }
 
       // Blur player buttons to remove :focus-visible state, preventing tooltips from staying visible
+      const buttonWithTooltipClasses = [
+        'shaka-play-button',
+        'shaka-fullscreen-button',
+        'shaka-mute-button',
+        'shaka-pip-button',
+        'full-window-button',
+        'theatre-button',
+        'screenshot-button',
+      ]
       function blurTooltipButtons() {
-        const buttonWithTooltipClasses = [
-          'shaka-play-button',
-          'shaka-fullscreen-button',
-          'shaka-mute-button',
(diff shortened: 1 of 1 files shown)


aditya0155:
I found a small edge case: 
1) Click the Mute button with mouse
2) Do not click anywhere else and press k or Spacebar 

https://github.com/user-attachments/assets/2b56834f-905b-4ea7-856f-f1c17d9ac95f

Shadorc:
Thanks for testing, it is indeed a bug! It comes from the fact that I only clear tooltip for the button corresponding to the shortcut key, but it could be a different tooltip shown. 
It could be fixed by either calling blur on all buttons each times, the safest approach but the more maintenance-needed one, or call blur on the whole `document.activeElement`, but it could prevent specific feature like navigation with keyboards. 
I will fix it using the first approach

aditya0155:
https://github.com/user-attachments/assets/71885b30-9817-4cc4-a38b-4a9a8b99064d

the edge case is fixed from my testing

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2291,13 +2291,33 @@ export default defineComponent({
...
+          'theatre-button',
+          'screenshot-button'
+        ]
+        for (const buttonClass of buttonWithTooltipClasses) {
+          const button = document.querySelector(`.${buttonClass}`)
+          if (button) {
+            button.blur()
+          }
+        }
+      }
```
PikachuEXE: Avoid creating a new array on every call
```suggestion
      // Blur player buttons to remove :focus-visible state, preventing tooltips from staying visible
      const buttonWithTooltipClasses = [
        'shaka-play-button',
        'shaka-fullscreen-button',
        'shaka-mute-button',
        'shaka-pip-button',
        'full-window-button',
        'theatre-button',
        'screenshot-button',
      ]
      function blurTooltipButtons() {
        for (const buttonClass of buttonWithTooltipClasses) {
          const button = document.querySelector(`.${buttonClass}`)
          if (button) {
            button.blur()
          }
        }
      }
```

PikachuEXE (changes requested):

PikachuEXE (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc:
@Shadorc Sorry for the late review and thanks for looking into so many things :)

Shadorc:
@efb4f5ff-1298-471a-8973-3d47447115dc No problem! I have some free times and I love to contribute 😄
And no hurry for the reviews, I totally understand

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2291,13 +2291,33 @@ export default defineComponent({
...
+        'theatre-button',
+        'screenshot-button',
+      ]
+      function blurTooltipButtons() {
+        for (const buttonClass of buttonWithTooltipClasses) {
+          const button = document.querySelector(`.${buttonClass}`)
+          if (button) {
+            button.blur()
+          }
+        }
```
absidue: Instead of searching through the entire document for all of the possible elements and calling blur on all of them every time a key is pressed, it is faster to get the focused element and check if that has any of the listed classes. Additionally please move the function outside of the `keyboardShortcutHandler` function (nesting functions should be avoided when possible, especially for frequently called outer functions, as the nested functions get recreated every time the outer function is called).
```suggestion
        const element = document.activeElement
        if (buttonWithTooltipClasses.some(className => element.classList.contains(className)) {
          element.blur()
        }
```

Shadorc: Indeed, that's better!
I have committed the changes.

absidue (approved):

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM!

Issue #7345, opened 2025-05-02
[Feature Request]: Add import/export buttons to Data Settings for the missing databases


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

https://docs.freetubeapp.io/usage/data-location/#notable-files lists 6 database files. We only provide import/export for Subscriptions, Playlists and History. An user might want to import/export the other files.

> [!NOTE]
Maybe we should also allow to export ~~History and~~ Playlist to prevent being locked in

### Proposed Solution

Provide import/export buttons for:
- [ ] `settings.db`
- [x] `search-history.db`
- [ ] `subscription-cache.db`

### Alternatives Considered

None.

### Issue Labels

ease of use improvement

absidue:
Importing the settings.db will be problematic if we try to do it in a similar way to our existing imports, as we block writing certain settings from inside a window for security reasons, such as the default download and screenshots folders (we should probably do something similar for the external player settings at some point too).

ArthurKun21:
wonder if it can be unified into single database file like

```json
{
  "settings" : ...,
  "subscriptions": ...,
  "history": ...
}
```

while still able to make it individually exported and imported

```json
{
  "settings" : ...,
}
```

```json
{
  "subscriptions": ...,
}
```

```json
{
  "history": ...
}
```

PR #9123, opened 2026-05-10
Add import/export settings buttons

Shadorc:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/1018#issuecomment-2629109797
https://github.com/FreeTubeApp/FreeTube/issues/7345

## Description

This pull request adds support for importing and exporting FreeTube settings.

Some settings are not exportable, they are represented by the `settingsNotTransferrable` set inside `settings.js`. They are OS/user-specific settings, experimental settings, and settings that depend on `process.env`.

I'm not entirely satisfied with having to duplicate setting keys, but the other solution that I thought of was to replace the state object values with a more complex structure, having additional fields like `isExportable`. However, in my opinion, it would have required too much refactoring.

## Screenshots 

image
image
image

#### Basic test
1. Change some settings (a and b)
2. Export settings
3. Change some other settings (c and d)
4. Import settings
5. Check that a and b have the same values as step 2 and that changes made on c & d in steps 3 have been reverted

#### Non-exportable setting test
1. Changes value of proxy settings 
2. Export settings
3. Revert change made on step 1
4. Import settings
5. Check that proxy settings are the same as step 3

#### UI test
1. Minimize the window to make it similar in size to a phone
2. Go to Settings > Data
3. Check that the title, tooltip button and import/export buttons are correctly displayed
4. Check that the tooltip bubble is not hidden by anything

### DIFF
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -65,9 +65,15 @@
         @click="showExportSearchHistoryPrompt = true"
       />
     </FtFlexBox>
-    <h4 class="groupTitle">
-      {{ t('Settings') }}
-    </h4>
+    <FtFlexBox class="titleTooltip">
+      <h4 class="groupTitle">
+        {{ t('Settings') }}
+      </h4>
+      <FtTooltip
+        class="selectTooltip"
+        :tooltip="t('Settings.Data Settings.Settings Tooltip')"
+      />
+    </FtFlexBox>
     <FtFlexBox class="box">
       <FtButton
         :label="t('Settings.Data Settings.Import Settings')"
@@ -111,6 +117,7 @@ import FtButton from '../FtButton/FtButton.vue'
 import FtFlexBox from '../ft-flex-box/ft-flex-box.vue'
 import FtPrompt from '../FtPrompt/FtPrompt.vue'
 import FtSettingsSection from '../FtSettingsSection/FtSettingsSection.vue'
+import FtTooltip from '../FtTooltip/FtTooltip.vue'
 
 import store from '../../store/index'
 
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -65,15 +65,14 @@
         @click="showExportSearchHistoryPrompt = true"
       />
     </FtFlexBox>
-    <FtFlexBox class="titleTooltip">
-      <h4 class="groupTitle">
-        {{ t('Settings') }}
-      </h4>
+    <h4 class="groupTitle">
+      {{ t('Settings') }}
       <FtTooltip
         class="selectTooltip"
+        position="top"
         :tooltip="t('Settings.Data Settings.Settings Tooltip')"
       />
-    </FtFlexBox>
+    </h4>
     <FtFlexBox class="box">
       <FtButton
         :label="t('Settings.Data Settings.Import Settings')"
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -66,7 +66,7 @@
       />
     </FtFlexBox>
     <h4 class="groupTitle">
-      {{ t('Settings') }}
+      {{ t('Settings.Settings') }}
       <FtTooltip
         class="selectTooltip"
         position="top"
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -1517,7 +1517,8 @@ async function importSettings() {
 
   const { content } = response
   const importedSettings = JSON.parse(content)
-  const currentSettings = transferableSettings.value
+  const currentTransferableSettings = transferableSettings.value
+  const currentSettings = store.state.settings
 
   for (const [importedKey, importedValue] of Object.entries(importedSettings)) {
     if (!Object.hasOwn(currentSettings, importedKey)) {
@@ -1526,7 +1527,13 @@ async function importSettings() {
       continue
     }
 
-    const currentValue = currentSettings[importedKey]
+    if (!Object.hasOwn(currentTransferableSettings, importedKey)) {
+      const message = `${t('Settings.Data Settings.Non-transferable setting key')}: ${importedKey}`
+      showToast(message)
+      continue
+    }
+
+    const currentValue = currentTransferableSettings[importedKey]
     const areValuesEqual = currentValue === importedValue ||
       (typeof importedValue === 'object' && JSON.stringify(currentValue) === JSON.stringify(importedValue))
     if (areValuesEqual) {
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -1529,7 +1529,6 @@ async function importSettings() {
 
   const { content } = response
   const importedSettings = JSON.parse(content)
-
   const currentSettings = exportableSettings.value
 
   for (const [importedKey, importedValue] of Object.entries(importedSettings)) {
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -1482,8 +1482,8 @@ async function exportYouTubeSearchHistory() {
 // #region settings
 
 /** @type {import('vue').ComputedRef<object>} */
-const settingsEntries = computed(() => {
-  return store.getters.getUserSettings
+const exportableSettingsEntries = computed(() => {
+  return store.getters.getExportableSettings
 })
 
 async function importSettings() {
@@ -1509,12 +1509,12 @@ async function importSettings() {
   }
 
   const { content } = response
-  const settings = JSON.parse(content)
+  const importedSettings = JSON.parse(content)
 
-  const settingsDb = settingsEntries.value
+  const exportableSettings = exportableSettingsEntries.value
 
-  for (const [key, value] of Object.entries(settings)) {
-    if (Object.hasOwn(settingsDb, key)) {
+  for (const [key, value] of Object.entries(importedSettings)) {
+    if (Object.hasOwn(exportableSettings, key)) {
       const updaterId = await store.dispatch('getDefaultUpdaterId', key)
       await store.dispatch(updaterId, value)
     } else {
@@ -1527,13 +1527,13 @@ async function importSettings() {
 }
 
 async function exportSettings() {
-  const settingsDb = JSON.stringify(settingsEntries.value)
   const dateStr = getTodayDateStrLocalTimezone()
-  const exportFileName = 'freetube-settings-' + dateStr + '.db'
+  const exportFileName = `freetube-settings-${dateStr}.db`
+  const exportableSettings = JSON.stringify(exportableSettingsEntries.value)
 
   await promptAndWriteToFile(
     exportFileName,
-    settingsDb,
+    exportableSettings,
     t('Settings.Data Settings.Settings File'),
     'application/x-freetube-db',
     '.db',
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -119,7 +119,7 @@ import FtSettingsSection from '../FtSettingsSection/FtSettingsSection.vue'
 import FtTooltip from '../FtTooltip/FtTooltip.vue'
 
 import store from '../../store/index'
-import { defaultUpdaterId } from '../../store/modules/settings'
+import { defaultUpdaterId, NON_TRANSFERABLE_SETTINGS } from '../../store/modules/settings'
 
 import { MAIN_PROFILE_ID } from '../../../constants'
 import { calculateColorLuminance, getRandomColor } from '../../helpers/colors'
@@ -1488,7 +1488,7 @@ async function exportYouTubeSearchHistory() {
 
 // #region settings
 
-/** @type {import('vue').ComputedRef<object>} */
+/** @type {import('vue').ComputedRef<object[]>} */
 const transferableSettings = computed(() => {
   return store.getters.getTransferableSettings
 })
@@ -1515,26 +1515,24 @@ async function importSettings() {
     return
   }
 
-  const { content } = response
-  const textDecode = content.split('\n')
+  const textDecode = response.content.split('\n')
   textDecode.pop()
 
-  const currentTransferableSettings = transferableSettings.value
   const currentSettings = store.state.settings
 
   textDecode.forEach((rawEntry) => {
     const entry = JSON.parse(rawEntry)
-    if (typeof entry._id !== 'string') {
+    if (typeof entry._id !== 'string' || !Object.hasOwn(entry, 'value')) {
       showToast(t('Settings.Data Settings.Setting object has insufficient data, skipping item'))
       console.error('Missing keys:', entry)
     } else if (!Object.hasOwn(currentSettings, entry._id)) {
       const message = `${t('Settings.Data Settings.Unknown setting key')}: ${entry._id}`
       showToast(message)
-    } else if (!Object.hasOwn(currentTransferableSettings, entry._id)) {
+    } else if (NON_TRANSFERABLE_SETTINGS.has(entry._id)) {
       const message = `${t('Settings.Data Settings.Non-transferable setting key')}: ${entry._id}`
       showToast(message)
     } else {
-      const currentValue = currentTransferableSettings[entry._id]
+      const currentValue = currentSettings[entry._id]
       const areValuesEqual = currentValue === entry.value ||
         (typeof entry.value === 'object' && JSON.stringify(currentValue) === JSON.stringify(entry.value))
       if (!areValuesEqual) {
@@ -1548,9 +1546,9 @@ async function importSettings() {
 }
 
 async function exportSettings() {
-  const settingDb = Object.entries(transferableSettings.value)
-    .map(([_id, value]) => JSON.stringify({ _id, value }))
-    .join('\n') + '\n'
+  const settingDb = transferableSettings.value.map((settingEntry) => {
+    return JSON.stringify(settingEntry)
+  }).join('\n') + '\n'
   const dateStr = getTodayDateStrLocalTimezone()
   const exportFileName = 'freetube-settings-' + dateStr + '.db'
 
--- a/src/renderer/components/DataSettings/DataSettings.vue
+++ b/src/renderer/components/DataSettings/DataSettings.vue
@@ -1496,8 +1496,8 @@ async function exportYouTubeSearchHistory() {
 // #region settings
 
 /** @type {import('vue').ComputedRef<object>} */
-const exportableSettings = computed(() => {
-  return store.getters.getExportableSettings
+const transferrableSettings = computed(() => {
+  return store.getters.getTransferrableSettings
 })
 
 const showRestartPrompt = ref(false)
@@ -1528,7 +1528,7 @@ async function importSettings() {
 
   const { content } = response
   const importedSettings = JSON.parse(content)
-  const currentSettings = exportableSettings.value
+  const currentSettings = transferrableSettings.value
 
   for (const [importedKey, importedValue] of Object.entries(importedSettings)) {
     if (!Object.hasOwn(currentSettings, importedKey)) {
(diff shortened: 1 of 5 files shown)


absidue:
> absidue was warning that it could be an issue with some settings, but I didn't experience problems with the mentioned settings.

You won't see a warning, it will purposefully silently drop them because they are not meant to be writeable by the window and you won't see any issues in the same window because in the same window you are showing the stuff in-memory in that window but it won't be saved to disk, synced to other windows or actually get used.

While I don't want to discourage new contributors, could I please ask that you put a bit more thought into your pull requests or at the very least put more effort into testing them, as this is the second one that you have submitted that doesn't actually work in practice.

Shadorc:
You're right, I'm sorry. I'm a bit too eager to contribute. I did test my changes, but I saw your warning about the folder fields a bit late, and I only tested the settings value after import rather than actually testing whether the folders were correct. 
I'll put this MR as draft and see if I can fix it and test it properly. I'll either close it if I can't fix it properly or mark it as ready otherwise. 
Sorry for wasting your time, I know that's not cool!

Shadorc:
Maybe folder paths should not be allowed to be imported, does it really make sense? It is user and OS dependent, in most case the imported path would be invalid.
It could be an exception, with a little help icon indicating which settings (corresponding to paths) would not be part of the export / import process 
What do you think?

PikachuEXE:
What probably should not be transferred: 
- OS specific like screenshot settings
  (at least path, not idea about others, related: https://github.com/FreeTubeApp/FreeTube/pull/8581)
- external player
- proxy
- experimental settings
- not sure about password...
- Any other?

Shadorc:
I checked all the settings, I think that settings impacted by env like `process.env.SUPPORTS_LOCAL_API` and `process.env.IS_ELECTRON`  should also be ignored. A custom logic for those settings could be implemented during import, but it's probably safer to not export them and it only concern one settings currently (if we consider that `screenshotAskPath` would be not exported anyway because it is a path).

Concerning passwords, they are not encrypted, so exporting them as plain text would not be an issue. As for whether this is relevant, I’d tend to say yes, but it could be changed if needed.

There are also settings that need the app to be restarted. I've made the implementation similar to the list of non-exportable settings for the declaration, and when they are imported, the user is prompted to restart the app or to revert the settings that needed a restart. It works but I will see if I find something more "elegant"

Shadorc:
I'm now satisfied with the state of this MR!
I have updated my initial MR description accordingly.
I have also written and tested all the test cases that I could think of, they all pass.

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1458,6 +1485,110 @@ async function exportYouTubeSearchHistory() {
...
+ */
+function handleRestart(value) {
+  showRestartPrompt.value = false
+
+  if (value === null || value === 'cancel') {
+    pendingSettings.clear()
+    return
+  }
+
+  if (process.env.IS_ELECTRON) {
```
PikachuEXE: Does this mean the prompt is shown for web build (even though not officially supported) and will do nothing?

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1458,6 +1485,110 @@ async function exportYouTubeSearchHistory() {
...
+
+  if (response === null) {
+    return
+  }
+
+  const { content } = response
+  const importedSettings = JSON.parse(content)
+  const currentSettings = exportableSettings.value
+
+  for (const [importedKey, importedValue] of Object.entries(importedSettings)) {
```
PikachuEXE: Probably need filtering here in case we got something missing or user is importing an manually modified file (especially from untrusted source, who knows)

src/renderer/store/modules/settings.js
```diff
@@ -414,10 +431,42 @@ const sideEffectHandlers = {
 
 const settingsWithSideEffects = Object.keys(sideEffectHandlers)
 
+const settingsNotExportable = new Set([
```
PikachuEXE:
```suggestion
const settingsNotTransferrable = new Set([
```

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1458,6 +1485,110 @@ async function exportYouTubeSearchHistory() {
...
+    const currentValue = currentSettings[importedKey]
+    const areValuesEqual = currentValue === importedValue ||
+      (typeof importedValue === 'object' && JSON.stringify(currentValue) === JSON.stringify(importedValue))
+    if (areValuesEqual) {
+      continue
+    }
+
+    if (settingsNeedingRestart.has(importedKey)) {
+      pendingSettings.set(importedKey, importedValue)
+      showRestartPrompt.value = true
```
PikachuEXE: Use a local bool var and set this once only after loop

Shadorc: I've used `pendingSettings.size > 0`  instead of local bool, should be equivalent

src/renderer/components/DataSettings/DataSettings.vue
Shadorc: I've asked myself the same question, but I've used the same template as everywhere else
https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/ThemeSettings.vue#L317
https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/ExperimentalSettings/ExperimentalSettings.vue#L68

absidue: @Shadorc The difference is that in the other places it will never even get to that code, the electron check is just there to ensure that the electron only code gets removed properly during the build (as deadcode removal isn't guaranteed with Vue in the mix) and makes it clearer to people reading the code that it is electron only.

E.g. for the smooth scrolling switch is not shown outside of electron: https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/ThemeSettings.vue#L20 and the entire experimental settings component is Electron only: https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/views/Settings/Settings.vue#L174-L181

So you can either make the entire settings import and export feature electron only or write it in a way that also has a non-Electron restart and flexible enough that an Android specific restart could be added in downstream FreeTubeAndroid.

src/renderer/store/modules/settings.js
Shadorc: Done
https://github.com/FreeTubeApp/FreeTube/pull/9123/changes/81806b801020f5ade51a8cb1e57c2eb561eced06

src/renderer/components/DataSettings/DataSettings.vue
Shadorc: I'm not sure what you mean by "filtering" because just below I'm checking `if (!Object.hasOwn(currentSettings, importedKey)) {` to verify that transfarrable settings contain the imported key.
Do you mean validating the `importedSettings`  object has a whole, by checking keys are strings, it is not null and so on, or do you mean something else?

src/renderer/components/DataSettings/DataSettings.vue
Shadorc: Thanks for the clarification.
I had searched for settings depending on the `IS_ELECTRON` env var, but I missed `disableSmoothScrolling` and `hideToTrayOnMinimize`.

Concerning your suggestions, I see one more option.
Instead of making the entire feature electron-only, all electron-only settings could be marked as not transferrable. 

It would not remove a lot of settings because, according to my searches, here are the settings depending on `process.env`:
```
/* Depends on process.env.IS_ELECTRON */
// ProxySettings
'useProxy',
'proxyProtocol',
'proxyHostname',
'proxyPort',
'proxyUsername',
'proxyPassword',
// ExternalPlayerSettings
'externalPlayer',
'externalPlayerExecutable',
'externalPlayerIgnoreWarnings',
'externalPlayerIgnoreDefaultArgs',
'externalPlayerCustomArgs',
'showAddedExternalPlayerCustomArgs',
// Others
'disableSmoothScrolling',
'hideToTrayOnMinimize',
'screenshotAskPath',
'screenshotFolderPath',

/* Depends on process.env.SUPPORTS_LOCAL_API */
'backendFallback',
'backendPreference',
'proxyVideos',
```

If we remove settings mentioned [here](https://github.com/FreeTubeApp/FreeTube/pull/9123#issuecomment-4418958092), we get: 

```
  /* Depends on process.env.IS_ELECTRON */
  // Others
  'disableSmoothScrolling',
  'hideToTrayOnMinimize',

  /* Depends on process.env.SUPPORTS_LOCAL_API */
  'backendFallback',
  'backendPreference',
  'proxyVideos',
```

The last three depend on `process.env.SUPPORTS_LOCAL_API`, they should probably not be transferrable. 
`hideToTrayOnMinimize` is OS-specific because it is not available on macOS. 
The remaining one is `disableSmoothScrolling`. 

So, making the whole feature electron-only would only be useful to support the smooth scrolling feature but would prevent users from importing and exporting settings on mobile.
It could still be possible to later implement your idea of having a flexible way to restart the app, no matter if it's electron or not, and add the missing support for `disableSmoothScrolling`.

I will update my MR to change the `settingsNotExportable` with all the settings mentioned above.
Let me know what you think, I've tried to imagine the best scenario for users, code maintainability and complexity.

src/renderer/components/DataSettings/DataSettings.vue
PikachuEXE: The block below with `if (!Object.hasOwn(currentSettings, importedKey)) {` checks for unknown setting
But there is no block for checking if the incoming setting should be allowed to be imported

Shadorc: My naming is probably unclear, but `const currentSettings = exportableSettings.value`, so it only contains keys that are exportable, and in my current implementation, exportable keys are also importable. If the implementation changes, allowing settings to be exported but not imported on all platforms, then it should definitely be checked here

PikachuEXE: I can see I missed that one as I was searching for `settingsNotTransferrable`
However there might be still one issue
The message is the same for actual unknown setting keys and non transferrable setting keys

Shadorc: From what I understand, the toast message is only passed to the `FtToast` component as `{{ toast.message }}`, it couldn't really lead to malicious code execution or some exploit. The worst things that I could think of is either breaking the toast UI by having a really long message or having a false message like "Visit this malicious website <url> for more info"
The only solution that I see is truncating or completely removing the `Unknown setting key` toast message parameter.
Is that what you are referring too?
If not, could you give example please? because I can't find what could be problematic

absidue: As I understand it Pika is asking for there to be two separate messages, one for settings that were not imported because they are unknown and another for settings that were not imported because they are not transferable. The former could point to an actual issue with the imported file, the latter is purely informational.

Shadorc: Thanks for clarifying, it makes sense
I thought that "The message is the same" was referring to Pika's initial message saying that it needed filtering, not the toast message. 
I'll change that!

PikachuEXE (dismissed):

src/renderer/store/modules/settings.js
```diff
@@ -414,10 +423,47 @@ const sideEffectHandlers = {
...
+  'externalPlayerIgnoreDefaultArgs',
+  'externalPlayerCustomArgs',
+  'showAddedExternalPlayerCustomArgs',
+  // Others
+  'disableSmoothScrolling',
+  'hideToTrayOnMinimize',
+  'screenshotAskPath',
+  'screenshotFolderPath',
+
+  /* Depends on process.env.SUPPORTS_LOCAL_API */
```
ghost, line 449: I'll let others chime in but I think we could keep the SUPPORTS_LOCAL_API settings transferable.

PikachuEXE: There are many references to these values and quite easy for the code to miss `SUPPORTS_LOCAL_API` check before using them
Better be safe first

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

absidue (commented):
I still need to do a full review but one major issue that I have already spotted is that you've invented your own export format rather than following the format that the data is actually saved in on the disk, like we do for subscriptions, playlists, watch history and search history. If the on-disk format was an sqlite database or some obscure binary format I would understand but it is just text.

Shadorc:
Indeed, there was no reason to change what already existed with other import/export formats, fixed.

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,87 @@ async function exportYouTubeSearchHistory() {
...
+    showToast(`${message}: ${err}`)
+    return
+  }
+
+  if (response === null) {
+    return
+  }
+
+  const { content } = response
+  const textDecode = content.split('\n')
```
absidue: The content variable is used once and even with it inlined the line is quite short, it would be better to inline it.
```suggestion
  const textDecode = response.content.split('\n')
```

src/renderer/store/modules/settings.js
```diff
@@ -415,10 +424,47 @@ const sideEffectHandlers = {
 
 const settingsWithSideEffects = Object.keys(sideEffectHandlers)
 
+const settingsNotTransferable = new Set([
```
absidue: Export and capitalise so that it is clear by the import alone that it is a constant.
```suggestion
export const NON_TRANSFERABLE_SETTINGS = new Set([
```

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,87 @@ async function exportYouTubeSearchHistory() {
...
+
+  textDecode.forEach((rawEntry) => {
+    const entry = JSON.parse(rawEntry)
+    if (typeof entry._id !== 'string') {
+      showToast(t('Settings.Data Settings.Setting object has insufficient data, skipping item'))
+      console.error('Missing keys:', entry)
+    } else if (!Object.hasOwn(currentSettings, entry._id)) {
+      const message = `${t('Settings.Data Settings.Unknown setting key')}: ${entry._id}`
+      showToast(message)
+    } else if (!Object.hasOwn(currentTransferableSettings, entry._id)) {
```
absidue: Instead of checking if the object has a specific property here, you can use the Set that with the names that you already created in the settings store.

```suggestion
    } else if (NON_TRANSFERABLE_SETTINGS.has(entry._id)) {
```

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,87 @@ async function exportYouTubeSearchHistory() {
...
+  const { content } = response
+  const textDecode = content.split('\n')
+  textDecode.pop()
+
+  const currentTransferableSettings = transferableSettings.value
+  const currentSettings = store.state.settings
+
+  textDecode.forEach((rawEntry) => {
+    const entry = JSON.parse(rawEntry)
+    if (typeof entry._id !== 'string') {
```
absidue: Might as well validate that the `value` exists too.
```suggestion
    if (typeof entry._id !== 'string' || !Object.hasOwn(entry, 'value')) {
```

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,87 @@ async function exportYouTubeSearchHistory() {
...
+      showToast(t('Settings.Data Settings.Setting object has insufficient data, skipping item'))
+      console.error('Missing keys:', entry)
+    } else if (!Object.hasOwn(currentSettings, entry._id)) {
+      const message = `${t('Settings.Data Settings.Unknown setting key')}: ${entry._id}`
+      showToast(message)
+    } else if (!Object.hasOwn(currentTransferableSettings, entry._id)) {
+      const message = `${t('Settings.Data Settings.Non-transferable setting key')}: ${entry._id}`
+      showToast(message)
+    } else {
+      const currentValue = currentTransferableSettings[entry._id]
```
absidue: There is no functional difference between reading the value from `currentSettings` or `currentTransferableSettings`, but reading from currentSettings means that the `currentTransferableSettings` variable can be removed.
```suggestion
      const currentValue = currentSettings[entry._id]
```

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,87 @@ async function exportYouTubeSearchHistory() {
...
+      }
+    }
+  })
+
+  showToast(t('Settings.Data Settings.All settings have been successfully imported'))
+}
+
+async function exportSettings() {
+  const settingDb = Object.entries(transferableSettings.value)
+    .map(([_id, value]) => JSON.stringify({ _id, value }))
```
absidue: As this is now the only reference to `transferableSettings` instead of creating an object in a loop in the store, then immediately calling `Object.entries` on it here and `Array.map`, it would make more sense to create the array of objects in the store getter and so you can just call `Array.join` here.

src/renderer/store/modules/settings.js
```diff
@@ -141,10 +150,10 @@ import { getSystemLocale, showToast } from '../../helpers/utils'
...
 // HELPERS
 const capitalize = str => str.charAt(0).toUpperCase() + str.slice(1)
-const defaultGetterId = settingId => 'get' + capitalize(settingId)
-const defaultMutationId = settingId => 'set' + capitalize(settingId)
-const defaultUpdaterId = settingId => 'update' + capitalize(settingId)
-const defaultSideEffectsTriggerId = settingId =>
+export const defaultGetterId = settingId => 'get' + capitalize(settingId)
+export const defaultMutationId = settingId => 'set' + capitalize(settingId)
+export const defaultUpdaterId = settingId => 'update' + capitalize(settingId)
+export const defaultSideEffectsTriggerId = settingId =>
```
absidue: Please only export what you actually use, as most linting rules and editor checks run on individual files (same goes for people working on the code), if this file is ever refactored and nothing in this file references the variable anymore, it will still be counted as used because it is exported. If the others are ever needed their exports can be added as needed in the future.

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,85 @@ async function exportYouTubeSearchHistory() {
...
+    const entry = JSON.parse(rawEntry)
+    if (typeof entry._id !== 'string' || !Object.hasOwn(entry, 'value')) {
+      showToast(t('Settings.Data Settings.Setting object has insufficient data, skipping item'))
+      console.error('Missing keys:', entry)
+    } else if (!Object.hasOwn(currentSettings, entry._id)) {
+      const message = `${t('Settings.Data Settings.Unknown setting key')}: ${entry._id}`
+      showToast(message)
+    } else if (NON_TRANSFERABLE_SETTINGS.has(entry._id)) {
+      const message = `${t('Settings.Data Settings.Non-transferable setting key')}: ${entry._id}`
+      showToast(message)
```
absidue: As these messages are displayed to the user and not all languages will have the value at the end of the string, could you please swap the string concatenation here for placeholders in the translation strings.

src/renderer/components/DataSettings/DataSettings.vue
```diff
@@ -1465,6 +1485,85 @@ async function exportYouTubeSearchHistory() {
...
+    }
+  })
+
+  showToast(t('Settings.Data Settings.All settings have been successfully imported'))
+}
+
+async function exportSettings() {
+  const settingDb = transferableSettings.value.map((settingEntry) => {
+    return JSON.stringify(settingEntry)
+  }).join('\n') + '\n'
```
absidue: As `NON_TRANSFERABLE_SETTINGS` is already imported, maybe it would be better to do the whole thing here (`Object.entries().map()`) and remove the getter in the store.This isn't a very performance critical code path so doing one extra interation and creating one small extra array is probably negligible, so I don't mind either way, up to you.

absidue (commented):
This should be the final review from my side, sorry for the many rounds, thank you for sticking around.

src/renderer/components/DataSettings/DataSettings.vue
Shadorc: It slightly reduces the amount of code changes and regroup what is used together in the same method, so I take this suggestion!

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #9134, opened 2026-05-13
Pass chapters to shaka-player

absidue:
## Pull Request Type

- [x] Feature Implementation

## Related issue

- closes #2829

## Description

As mentioned in #8772 shaka-player v5 added better support for chapters. This pull request passes the chapters to shaka-player so that chapter name appears when you hover over the seekbar and making it possible to use their in-player chapters menu (not a replacement for the main chapters menu, it just lets you easily change chapters in full screen and full window modes).

Unfortunately I ran into some issues while trying to use shaka-player markers, on short videos with few chapters such as Me at the Zoo, they displayed fine, but on longer videos with lots of chapters such as the WAN show they were a blurry mess, so this pull request still uses FreeTube's own chapter markers on the seek bar. As far as I could tell from my experimentation it seems like CSS linear-gradients were designed to either be used with only percentages or absolute pixel values, trying to combine both with calc like shaka-player does causes the blurring issues.

## Screenshots

player menu with chapters button

open chapters menu inside the player menu

chapter title shown on seek bar hover

## Testing

Open a video with chapters and check that the chapters show up on the seek bar as markers, in the tooltip when you hover over the seek bar and as a menu option in the player overflow menu. Check that all 3 things happen for audio, dash and legacy formats.

## Desktop

- **OS:** Windows
- **OS Version:** 11

PR #9136, opened 2026-05-14
Fix "replace http cache" not working in dev

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

N/A

## Description

Dunno why CORS started to be not working today when testing https://github.com/FreeTubeApp/FreeTube/pull/9134 (audio format)
Found this in console
image

## Testing

- Start dev
- Enable Replace HTTP Cache
- Start random video switch audio format
- Ensure thumbnail loaded

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Dunno if new electron version or whatever causing this

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -73,19 +73,17 @@ function runApp() {
         corsEnabled: true
       }
     },
+    ...(process.env.NODE_ENV === 'production'
+      ? [{
+          scheme: 'app',
+          privileges: {
+            standard: true,
+            secure: true,
+            supportFetchAPI: true
+          }
+        }]
+      : []),
   ])
-  if (process.env.NODE_ENV === 'production') {
-    protocol.registerSchemesAsPrivileged([
-      {
-        scheme: 'app',
-        privileges: {
-          standard: true,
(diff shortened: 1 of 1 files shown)


absidue (commented):
From the docs:

> This method can only be used before the ready event of the app module gets emitted and can be called only once.

https://www.electronjs.org/docs/latest/api/protocol#protocolregisterschemesasprivilegedcustomschemes

I would suggest moving it to the else block on the if prod, that way for release builds it will be called once with both schemes and in dev once with one scheme

absidue:
> Dunno why CORS started to be not working today

Electron 42, I only fixed the release builds in the electron pr, i should have fixed dev at the same time, sorry.

PikachuEXE:
OK fixed

absidue (approved):

ghost (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #8653, opened 2026-02-12
[Feature Request]: Add fallback behavior for RSS ratelimiting


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Problem Description

When YouTube is having a RSS outage on the RSS feeds users are met with the following on our Subscription page

* Your subscribed channels currently does not have any videos
* Channels with errors

While technically correct it isnt very user friendly.

### Proposed Solution

Show the videos that were listed before refreshing the subscriptions and show a toast that mentions that fetching failed instead of the things listed above

### Alternatives Considered

None.

### Issue Labels

ease of use improvement, display more information to user, visual improvement

ArthurKun21:
I would like to be able to edit/update profile that I could select channels up to 100 from the same profile as priority/pinned/(I don't have a good word to describe this) such that when RSS rate limiting hits it would fallback to using normal api to fetch those 100 instead.

EvilGremlin:
@ArthurKun21 Proper way is to fetch subscriptions in <100 chunks automatically, i don't know why no one implementing it still, even with RSS constant failing. Making multiple profiles is NOT a solution.

Zazie-Lavender:
Would it be possible to implement this in a way that allows the user to completely disable RSS fetching; and therefore fetching more than 100 channels per profile? Obviously enabling a setting should come with a banner that tells the user clearly that profiles that are "too big" won't fetch all channels in this mode.

I liked the caching idea as well...perhaps a banner that tells the user the displayed video feed is stale, how old it is; and explains the constant ongoing YouTube outage? (This may help with queries about #8643 as well.)

[...]

PR #9157, opened 2026-05-19
Feature: Add fallback behavior for RSS ratelimiting 

laclcia:
## Pull Request Type
- [x] Feature Implementation

## Related issue
[8653](https://github.com/FreeTubeApp/FreeTube/issues/8653)

## Description
this add's a toggle in the subscription section of the settings page (default off) witch when enabled will change the behavior when fetching subscriptions for large (above 125) lists. instead of falling back to RSS witch often has problems as YT's rss goes down often it will instead chunk fetch requests in chunk's of 80 separated by 2 seconds (2000 ms) to avoid rate limiting without relying on rss. 

## Screenshots 
image

## Testing
1 launch app and refresh a large subscription list. the RSS fetch should still work.
2 go to settings tab and the subscription subsection and check "limit request fallback without RSS" 
3 go back to subscriptions and refresh. you should notice the bottom loading bar stopping for 2 sec then resuming (and stopping again since it fetches in chunk's of 80) a few times.
4 once all subscriptions are fetched they should all have timestamps confirming it worked without rss.  

## Desktop

- **OS:Cachyos
- **OS Version:X86_64
- **FreeTube version:v0.24.0 Beta

## Additional context
i originally created a proof of concept for this about a month ago and posted the diff in the original feature request thread here https://github.com/FreeTubeApp/FreeTube/issues/8653#issuecomment-4315729687 . i have been running that fix daily on my computers (and some friends) since then and it work's without problem.
English is a second language for me and i am not a developer by trade. this is the second pull request i have made (the first was for a one line fix for another most likely abandoned project). i have tried to the best of my ability to follow the guidelines in the "Code Contributions" file.

### DIFF
--- a/src/renderer/components/SubscriptionSettings/SubscriptionSettings.vue
+++ b/src/renderer/components/SubscriptionSettings/SubscriptionSettings.vue
@@ -24,6 +24,13 @@
           compact
           @change="updateUnsubscriptionPopupStatus"
         />
+        <FtToggleSwitch
+          :label="$t('Settings.Subscription Settings.Limit Request Fallback Without RSS')"
+          :default-value="limitRequestFallbackWithoutRss"
+          :tooltip="$t('Tooltips.Subscription Settings.Limit Request Fallback Without RSS')"
+          compact
+          @change="updateLimitRequestFallbackWithoutRss"
+        />
       </div>
       <div class="switchColumn">
         <FtToggleSwitch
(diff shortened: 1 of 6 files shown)


PikachuEXE:
YT/FT broken on getting videos on sub just this hour (no issue a few hours ago
Will test again once it's resolved

PikachuEXE (commented):
Briefly tested with dev merged, a few things I can think of:
- Might be better to place the new setting below/above the RSS one (no sure about which one is better)
- New setting should be disabled when RSS one enabled

Also still wondering if we need a new setting instead of:
If size > 125 then use new batching logic instead of RSS
Else all done in one batch
No sure why we still need to auto switch to RSS with this new method added

ghost:
I think it'd make sense to implement some of this logic for SubscriptionPosts (which doesn't support RSS but could support this chunking of requests when RSS is force enabled and Limit Fallback Request is set)

efb4f5ff-1298-471a-8973-3d47447115dc:
@laclcia  Please rebase into dev branch

PR #9203, opened 2026-05-28
Fix electron extract issue in dev with latest NodeJS

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

- https://github.com/electron/electron/issues/51619 (The one reported in electron)
- https://github.com/max-mapper/extract-zip/issues/154 (The one reported in the library electron uses)
- https://github.com/thejoshwolfe/yauzl/pull/170 (The actual fix)

## Description

Cannot run dev after NodeJS updated to 24.16.0
See reason in issue above
Since electron uses `extract-zip` (a thin wrapper around `yauzl`) and it locks `yauzl` on 2.x
The fix cannot be included until we override its version to 3.x

## Screenshots 

Failure looks like
image

## Testing

- Update NodeJS to 24.16.0+ (or 26.1.0+ if you want but FT uses 24.x in CI)
- `rm -rf node_modules` (necessary to emulate the installation of new electron versions)
- `pnpm install && pnpm dev`
- Ensure app starts

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**


absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
Added `lefthook` packages to `minimumReleaseAgeExclude` due to recent version updated via dependabot (and it's using older pnpm = ignore `minimumReleaseAge`
To be removed after pnpm 11 support added

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

absidue (dismissed):

Dinsi (approved):

ghost (dismissed):

pnpm-workspace.yaml
```diff
@@ -10,3 +10,8 @@ allowBuilds:
 
 overrides:
   uuid@<11.1.1: '>=11.1.1'
+  yauzl: '>=3.3.1'
+
+minimumReleaseAgeExclude:
+- "lefthook"
+- "lefthook-*"
```
absidue: Still don't understand why these lefthook changes are here.

absidue (dismissed):

pnpm-workspace.yaml
PikachuEXE: Last time I got merge conflict due to lefthook updated, and it's updated with older pnpm 10 = it ignores `minimumReleaseAge` = pnpm install will fail if not added to `minimumReleaseAgeExclude` 

https://github.com/FreeTubeApp/FreeTube/pull/9203#issuecomment-4580719202

absidue: It's been quite a while since then though, so it should be working fine now right? If you are leaving it there because it could happen again, then it could also happen to any dependency, the proper solution is to configure the cooldown option in the dependabot  config: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown- it doesn't require a pnpm update or manually adding excludes for every dependency.

PikachuEXE:
```suggestion
```

PikachuEXE: Updated but not adding cooldown in this PR as it's not related

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ghost (approved):

PR #9208, opened 2026-05-29
Fix: Hide 'Minimize to system tray' setting on Wayland

### LINKED ISSUE #9187: [Bug]: 'Minimize to Tray' does not work with Wayland
### Guidelines

### Describe the bug

1. Enable "Minimize to system tray" setting
2. Minimize window
3. The window is not minimized to tray

### Expected Behavior

Window should be minimized to tray

### Issue Labels

usability issue

### FreeTube Version

v0.24.0-beta

### Operating System Version

Bazzite 44 NVIDIA Edition (Wayland)

### Installation Method

Flathub

### Primary API used

Local API

_No response_

_No response_

### Additional Information

After some investigation, I found that the 'minimize' event is never triggered by `BrowserWindow`.

I searched to see if this was an Electron issue and found this issue: https://github.com/electron/electron/issues/44586. It's not directly related since it concerns macOS, but I tried the sample used to reproduce the bug. When using the sample as-is with `"electron": "31.2.1"`, I see the 'minimize' log being displayed. However, when I update to `"electron": "42.0.1"` (the same version as FreeTube), it does not work and displays the following error: `'--ozone-platform=wayland' is not compatible with Vulkan. Consider switching to '--ozone-platform=x11' or disabling Vulkan`.

If I add `--ozone-platform=x11` to FreeTube's Electron args, the 'minimize' event is correctly triggered and the tray icon works.

I'm not sure whether the issue originates from FreeTube, Electron, or Chromium, so feel free to close it if you consider it to be upstream.

Edit: I've open an issue in electron project.

Interesting links:
- https://github.com/brave/brave-browser/issues/43097
- https://issues.chromium.org/issues/41392107

### Nightly Build

- [X] I have encountered this bug in the latest **official** [nightly build](https://docs.freetubeapp.io/development/nightly-builds), do not check this box if you are using an unofficial or custom build.

Shadorc:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/9187

## Description

Wayland does not have a 'minimize' event (see https://github.com/electron/electron/issues/51766#issuecomment-4536275295). 
FreeTube can't really do anything on its side, nor can Electron.
Similar to MacOS, the fix is to hide this setting if Wayland is detected.

## Testing
**Linux Wayland**
1. Execute FreeTube on Linux Wayland
2. Check that 'Minimize to tray' setting is hidden

**Linux Wayland with compatibility mode**
1. Execute FreeTube on Linux Wayland with `--ozone-platform=x11`
2. Check that 'Minimize to tray' setting is visible and working as intended

**Windows**
1. Execute FreeTube on Windows
2. Check that 'Minimize to tray' setting is visible and working as intended


efb4f5ff-1298-471a-8973-3d47447115dc:
im wondering if we could just hide the setting when wayland is detected

Shadorc:
I've spent some time searching to find out how to detect if we are using Wayland, but there doesn't seem to be a simple solution. 
The standard approach is to rely on `XDG_SESSION_TYPE`, but as I mentioned in my issue, you can run Electron with `--ozone-platform=x11` and it will work. This means we should also check for this argument. 
There are people https://github.com/FreeTubeApp/FreeTube/issues/9187#issuecomment-4527296183 that use Wayland but seem to have their OS running FreeTube with X11 compatibility mode or something. 
You can also use software like XWayland to run Wayland apps on an X11 server. It might change the `XDG_SESSION_TYPE` value correctly, but that seems like too many scenarios to implement something simple and reliable.

Shadorc:
I've tested with two electron-based apps that I use: Discord and Signal. 
They have a different approach, they both consider that "Minimize to system tray" can only occur when a user presses the close button, removing the app from the taskbar, not when minimizing. 
According to https://github.com/signalapp/Signal-Desktop/issues/5840#issuecomment-1590340967, it is the same behaviour for Battle.net

absidue:
The setting in FreeTube is explicitly designed to be minimize to system tray, not close to system tray. As for detecting wayland you can use the same approach as the existing code: https://github.com/FreeTubeApp/FreeTube/blob/development/src/main/index.js#L1163.

Shadorc:
I apologize if I wasn't clear. I wasn't suggesting that FreeTube should mimic "close to system tray" behavior. I was simply providing examples of how other Electron apps handle the "Minimize to tray" setting. If any of those apps successfully minimized to the system tray like FreeTube does on X11, it could have indicated that a solution was possible.

Thanks for the code reference. As I mentioned in my concerns [here](https://github.com/FreeTubeApp/FreeTube/pull/9208#issuecomment-4573458659), I wasn't sure it would cover all possible scenarios, but if you're confirming it does, that's great! I'll use it to hide the setting when it's unavailable.

src/main/index.js
```diff
@@ -1346,6 +1346,12 @@ function runApp() {
     }
   })
 
+  ipcMain.handle(IpcChannels.GET_CMD_SWITCH_VALUE, (event, name) => {
+    if (isFreeTubeUrl(event.senderFrame.url)) {
+      return app.commandLine.getSwitchValue(name)
+    }
+  })
```
absidue: IPC handlers and the preload script grant the page more power than a normal website would have, so they should always be as narrowly scoped as possible. In this case the page only needs to know whether the app is running on wayland or not, it does not require arbitrary access to all command line switches. Please change this handler to just return a boolean whether the app is running on wayland or not, it also makes it easier to maintain because you don't need to look at multiple different files to figure out how FreeTube determines that it is running on wayland or not.

src/renderer/components/GeneralSettings/GeneralSettings.vue
```diff
@@ -42,7 +42,7 @@
           @change="updateOpenDeepLinksInNewWindow"
         />
         <FtToggleSwitch
-          v-if="!IS_MAC && USING_ELECTRON"
+          v-if="!IS_MAC && !isLinuxWayland && USING_ELECTRON"
```
ghost, line 45: My only concern would be if someone on Wayland enabled this setting before this PR gets merged then the setting would still be enabled but hidden

Shadorc: I could add a setting migration, to disable it when updating freetube, but I'm not sure if this is really necessary.

Edit: After thinking about it, it could be problematic indeed if one day Wayland decides to support this state, people wouldn't be able to change FreeTube's behaviour. I will fix that

Shadorc: Fixed in https://github.com/FreeTubeApp/FreeTube/pull/9208/changes/5e262c70f9331baaf24fc776f3dfc521c90c51bf

Shadorc:
I'm not sure what is happening with the CI, it builds on my side and `yarn run lint-fix` does not show any warning/error. It indicates "Module not found" for three different modules. I've rebased on development branch with no luck.

src/renderer/components/GeneralSettings/GeneralSettings.vue
absidue: Please revert the migration, the settings value should stay as people are known to share their FreeTube data between different machines. You should just skip the relevant code paths on wayland, so everywhere that there is an existing macOS check for the tray minimize code in the `src/main/index.js` file you also need to add a linux + wayland guard.

src/datastores/handlers/base.js
```diff
@@ -1,3 +1,4 @@
+import { app } from 'electron'
```
absidue: This import is the reason that the web build is failing, because you are importing the `'electron'` module in a file that gets bundled into both the Electron main process and the "web frontend" build (that's a smoke test for downstream FreeTubeAndroid). However as mentioned above the migration should be removed, so this issue will disappear too.

src/renderer/components/GeneralSettings/GeneralSettings.vue
```diff
@@ -190,6 +190,13 @@ const IS_MAC = process.platform === 'darwin'
 const { t } = useI18n()
 const router = useRouter()
 
+// The 'minimize' event doesn't fire on wayland
+// https://github.com/electron/electron/issues/51766
+const isLinuxWayland = ref(false)
+onMounted(async () => {
+  isLinuxWayland.value = process.platform === 'linux' && await window.ftElectron.getWaylandPlatform()
+})
```
absidue:
```suggestion
if (process.env.IS_ELECTRON && process.platform === 'linux') {
  onMounted(async () => {
    isLinuxWayland.value = await window.ftElectron.getWaylandPlatform()
  })
}
```

src/preload/interface.js
```diff
@@ -30,6 +30,13 @@ export default {
     return ipcRenderer.invoke(IpcChannels.GET_SYSTEM_LOCALE)
   },
 
+  /**
+   * @returns {Promise<boolean>}
+   */
+  getWaylandPlatform: () => {
+    return ipcRenderer.invoke(IpcChannels.GET_WAYLAND_PLATFORM)
```
absidue: As the value is a boolean I would prefer `is` to `get` e.g. `isWayland` and `IS_WAYLAND`

src/datastores/handlers/base.js
Shadorc: Thanks for the explanation!

src/preload/interface.js
Shadorc: I agree, I was a bit confused by `GET_REPLACE_HTTP_CACHE`, which is also a boolean, and decided to go with what seemed to be the current "naming convention". I will change that

absidue:
Looks like you accidentally added back the yarn.lock file but the development branch switched to pnpm a few weeks ago, so you probably need to rebase again after your reverts.

src/renderer/components/GeneralSettings/GeneralSettings.vue
Shadorc: Done! The remaining `process.platform !== 'darwin'` occurrences do not seem to concern tray

efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM! Tested on Fedora and Windows

Sorry for the wait.

absidue (approved):

PikachuEXE (approved):
Tested on windows

PR #9253, opened 2026-06-08
Set @parcel/watcher to false in the pnpm allowBuilds setting

absidue:
## Pull Request Type

- [x] Other

## Description

`@parcel/watcher` now has an install script that compiles it from source if the `npm_config_build_from_source` npm setting is enabled, as we are fine using the pre-built binaries we can disable that postinstall script (also stops pnpm complaining about it).

- https://npmx.dev/package-code/@parcel/watcher/v/2.5.6/package.json#L35
- https://npmx.dev/package-code/@parcel/watcher/v/2.5.6/scripts%2Fbuild-from-source.js

## Testing

1. `pnpm install`
2. `pnpm run dev`

If the pnpm install works and the dev server starts correctly this change works and doesn't break anything.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,6 +4,7 @@
 
 # https://pnpm.io/next/settings#allowbuilds
 allowBuilds:
+  '@parcel/watcher': false # Only required when you want to build @parcel/watcher from source
   electron-winstaller: false # We build NSIS installers on Windows, not squirrel ones
   lefthook: true # Installing git hooks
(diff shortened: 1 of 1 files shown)


PikachuEXE:
I tried that locally before and I re-tested now the warning still shows and I dunno why
image

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
For me to get rid of the warning
I need to run `pnpm approve-builds` (even though config is changed already)
Dunno why (no file changed after running it

ghost (approved):

PR #9261, opened 2026-06-09
Fix proxy config not working on startup

bunchy7ss:
i built it for .deb only and it worked, on my system atleast (No reason this shouldnt work on other systems.)
i did not see an existing issue for this, therefore i cannot link one

## Pull Request Type
- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Description
session.setProxy() was being used to configure the proxy during app startup, but it doesn't reliably take effect because electrons network service isn't always ready at that point, So, i switched it to app.commandLine.appendSwitch which sets the proxy at the chromium level before any weird electrion networking happens.

The runtime proxy toggle (settings page) still uses session.setProxy() and that works fine because the network service is already running atp..

## Testing
Setup a (atleast in my usage) a microsocks proxy, 
configure it in the settings, 
close the app, re open it, it should fail to connect to the youtube api. 
go to the settings, 
change the proxy setting, click "test", (it wont work if setting are invalid) 
change them back, then it should work.

## Desktop
- **OS: Linux Mint 22.2 x86_64 **
- **OS Version: 6.18.21-x64v3-xanmod1 **
- **FreeTube version: 0.24.1**

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -541,9 +541,7 @@ function runApp() {
     if (useProxy) {
       proxyUrl = `${proxyProtocol}://${proxyHostname}:${proxyPort}`
 
-      session.defaultSession.setProxy({
-        proxyRules: proxyUrl
-      })
+      app.commandLine.appendSwitch('proxy-server', proxyUrl)
     }
 
     const fixedUserAgent = session.defaultSession.getUserAgent()


bunchy7ss:
Crap, i shouldve previewed the PR before sending it, sorry that it dosent align with the template 100%

PR #9315, opened 2026-06-22
Fix keyboard shortcuts arent't working during "SABR backoff"

### LINKED ISSUE #9101: [Bug]: Multiple Player keyboard shortcuts arent't working during "SABR backoff" timer
### Guidelines

### Describe the bug

1. Open any video
2. There is always a SABR timer below the video (4.8 seconds, but the .8 seem to be ignored)
3. During this timer, press the F button to go fullscreen
4. Notice that it doesn't go fullscreen
5. Before the timer finishes, click the button to go fullscreen
6. This does work

### Expected Behavior

When the button to go fullscreen is actually functional during this timer, the keybind to do the same should still respond as well.

### Issue Labels

accessibility issue, inconsistent behavior, keyboard control not working

### FreeTube Version

v0.24.0 Beta

### Operating System Version

Windows 10 Home 22H2

### Installation Method

Portable

### Primary API used

Local API

_No response_

_No response_

### Additional Information

Note this keybind works perfectly fine during video playback, and before the SABR implementation (since there wasn't any video player on screen yet during this timer to make fullscreen, I guess).

### Nightly Build

ozrendev:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #9101 

## Description

Re-orders keyboard shortcuts that are not dependent on video playback so they can be triggered before "SABR backoff" has finished

Thanks to @Shadorc for pointing out where the keyboard shortcuts are handled

## Testing

Open any video and while "SABR backoff" is visible, try to trigger the following keyboard shortcuts: Theater Mode (t), Full Window (s) and Full screen (f)

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -2280,6 +2280,44 @@ export default defineComponent({
      * @param {KeyboardEvent} event
      */
     function keyboardShortcutHandler(event) {
+      switch (event.key.toLowerCase()) {
+        case KeyboardShortcuts.VIDEO_PLAYER.GENERAL.FULLSCREEN:
+          // Toggle full screen
+          event.preventDefault()
+          ui.getControls().toggleFullScreen()
+          blurTooltipButtons()
+          break
+        case 'escape':
+          // Exit full window
+          if (fullWindowEnabled.value) {
+            event.preventDefault()
+
+            events.dispatchEvent(new CustomEvent('setFullWindow', {
+              detail: false
+            }))
+          }
+          break
+        case KeyboardShortcuts.VIDEO_PLAYER.GENERAL.FULLWINDOW:
+          // Toggle full window mode
+          event.preventDefault()
+          events.dispatchEvent(new CustomEvent('setFullWindow', {
+            detail: !fullWindowEnabled.value
+          }))
+          blurTooltipButtons()
+          break
+        case KeyboardShortcuts.VIDEO_PLAYER.GENERAL.THEATRE_MODE:
+          // Toggle theatre mode
+          if (props.theatrePossible) {
+            event.preventDefault()
+
+            events.dispatchEvent(new CustomEvent('toggleTheatreMode', {
+              detail: !props.useTheatreMode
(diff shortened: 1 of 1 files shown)


Shadorc:
What about early exit conditions that are no longer evaluated, is it fine to skip them?
For example 
```
if (document.activeElement.classList.contains('ft-input') || event.altKey) {
    return
}
```

src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
```diff
@@ -2280,6 +2280,44 @@ export default defineComponent({
      * @param {KeyboardEvent} event
      */
     function keyboardShortcutHandler(event) {
+      switch (event.key.toLowerCase()) {
```
PikachuEXE: A potential issue is that this not just skips `!player || !hasLoaded.value` check but also other checks like `document.activeElement.classList.contains('ft-input') || event.altKey`

ozrendev:
Yeah, you're both right. I just saw `!hasLoaded.value` and got ahead of myself. I've re-ordered the `switch` statement now.

I think by isolating `!hasLoaded.value` it should be okay, right? This is the only check that shouldn't need to be executed for the modified shortcuts

PikachuEXE (dismissed):

absidue (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Should the following be included in here?

i
c
m

absidue:
- m (toggle mute): yes
- c (toggle captions): no, because shaka-player only considers the player loaded after both the manifest and first segment have loaded and as we get the backoff time as part of getting the first segment and wait in there. It would be better to not mess with player internal stuff while we are actively stalling its loading process.
- i (pip): if it works yes (doesn't work with audio only so that guard would need to remain in place)

ozrendev:
Latest commit adds (m) mute. I've also tested it and it looks like pip doesn't work at all

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

PR #9321, opened 2026-06-23
Fix: publishedText is missing but the video isn't live or upcoming

### LINKED ISSUE #8085: [Bug]: Posts tab throws silent error `publishedText is missing but the video isn't live or upcoming`
### Guidelines

### Describe the bug

My Post tab on the Subscription page and even some channels were throwing these errors in the console and i finally found out why.

1. Go to http://youtube.com/post/Ugkxd4Opnnau70s5l6q9WLBzLoYY2toMv78N?feature=shared
2. See that video has been taken down

Sometimes this also occurs on the Search page but there isnt a removed video thumbnail visbile

### Expected Behavior

This doesnt break the Posts tab but would be nice to not receive the error anymore

### Issue Labels

race condition

### FreeTube Version

v0.23.9-nightly-6476 Beta

### Operating System Version

### Installation Method

.exe

### Primary API used

Local API

_No response_

### Additional Information

_No response_

### Nightly Build

- [x] I have encountered this bug in the latest [nightly build](https://docs.freetubeapp.io/development/nightly-builds).

Shadorc:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/8085

## Description

It is a valid state to have an empty `publishedText` when the video is not live/premiere/upcoming.  

A good example is the [paramountplus channel](https://www.youtube.com/@paramountplus) when using an IP located in USA:

**YouTube:**
image

**FreeTube:**
image

Differentiating valid and unvalid cases does not seem possible. It does not break FreeTube UI and does not impact the user when a video is returned without `publishedText`, so I suggest to remove the log.

## Testing

1. Go to [paramountplus channel](https://www.youtube.com/@paramountplus) using an IP located in USA
2. Check that there is no error

### DIFF
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -69,7 +69,6 @@ export function calculatePublishedDate(publishedText, isLive = false, isUpcoming
   }
 
   if (!publishedText) {
-    console.error("publishedText is missing but the video isn't live or upcoming")
     return undefined
   }
 


PikachuEXE (approved):

absidue:
Most of the time when that message appears it is an actual issue that needs to be looked into, but I guess you'll just have to rely on noticing the parsing issues yourself. In this case with those unwatchable ads for a paid subscription service it is valid for the published date to be missing because they are not YouTube videos (the more correct solution would probably be to get rid of those ads like we do with members only videos, but for the moment I guess removing some genuinely useful logging will have to do).

Shadorc:
I haven't found a way to distinguish these videos (the ones without `publishText`) from the others, but maybe there is one. 
I will try searching again to see if I find something.

If you think it’s better to keep the error message because it’s more important to get an error when a viewable video has a parsing issue than to have no errors for a video that can’t be viewed, I understand, and this MR can be closed.

absidue:
@PikachuEXE @efb4f5ff-1298-471a-8973-3d47447115dc Which would you prefer: Silent failure in situations where it is an actual problem or errors in situations where it isn't actually an error?

absidue:
It's probably fine to remove, for the rest of FreeTube we also go with "the users will inevitably complain when something breaks and even when it doesn't" approach, so using that here too is probably fine.

PikachuEXE:
If the error only appears in console, I guess it's fine to get a warning (instead of current error
No idea what's the ratio of valid vs actual issue current date, but I guess if it's warning in console only normal user won't see it?

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #9327, opened 2026-06-25
Add limit the number of videos displayed for each channel should be set to a max of 15 when RSS is enabled

### LINKED ISSUE #8105: [Feature Request]: `Limit the number of videos displayed for each channel` should be set to a max of 15 when RSS is enabled/forced
### Guidelines

### Problem Description

In Settings -> Subscription there is a setting called `Limit the number of videos displayed for each channel` and can be set to a maximum of 30. If RSS is enabled or enforced the maximum amount of videos shown per channel is 15. The setting still allows the user to set the limit to 30 even when the subscription page only shows 15 due to RSS being used.

### Proposed Solution

Set a cap of 15 when RSS is enabled or forced

### Alternatives Considered

None

### Issue Labels

visual improvement, display more information to user

### Additional Information

_No response_

ozrendev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #8105 

## Description

Updates the UI so that when fetching feeds via RSS is enabled, the max video slider value is capped at 15 (technical limitation of RSS feed fetching)

## Testing

Go to Settings -> Subscription
Ensure 'Limit the number of videos displayed for each channel' is enabled
Toggle 'Fetch feeds from RSS' and change the value slider
See what the limit is 15 when RSS is enabled and the limit is 30 when RSS is disabled
In all cases, confirm the number of videos displayed on the Subscription page matches the value slider

## Additional context

Was unsure if the RSS disabled/enabled values should be set as variables, if anyone has feedback on this?

### DIFF
--- a/src/renderer/components/SubscriptionSettings/SubscriptionSettings.vue
+++ b/src/renderer/components/SubscriptionSettings/SubscriptionSettings.vue
@@ -38,7 +38,7 @@
             :default-value="onlyShowLatestFromChannelNumber"
             :disabled="!onlyShowLatestFromChannel"
             :min-value="1"
-            :max-value="30"
+            :max-value="useRssFeeds ? 15 : 30"
             :step="1"
             @change="updateOnlyShowLatestFromChannelNumber"
           />
@@ -75,6 +75,10 @@ const useRssFeeds = computed(() => store.getters.getUseRssFeeds)
  */
 function updateUseRssFeeds(value) {
   store.dispatch('updateUseRssFeeds', value)
+  const max = value ? 15 : 30
+  if (onlyShowLatestFromChannelNumber.value > max) {
+    updateOnlyShowLatestFromChannelNumber(max)
+  }
 }
 
 /** @type {import('vue').ComputedRef<boolean>} */
--- a/src/renderer/components/SubscriptionSettings/SubscriptionSettings.vue
+++ b/src/renderer/components/SubscriptionSettings/SubscriptionSettings.vue
@@ -78,6 +78,8 @@ function updateUseRssFeeds(value) {
   const max = value ? 15 : 30
   if (onlyShowLatestFromChannelNumber.value > max) {
     updateOnlyShowLatestFromChannelNumber(max)
+  } else if (!value && onlyShowLatestFromChannelNumber.value === 15) {
+    updateOnlyShowLatestFromChannelNumber(max)
   }
 }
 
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Im wondering if we should update the value to 30 again if rss is set from enabled to disabled or to the value it was set previously (more a convenience thing)

https://github.com/user-attachments/assets/253cc0bb-d453-4617-9331-bb3405e17536

ozrendev:
I think it should remember the user's last value, while the settings are linked, it still is technically a separate setting. Although maybe if the value is already at the max, it could be safely assumed the user would like it maximized either way? It also makes it clearer the maximum value has changed

https://github.com/user-attachments/assets/37bec129-83d5-4c22-b07f-1c3bd916013f

efb4f5ff-1298-471a-8973-3d47447115dc:
Behaviour explained and shown in screencap seems fine to me, so lets go with that

efb4f5ff-1298-471a-8973-3d47447115dc:
Oh hmm lets say i deliberately choose to only see 20 (or anything above 15) without rss. Now i enable/disable rss and it will jump to 30. Isnt that undesired?

ozrendev:
Just getting around to this again. Still deciding what should be the default action of the slider. Maybe it could remember the last selected value, but not sure if that just adds unnecessary complexity?

Shadorc:
Remembering the last value seems the more user friendly, but also the more complex to implement 
Imo, truncating to 15 and leaving the value to 15, even if 30 is the new maximum, seems the easier to implement while still addressing the original feature request.

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE @absidue thoughts?

PikachuEXE:
I don't even understand why there are 2 values for RSS on/off...

absidue:
Because YouTube's RSS feeds contain 15 videos whereas one page in the API responses for the channel videos/live/shorts channels tabs contain 30 videos.

PikachuEXE:
I think it's fine to just use current value when limit 15 > 30
Maybe add a tooltip if needed

efb4f5ff-1298-471a-8973-3d47447115dc:
So i think that my suggestion in https://github.com/FreeTubeApp/FreeTube/pull/9327#pullrequestreview-4581374259 added unintended complexity, Maybe we should revert to previous state?

ozrendev:
Yeah, I think it's best to revert it. I don't see it being a big issue either way and this way we avoid added complexity.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

Shadorc (approved):

PR #9325, opened 2026-06-25
Fix online playlist parsing & revert workaround with youtubei.js update

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

Replaces https://github.com/FreeTubeApp/FreeTube/pull/9291
Reverts workaround introduced in https://github.com/FreeTubeApp/FreeTube/pull/9227
Closes https://github.com/FreeTubeApp/FreeTube/issues/8639
Closes https://github.com/FreeTubeApp/FreeTube/issues/9263

## Description

Since youtubei.js 17.2.0 released, workaround not needed anymore
Long playlist continuation also fixed = can load whole playlist with item > 200
Also fixed online playlist item (LockupView) handling

## Testing

See test cases in linked PRs

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Need to add minimumReleaseAgeExclude for youtubei.js coz the latest version is only out for a day or less

PR #9334, opened 2026-06-27
Fix video view all playlist

### LINKED ISSUE #629: Play All button
Some channels I subscribe to upload multiple videos a day. When I play the newest video the next on the queue is usually from a totally different channel. So what I used to do in YouTube was go to their channel and click the Videos section and there's a Play All button. This would play their videos from the latest to the oldest so I don't have to go back after every video and look for the next one to play.

r3dArch:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #629 

## Description

Creates a path for the view playlist button based on channel ID, allowing users to play all of a channel's videos.

## Screenshots 

Before
Grimshot 2026-06-26 21-53-12
After
Grimshot 2026-06-26 21-53-26

## Testing

- Go to a channel
- Go to home (if not there already)
- View Playlist should be visible
- Click View Playlist and it should load a playlist with all of the channel's videos (requires #9325)

## Desktop

- **OS:** Arch Linux
- **OS Version:** Kernel 7.0.10-zen-1-zen
- **FreeTube version:** 0.24.1

## Additional context

None

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -4,6 +4,7 @@ const IpcChannels = {
   DISABLE_PROXY: 'disable-proxy',
   GET_SYSTEM_LOCALE: 'get-system-locale',
   GET_NAVIGATION_HISTORY: 'get-navigation-history',
+  IS_WAYLAND_PLATFORM: 'is-wayland-platform',
   STOP_POWER_SAVE_BLOCKER: 'stop-power-save-blocker',
   START_POWER_SAVE_BLOCKER: 'start-power-save-blocker',
   CREATE_NEW_WINDOW: 'create-new-window',
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -290,6 +290,7 @@ function runApp() {
   let trayOnMinimize = false
   let trayWindows = []
   const trayMaximizedWindows = {}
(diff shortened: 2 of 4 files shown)


r3dArch:
After doing some more testing I found a better way to fix this so that it uses the getChannelPlaylistId function and works for popular videos and shorts. I had to change the type for finding the continuation, but I'm not sure if that is the correct way to fix that; and it seems as though the videos load all at once rather than lazily (that's what the continuation is for, right?). I also need to figure out why the shorts only load the first 100 videos, but that will have to wait until tomorrow.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @r3dArch thanks for looking into this issue. 

There is a better way to address this one by doing this. 

Add play all buttons inside the relevant tabs and auto-pick the correct playlist based on the current tab and sort direction. 

This has the advantage of it always showing instead of only when the channel has the videos, live or shorts sections added to their home tab.

Would it be possible for you to implement these changes?

r3dArch:
Hi @efb4f5ff-1298-471a-8973-3d47447115dc , I think I understand what you mean and will try to implement that soon. Would you still want the current changes as well? Thank you.

efb4f5ff-1298-471a-8973-3d47447115dc:
Hi @r3dArch the requested changes will replace current so no need to keep current changes

PR #9339, opened 2026-06-28
Build static runtime AppImages with electron-builder

absidue:
## Pull Request Type

- [x] Other

## Description

Currently we build AppImages with the old AppImage runtime with electron-builder, then in a second step we unpack those appimages and recreate them with the static runtime using appimagetool. There are a few issues with that approach firstly recent-ish releases of electron-builder received support for building AppImages with the static runtime itself (will be the default in electron-builder 27), secondly appimagetool doesn't have versionsed releases so we are currently just downloading whichever the latest version is when the workflow runs in the build and release workflows, which is not great for security and lastly it means that if you build FreeTube locally you need to manually perform those extra steps to get a static runtime AppImage of FreeTube.

This pull request switches to telling electron-builder to create static runtime AppImages and removes the extra unpack and rebuild steps. In the release workflow we were also using appimagetool to inject the AppImageUpdate update information, as that is just writing to a specific offset in the binary and calling zsyncmake it didn't seem worth keeping appimagetool just for that, so I replicated it with a few bash commands (find the offset of the `.upd_info` section, open the file, write the info to that offset, close the file, run zsyncmake). Even with doing those steps ourselves in bash it still speeds up the workflow as that only takes 1 second, whereas the previous extra steps averaged around 40 seconds).

Also yes the node eval call is a bit silly but I did try to use `objcopy` and `dd` to write the update information but both resulted in appimageupdatetool saying that the AppImage was corrupted, so I went with that node JavaScript line instead.

## Testing

- Release workflow run before (with appimageupdatetool debugging): https://github.com/absidue/FreeTube/actions/runs/28318560107
- Release workflow run afterwards (with appimageupdatetool debugging): https://github.com/absidue/FreeTube/actions/runs/28326479155
- Build workflow run afterwards: https://github.com/absidue/FreeTube/actions/runs/28327514309

I don't have a Linux machine to test if the AppImage runs correctly (I would assume so but it is probably still worth testing), so if someone with a Linux VM could test the AppImage (just check if it launches) from the linked build workflow that would be very helpful.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,22 +154,6 @@ jobs:
       shell: bash
       run: pnpm run build:arm64
 
-    - name: Convert X64 AppImage to static runtime
-      if: startsWith(matrix.os, 'ubuntu') && startsWith(matrix.runtime, 'linux-x64')
-      shell: bash
-      env:
-        VERSION: ${{ steps.versionNumber.outputs.version }}
-      run: |
-        sudo apt install desktop-file-utils
-        cd build
-        appimage="FreeTube-${VERSION}.AppImage"
-        wget "https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage" -O ./appimagetool.AppImage
-        chmod +x ./"$appimage" ./appimagetool.AppImage
-        ./"$appimage" --appimage-extract && rm -f ./"$appimage"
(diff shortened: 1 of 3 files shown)


PikachuEXE (approved):
Launched in a linux VM (using `Build workflow run afterwards`

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

ghost (approved):

PR #9340, opened 2026-06-28
Add play all button

### LINKED ISSUE #629: Play All button

r3dArch:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #629 

## Description

This PR adds a button to the home tab, videos tab, shorts tab, and live tab of a channel to play all of the respective content based on the sorting selected. It also left justifies the content of the select container.

### Before
Grimshot 2026-06-28 13-15-48
Grimshot 2026-06-28 13-15-52

### After

Grimshot 2026-06-28 13-12-09
Grimshot 2026-06-28 13-12-13
Grimshot 2026-06-28 13-12-20
Grimshot 2026-06-28 13-12-31
Grimshot 2026-06-28 13-12-40
Grimshot 2026-06-28 13-12-46
Grimshot 2026-06-28 13-12-56
Grimshot 2026-06-28 13-13-02

## Testing

- Requires #9325
- Go to any channel
- Home tab should have a play all button that opens a playlist with all channel content ordered by newest
- Videos, shorts, and live tabs should have a button that works similarly with their respective content and sorting
- The videos, shorts, and live tabs play all button should be hidden when not sorting by newest or most popular as youtube only has playlists for sorting by newest or most popular

## Desktop

- **OS:** Arch Linux
- **OS Version:** Kernel 7.0.10-zen1-1-zen
- **FreeTube version:** 0.24.1

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -775,6 +775,7 @@ Channel:
   This channel does not allow searching: This channel does not allow searching
   This channel is age-restricted and currently cannot be viewed in FreeTube.: This channel is age-restricted and currently cannot be viewed in FreeTube.
   Channel Tabs: Channel Tabs
+  Play All: Play All
   Videos:
     Videos: Videos
     This channel does not currently have any videos: This channel does not currently
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -775,7 +775,7 @@ Channel:
   This channel does not allow searching: This channel does not allow searching
   This channel is age-restricted and currently cannot be viewed in FreeTube.: This channel is age-restricted and currently cannot be viewed in FreeTube.
   Channel Tabs: Channel Tabs
-  Play All: Play All
+  View All: View All
   Videos:
     Videos: Videos
     This channel does not currently have any videos: This channel does not currently
--- a/src/renderer/views/Channel/Channel.css
+++ b/src/renderer/views/Channel/Channel.css
@@ -15,6 +15,8 @@
   display: flex;
   flex-direction: row;
   justify-content: flex-end;
+  align-items: baseline;
+  gap: 10px;
   margin-block-end: 10px;
 }
 
--- a/src/renderer/views/Channel/Channel.vue
+++ b/src/renderer/views/Channel/Channel.vue
@@ -40,10 +40,10 @@
       />
       <div class="select-container">
         <FtButton
-          v-if="showPlayAllButton"
+          v-if="showViewAllButton"
           style="margin-top: 33px;"
-          :label="$t('Channel.Play All')"
-          @click="router.push(currentTabPlayAllRoute)"
+          :label="$t('Channel.View All')"
+          @click="router.push(currentTabViewAllRoute)"
         />
         <FtSelect
           v-if="showVideoSortBy"
@@ -517,7 +517,7 @@ const tabInfoValues = computed(() => {
   return values
 })
 
-const showPlayAllButton = computed(() => {
+const showViewAllButton = computed(() => {
   switch (currentTab.value) {
     case 'videos': return (videoSortBy.value === 'newest' || videoSortBy.value === 'popular') && (showFetchMoreButton.value || filteredVideos.value.length > 1)
     case 'shorts': return (shortSortBy.value === 'newest' || shortSortBy.value === 'popular') && (showFetchMoreButton.value || filteredShorts.value.length > 1)
@@ -526,7 +526,7 @@ const showPlayAllButton = computed(() => {
   }
 })
 
-const currentTabPlayAllRoute = computed(() => {
+const currentTabViewAllRoute = computed(() => {
   switch (currentTab.value) {
     case 'videos': return `/playlist/${getChannelPlaylistId(id.value, 'videos', videoSortBy.value)}`
     case 'shorts': return `/playlist/${getChannelPlaylistId(id.value, 'shorts', shortSortBy.value)}`
--- a/src/renderer/views/Channel/Channel.vue
+++ b/src/renderer/views/Channel/Channel.vue
@@ -39,6 +39,11 @@
         :related-channels="relatedChannels"
       />
       <div class="select-container">
+        <FtButton
+          v-if="showPlayAllButton"
+          :label="$t('Channel.Play All')"
+          @click="router.push(currentTabPlayAllRoute)"
+        />
         <FtSelect
           v-if="showVideoSortBy"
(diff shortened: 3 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
#9325 is merged, please rebase into dev branch

r3dArch:
@efb4f5ff-1298-471a-8973-3d47447115dc I think my changes are good, but something seems to be broken. Even using the main development branch the sort selection drop down doesn't show up. I'm going to look into it now.

r3dArch:
I forgot to pull the latest changes from the main repo. Everything seems to be good now.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Please revert the placement of the Sort By dropdown.
Also put the Play All button the the left of the dropdown

r3dArch:
How is this?
Grimshot 2026-07-02 17-56-33
Grimshot 2026-07-02 17-56-38

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
The placement on the Home tab feels a bit weird to me. It being to the right is fine but its too high. I think aligning the height of the button to match the height of the header of the Videos section would be better.

Before you make this change. Could you please answer the following question (it could be the change above isnt needed depending on the answer). What is the reason you placed the Play All button on the Home page in the first place? The Videos tab Play All button brings me to the same playlist so it feels a bit unnecessary but i could be missing something here.

r3dArch:
The play all button on the home tab should play all of a channels content; so if they have videos, shorts, and live content it should all be mixed together. I see what you mean about the look, but I don't quite understand your suggested change. Are you saying to have the button at the same level as the videos text?

efb4f5ff-1298-471a-8973-3d47447115dc:
> The play all button on the home tab should play all of a channels content; so if they have videos, shorts, and live content it should all be mixed together

You are right. I somehow didnt notice it the first time and thought it was the same.

> Are you saying to have the button at the same level as the videos text?

Correct

r3dArch:
To make this change I've been looking at the channel home more closely and it seems that each shelf usually has a playlist button which would link to the respective content, so it might be confusing if it was positioned there. What do you think about putting it in the channel details since it plays all of a channels content?
Grimshot 2026-07-02 22-24-04

absidue:
Most people will want just the videos, shorts or live streams, not everything mixed together, so I would get rid of the play all uploads button entirely.

Additionally instead of having a separate button for each tab, I would simplify it to one button in the template, with a `v-if` (we only use `v-show` on the `FtSelect` components so that they maintain their state correctly when the user switches tabs) that points to a computed that decides whether that one button should be shown or not (e.g. with a switch statement) and then instead of defining the click handler inline it could be single function in the script section which uses the current tab and sort by information to send it to the correct URL. That makes the template easier to read and avoids Vue having to re-calculate and re-render the button every time the tab is changed (it will only need to make changes when the button gets shown or hidden).

r3dArch:
@absidue I made the requested changes, please let me know if you want it done any differently. Thank you for the explanation of `v-if` and `v-show`; I had not worked with Vue before this.

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
Im not seeing the button appear if Oldest sorting is selected

https://github.com/user-attachments/assets/ed33e4fb-e3ec-46c1-9143-fa32c8f0c113


Also question why does the Popular playlist only contain 200 videos?

r3dArch:
Youtube doesn't have a playlist that sorts by oldest. I would guess that Youtube's popular playlist only shows the first 200 videos. This could be fixed by creating a local playlist and sorting it locally, but right now these buttons link to playlists that Youtube automatically generates for every channel.

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Not sure if related to this PR or not but Short playlists only show 100 videos regardless of sorting selection

https://github.com/user-attachments/assets/1111f7c7-4a09-49f5-83cb-81c9e7fb26d2

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Should we maybe move the placement of the Play All button when there is no dropdown? It feels like it sticks too much to the sides now. Maybe same placement height as dropdown?

https://github.com/user-attachments/assets/0abb6f42-ffee-4320-9078-6863e1fdfd48

r3dArch:
Here are some screenshots of what that looks like.
Grimshot 2026-07-07 18-29-23
Grimshot 2026-07-07 18-29-36
I feel like it doesn't look much better, but I don't have any ideas to make it look good.

efb4f5ff-1298-471a-8973-3d47447115dc:
Need thoughts from others so ill wait for them to chime in

PikachuEXE:
Consistent spacing looks better to me

r3dArch:
I added a top margin to the button to keep the spacing consistent. If there is a more preferred way to do it, please let me know.

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

static/locales/en-US.yaml
```diff
@@ -775,6 +775,7 @@ Channel:
   This channel does not allow searching: This channel does not allow searching
   This channel is age-restricted and currently cannot be viewed in FreeTube.: This channel is age-restricted and currently cannot be viewed in FreeTube.
   Channel Tabs: Channel Tabs
+  Play All: Play All
```
PikachuEXE: Just wondering if we should label the button as `Play All` or `View All` coz when I press it I view the playlist not starts playing it
Not blocking

r3dArch: Maybe `Open Playlist`? I think the context would imply that it opens a playlist of the current channel tab.

PikachuEXE: Problem is when user (or just me) sees `Open Playlist` on a tab he would be confused first on why "playlist" is relevant here
View all means view all entries (videos/lives/shorts) and whether it's in playlist form or what is something else but the desired effect is "view all" not "open playlist" (that's implementation)

r3dArch: Should be good now. Thank you for the feedback.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

Issue #7262, opened 2025-04-24
[Feature Request]: `Fetch feed automatically for each profile` setting


Victor239:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.

### Problem Description

This is a follow-up on [this bug report](https://github.com/FreeTubeApp/FreeTube/issues/6089) as it's apparently intended behaviour.

The current setting `Fetch feed automatically` only automatically fetches the default profile. There is currently no way to make other profiles be automatically fetched when switched to. 

### Proposed Solution

This new setting is so that when a user switches to a different profile then it'll automatically fetch. This should probably only occur once for each profile in case a user is switching back and forth. If FreeTube is fully closed and then reopened however then this should reset and automatically fetch each profile again when switched to.

### Alternatives Considered

N/A

### Issue Labels

ease of use improvement, improvement to existing feature, new optional setting

### Additional Information

N/A

valhalla-creator:
hi

PR #9360, opened 2026-07-02
`Fetch Feed Automatically` will now auto-fetch each profile's feed once per session, not just the first profile viewed

### LINKED ISSUE #6089: [Bug]: Switching profiles does not refresh feed
### Guidelines

### Describe the bug

In Subscription settings I have enabled `Fetch feed automatically`, which says it will automatically fetch when switching profile and when opening a new window. In the current version neither of those seem to work. Opening a new window will however refresh my one `default profile`, as my `Default landing page` is `Subscriptions`.

In the previous version changing profile would refresh that profile, but this would only happen once per profile. This worked well for me because I swap between profiles constantly, I don't need a refresh on every single switch. 

Closing the application and reopening it used to refresh the default profile again and would allow the one-time refresh of each profile again, so I did this every few hours to get a refresh of all profiles with minimal clicks. 

As you can see the behaviour of this setting was quite different from the description in my experience, though I have an idea of a better overall behaviour. 

### Expected Behavior

Switching profiles or opening a new window should refresh fetch the feed again if `Fetch feed automatically` is enabled, but it should also avoid fetching the feed if it's been fetched recently.

This should be a separate setting for how long to wait until fetching feed, as a default set it to 60 minutes. 

### Issue Labels

feature stopped working

### FreeTube Version

v0.22.0 Beta 

### Operating System Version

Kubuntu 24.10

### Installation Method

Flathub

### Primary API used

Local API

v0.21.0 Beta 

### Additional Information

_No response_

### Nightly Build

VictorVow:
# Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #6089, #7262

## Description

The `Fetch Feed Automatically` setting is described as fetching the subscription feed "on startup and when a new window is opened". Since #5743 (first shipped in v0.22.0, which matches the "Last Known Working Version: v0.21.0" in the linked issue), the flag tracking whether the first automatic fetch has already run is a single boolean per window. The consequence: only the profile you view first ever gets an automatic fetch. When you switch to another profile, the flag is already consumed, so that profile silently shows whatever is in the persistent disk cache (which can be days old) and its feed is never automatically fetched at all during that session.

The issue was closed with the reasoning that showing cached results is intended, but I believe that misses what's actually going wrong. This is not about wanting a re-fetch on every profile switch - the problem is that for every profile except the first one, the automatic fetch promised by the setting never happens, not even once. If a profile's feed is never fetched automatically, then `Fetch Feed Automatically` simply isn't doing what its name and tooltip promise for that profile, and it's a regression from v0.21.0 where each profile got one automatic fetch per session.

This PR scopes the "first auto-fetch has run" tracking per profile (a `Set` of profile ids per tab type, instead of a boolean). Each profile now receives exactly one automatic remote fetch per session, the first time its feed is viewed. It deliberately does **not** re-fetch on every switch: returning to an already-fetched profile still shows the session cache, so the caching behaviour is preserved and there are no additional network requests compared to v0.21.0.

The one-time-fetch check was also moved above the cache-presence check in `loadVideosFromCacheSometimes` / `loadPostsFromCacheSometimes`, mirroring what #5743 did for the new-window path - otherwise the persistent disk cache would again take priority over the first automatic fetch.

## Screenshots

N/A - no visual changes.

## Testing

1. Create two profiles, A and B, with overlapping subscriptions: A subscribed to channels 1 and 2, B subscribed to channel 1 only.
2. Enable Settings > Subscription Settings > `Fetch Feed Automatically`.
3. Restart FreeTube with profile A active and open the Subscriptions page: channels 1 and 2 are fetched from remote as before.
4. Switch to profile B:

   - Without this PR: the page shows the stale feed from the disk cache and no fetch happens; B's feed is never fetched automatically for the rest of the session.

   - With this PR: B's feed is fetched once, and channel 1 is served from the session cache populated by profile A's fetch rather than being re-fetched from remote.
5. Switch back to profile A, then to profile B again: both load instantly from the session cache - no re-fetching on subsequent switches, and no re-fetching of channel 1 despite the overlap.
6. Repeat on the Shorts, Live and Posts tabs; each tab tracks its first fetch per profile independently, and each avoids re-fetching channels already cached by another profile.
7. As a control, create a third profile C with no overlap with A or B (e.g. channel 3) and switch to it: confirm channel 3 is fetched from remote normally, same as before this PR.

### DIFF
--- a/src/renderer/components/SubscriptionsLive.vue
+++ b/src/renderer/components/SubscriptionsLive.vue
@@ -59,6 +59,8 @@ const fetchSubscriptionsAutomatically = computed(() => store.getters.getFetchSub
 
 const activeSubscriptionList = computed(() => store.getters.getActiveProfile.subscriptions)
 
+const activeProfileId = computed(() => store.getters.getActiveProfile._id)
+
 const cacheEntriesForAllActiveProfileChannels = computed(() => {
   const liveCache = store.getters.getLiveCache
   const entries = []
@@ -140,22 +142,25 @@ function loadVideosFromRemoteFirstPerWindowSometimes() {
 
   alreadyLoadedRemotely = true
   loadVideosForSubscriptionsFromRemote()
-  store.commit('setSubscriptionForLiveStreamsFirstAutoFetchRun')
+  store.commit('setSubscriptionForLiveStreamsFirstAutoFetchRun', activeProfileId.value)
 }
 
 function loadVideosFromCacheSometimes() {
   // Can only load reliably when cache ready
   if (!subscriptionCacheReady.value) { return }
 
-  // This method is called on view visible
-  if (videoCacheForAllActiveProfileChannelsPresent.value) {
-    loadVideosFromCacheForAllActiveProfileChannels()
+  // Check if this profile needs to be auto-fetched for the first time
+  if (fetchSubscriptionsAutomatically.value && !store.getters.getSubscriptionForLiveStreamsFirstAutoFetchRun) {
+    // `isLoading.value = false` is called inside `loadVideosForSubscriptionsFromRemote` when needed
+    alreadyLoadedRemotely = true
+    loadVideosForSubscriptionsFromRemote()
+    store.commit('setSubscriptionForLiveStreamsFirstAutoFetchRun', activeProfileId.value)
     return
   }
 
-  if (fetchSubscriptionsAutomatically.value) {
-    // `isLoading.value = false` is called inside `loadVideosForSubscriptionsFromRemote` when needed
-    loadVideosForSubscriptionsFromRemote()
+  // This method is called on view visible
+  if (videoCacheForAllActiveProfileChannelsPresent.value) {
+    loadVideosFromCacheForAllActiveProfileChannels()
     return
   }
 
--- a/src/renderer/components/SubscriptionsLive.vue
+++ b/src/renderer/components/SubscriptionsLive.vue
@@ -151,10 +151,20 @@ function loadVideosFromCacheSometimes() {
 
   // Check if this profile needs to be auto-fetched for the first time
   if (fetchSubscriptionsAutomatically.value && !store.getters.getSubscriptionForLiveStreamsFirstAutoFetchRun) {
-    // `isLoading.value = false` is called inside `loadVideosForSubscriptionsFromRemote` when needed
     alreadyLoadedRemotely = true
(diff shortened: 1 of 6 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
- doesnt this close your own issue #7262?
- Also should the tooltip description be changed to clarify behavior?

PikachuEXE:
I would not consider this as a "fix" as the tooltip clearly indicates that it only does that on startup and new window (or just new window coz one is created on start up)
image

I am not rejecting this PR but its type is a feature change (improvement) and we generally review/test feature changes longer than fixes (well the small ones)

I think the Testing section needs to include all sorts of profile overlap types instead of the current one (A & B got some overlap only) first so we can test the changes

e.g. When profile A includes B but not reverse (A: ch 1,2 + B: ch 1) 
Switching profile would re-fetch ch 1 and that doesn't seem right to me (worse when some profiles got a lot of overlaps)

VictorVow:
Good questions.

- This does close #7262 - it's got a slightly larger scope the solution I theorised there but this PR has the minimal scope needed to resolve the core issue.
- Agreed on the tooltip - it currently doesn't mention profile behavior at all, which is part of why this reads as a feature change rather than a restoration. Pushed a commit updating it to say each profile gets one automatic fetch the first time it's viewed in a session.

@PikachuEXE 
Fair enough, happy to have this treated as a feature change rather than a straight fix given it does more than the existing tooltip promises.

You're right about the overlap case, and I confirmed it can occur: `loadXForSubscriptionsFromRemote` always fetches every channel in the active profile, with no check against the already-populated session cache. So switching A to B when B's channels are a subset of A's does re-fetch channels that were just fetched.

I've pushed a follow-up commit that filters the auto-fetch-on-switch path down to only channels missing from the session cache (manual refresh still force-refetches everything, since that's explicit user intent). I'll also expand the Testing section with the overlap scenario you described so it's easy to verify.

PikachuEXE:
Testing needs cases for 4 profiles:
A: Default, 2-3 channels maybe
B: 1-2 more channels than A (includes all channels from A)
C: 1-2 Less channels than A
D: No overlap at all

Expected:
Switching from A-D in order:
A & D got no. of channels requests, B = diff number, C = 0

What I got:
A = expected, B-D = 0

Seems still need fixing

PR #9374, opened 2026-07-05
Handle compact published dates e.g. 1w ago in LockupViews

absidue:
## Pull Request Type

- [x] Bugfix

## Description

YouTube is currently A/B testing compact published dates in the up next section e.g. "1 week ago" -> "1w ago". The fix was just making the space in the regex optional.

While I was at it I also removed the `isViewOrWaitingCountText` function which was only used in one place, as that checked for "watching" and then called `isViewCountText` which checked for `watching` and various other things, in favour of just calling `isViewCountText` directly and added `b` for billions in the `VIEWS_IN_NUMBER_ONLY` regex. Those changes seemed minor enough to not be worth sticking in their own pull request.

## Testing

If you are unable to get it on your side after lots of tries, try setting your VPN to France or the Netherlands as it seems to appear more often in those two locations.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/helpers/api/local.js
+++ b/src/renderer/helpers/api/local.js
@@ -1533,8 +1533,7 @@ export function parseLocalListVideo(item, channelId, channelName) {
 }
 
 const VIEWS_OR_WATCHING_REGEX = /views?|watching|waiting/i
-const WAITING_REGEX = /waiting/i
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Image

ghost (approved):

PR #9379, opened 2026-07-06
Use batched fetching instead of forcing RSS with sub count > N

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Taken from https://github.com/FreeTubeApp/FreeTube/pull/9157 without the new setting

closes #9157

## Description

Currently FT forces user to use RSS when subscription count within the current profile is > 125
This PR makes FT just fetch videos, live, community posts in batches of 80 regardless (sub count <= 80 = same as before)
No new setting at all, `Posts` tab can be shown now as sub count shouldn't matter

No idea what batch size/timeout should be used, all taken from https://github.com/FreeTubeApp/FreeTube/pull/9157

## Screenshots 

Only shows video one, test the rest yourself

https://github.com/user-attachments/assets/9cfe3658-8c4d-42a7-9434-4c8f60382520

## Testing

Preparation: get 125+ subscriptions
- Refresh sub in video tab
- Refresh sub in live tab
- Refresh sub in posts tab (the tab should be visible

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/SubscriptionsLive.vue
+++ b/src/renderer/components/SubscriptionsLive.vue
@@ -185,14 +185,7 @@ async function loadVideosForSubscriptionsFromRemote() {
   let channelCount = 0
   isLoading.value = true
 
-  let useRss = useRssFeeds.value
-  if (channelsToLoadFromRemote.length >= 125 && !useRss) {
-    showToast(
-      t('Subscriptions["This profile has a large number of subscriptions. Forcing RSS to avoid rate limiting"]'),
-      10000
-    )
-    useRss = true
-  }
+  const useRss = useRssFeeds.value
 
   store.commit('setShowProgressBar', true)
   store.commit('setProgressBarPercentage', 0)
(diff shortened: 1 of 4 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc:
should this be modified to not include posts anymore?
https://github.com/FreeTubeApp/FreeTube/blob/02cf8fbdfcee404866ffe13ef2e07b393d412eee/static/locales/en-US.yaml#L1086-L1088

PikachuEXE:
When `Fetch Feeds from RSS` enabled posts still can't be used (and tab hidden

efb4f5ff-1298-471a-8973-3d47447115dc:
Works good. Question before approval (because other PR author isnt responsive), 

* i know the 80 number comes from the other PR but could you maybe try to explain why you think this number is picked?
* How are we sure that this wont trigger the 429 errors?

PikachuEXE:
I guess 80 comes from ~2/3 of 125 limit (and where does 125 comes from? dunno
And I am not sure if this works great (nob 429) since we can never be sure about how YT setup their rate limiting (not to mention some people use VPN which would might make it worse but I assume no VPN case first)
Since there is a fallback to use RSS (even it requires user setting it manually) I consider this safe-ish but adding a setting for batch interval might be needed if users start getting 429 (but doesn't have to be now)

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PikachuEXE:
@ChunkyProgrammer ~

ghost (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE could you create a PR to reflect current changes in docs

Issue #3541, opened 2023-05-14
[Bug]: Duplicates when click spam loading elements in FT


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [X] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the bug report I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [X] This issue contains only one bug.

### Describe the bug

Only thing u need to do to reproduce is clicking very fast on the element

Following videos are recorded on all the channel tabs of the channel page

https://github.com/FreeTubeApp/FreeTube/assets/73130443/a3b350b7-f2cb-4b44-9d3f-c45bef1d6017

https://github.com/FreeTubeApp/FreeTube/assets/73130443/3f7686d0-eff6-4f62-ab81-a24dff0cf485

https://github.com/FreeTubeApp/FreeTube/assets/73130443/5adc3fc2-c03c-40db-ac46-b359f0cb61bf

https://github.com/FreeTubeApp/FreeTube/assets/73130443/8e2ea4f4-2b87-4c89-94d5-fe59f7ef361e

Following 2 videos are recorded in the comment section

https://github.com/FreeTubeApp/FreeTube/assets/73130443/495263e5-0475-49e1-97cb-9879b72fcbb3

https://github.com/FreeTubeApp/FreeTube/assets/73130443/04f451d1-b99a-4738-918e-166da426a27c

Following video is recorded on the search page

https://github.com/FreeTubeApp/FreeTube/assets/73130443/c15601ba-0aa8-421c-b1b6-a72f73b994f2

### Expected Behavior

1. No duplicate videos on all the channel tabs when i click very fast on the load more button
2. No duplicate comments when i click very fast on the load more comment/replies button
3. No duplicate videos when i click very fast on the load more button on the search page

### Issue Labels

inconsistent behavior, visual bug

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/a4d45b5fa8a8a1726808c9eebe307643f8dde9c9

### Operating System Version

[...]

PR #9383, opened 2026-07-07
Fix: Duplicates when click spam loading elements in FT (1/3)

Shadorc:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

partially addresses https://github.com/FreeTubeApp/FreeTube/issues/3541 (channel tabs)

## Description

There are no safeguards to prevent multiple queries from being sent simultaneously with the same data, which causes duplicate results to be displayed.

Most components already have `isLoading` bools, but they have side effects on page content. They're designed to detect initial loading rather than any loading state, which is why I opted for a separate variable.

I also chose to hide the button during loading, but we could alternatively disable it with CSS or display a loading icon. It won't make much difference here since the request usually completes too quickly for it to appear, but that won't be the case for the other two fixes.

I considered grouping all three fixes in one MR, but some are more complex and would probably need a separate MR for easier review.

## Screenshots 

The following video is recorded with a 2s delay applied to each request.
https://github.com/user-attachments/assets/8fe0b174-2cca-402a-90b4-d63a439d821d

## Testing

Testing can be simplified by simulating a slow connection. To do this, simply add `await new Promise((resolve) => setTimeout(resolve, 2000))` to the `Channel#handleFetchMore` method after `isFetchMoreLoading.value = true`

1. Open any channel with enough videos/shorts/playlists to display "Fetch more results"
2. Click very fast on the "Fetch more results" button in any of the tab
3. Check that the button disappears when the results are loading 
4. Check that no duplicated videos/shorts/playlists are shown once loaded

## Desktop

- **OS:** Bazzite
- **OS Version:**
- **FreeTube version:** 0.24.1-beta

### DIFF
--- a/src/renderer/views/Channel/Channel.vue
+++ b/src/renderer/views/Channel/Channel.vue
@@ -230,7 +230,7 @@
           </p>
         </FtFlexBox>
         <FtAutoLoadNextPageWrapper
-          v-if="showFetchMoreButton"
+          v-if="showFetchMoreButton && !isFetchMoreLoading"
           @load-next-page="handleFetchMore"
         >
           <div
@@ -338,6 +338,7 @@ let mayContainContentFromOtherChannels = false
 const isLoading = ref(true)
 const isElementListLoading = ref(false)
 const isSearchTabLoading = ref(false)
+const isFetchMoreLoading = ref(false)
 const currentTab = ref('videos')
 
 const isCurrentTabLoading = computed(() => {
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):
LGTM

Shadorc:
I could wrap the `isFetchMoreLoading.value = false` inside a `try / finally` block to make sure the value does not get locked to true in case of exceptions, but all the calls inbetween are already wrapped with `try / catch`

PikachuEXE (approved):

absidue (approved):

PR #9395, opened 2026-07-10
Fix Disable Channel Links in Channel Sidenav Menu

ozrendev:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

relates to #8502

## Description

These links were missed in the original PR, so this disables them as originally intended

## Testing

Go to Settings -> Parental Control -> Enabled "Disable Channel Link on Videos"
Go to Channels sidenav item, see links are disabled for Channel thumbnail and name

### DIFF
--- a/src/renderer/components/SideNav/SideNav.css
+++ b/src/renderer/components/SideNav/SideNav.css
@@ -152,6 +152,10 @@
   block-size: 1.3em;
 }
 
+.disabledIcon {
+  pointer-events: none;
+}
+
 @media only screen and (width > 680px) {
   ::-webkit-scrollbar {
     inline-size: 10px;
--- a/src/renderer/components/SideNav/SideNav.vue
+++ b/src/renderer/components/SideNav/SideNav.vue
@@ -186,7 +186,8 @@
         v-if="!hideActiveSubscriptions"
         class="mobileHidden"
       >
-        <router-link
+        <component
+          :is="enableChannelLinks ? 'router-link' : 'span'"
           v-for="channel in activeSubscriptions"
           :key="channel.id"
           :to="`/channel/${channel.id}`"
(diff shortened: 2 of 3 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE:
Should we also disable the icons on side bar? (PR disable channel links in Channel List
image

ozrendev:
> Should we also disable the icons on side bar? (PR disable channel links in Channel List 

Yes, good catch! Added that now

efb4f5ff-1298-471a-8973-3d47447115dc:
Umm i dont know if thats really necessary as there is already a setting to hide them. Making them non functional feels not logical to me but ofcourse my feeling could be wrong here

PikachuEXE:
Well my reason would be the existing setting is under "distraction free" not parental control
And we should provide less surprise for users ("what? I need 2 settings to do what I want?
Now whether those sidebar icons should be displayed at all when the new setting is enabled is something else (I have no idea

efb4f5ff-1298-471a-8973-3d47447115dc:
I guess you're right. I didnt put myself in users shoes enough

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
should the hover and click animation be disabled too here?

https://github.com/user-attachments/assets/33ca300e-47ff-488a-ba59-ab8b52c7f873

PikachuEXE:
> should the hover and click animation be disabled too here?

ya

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

Issue #9428, opened 2026-07-17
[Bug]: "Freetube is damaged and can't be opened. You should move it to the trash" (arm64, MacOS)


1fichier-dot-info:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads). **The only official downloads are from the [https://freetubeapp.io](https://freetubeapp.io/#download) website, the [FreeTubeApp/FreeTube GitHub repository](https://github.com/FreeTubeApp/FreeTube/releases) and the [io.freetubeapp.FreeTube flatpak](https://flathub.org/en/apps/io.freetubeapp.FreeTube) from Flathub.**
- [x] I have [searched the issue tracker for **open and closed** issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

1. Go to: https://freetubeapp.io/#download
2. Download .zip (arm64)
3. Upon trying to open the file, get the error message "Freetube is damaged and can't be opened. You should move it to the trash."

### Expected Behavior

To be able to open the file and proceed to the installation of Freetube.

### Issue Labels

usability issue

### FreeTube Version

0.25.0 Beta

### Operating System Version

macOS Tahoe 26.5.2

### Installation Method

.zip / .7z

### Primary API used

Local API

### Additional Information

The .dmg and the .7z installations have the same issue. The .x64 version for MacOS does not seem to be affected by this.

### Nightly Build

- [ ] I have encountered this bug in the latest **official** [nightly build](https://docs.freetubeapp.io/development/nightly-builds), do not check this box if you are using an unofficial or custom build.

Shadorc:
Hello!
You can find a section in the FAQ addressing this issue: https://docs.freetubeapp.io/faq/#macos-freetube-is-damaged-and-cant-be-opened-you-should-move-it-to-the-trash

1fichier-dot-info:
[...]

PR #9444, opened 2026-07-17
Explicitly configure ad-hoc codesigning after electron-builder changes

absidue:
## Pull Request Type

- [x] Bugfix

## Related issue

- #9428
- Comments on #9425

## Description

For a while electron-builder was ad-hoc signing by default when no code signing identity was provided but in version [26.15.0](https://github.com/electron-userland/electron-builder/releases/tag/electron-builder%4026.15.0) (PR: https://github.com/electron-userland/electron-builder/pull/9822) they stopped doing that and made it an opt-in thing. Without the ad-hoc codesigning macOS refuses to launch saying that the app is broken, with the ad-hoc code signing it still refuses to launch by default but it only complains about not being able to verify the signature which is a lot less alarming to users and easier to bypass (checking boxes in settings GUIs instead of having to remove the quarantine attribute with a command).

The difference in the dialogs that macOS shows can be seen here: https://github.com/electron-userland/electron-builder/pull/9007

## Testing

Test build here: https://github.com/absidue/FreeTube/actions/runs/29598510452

As I don't have a macOS machine I haven't been able to test it myself but looking at the code changes on the electron-builder side, this should be enough to get us back to how it was before.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/_scripts/ebuilder.config.mjs
+++ b/_scripts/ebuilder.config.mjs
@@ -96,7 +96,12 @@ export default {
       NSBluetoothPeripheralUsageDescription: undefined,
       NSCameraUsageDescription: undefined,
       NSMicrophoneUsageDescription: undefined,
-    }
+    },
+
+    // Enable ad-hoc signing
+    // If we skip signing entirely, macOS says that the application is damaged, which makes users open bug reports.
+    // With an ad-hoc signature it still refuses to launch by default but with the reason that it cannot verify the signature
+    identity: '-'
   },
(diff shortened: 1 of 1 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):
Tested with macOS 15.7.7

ghost (approved):

PR #9464, opened 2026-07-19
Change: Add information about current/default invidious instance

Shadorc:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [X] Other

## Description

I've been confused about the difference between the current Invidious instance and the default one for a while.
I'm proposing to clarify this distinction in the tooltip.

I may be the only one confused about this, and it may not be worth adding it. In this case, feel free to close this MR!

## Screenshots 

image

### DIFF
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -1059,7 +1059,7 @@ Tooltips:
     Thumbnail Preference: All thumbnails throughout FreeTube will be replaced with
       a frame of the video, blurred or hidden instead of the default thumbnail.
     Invidious Instance: The Invidious instance that FreeTube will connect to for API
-      calls.
+      calls. This setting applies to the current window only, other windows use the default instance.
     Region for Trending: The region of trends allows you to pick which country's trending
       videos you want to have displayed.
     External Link Handling: |
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -1059,7 +1059,7 @@ Tooltips:
     Thumbnail Preference: All thumbnails throughout FreeTube will be replaced with
       a frame of the video, blurred or hidden instead of the default thumbnail.
     Invidious Instance: The Invidious instance that FreeTube will connect to for API
-      calls. Changes to this setting apply to the current window only, unless set as default.
+      calls. Changes to this setting apply only to the current window unless you set this instance as the default.
     Region for Trending: The region of trends allows you to pick which country's trending
       videos you want to have displayed.
     External Link Handling: |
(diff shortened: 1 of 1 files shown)


PikachuEXE:
That's not the logic
You can search for `InvidiousInstance.value === ''` to see
If there is no default instance set, a random one is set from a list, otherwise do nothing (use whatever set)
A new window might make it set a new one (not sure about this one has not test yet)
But if one window set a default instance all window should use the same one
Let me know if you got different test result coz I rarely use and test IV

Shadorc:
According to absidue, this works like this:
- Nothing set: pick a random instance when the window opens (made a lot more sense in the past when there was a large list of public Invidious instances with their APIs enabled)
- Current instance set: uses that instance in that window, not used in any other window and not saved to the file system
- Default instance set: Always use that instance in all windows by default, saved to the disk, can still be temporarily overridden on a per window basis by setting the current instance

So, if I understand this correctly, `Current instance` has always the priority for the current window, otherwise it's the default one, and if none is set, a random url has been chosen. 

It could be rephrased like "other windows use the default instance _, or their current one if set_" but I wanted something consice

Edit: By "This setting", I mean the current Invidious instance field

PikachuEXE:
Current instance can be set by FT (random when no default set, otherwise from default), set by user (then limited to current window)
So `This setting applies to the current window only` reads a bit off to me is due to it incorrectly covers the (FT set from default without user changing it case)
I am guessing a better message might be "if user changes it, change limited to current window only, unless they also press `Set Current as Default`"

You can see there is text below the input field which changes according to default instance
Maybe that can be updated instead?
image
image

Shadorc:
Ok, I agree
What about this (it's what you're proposing, a little bit more consice): "Changes to this setting apply to the current window only, unless set as default."?

PikachuEXE:
Just show an extra line when current != default?
image

image
image

efb4f5ff-1298-471a-8973-3d47447115dc:
I think that ill prefer the tooltip for the sake of being consistent on how we distribute info to our users

Shadorc:
I think I prefer the tooltip, too. Otherwise, it feels like there's too much text displayed. And, as absidue said, it seems that it's not really confusing for most of the users, so it probably doesn't need to stay visible.

Edit: I've modified my MR and the screenshot with my last proposal, but feel free to suggest something better

PikachuEXE (dismissed):

static/locales/en-US.yaml
```diff
@@ -1059,7 +1059,7 @@ Tooltips:
     Thumbnail Preference: All thumbnails throughout FreeTube will be replaced with
       a frame of the video, blurred or hidden instead of the default thumbnail.
     Invidious Instance: The Invidious instance that FreeTube will connect to for API
-      calls.
+      calls. Changes to this setting apply to the current window only, unless set as default.
```
efb4f5ff-1298-471a-8973-3d47447115dc: My brain felt like its worded a bit weird so made a little tweak. please lmk if this isnt desired 

```suggestion
      calls. Changes to this setting apply only to the current window unless you set this instance as the default.
```

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

absidue (approved):

PR #9460, opened 2026-07-19
Update electron-builder to fix Windows installers

absidue:
## Pull Request Type

- [x] Bugfix

## Related issue

- closes #9427

## Description

As per the linked issue the Windows installers are installing all files except the FreeTube.exe file on ARM64 machines. This pull request updates electron-builder to the latest version as the changelog for version 26.15.6 contains an entry that says `fix(nsis): reliably install the main executable and native binaries on x64 and arm64` which sounds like the exact issue that the users reported. The reason why dependabot didn't do the update is because 26.15.3 is currently marked as the latest on npmjs, as the electron-builder maintainers have a policy of only changing the latest tag on npmjs when it surpasses a certain number of downloads, which is a good policy but in this case I feel like fixing our broken Windows installers is more important.

To hopefully keep conflicts to a minimum I performed the following steps to update it:
1. Change the version in the `package.json` file.
2. Run `pnpm install`
3. Manually edit the lockfile to remove all unrelated changes.
4. Run `pnpm install --frozen-lockfile` to ensure that pnpm would accept my manual changes.

electron-builder changelogs:
- https://github.com/electron-userland/electron-builder/releases/tag/electron-builder%4026.15.4
- https://github.com/electron-userland/electron-builder/releases/tag/electron-builder%4026.15.5
- https://github.com/electron-userland/electron-builder/releases/tag/electron-builder%4026.15.6
- https://github.com/electron-userland/electron-builder/releases/tag/electron-builder%4026.15.7

## Testing

I haven't actually been able to test the new installers myself as I don't have a Windows ARM64 laptop, but the build didn't fail and I am trusting the electron-builder changelog on this one.

https://github.com/absidue/FreeTube/actions/runs/29689739729

## Desktop

- **OS:** Windows
- **OS Version:** 11

PR #9466, opened 2026-07-20
Update release workflow to use GH CLI instead of deprecated action

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [x] Other

## Related issue

This PR includes https://github.com/FreeTubeApp/FreeTube/pull/9460 so that assets can be tested properly
That PR should be merged first

## Description

Spotted deprecation message in last run - https://github.com/FreeTubeApp/FreeTube/actions/runs/29556083929
Found `gh release upload` - https://github.com/marketplace/actions/publish-release (bottom, example included)
Doc for command & env var:
- https://cli.github.com/manual/gh_release_create
- https://cli.github.com/manual/gh_help_environment
No rename/upload with different name option yet so rename using powershell needed (found in https://stackoverflow.com/questions/70515210/can-i-use-a-github-action-to-rename-a-file

## Screenshots 

Build log screenshot (only the changed part
image

## Testing

Download assets (all versions, archive and installer) for each OS to see if app can be extracted/installed and run
https://github.com/PikachuEXE/FreeTube/releases/tag/v0.25.0-beta-release-test-202607201605
Assets uploaded by https://github.com/PikachuEXE/FreeTube/actions/runs/29726731446

In addition you might want to test https://github.com/FreeTubeApp/FreeTube/pull/9444 & https://github.com/FreeTubeApp/FreeTube/pull/9460

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

## Additional context

Already tested macos x64 + arm64 on m1 mac w/ macOS 15.7.7

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,10 +7,10 @@ name: Upload Release
 on:
   workflow_dispatch:
     inputs:
-      releaseId:
+      tag:
         type: string
         required: true
-        description: Release ID
+        description: Tag
 
 permissions: {}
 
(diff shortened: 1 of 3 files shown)


absidue:
We cannot use the GitHub CLI for this as it uses the wrong mimetypes for the assets and the maintainers have explicitly said that they will not allow the caller to customise the mimetype as they consider that a security risk https://github.com/cli/cli/issues/11795. If you want to get rid of the deprecated action you will need to do it some other way, such as using curl to upload them, I was going to write a bash script to do it with curl so we could get rid of all the boiler plate of defining a separate step for each asset, but I haven't got around to doing that yet.

Additionally please don't use GitHub Actions templating in shell commands, instead set the values as environment variables using `env` and then use the environment variables in the script. See https://docs.zizmor.sh/audits/#template-injection

PikachuEXE:
You go ahead with `bash script` coz I have no idea how to get rid of all the boiler plate of defining a separate step for each asset 😅

PR #9494, opened 2026-07-25
Combine in-template keydown listeners that call the same handler

absidue:
## Pull Request Type

- [x] Performance improvement

## Description

When we write multiple keydown listeners for different keys in a Vue template that call the same handler (e.g. spacebar and enter), the Vue compiler is clever enough to recognise that it is the same event and only registers a single keydown event listener in the DOM but doesn't recognise that they call the same handler, instead looping over the filters and handlers. This pull request combines those cases in the template so that the Vue compiler generates a single filter and handler call, we already use that pattern in various places in the code base (e.g. channel and trending tabs), so this just expands it to the remaining places.

## Screenshots

Formatted snippet of the `dist/renderer.js` file showing the before and after difference for the search page fetch more results button when running `pnpm run pack:renderer` (`w` is our `nextPage` function).

```diff
  xo(
    "div",
    {
      class: "getNextPage",
      role: "button",
      tabindex: "0",
      onClick: w,
-     onKeydown: [
-       tc(Ql(w, ["prevent"]), ["enter"]),
-       tc(Ql(w, ["prevent"]), ["space"]),
-     ],
+     onKeydown: tc(Ql(w, ["prevent"]), ["enter", "space"]),
    },
    [
      Eo(Vt($g), { icon: ["fas", "search"] }),
      Mo(" " + ie(Vt(t)("Search Filters.Fetch more results")), 1),
    ],
    40,
    JX
  )
```

## Testing

Try various buttons that were affected by the changes in this pull request such as expanding and collapsing video descriptions or selecting chapters in the chapter selector.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/renderer/components/CommentSection/CommentSection.vue
+++ b/src/renderer/components/CommentSection/CommentSection.vue
@@ -12,8 +12,7 @@
         role="button"
         tabindex="0"
         @click="showComments = false"
-        @keydown.space.prevent="showComments = false"
-        @keydown.enter.prevent="showComments = false"
+        @keydown.enter.space.prevent="showComments = false"
       >
         {{ $t("Comments.Hide Comments") }}
       </span>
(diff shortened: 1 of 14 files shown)


radmorecameron (approved):
Seems to work as expected for the places I tried it 😄

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

PR #9508, opened 2026-07-26
Feat: Add SponsorBlock channels exclusion list

### LINKED ISSUE #4991: [Feature Request]: Allow whitelisting of channels from Sponsorblock skips
### Guidelines

### Problem Description

You can only set a global setting for Sponsorblock, even if you want to see sponsor segments of some YouTubers.

### Proposed Solution

Add a list of channels excluded from Sponsorblock, either for each segment category by itself or for the whole Sponsorblock functionality.

### Alternatives Considered

None as far as I know

### Issue Labels

### Additional Information

_No response_

Shadorc:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/4991

## Description

Add an exclusion list setting for SponsorBlock which works the same as the 'Hide Channel' setting.

## Screenshots 

image
image

## Testing

1. Go to `Settings > SponsorBlock` and disable SponsorBlock
2. Go to any video with sponsor segments
3. Check that SponsorBlock is not enabled (watch progress shouldn't display any sponsor segments and the video should play entirely)
4. Go to `Trending` or `Subscriptions` and check that the dropdown menu on any video does not have the option "Exclude Channel from SponsorBlock" 
5. Go to `Settings > SponsorBlock` and enable SponsorBlock
6. Go to any video with sponsor segments
7. Check that SponsorBlock is enabled (watch progress should display sponsor segments and the video should skip those segments)
8. Go to `Trending` or `Subscriptions` and check that the dropdown menu on any video has the option "Exclude Channel from SponsorBlock" 
9. Go to `Settings > SponsorBlock` and add a channel to `Excluded Channels`
10. Check that the channel appears in the added item
11. Check that `Show Added Items` toggle works as expected
12. Go to any video from the excluded channel with sponsor segments
13. Check that the watch progress displays sponsor segments 
14. Check that the video plays entirely, no matter the "Skip Options" defined in SponsorBlock settings

## Desktop

- **OS:** bazzite
- **OS Version:**

### DIFF
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -308,7 +308,7 @@ const state = {
   useDeArrowThumbnails: false,
   deArrowThumbnailGeneratorUrl: 'https://dearrow-thumb.ajay.app',
   sponsorBlockExcludedChannels: '[]',
-  sponsorBlockShowAddedExcludedChannels: true,
+  sponsorBlockShowExcludedChannels: true,
   // This makes the `favorites` playlist uses as quick bookmark target
   // If the playlist is removed quick bookmark is disabled
   quickBookmarkTargetPlaylistId: 'favorites',
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -307,6 +307,8 @@ const state = {
   useDeArrowTitles: false,
   useDeArrowThumbnails: false,
   deArrowThumbnailGeneratorUrl: 'https://dearrow-thumb.ajay.app',
+  sponsorBlockChannelsAllowed: '[]',
+  sponsorBlockShowAddedChannelsAllowed: true,
   // This makes the `favorites` playlist uses as quick bookmark target
   // If the playlist is removed quick bookmark is disabled
   quickBookmarkTargetPlaylistId: 'favorites',
--- a/src/renderer/store/modules/settings.js
+++ b/src/renderer/store/modules/settings.js
@@ -307,8 +307,8 @@ const state = {
   useDeArrowTitles: false,
   useDeArrowThumbnails: false,
   deArrowThumbnailGeneratorUrl: 'https://dearrow-thumb.ajay.app',
-  sponsorBlockChannelsAllowed: '[]',
-  sponsorBlockShowAddedChannelsAllowed: true,
+  sponsorBlockExcludedChannels: '[]',
+  sponsorBlockShowAddedExcludedChannels: true,
   // This makes the `favorites` playlist uses as quick bookmark target
   // If the playlist is removed quick bookmark is disabled
   quickBookmarkTargetPlaylistId: 'favorites',
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -876,8 +876,8 @@ Video:
   Copy Invidious Channel Link: Copy Invidious Channel Link
   Hide Channel: Hide Channel
   Unhide Channel: Show Channel
-  SponsorBlock Allow Channel: Allow Channel in SponsorBlock
-  SponsorBlock Disallow Channel: Disallow Channel in SponsorBlock
+  SponsorBlock Include Channel: Include Channel to SponsorBlock
+  SponsorBlock Exclude Channel: Exclude Channel from SponsorBlock
   Views: Views
   Loop Playlist: Loop Playlist
   Shuffle Playlist: Shuffle Playlist
@@ -1154,8 +1154,8 @@ Screenshot Error: Screenshot failed. {error}
 Screenshot Clipboard Error: Screenshot copy to clipboard failed
 Channel Hidden: '{channel} added to channel filter'
 Channel Unhidden: '{channel} removed from channel filter'
-SponsorBlock Channel Allowed: '{channel} added to SponsorBlock allowed list'
-SponsorBlock Channel Disallowed: '{channel} removed from SponsorBlock allowed list'
+SponsorBlock Channel Excluded: '{channel} excluded from SponsorBlock filter'
+SponsorBlock Channel Included: '{channel} included in SponsorBlock filter'
 Trimmed input must be at least N characters long: Trimmed input must be at least 1 character long | Trimmed input must be at least {length} characters long
 Tag already exists: '"{tagName}" tag already exists'
 
--- a/static/locales/en-US.yaml
+++ b/static/locales/en-US.yaml
@@ -689,6 +689,8 @@ Settings:
       Invalid Channel: Channel ID provided was invalid
       API Error: Error retrieving user with the ID provided. Please check again if the ID is correct.
       Already Exists: Channel ID already exists
+      Tooltip: Enter a channel ID to not use SponsorBlock on. Sponsor segments in videos from this channel won't be skipped.
+        The channel ID entered must be a complete match and is case sensitive.
   Parental Control Settings:
     Parental Control Settings: Parental Control
     Hide Unsubscribe Button: Hide Unsubscribe Button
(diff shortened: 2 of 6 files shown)


static/locales/en-US.yaml
```diff
@@ -682,6 +682,12 @@ Settings:
       Prompt To Skip: Prompt To Skip
       Do Nothing: Do Nothing
     Category Color: Category Color
+    Allowed Channels Disabled Message: Some channels were allowed using ID and weren't processed. Feature is blocked while those IDs are updating
+    Allowed Channels: Allowed Channels
+    Allowed Channels Placeholder: Channel ID
+    Allowed Channels Invalid: Channel ID provided was invalid
+    Allowed Channels API Error: Error retrieving user with the ID provided. Please check again if the ID is correct.
+    Allowed Channels Already Exists: Channel ID already exists
```
PikachuEXE: Please just group these new entries like `Watched Progress Saving Mode`

static/locales/en-US.yaml
```diff
@@ -1109,6 +1117,8 @@ Tooltips:
   SponsorBlock Settings:
     UseDeArrowTitles: Replace video titles with user-submitted titles from DeArrow.
     UseDeArrowThumbnails: Replace video thumbnails with thumbnails from DeArrow.
+    Allowed Channels: Enter a channel ID to not use SponsorBlock on. Sponsor segments in videos from this channel won't be skipped.
```
PikachuEXE: Generally I starts to put tooltips within the new "group" instead of under `Tooltips`, but this is not blocking.
```suggestion
    Allowed Channels: Enter a channel ID to have SponsorBlock disabled. Sponsor segments in videos from this channel won't be skipped.
```

PikachuEXE (commented):
1. The new action is in different category from Hide/Show Channel, please add separator
2. The actual meaning is "channels excluded from SponsorBlock" or "channels with SponsorBlock disabled" or "SponsorBlock disabled channels" not "allowed by SponsorBlock". So the actions should be something like "Exclude channel from SponsorBlock" / "Disable SponsorBlock on channel"

Shadorc:
@PikachuEXE Thanks for your suggestions. I was not happy with allow/disallow but I didn't think about include/exclude

Also I've added a non-suggested change: the dropdown option is not added anymore if sponsor block is disabled.

Shadorc:
I've found a bug but it seems to concern `FtInputTags`, this is not related to my MR.
If you spam enter while entering channel ID, it is added multiple times
image

Separate issue open: https://github.com/FreeTubeApp/FreeTube/issues/9510

static/locales/en-US.yaml
```diff
@@ -870,6 +879,8 @@ Video:
   Copy Invidious Channel Link: Copy Invidious Channel Link
   Hide Channel: Hide Channel
   Unhide Channel: Show Channel
+  SponsorBlock Include Channel: Include Channel to SponsorBlock
```
PikachuEXE: Though it reads rightist for `Exclude Channel from SponsorBlock`, `Include Channel to SponsorBlock` feels odd
What about `Disable/Enable SponsorBlock on Channel` or... `Re-Include Channel to SponsorBlock`? (no good idea yet

src/renderer/store/modules/settings.js
```diff
@@ -307,6 +307,8 @@ const state = {
   useDeArrowTitles: false,
   useDeArrowThumbnails: false,
   deArrowThumbnailGeneratorUrl: 'https://dearrow-thumb.ajay.app',
+  sponsorBlockExcludedChannels: '[]',
+  sponsorBlockShowAddedExcludedChannels: true,
```
PikachuEXE: Dunno why we keep adding `Added` to "show array like settings" settings
```suggestion
  sponsorBlockExcludedChannels: '[]',
  sponsorBlockShowExcludedChannels: true,
```

Shadorc: I've tried to follow what alreay existed, but I agree this is a strange naming

static/locales/en-US.yaml
Shadorc: `Disable/Enable SponsorBlock on Channel` is fine
It could also be `Don't use SponsorBlock on Channel`/`Use SponsorBlock on Channel`

What do you think?

Edit: "Disable/Enable" is shorter, and more common, I think it's a good phrasing, I will go with that

efb4f5ff-1298-471a-8973-3d47447115dc:
This copies the behavior from Hide Channel of not showing the entry on the Subscription page. I would argue that this needs to be shown on the Subscriptions page. 

The Hide Channel entry was excluded from the Subscriptions page as it didnt really made any sense to block a channel you are subscribed to. On the other hand its totally possible that the user is subscribed to certain channels they dont mind watching the SP segments

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Should we maybe reuse these instead. This way translators dont need to basically copy paste the strings

https://github.com/FreeTubeApp/FreeTube/blob/6bd8b322fe1f07cc03fdb7441e3d5edd340286f4/static/locales/en-US.yaml#L573-L577

Shadorc:
@efb4f5ff-1298-471a-8973-3d47447115dc Those strings are in `Settings.Distraction Free Settings`, are you ok if I move them up inside `Settings` to use them ? Or I can let them here, and use them as is, as you prefer.

efb4f5ff-1298-471a-8973-3d47447115dc:
Not sure what the best way togo is tbh, so ill wait on the others for direction

Shadorc:
I will use them as is for now, it prevents at least having duplicated strings to localize. Let me know if I should re-order them too.

PikachuEXE:
Just use them until we got 3 places using them (in that case we can move them out of group

PikachuEXE:
Re-read https://github.com/FreeTubeApp/FreeTube/issues/4991 and wondering if it's asking for total disable or just disable the skipping...

> Problem: You can only set a global setting for Sponsorblock, even if you want to see sponsor segments of some YouTubers.

Not saying this PR doesn't deal with the problem, just wondering if it's better to show the segments or not (maybe it's much easier to just turn the whole thing off I dunno)

Shadorc:
I agree that it's probably more useful to show the sponsor segments, but not skip them.
This would become an overwrite of all the "Skip Option" settings to "Do Nothing", instead of an overwrite of the "Enable SPonsorBlock" to disable

@efb4f5ff-1298-471a-8973-3d47447115dc What do you think?

efb4f5ff-1298-471a-8973-3d47447115dc:
So the only thing i could find regarding this is https://wiki.sponsor.ajay.app/w/Advanced_skip_options. So if this stays the only data point then we can basically do whatever we want. I have to think some more on it what would be "better". Show segments or not showing anything at all.

Feel free to persuade me with more arguments though

Edit: i could also ping everybody that is involved in that issue and let them fight it out

PikachuEXE:
The issue title is `Allow whitelisting of channels from Sponsorblock skips` so just skips not the segment displays
I don't see benefit in not showing segments for some channels (since they want to watch it, but maybe they also want to skip when some feel boring and it's easier to manually skip those when segments are shown

efb4f5ff-1298-471a-8973-3d47447115dc:
Thats making an assumption based on the title but they never clarified in the issue if they actually would want to still see the segments appear on the timeline or not

edit: i think i agree with your argument though

Shadorc:
I agree with PikachuEXE's arguments too 
Maybe we could ask users, just to be sure, maybe there will be a consensus, who knows!

efb4f5ff-1298-471a-8973-3d47447115dc:
If we want I can do a poll on our Discussions page, ping the users that are involved with that issue and throw a link in the matrix general chat. 

edit: and hope that users reach a consensus instead a close 40/60 split or something similar

absidue:
Doing a poll for this just sounds like unnecessary admin work, showing the segments without skipping is non-invasive and as Pika mentioned allows people to still manually skip segments if they want to.

Shadorc:
I've updated the MR to change the behaviour as discussed.
I've not updated texts because they are still pretty aligned with the new behaviour imo

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Issue #9468, opened 2026-07-20
[Bug]: Playlist Video Draggable Area Disappears When Removing Video


deepspaceaxolotl:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads). **The only official downloads are from the [https://freetubeapp.io](https://freetubeapp.io/#download) website, the [FreeTubeApp/FreeTube GitHub repository](https://github.com/FreeTubeApp/FreeTube/releases) and the [io.freetubeapp.FreeTube flatpak](https://flathub.org/en/apps/io.freetubeapp.FreeTube) from Flathub.**
- [x] I have [searched the issue tracker for **open and closed** issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

When removing a video from a playlist in grid mode, the draggable area above each video disappears for a few seconds before returning.

### Expected Behavior

The draggable area should not disappear when a video is removed.

### Issue Labels

visual bug

### FreeTube Version

v0.25.0 Beta

### Operating System Version

Windows 10, version 22H2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [ ] I have encountered this bug in the latest **official** [nightly build](https://docs.freetubeapp.io/development/nightly-builds), do not check this box if you are using an unofficial or custom build.

efb4f5ff-1298-471a-8973-3d47447115dc:
This is intend behavior so it isnt possible to mess with the database while there is still a recovery period going on (the toast message to undo event)

deepspaceaxolotl:
Good to know! Wondering if it would be possible to make it less visually disruptive though, like maybe just greying out the drag symbol rather than having the whole element disappear and have everything jump around?

efb4f5ff-1298-471a-8973-3d47447115dc:
[...]

PR #9515, opened 2026-07-28
Update user playlist view to show grab bar disabled when items being deleted

PikachuEXE:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/9468#issuecomment-5069220936 (already closed, but the suggestion is from a comment on the issue

## Description

> Wondering if it would be possible to make it less visually disruptive though, like maybe just greying out the drag symbol rather than having the whole element disappear and have everything jump around?

This PR makes grab bar remain visible but disabled (this is about grid mode only, list mode unchanged)

## Screenshots 

https://github.com/user-attachments/assets/cb231984-22cd-4c49-be7b-3f5e6322d2a9

## Testing

- Set `Video View Type` to `Grid`
- Spawn a user playlist with 3+ items
- Remove an item
- Ensure grab bar still visible and look disabled and cannot be dragged

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/FtElementList/FtElementList.vue
+++ b/src/renderer/components/FtElementList/FtElementList.vue
@@ -24,8 +24,8 @@
       :playlist-type="playlistType"
       :playlist-item-id="result.playlistItemId"
       :dragged-video="draggedVideo"
-      :is-sort-order-custom="isSortOrderCustom"
       :is-video-dragging="isVideoDragging"
+      :video-dragging-possible="videoDraggingPossible"
       @drag-video="dragVideo"
       @move-dragged-video="moveDraggedVideo"
       @drag-video-end="afterDrag"
@@ -121,11 +121,11 @@ const props = defineProps({
     type: Object,
     default: () => ({ videoId: null, playlistItemId: null }),
   },
-  isSortOrderCustom: {
+  isVideoDragging: {
     type: Boolean,
     default: false,
   },
-  isVideoDragging: {
+  videoDraggingPossible: {
     type: Boolean,
     default: false,
   },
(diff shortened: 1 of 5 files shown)


efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Two things i noticed. 

The first one is that the user can drag the grab bar when its disabled. Im not sure if we need todo anything about this though.

Second, grab bar shouldnt be available when there is only one video in playlist

https://github.com/user-attachments/assets/8774f18e-97d6-4f9f-aba2-38d9ccab0981

PikachuEXE:
Fixed grab bar draggable when disabled (unnecessary change from me)
However for grab bar it seems more consistent to show grab bar on custom sort regardless of no. of videos (just like when videos being deleted

Shadorc:
I've tested the MR with different scenarios and it works fine

I would tend to say that disabling the grab bar would have made sense if the drop-down sort menu was always displayed. It would have indicated that, indeed it's a custom sort, but you can't sort one video.
With the "Sort By" dropdown being hidden with one video, it feels more natural to also hide the grab bar.

PikachuEXE:
Did not test playlist with 1 items only
Now grab bar hidden like sort control

Shadorc:
LGTM!

efb4f5ff-1298-471a-8973-3d47447115dc:
Should we enforce the same behavior in list mode to keep the UX consistent?

PikachuEXE:
I think this PR is necessary to avoid the jumping around thing, not on list mode side
And user can't use both mode at the same time...

https://github.com/FreeTubeApp/FreeTube/issues/9468#issuecomment-5068049916
> Wondering if it would be possible to make it less visually disruptive though, like maybe just greying out the drag symbol rather than having the whole element disappear and have everything jump around?

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

Shadorc (approved):

PR #9525, opened 2026-07-29
Fix: Opening a playlist from a user playlist reloads the view twice

### LINKED ISSUE #9513: [Bug]: Opening an invalid playlist from a user playlist breaks UI
### Guidelines

### Describe the bug

1. Go to a user playlist
2. Enter an invalid playlist URL in the search bar (e.g. https://www.youtube.com/playlist?list=foobar)
3. Observe that the UI is broken

### Expected Behavior

The page shows an empty playlist, or better, an unviewable playlist (https://github.com/FreeTubeApp/FreeTube/issues/9518)

### Issue Labels

visual bug

### FreeTube Version

v0.25.1-beta

### Operating System Version

Fedora

### Installation Method

.zip / .7z

### Primary API used

Local API

_No response_

[before.png](https://github.com/user-attachments/assets/0c802bd5-188a-4757-bc5b-07347756c537)
[after.png](https://github.com/user-attachments/assets/f4f3d7c7-5109-43d3-aa9c-c5d40ee46a8a)

### Additional Information

_No response_

### Nightly Build

- [x] I have encountered this bug in the latest **official** [nightly build](https://docs.freetubeapp.io/development/nightly-builds), do not check this box if you are using an unofficial or custom build.

### LINKED ISSUE #9517: [Bug]: Opening a playlist from a user playlist reloads the view twice
### Guidelines

### Describe the bug

1. Open a user playlist 
2. Enter a (valid) playlist URL in the search bar
3. Observe that the UI reloads after the first load and causes a flicker

### Expected Behavior

UI should not reload

### Issue Labels

visual bug

### FreeTube Version

v0.25.1-beta

### Operating System Version

Fedora

### Installation Method

Flathub

### Primary API used

Local API

_No response_

_No response_

### Additional Information

In the 'after' screenshot from https://github.com/FreeTubeApp/FreeTube/issues/9513, we can see that the error "This playlist type is unviewable" is throw twice, even if I've only loaded the page once.

When we change route from playlist page, `getPlaylistInfo` is called twice.
The first time due to the `playlistId` change: https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/views/Playlist/Playlist.vue#L591

The second time due to the `selectedUserPlaylistVideoCount` change: https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/views/Playlist/Playlist.vue#L606

They are separated by 100ms due to one being debounced and not the other.
They both change reactive variables, which updates the UI.

This causes a flicker on load after 100ms

### Nightly Build

Shadorc:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/9517
closes https://github.com/FreeTubeApp/FreeTube/issues/9513 (side effect)

## Description

Early exit from watchers that are triggering redundant calls to `getPlaylistInfo` (debounced or not).
For the fix to work, the state of the playlist needs to be reset, otherwise everyting break on route change. This has the added benefit of fixing issue https://github.com/FreeTubeApp/FreeTube/issues/9513 too.

## Testing

**Valid playlist**
1. Open a user playlist
2. Enter a valid playlist URL in the search bar
3. Check that the UI does not reload after the first load and does not cause a flicker

**Invalid playlist**
1. Open a user playlist
2. Enter an invalid playlist URL in the search bar (e.g. https://www.youtube.com/playlist?list=foobar)
3. Check that the UI does not break and display `The playlist currently has no videos.`

## Desktop

- **OS:** Bazzite
- **OS Version:**

### DIFF
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -1077,7 +1077,7 @@ export function debounce(func, wait) {
   // Using a fully fledged function here instead of an arrow function
   // so that we can get `this` and pass it onto the original function.
   // Vue components using the options API use `this` alot.
-  return function(...args) {
+  return function (...args) {
     const context = this
 
     clearTimeout(timeout)
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -1066,15 +1066,10 @@ export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle
 }
 
 /**
- * @template {Function} U
- * @typedef {{cancel: () => void} & U} ReturnedFunc
- */
-
-/**
  * @template {Function} T
  * @param {T} func
  * @param {number} wait
- * @returns {ReturnedFunc<T>}
+ * @returns {T}
  */
 export function debounce(func, wait) {
   let timeout
@@ -1082,7 +1077,7 @@ export function debounce(func, wait) {
   // Using a fully fledged function here instead of an arrow function
   // so that we can get `this` and pass it onto the original function.
   // Vue components using the options API use `this` alot.
-  function debounceFunc(...args) {
+  return function(...args) {
     const context = this
 
     clearTimeout(timeout)
@@ -1092,13 +1087,6 @@ export function debounce(func, wait) {
       func.apply(context, args)
     }, wait)
   }
-
-  debounceFunc.cancel = function () {
-    clearTimeout(timeout)
-    timeout = null
-  }
-
-  return debounceFunc
 }
 
 /**
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -1069,7 +1069,7 @@ export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle
  * @template {Function} T
  * @param {T} func
  * @param {number} wait
- * @returns {T}
+ * @returns {{ (...args: any[]): any, cancel: () => void }}
  */
 export function debounce(func, wait) {
   let timeout
@@ -1077,7 +1077,7 @@ export function debounce(func, wait) {
   // Using a fully fledged function here instead of an arrow function
   // so that we can get `this` and pass it onto the original function.
   // Vue components using the options API use `this` alot.
-  return function (...args) {
+  function debounceFunc(...args) {
     const context = this
 
     clearTimeout(timeout)
@@ -1087,6 +1087,13 @@ export function debounce(func, wait) {
       func.apply(context, args)
     }, wait)
   }
+
+  debounceFunc.cancel = function () {
+    clearTimeout(timeout)
+    timeout = null
+  }
+
+  return debounceFunc
 }
 
 /**
--- a/src/renderer/helpers/utils.js
+++ b/src/renderer/helpers/utils.js
@@ -1066,10 +1066,15 @@ export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle
 }
 
 /**
+ * @template {Function} U
+ * @typedef {{cancel: () => void} & U} ReturnedFunc
+ */
+
+/**
  * @template {Function} T
  * @param {T} func
  * @param {number} wait
- * @returns {{ (...args: any[]): any, cancel: () => void }}
+ * @returns {ReturnedFunc<T>}
  */
 export function debounce(func, wait) {
   let timeout
--- a/src/renderer/views/Playlist/Playlist.vue
+++ b/src/renderer/views/Playlist/Playlist.vue
@@ -461,7 +461,7 @@ function resetState() {
   playlistThumbnail.value = ''
   viewCount.value = 0
   videoCount.value = 0
-  lastUpdated.value = ''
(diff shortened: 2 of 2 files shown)


src/renderer/helpers/utils.js
```diff
@@ -1069,15 +1069,15 @@ export function localizeAndAddKeyboardShortcutToActionTitle(localizedActionTitle
  * @template {Function} T
  * @param {T} func
  * @param {number} wait
- * @returns {T}
+ * @returns {{ (...args: any[]): any, cancel: () => void }}
  */
```
PikachuEXE, line 1073: Modified from https://stackoverflow.com/questions/68738823/jsdoc-typedef-for-function-with-sub-function
```suggestion

/**
 * @template {Function} U
 * @typedef {{cancel: () => void} & U} ReturnedFunc
 */

/**
 * @template {Function} T
 * @param {T} func
 * @param {number} wait
 * @returns {ReturnedFunc<T>}
 */
```

Image

Image

Shadorc: Thanks!

PikachuEXE:
Saw from #9517 

> Open a playlist
Enter another playlist URL in the search bar (e.g. https://www.youtube.com/playlist?list=foobar), can be valid or invalid
Observe that the UI reloads after the first load and causes a flicker

But is it online playlist or user playlist for 1 & 2? I want to reproduce it first

Shadorc:
> But is it online playlist or user playlist for 1 & 2? I want to reproduce it first

User playlists are playlists specific to FreeTube, available in the `Playlists` tab, right? 
If so, is it possible to open a user playlist from a URL? 

I have updated this MR title and the issue title + repro steps to be clearer

It does not happen with two YouTube playlists because the issue comes from a change in the `selectedUserPlaylistVideos` array (it becomes empty when switching to a YouTube playlist).
If it's possible to switch to a user playlist from another user playlists, then the issue would also happen if the two playlists have a different number of videos.

src/renderer/views/Playlist/Playlist.vue
```diff
@@ -609,6 +609,12 @@ watch(selectedUserPlaylistVideoCount, async () => {
   // So length is monitored instead
   // Assuming in user playlist video cannot be swapped without length change
 
+  // This watch function can be triggered while a debounced `getPlaylistInfoDebounce`
+  // call is still pending from another watch function. If we let it execute, we'd have
+  // two close calls to fetch playlist info in quick succession.
+  // See https://github.com/FreeTubeApp/FreeTube/issues/9517
+  getPlaylistInfoDebounce.cancel()
+
```
PikachuEXE: After testing locally for a while I think the fix is to NOT call `getPlaylistInfo` when the change is due to switching to a YT playlist
So might be better to change computed to
```js
/** @type {import('vue').ComputedRef<any[]>} */
const selectedUserPlaylistVideos = computed(() => selectedUserPlaylist.value?.videos)

const selectedUserPlaylistVideoCount = computed(() => selectedUserPlaylistVideos.value?.length)
```

And change watcher to
```js
watch(selectedUserPlaylistVideoCount, async (newVal) => {
  // Monitoring `selectedUserPlaylistVideos` makes this function called
  // Even when the same array object is returned
  // So length is monitored instead
  // Assuming in user playlist video cannot be swapped without length change

  // Only do refetch when it's user playlist
  if (newVal != null) {
    // Re-fetch from local store when current user playlist videos updated
    // MUST NOT use `getPlaylistInfoDebounce` as it will cause delay in data update
    // Causing deleted videos to reappear for one frame
    getPlaylistInfo()
  }
})
```

So that if there is a number in new value it's user playlist and should call `getPlaylistInfo` and not call it otherwise

Update 1: Somehow this might be needed, apply if needed
```js
const userPlaylistDuplicateItemCount = computed(() => {
  if (userPlaylistUniqueVideosCount.value === 0) { return 0 }
  if (selectedUserPlaylist.value == null) { return 0 }

  return selectedUserPlaylist.value.videos.length - userPlaylistUniqueVideosCount.value
})
```

Shadorc: If I try your approach, the route completely breaks.
I don't understand what's causing this...

Shadorc: I was not able to figure out what was happening. The state of the playlist was broken.

The only way to apply your suggestions, was to properly reset the state of the playlist on route change, which has the side effect of fixing https://github.com/FreeTubeApp/FreeTube/issues/9513

I also discovered that we had several debounced calls only useful for user playlists, so I added early exits.
Now that the state is correctly reset, it is possible to apply the same early exit for the `selectedUserPlaylistVideoCount` watch.

Finally, I ended up straying completely from your suggestions, but I think the fix is more consistent with what we do in the other watch-er, and it has the added benefit of solving another bug.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

Issue #4323, opened 2023-11-13
[Feature Request]: Support for clips


CalAlaera:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

YouTube clips have a URL that looks like https://youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_ and point to a short segment of a much longer video (in this case, a 13-second snippet out of a several-hours-long video).

They're a way for people to easily share moments from longer videos with their friends.

It would be great if FreeTube could support them, please.

### Proposed Solution

It would be sufficient for FreeTube to discover the larger video's URL and the clip's start timestamp, and start playing the larger video at the appropriate time using existing features.

### Alternatives Considered

The way YouTube handles this is to pause at the end of the clip (whether that's stored as a timestamp or as a duration, I don't know) and allow the user to click to continue playing the video from there. This could be implemented by automatically pausing the video at the end of the clip. However, this would merely be a "nice-to-have", as clips cannot be longer than I think a minute - short enough that it should be obvious from context what part of the video the sharer intended to share.

### Issue Labels

new feature

absidue:
The provided link shows a page not available error page when visited in a browser. Could you please provide a working URL example?

CalAlaera:
@absidue Github chopped off the trailing underscore. Try [https://youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_](https://youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_).

PR #9532, opened 2026-07-29
feat: Add initial support for YouTube clip links

radmorecameron:
## Pull Request Type
- [x] Feature Implementation

## Related issue
#4323 

## Description
This PR adds initial support for clip links.

## Testing

- Enter [https://www.youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_](https://www.youtube.com/clip/UgkxxPM3BRphCAPLP88YoUGuj79KXPfpNNO_) into top nav url
- Get redirected to the correct video at the correct time stamp (end timestamp + clip title are parsed but currently not displayed)

## Desktop

- **OS:** Fedora Linux
- **OS Version:** 42

### DIFF
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -863,6 +863,7 @@ export async function getHashtagInvidious(hashtag, page = 1) {
 export async function getClipInvidious(clipId) {
   // todo: invidious does have a clip endpoint for the api but
   // it currently makes an expensive get_video request that we might want to avoid.
+  // https://github.com/iv-org/invidious/pull/5871
   const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
   return response
 }
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -860,6 +860,13 @@ export async function getHashtagInvidious(hashtag, page = 1) {
   return response.results
 }
 
+export async function getClipInvidious(clipId) {
+  // todo: invidious does have a clip endpoint for the api but
+  // it currently makes an expensive get_video request that we might want to avoid.
+  const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
+  return response
+}
+
 /**
  * Generates a DASH manifest locally from Invidious' adaptive formats and manifest,
  * doing so allows us to support multiple audio tracks, which Invidious doesn't support yet
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -864,7 +864,7 @@ export async function getHashtagInvidious(hashtag, page = 1) {
 export async function getClipInvidious(clipId) {
   const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
 
-  const parsedParams = ClipParams.decode(decodeURIComponent(Utils.base64ToU8(response.params)))
+  const parsedParams = ClipParams.decode(Utils.base64ToU8(decodeURIComponent(response.params)))
   return {
     videoId: response.videoId,
     startTime: parsedParams.clipParamData.startTime / 1000, // convert to seconds
--- a/src/renderer/helpers/api/invidious.js
+++ b/src/renderer/helpers/api/invidious.js
@@ -2,7 +2,8 @@ import store from '../../store/index'
 import { calculatePublishedDate, getRelativeTimeFromDate } from '../utils'
 import { isNullOrEmpty } from '../strings'
 import autolinker from 'autolinker'
-import { FormatUtils, Misc, Player } from 'youtubei.js'
+import { FormatUtils, Misc, Player, Utils } from 'youtubei.js'
+import { ClipParams } from '../../../../node_modules/youtubei.js/dist/protos/generated/misc/params'
 
 /** @typedef {{url: string, width: number, height: number}} InvidiousImageObject */
 /** @typedef {{quality: string, url: string, width: number, height: number}} InvidiousThumbnailObject */
@@ -861,11 +862,16 @@ export async function getHashtagInvidious(hashtag, page = 1) {
 }
 
 export async function getClipInvidious(clipId) {
-  // todo: invidious does have a clip endpoint for the api but
-  // it currently makes an expensive get_video request that we might want to avoid.
-  // https://github.com/iv-org/invidious/pull/5871
   const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
-  return response
+
+  const parsedParams = ClipParams.decode(decodeURIComponent(Utils.base64ToU8(response.params)))
+  return {
+    videoId: response.videoId,
+    startTime: parsedParams.clipParamsData.startTime / 1000, // convert to seconds
+    endTime: parsedParams.clipParamsData.endTime / 1000, // convert to seconds
+    clipTitle: parsedParams.clipParamsData.clipTitle,
+    clipMetadata: parsedParams.clipParamsData.clipMetadata
+  }
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/renderer/helpers/api/local.js
```diff
@@ -2268,3 +2268,18 @@ export async function getLocalCommunityPostComments(postId, channelId) {
...
 }
+
+export async function getLocalClip(clipId) {
+  const innertube = await createInnertube()
+
+  const clipResponse = await innertube.resolveURL('https://www.youtube.com/clip/' + clipId)
+
+  const videoId = clipResponse?.payload?.videoId
+
+  // todo: figure out how we should parse additional properties from payload params protobuf
```
radmorecameron: @absidue any ideas on how we should be parsing the protobuf params (the params contain the clip title, start time and end time)? I'm familiar with how it's done in Invidious but am unsure on how we should parse it here (should it be done in YouTube.js somewhere? should we do it in this repo somehow?)

absidue: By parsing the protobuf, if YouTube.js already has it you can import the protobuf decoder for it (see the PR i did a while back here with FormatXTags).

efb4f5ff-1298-471a-8973-3d47447115dc:
Can be undrafted :)

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
navigating to the url throws an error

Image

Shadorc:
Navigating to the URL with Invidious API throws too
image

src/renderer/helpers/api/invidious.js
```diff
@@ -860,6 +861,19 @@ export async function getHashtagInvidious(hashtag, page = 1) {
...
+export async function getClipInvidious(clipId) {
+  const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
+
+  const parsedParams = ClipParams.decode(decodeURIComponent(Utils.base64ToU8(response.params)))
+  return {
+    videoId: response.videoId,
+    startTime: parsedParams.clipParamsData.startTime / 1000, // convert to seconds
+    endTime: parsedParams.clipParamsData.endTime / 1000, // convert to seconds
+    clipTitle: parsedParams.clipParamsData.clipTitle,
+    clipMetadata: parsedParams.clipParamsData.clipMetadata
```
radmorecameron:
```suggestion
    clipMetadata: parsedParams.clipParamData.clipMetadata
```

src/renderer/helpers/api/invidious.js
```diff
@@ -860,6 +861,19 @@ export async function getHashtagInvidious(hashtag, page = 1) {
...
 }
 
+export async function getClipInvidious(clipId) {
+  const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
+
+  const parsedParams = ClipParams.decode(decodeURIComponent(Utils.base64ToU8(response.params)))
+  return {
+    videoId: response.videoId,
+    startTime: parsedParams.clipParamsData.startTime / 1000, // convert to seconds
+    endTime: parsedParams.clipParamsData.endTime / 1000, // convert to seconds
```
radmorecameron:
```suggestion
    endTime: parsedParams.clipParamData.endTime / 1000, // convert to seconds
```

src/renderer/helpers/api/invidious.js
```diff
@@ -860,6 +861,19 @@ export async function getHashtagInvidious(hashtag, page = 1) {
   return response.results
 }
 
+export async function getClipInvidious(clipId) {
+  const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
+
+  const parsedParams = ClipParams.decode(decodeURIComponent(Utils.base64ToU8(response.params)))
+  return {
+    videoId: response.videoId,
+    startTime: parsedParams.clipParamsData.startTime / 1000, // convert to seconds
```
radmorecameron:
```suggestion
    startTime: parsedParams.clipParamData.startTime / 1000, // convert to seconds
```

src/renderer/helpers/api/invidious.js
```diff
@@ -860,6 +861,19 @@ export async function getHashtagInvidious(hashtag, page = 1) {
...
+    endTime: parsedParams.clipParamsData.endTime / 1000, // convert to seconds
+    clipTitle: parsedParams.clipParamsData.clipTitle,
```
radmorecameron:
```suggestion
    clipTitle: parsedParams.clipParamData.clipTitle,
```

radmorecameron:
Both the Invidious and local api should be working now. Issue was a mix of typo + updating the name of a param in my YouTube.js PR and forgetting to reflect that change here.

src/renderer/helpers/api/local.js
```diff
@@ -2486,3 +2487,21 @@ export async function getLocalCommunityPostComments(postId, channelId) {
...
+
+  const parsedParams = ClipParams.decode(Utils.base64ToU8(decodeURIComponent(clipResponse.payload.params)))
+
+  return {
+    videoId,
+    startTime: parsedParams.clipParamData.startTime / 1000, // convert to seconds
+    endTime: parsedParams.clipParamData.endTime / 1000, // convert to seconds
+    clipTitle: parsedParams.clipParamData.clipTitle,
+    clipMetadata: parsedParams.clipParamData.clipMetadata
+  }
```
absidue: As this section of the code is the same for the local and Invidious APIs apart from where the video ID and params string is read from, could you please extract it into a shared function that takes the video ID and params text as the input, decodes it and returns the object.

src/renderer/helpers/api/invidious.js
```diff
@@ -860,6 +861,19 @@ export async function getHashtagInvidious(hashtag, page = 1) {
   return response.results
 }
 
+export async function getClipInvidious(clipId) {
+  const response = await resolveUrl('https://www.youtube.com/clip/' + clipId)
+
+  const parsedParams = ClipParams.decode(Utils.base64ToU8(decodeURIComponent(response.params)))
```
absidue: As we cannot treeshake YouTube.js because of Options API components, we exclude the entire YouTube.js dependency in builds where `SUPPORTS_LOCAL_API` is `false` (that's also why we use the `repairInvidiousManifest` function in those builds instead of generating a DASH manifest with YouTube.js). So you either need to gate the Invidious functionality here behind `SUPPORTS_LOCAL_API` too or use the Invidious clips API endpoint (I know that you originally avoided it here because of performance concerns on their side but using it as a fallback is probably okay).

radmorecameron: Done :smile:

src/renderer/helpers/api/local.js
radmorecameron: I created a new `shared.js` file in the api folder. I felt like the current `utils.js` wasn't a great fit for this shared code since it wouldn't have much use outside of the local.js and invidious.js files.

Shadorc (dismissed):
Tested with local API, Invidious API and `SUPPORTS_LOCAL_API` set to false

src/renderer/helpers/api/shared.js
```diff
@@ -0,0 +1,19 @@
+// code shared between invidious and local api (like parsing or types)
+export function parseVideoClipsParams(videoId, params) {
+  if (process.env.SUPPORTS_LOCAL_API) {
+    const { Utils } = require('youtubei.js')
+    const { ClipParams } = require('../../../../node_modules/youtubei.js/dist/protos/generated/misc/params')
```
absidue, line 5: Please use proper ESM static imports like we do in all other files, importing with require forces webpack to treat the entirety of YouTube.js as commonjs which means larger bundles and slower performance, as every reference to YouTube.js across the entire code base will go through the simulated commonjs runtime.

absidue (changes requested):

src/renderer/helpers/api/shared.js
radmorecameron: Makes sense, for some reason I was thinking we needed to gate it like we do for electron imports but that doesn't make sense since we import it in other files without gating it.

Updated :smile:

Shadorc (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #9551, opened 2026-08-01
Style context menus as light or dark depending on the base theme

absidue:
## Pull Request Type

- [x] Feature Implementation

## Description

Currently the context menus in FreeTube follow the system level dark/light mode, which means you can get dark menus with a light theme selected in FreeTube or get flashbanged by white menus when you have a dark theme set in FreeTube. With this pull request FreeTube overrides [Electron's `themeSource` setting](https://www.electronjs.org/docs/latest/api/native-theme#nativethemethemesource) to `"dark"`, `"light"` or `"system"` as appropriate. It also affects the dev tools and has some operating system specific effects such as changing the title bar and/or app menu colours, so most parts of the Electron UI should match as well as operating system level UI that Electron is able to override.

### Black theme
black theme with dark context menu

### Solarized Light theme
solarized light them with light context menu

## Testing

Test various themes in FreeTube and check that the context menu and devtools theming matches, additionally confirm that it correctly switched back to the operating system theme when the system theme is selected in FreeTube.

## Desktop

- **OS:** Windows
- **OS Version:** 11

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -262,6 +262,32 @@ const MIXED_SEARCH_HISTORY_ENTRIES_DISPLAY_LIMIT = 4
 // Displayed on the about page and used in the main.js file to only allow bitcoin URLs with this wallet address to be opened
 const ABOUT_BITCOIN_ADDRESS = '1Lih7Ho5gnxb1CwPD4o59ss78pwo2T91eS'
 
+const LIGHT_BASE_THEMES = [
+  'light',
+  'pastelPink',
+  'catppuccinLatte',
+  'everforestLightHard',
+  'everforestLightMedium',
+  'everforestLightLow',
+  'gruvboxLight',
+  'solarizedLight'
+]
+
+const DARK_BASE_THEMES = [
+  'dark',
+  'black',
+  'nordic',
+  'hotPink',
+  'catppuccinFrappe',
+  'catppuccinMocha',
+  'dracula',
+  'everforestDarkHard',
+  'everforestDarkMedium',
+  'everforestDarkLow',
+  'gruvboxDark',
+  'solarizedDark',
+]
+
 export {
   IpcChannels,
   DBActions,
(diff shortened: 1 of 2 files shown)


src/constants.js
```diff
@@ -262,6 +262,32 @@ const MIXED_SEARCH_HISTORY_ENTRIES_DISPLAY_LIMIT = 4
...
 
+const LIGHT_BASE_THEMES = [
+  'light',
+  'pastelPink',
+  'catppuccinLatte',
+  'everforestLightHard',
+  'everforestLightMedium',
+  'everforestLightLow',
+  'gruvboxLight',
+  'solarizedLight'
```
PikachuEXE: Just have trailing commas (I will update this myself
```suggestion
  'solarizedLight',
```

PikachuEXE (approved):

Shadorc:
On Linux Fedora, KDE Plasma, Wayland, it works for the dev tools but not for the context menus
image

absidue:
@Shadorc Is that just after launch or does it also happen if you change the base theme while FreeTube is running?

Potentially related upstream items:
- https://github.com/electron/electron/issues/28887
- https://github.com/electron/electron/pull/52575

Shadorc:
@absidue It does not work while FreeTube is running **and** after launch. I'm never able to get the context menus with a light theme (my OS theme is dark, it's probabbly using this no matter what). 

Edit: I confirm that the context menus theme can only be changed by changing my OS theme, FreeTube's base theme has no impact.

absidue:
If you press <kbd>Alt</kbd> to show the app menu does that at least match?

The fact that the devtools get updated means it is at least overriding the `prefers-color-scheme` CSS media query.

Shadorc:
It does not work with `Alt` (not in real time nor after a restart)
image

absidue:
In that case it seems like it is an Electron or Chromium bug to me (if you could create a standalone repro in Electron Fiddle and report it in the Electron repo that would be great). As it seems to work fine on Windows and macOS, as long as this pull request doesn't make it any worse than before on Linux, it would still be good to see this merged (it should hopefully start working on Linux automatically once it is fixed upstream).

radmorecameron (approved):

Shadorc:
I agree! 
I will try to reproduce with Electron Fiddle. 
This MR can be merge as is for me, it does not change the linux experience

Edit: Reproduced with Electron Fiddle, upstream bug confirmed to be https://github.com/electron/electron/issues/28887

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PR #9555, opened 2026-08-02
Implement threaded replies with YouTube.js 18.0.0

### LINKED ISSUE #9474: [Bug]: Comments not loading
### Guidelines

### Describe the bug
1. Go to any videos with comments
2. Click to load comments
3. Observe error: `TypeError: Cannot read properties of undefined (reading 'endpoint')`

### Expected Behavior

Comments should load

### Issue Labels

content not loading

### FreeTube Version

v0.25.1-beta

### Operating System Version

Fedora

### Installation Method

Flathub

### Primary API used

Local API

_No response_

_No response_

### Additional Information

_No response_

### Nightly Build

- [X] I have encountered this bug in the latest **official** [nightly build](https://docs.freetubeapp.io/development/nightly-builds), do not check this box if you are using an unofficial or custom build.

### LINKED ISSUE #9661: [Bug]:  Comments not loading using Invidious API Backend
### Guidelines

### Describe the bug

1. Go to Settings → General and set the backend to "Invidious".
2. Open any video that has comments.
3. Scroll down to the comment section.
4. See error: comments/avatars don't load (console: TypeError: Cannot read properties of undefined (reading 'at')).

### Expected Behavior

 Comments load normally.

### Issue Labels

content not loading

### FreeTube Version

v0.25.2 Beta

### Operating System Version

archlinux 7.1.8-arch1-3

### Installation Method

Flathub

### Primary API used

Invidious API

_No response_

_No response_

### Additional Information

Root cause:
In src/renderer/helpers/api/invidious.js, parseInvidiousCommentData reads comment.authorThumbnails.at(-1).url, but the Invidious comments API returns the avatar as authorThumbnail (a String URL), not authorThumbnails (an array). So it's always undefined.
Docs: https://docs.invidious.io/api/ — GET /api/v1/comments/:id returns authorThumbnail: String.
Fix:
authorThumb: toAbsoluteInvidiousUrl(youtubeImageUrlToInvidious(comment.authorThumbnail)),

### Nightly Build

CL-Jeremy:
## Pull Request Type

- [x] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #9474
closes #9661

## Description

This PR makes use of the recent 18.0.0 release of YouTube.js to implement threaded replies for local API (which would be the way forward based on the speculation in https://github.com/LuanRT/YouTube.js/issues/1214), without affecting Invidious (flattened replies now work as before the issue emerged).

Summary:
- Major refactoring: Not storing fetched replies and UI show/hide states in threads or swapping out `replyToken` in the returned data structures from the API (both local and Invidious). These are now immutable. States are managed exclusively in Vue components.
  * This has the side effect of not persisting states in deeply nested components when the the DOM sub-tree visibility is toggled, triggering GC. Currently less of a problem IMHO, but suggestions are appreciated.
  * A few obvious fixes to code and type errors.
- CSS reworks to adapt to nested replies (proved to work fine even when screen is narrow, with a 420px grid width, as replies are nested at most 3 levels below the main comment).
  * (New) Responsive design for mobile.
- When replies are toggled on, remove the counter from the show/hide link (counter-intuitive when nested threads are not shown) and add a countdown to the “load more” link.
  * Add translations to a few languages I am familiar with (some cross-checked with localization of Amazon.com), as well as a few low-hanging fruits where existing translations can be used for analogy.
  * A few minor corrections in translations as I saw fit.

## Screenshots 
comments-tablet
comments-mobile

## Testing

Open any video with comments and try to load them. Replies should also work.
## Desktop

- **OS:** macOS
- **OS Version:** 15.7.9
- **FreeTube version:** 0.25.2 Beta

## Additional context

- While loading comments now requires a full session (in order to get sub-threads) just like getting video info, the session is bound to the initial comment request, so continuations should still be fast. Note the session is reused between continuations of the same comment thread request, i. e. one per video page refresh. Question is whether that should be of concern.


CL-Jeremy:
This is largely ready for use if the functionality (viewing comments) is all that's needed.

Tested working when built against latest edge version of YouTube.js, unfolding at most 2 layers per load.
<details><summary>Screenshots</summary>
ytjs-subthreads-1
ytjs-subthreads-2
</details> 

During the refactor, I chose to store all the state in the (newly factored-out) local Vue component instead of mutating/emitting values that the current level doesn't own. This has the negative effect of not caching any loaded data or continuation state when the DOM subtree is removed after pressing "Hide <N> replies" (except top level). I currently plan to explore `<KeepAlive>` tag, but I might get reactions on the existing changes such as having done too much unnecessary/premature refactoring, etc., so I might as well wait a bit and see how UX may be improved (but I do stand behind my code with 0 AI usage up till now, if that matters).

Also tested (largely, can't exactly confirm regression) working with latest dev version of Invidious (https://inv.nadeko.net, currently only comments via API, it's for testing anyway) with flattened replies as before. I can only say I fixed a few obvious mistakes based on testing/upstream docs.

CL-Jeremy:
pnpm 11 makes a good point about risks of supply chain attacks, though this is currently irrelevant for this draft PR.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "freetube",
   "productName": "FreeTube",
   "description": "A private YouTube client",
-  "version": "0.25.1",
+  "version": "0.25.2",
```
efb4f5ff-1298-471a-8973-3d47447115dc, line 5: i think you didnt properly merge branch into dev

CL-Jeremy: Do I merge into dev or the other way around? I'm now rebasing anyway to get rid of the extraneous commits.

efb4f5ff-1298-471a-8973-3d47447115dc: you merge dev branch into yours

static/locales/es.yaml
```diff
@@ -41,6 +41,7 @@ Global:
     Watching Count: 1 espectador | {count} espectadores
     Comment Count: 1 comentario | {count} comentarios
     Like Count: 1 me gusta | {count} me gustan
+    Replies Remaining: "{' '}(queda 1) | {' '}(quedan {count})"
```
PikachuEXE, line 44: How do you translate these new entries?
Also what's the `{' '}` value here?

src/renderer/components/FtComment/FtComment.vue
```diff
@@ -0,0 +1,357 @@
...
+const showMoreRepliesText = computed(() => {
+  const numLoadedReplies = replies.value.reduce((sum, reply) => sum + 1 + reply.numReplies, 0)
+  const count = props.comment.numReplies - numLoadedReplies
+  return t('Comments.Show More Replies') + t('Global.Counts.Replies Remaining', { count }, count)
+})
+
+const toggleCommentRepliesText = computed(() => {
+  const { channelName, comment: { hasOwnerReplied, numReplies: replyCount } } = props
+  if (showReplies.value) {
+    return t('Comments.Hide Replies', { replyCount }, replyCount)
```
PikachuEXE: If the intention is to NOT show reply count any more, why pass `replyCount`

static/locales/ko.yaml
```diff
@@ -453,7 +454,7 @@ Channel:
   Subscribe: '구독'
   Unsubscribe: '구독 취소'
   Channel has been removed from your subscriptions: '채널이 구독에서 제거되었습니다'
-  Removed subscription from {count} other channel(s): '{count} 다른 채널에서 구독을 제거했습니다'
+  Removed subscription from {count} other channel(s): '{count}개의 다른 채널에서 구독을 제거했습니다'
```
PikachuEXE: Again how do you get this...

PikachuEXE (changes requested):

static/locales/es.yaml
CL-Jeremy: I happen to speak many languages, some better than others. :)

I did some translations just by myself (German, French, Japanese), a few others with extensive, technical dialogues asking for confirmations from AI/translators to my initial attempts (like settling on a verb instead of an adjective for Spanish since that fits this case better), and others through analogy (for example, I speak some Czech, and then looked into the correspondents matching the case, number and aspects in other Slavic languages – I even know there is room for improvement, since the plural system is three-tier, for Slovenian even 4-tier, but deem that out of scope of this PR). I have a degree specialized in linguistics.

This `{' '}` thing is used to escape space (U+0020) without relying on a non-breakable space (U+00A0). This is because parentheses in Chinese and Japanese are full-width and comes with spacing pre-attached, so I had to introduce this leading space and prevent it from being trimmed. I could have modified the `Show More Replies` string directly, but find this solution less invasive, as leaving this one out leaves the user with the bare text without a counter, which is better than just nothing.

src/renderer/components/FtComment/FtComment.vue
CL-Jeremy: An obvious oversight by me.

static/locales/ko.yaml
CL-Jeremy: An obvious typo here. The original is not entirely wrong, but not formal as in suitable for use in UI without the `개의`. Evidence could be found in the file by searching for this string.

Did the changes basically in one go along with the spacing corrections. Now that I think about it, if I understand the text correctly, it should be `다른 {count}개의` instead of `{count}개의 다른` if this is like "X additional channels". Currently this would mean "X different channels" with or without `개의`.

src/renderer/components/CommentSection/CommentSection.vue
```diff
@@ -851,34 +419,23 @@ async function getPostCommentsInvidious() {
...
-    if (continuation) {
-      replyTokens.set(comment.id, continuation)
-      comment.hasReplyToken = true
-    } else {
-      replyTokens.delete(comment.id)
-      comment.hasReplyToken = false
-    }
-
-    isLoading.value = false
+    return !props.isPostComments
```
Shadorc, line 437: (minor) it's easier to read a positive condition than a negative one 
Could be `return props.isPostComments ? await getInvidiousCommunityPostCommentReplies({ postId: props.id, replyToken, authorId: props.postAuthorId }) : await invidiousGetCommentReplies({ id: props.id, replyToken })`

static/locales/fr-FR.yaml
```diff
@@ -897,7 +898,7 @@ Comments:
   View {replyCount} replies: Voir 1 réponse | Voir {replyCount} réponses
   Subscribed: Abonné(e)
   There are no comments available for this post: Il n'y a pas de commentaire associé à ce message
-  Hide {replyCount} replies: Masquer 1 réponse | Masquer {replyCount} réponses
+  Hide Replies: Masquer réponses
```
Shadorc:
```suggestion
  Hide Replies: Masquer les réponses
```

Shadorc:
Tested with Invidious and Local API, seems to work fine

PikachuEXE:
I haven't tested this with IV yet coz my IV instance not working even updated to latest
But local one seems fine

src/renderer/components/CommentSection/CommentSection.vue
CL-Jeremy: Yeah this has not been consistent throughout the file, but it happens to be like this for every thin wrapper around two functions, and the other way around whenever the logic is more complex. I guess it’s more natural to have the heavier choice at the last place.

src/renderer/components/FtComment/FtComment.vue
```diff
@@ -0,0 +1,357 @@
...
+    type: String,
+    required: true
+  },
+  channelThumbnail: {
+    type: String,
+    required: true
+  },
+  autoloadThisReplyLevel: {
+    type: Boolean
+  },
```
absidue, line 205: Please add a default prop value like we do throughout the rest of the code base.

```suggestion
  autoloadThisReplyLevel: {
    type: Boolean
    default: false
  },
```

src/renderer/components/FtComment/FtComment.vue
```diff
@@ -0,0 +1,357 @@
...
+}
+
+const showMoreRepliesText = computed(() => {
+  const numLoadedReplies = replies.value.reduce((sum, reply) => sum + 1 + reply.numReplies, 0)
+  const count = props.comment.numReplies - numLoadedReplies
+  return t('Comments.Show More Replies') + t('Global.Counts.Replies Remaining', { count }, count)
+})
+
+const toggleCommentRepliesText = computed(() => {
+  const { channelName, comment: { hasOwnerReplied, numReplies: replyCount } } = props
```
absidue: The rest of the codebase doesn't use prop destructuring so please don't do it here either.

src/renderer/components/FtComment/FtComment.vue
```diff
@@ -0,0 +1,357 @@
...
+        :can-fallback-to-invidious="canFallbackToInvidious"
+        :get-invidious-comment-replies="getInvidiousCommentReplies"
+      />
+      <div
+        v-if="replyToken && !repliesLoading"
+        class="showMoreReplies"
+        role="button"
+        tabindex="0"
+        @click="getCommentReplies()"
+        @keydown.enter.space.prevent="getCommentReplies()"
```
absidue: It is better to reference the functions in the template directly when you can, as calling them in the template usually results in the Vue compiler wrapping them in an extra arrow function.
```suggestion
        @click="getCommentReplies"
        @keydown.enter.space.prevent="getCommentReplies"
```

absidue (commented):
Please check the layout on mobile/narrow window sizes, using indentation was fine before as comments only had one level of replies but I can imagine that there will be layout issues on deeply nested comment chains now.

Additionally please only add translations for languages that you, a human, can fluently speak/write (without translation tools). Translations are generally managed through weblate where volunteers that actually know the language will add proper translations.

absidue:
Additionally please update the pull request description to reflect the current state of this pull request (prune the outdated sections add and new information where relevant).

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
So the behavior i have seen follows YouTube behavior quite good. I tested with the largest edge case i could think of, A MrBeast video. In cases like that i feel like it would be really nice to have the thread lines. Would it maybe be possible to add those?

Shouldnt this say hide reply?

Image

Having the count is nice to see but idk if we really need to show it. No real opinion on whether is should stay or not just maybe worth to discuss

Image

Bug found when trying to load replies on a Post

https://github.com/user-attachments/assets/57c734e8-a7b0-4a09-addc-017f78607f68

src/renderer/components/CommentSection/CommentSection.vue
Damus765: YouTube.js v18.0.0 changed the no comment error message in https://github.com/LuanRT/YouTube.js/pull/1194 to `The comments page did not have any content`. This change breaks FreeTube's no comment detection. It can be reproduced with https://youtu.be/87DyyMV0kCY.

Shadorc: Good catch 
It's weird because I tested with a video without comments and it worked...

Damus765: Maybe this exception is only thrown when comments are disabled.

PikachuEXE:
I found that timestamps in comments (top level, replies not tested yet but probably the same) does not respond to clicks like before (jump player current time to clicked timestamp)
Those in description still works

Update 1: Found a timestamp in replies: https://youtu.be/t6MAZW-joCo

image

image

Update 2: Can be fixed by reconnecting the event handlers back
image

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Threadlines in FreeTube make it seems like that all parent comments are connected

Image

Make threadlines only connect with their childs and not other parents like YT
Image

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
Sorting options placement seem unexpected, idk if this happened before this PR though

https://github.com/user-attachments/assets/9e75ed3d-a6e4-4b21-97e7-29182e5c4ded

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

Damus765 (commented):
What was the issue with the reply count exceeding 1000? I can't quite tell from the video you linked.

CL-Jeremy:
> What was the issue with the reply count exceeding 1000? I can't quite tell from the video you linked.

Both Invidious and YouTube.js return count above 1000 rounded to the nearest 100 (just like number of likes, which is now parsed from text for YouTube.js to match Invidious, see updated screenshots in OP, 2400 was previously shown as 2.4K), so the counter would be inaccurate. YouTube doesn't officially support this case anyway (even considered as a glitch by many).

Do note that on youtube.com a (presumably) cached value is somehow returned directly for the UI element, which is not the same as the rounded value used by 3rd party APIs. No idea where to get that, nor is that value particularly useful for the vast majority of videos without this glitch.

src/renderer/components/FtComment/FtComment.vue
PikachuEXE: This seems still not updated

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

CL-Jeremy:
LOL so many of these things are only findable with `"checkJs": true` in `jsconfig.json`, but can't have that on all the time, way too much work.

absidue:
The jsdoc comments in the codebase currently only exist for the sake of auto-complete, not for proper type checking, so the fact that typechecking fails in a bunch of places is entirely expected. As this pull request is already quite large just with the code changes, could you please revert all unrelated type/jsdoc changes, if you want to submit them in a separate pull request I would be alright with that but I would prefer this current pull request not to be a catch-all for all things that you want to change about the codebase.

PikachuEXE (approved):
No issue found on daily use so far

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

absidue (approved):

PR #9571, opened 2026-08-06
Add disable hardware acceleration setting to experimental settings

### LINKED ISSUE #2296: [Feature Request]: Ability to toggle hardware acceleration off
### Guidelines

### Problem Description

my PC for whatever reason does not like hardware acceleration on some apps, freetube included, it will make the screen all glitchy and spastic, sometimes go fully black in the app. I know it's caused by hardware acceleration because my browser did the same thing on youtube's website until I disabled hardware acceleration in the settings.

### Proposed Solution

Simple toggle in the Player Settings

### Alternatives Considered

I could not find any other solution to this issue

### Issue Labels

### Additional Information

I have a 3060 GPU

(Not my post, just an example)
https://community.brave.com/t/hardware-acceleration-broken-youtube-videos-flickering-anti-aliasing-broken/319307

The app works fine on my laptop but does the above listed issues on my main PC.

r3dArch:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #2296 

## Description

This PR adds a "Disable Hardware Acceleration" setting to the experimental settings section of the settings page.

### Before
ft_before

### After
ft_after

## Testing

I tested that the disable hardware acceleration is actually working using the `electron.app.isHardwareAccelerationEnabled()` function which I temporarily added to main/index.js, but I couldn't find a way to determine if it is working using the electron JS console.

## Desktop

- **OS:** Arch Linux
- **OS Version:** Kernel 7.1.5-zen1-2-zen
- **FreeTube version:** 0.25.1

## Additional context

I based a lot of this code on the implementation of the Replace HTTP Cache setting.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -36,6 +36,9 @@ const IpcChannels = {
   GET_REPLACE_HTTP_CACHE: 'get-replace-http-cache',
   TOGGLE_REPLACE_HTTP_CACHE: 'toggle-replace-http-cache',
 
+  GET_DISABLE_HARDWARE_ACCELERATION: 'get-disable-hardware-acceleration',
+  TOGGLE_DISABLE_HARDWARE_ACCELERATION: 'toggle-disable-hardware-acceleration',
+
   PLAYER_CACHE_GET: 'player-cache-get',
   PLAYER_CACHE_SET: 'player-cache-set',
 
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -306,6 +306,12 @@ function runApp() {
     app.commandLine.appendSwitch('disable-http-cache')
   }
 
+  const DISABLE_HARDWARE_ACCELERATION_PATH = `${userDataPath}/experiment-disable-hardware-acceleration`
+  const disableHardwareAcceleration = existsSync(DISABLE_HARDWARE_ACCELERATION_PATH)
+  if (disableHardwareAcceleration) {
+    app.commandLine.appendSwitch('disable-gpu')
+  }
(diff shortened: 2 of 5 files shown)


PikachuEXE (approved):
I tested in dev only

Image

Image

efb4f5ff-1298-471a-8973-3d47447115dc:
Just wondering, why did you choose to place this in the Experimental Settings?

r3dArch:
It's mostly just because that is where the Replace HTTP Cache setting is, which I based most of my code on. Although it is something that most users probably shouldn't use since it will hurt performance. Looking through the settings headers, I think the only other spot that makes sense would be "General". Do you have somewhere else you'd prefer this setting to be?

Shadorc:
I tested in dev too and was able to confirm that the GPU is not used with PikachuEXE code + linux processus manager showing no GPU usage for electron when hardware acceleration is disabled in FreeTube settings.

Shadorc (approved):

absidue (approved):
Reminder that we should remember to ask everyone that has any graphical glitches in FreeTube to make sure that they have this setting disabled, as this is not a setup that most people should be using, for the obvious reason that one of the core features of FreeTube is playing video and that the software rendering pipeline is considered a fallback for machines and environments that don't have a GPU, so if it works worse that is to be expected.

Approving under the condition that this **never** becomes a properly supported setting/never moves out of the experimental settings section.

Issue #9572, opened 2026-08-07
[Feature Request]: Bring back Download capability, but only as a frontend for yt-dlp


KerballOne:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Problem Description

> The built-in download feature has been removed. This functionality was already partially broken and became completely unusable following the introduction of SABR. In its existing state, it lacked many features users typically expect from a YouTube downloader, and restoring it would have required a significant rewrite just to bring it back to a minimal, maintenance-level condition. Given the scope of work required and the availability of more robust, dedicated tools such as [yt-dlp](https://github.com/yt-dlp/yt-dlp) and various online download services, we determined that removing the built-in downloader was the most practical decision.

### Proposed Solution

Just like PipePipe beta 5.0 did when it got tedious keeping up with YT SABR for downloads.... when they told users to install a 3rd party app that manages yt-dlp (Seal), Freetube should bring back the Download button, that does 1 of 2 things.  
1)  If yt-dlp is not installed, inform the user to install it.
2)  If yt-dlp is installed, call it directly.

This keeps this dev team from chasing the issues and doing a lot of work for limited returns (the main concern that removed the Downloading function).
Also this lets the user know that yt-dlp is the downloader program and they should look to them for **updates** and **issues**.  This takes responsibility off the FreeTube team.

This provides users with a GUI frontend that is easy to use and familiar so they don't have to use online services which can be suspicious, or deal with command line.

### Alternatives Considered

Users would need to find a good yt-dlp frontend that is easy to use, but need to cut and paste URLs from Freetube.  Online services exists but can be shady and host malvertising.

### Issue Labels

new feature, support for external software

### FreeTube Version

0.25

### Operating System Version

Windows 11

### Installation Method

.exe

Issue #8966, opened 2026-04-04
[Feature Request]: Downloading via yt-dlp


truelockmc:
### Guidelines

- [x] I have [searched the issue tracker for open and closed issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the feature request I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [x] This issue contains only one feature request.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Problem Description

Downloading has been removed because it was to difficult too maintain, partially broken and there are better tools for doing that, like yt-dlp.

It would be really cool tho if i could download videos to watch them offline, In-App, keeping Progress tracking and a nice UI.

### Proposed Solution

The download button should be turned off by default.
You would have to enable it in settings and set a yt-dlp executable path there.
(That way it would also be a Feature that's more optional and doesn't have to be in the focus of development)

When enabled you would get a download button under videos and another section in the sidebar named Downloads where you could manage and watch your downloaded contents.

Downloads should happen via yt-dlp and download progress tracking could be realized with command line parsing.

I am unsure about the implementation of the In-App Player tho, i don't know if it would be better if you could only watch downloaded content via the download section or also while browsing normally.
But the watch Progress should definitely be tracked somehow.

### Alternatives Considered

Alternatively you could also download via yt-dlp and then only offer the option to open it in an external player instead of In-App. 
But rather than doing that i would just leave the Feature out.

### Issue Labels

new feature

### FreeTube Version

v.0.24.0

### Operating System Version

Debian Trixie

### Installation Method

.AppImage

Issue #4178, opened 2023-10-17
[Feature Request]: Add yt-dlp intergration


KevinReems:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

Currently you can only download a limited selection of video qualities with audio included.

### Proposed Solution

yt-dlp is capable of downloading and then muxing high resolution video with high bitrate audio streams.

For example: yt-dlp.exe -f bestvideo[height<=1080]+bestaudio will output a 1080p files with the best available audio stream available to download. Since nothing is actually encoded, this process is extremely fast.

A few menu options could be added to Settings to specify the path of yt-dlp, output directory, desired/maximum resolution etc.

### Alternatives Considered

There may be other command line downloaders but since yt-dlp seems to be the clear winner it would be best to start there.

### Issue Labels

support for external software

absidue:
To avoid us having to implement loads of individual settings, especially considering the amount of possible options even just for yt-dlp's `-f` flag, I would suggest that this be implemented similar to the external player settings. One field to specify the path to yt-dlp, another field to set the command line options for yt-dlp, a link to yt-dlp's README so people can find out what options they can specify and we can reuse the existing download folder setting.

FreeTube would then call yt-dlp like this: `[path to binary] [your specified commandline options] https://www.youtube.com?v=[video id]`.

Aphexus:
I'd only say it should be more general. Instead of working it as if one must yt-dlp, just call it "external downloader" or "custom downloader". One should be able to wrap any calls to yt-dlp in a batch file as long as they know the link. If batch downloading is added(such as downloading a watch list) then either this could be iterated or an additional parameter can be passed that signals it is a list(since yt-dlp allows for lists).

E.g., if one has a text field then one can just type in the command to execute: "C:\yt-dlp.exe -f best %1" or whatever and all freetube has to do is scan for the tokens to replace(e..g, %1 or maybe <url>).


[...]

PR #9588, opened 2026-08-07
[Feature Request]: Add yt-dlp intergration

KerballOne:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/4178 https://github.com/FreeTubeApp/FreeTube/issues/8966 https://github.com/FreeTubeApp/FreeTube/issues/9572 https://github.com/FreeTubeApp/FreeTube/pull/8543

## Description

This adds the feature request from #4178 which was closed when it was decided to remove the built in download capability.  But the feature request is now needed MORE because the native download feature is gone.

This implementation ~~follows~~ was initially based on @absidue's last comment on the subject https://github.com/FreeTubeApp/FreeTube/issues/4178#issuecomment-2325219214

## Screenshots 

image
Screenshot 2026-08-10 133938
Screenshot 2026-08-10 133947
image

## Testing

- If yt-dlp is not installed, pressing a button action will show a toast error and clicking that will take you to the new settings page.  

Configure:
1. Go to Settings > External Downloader
2. Enable Downloads
3. (optional) Click the Download yt-dlp button
4. (if not detected in $PATH) Choose the executable file
5. (optional) Set an output directory
6. (optional) Set custom arguments (blank will work with yt-dlp defaults)

Download:
1. Click the Download button in the Player window
2. (optional) Choose a time span option
3. Click Video or Audio button
4. (if output directory not set) Choose a Download directory from the popup
5. View the terminal output for any errors or warnings
6. Go to the Output folder to check the file that was created

## Desktop

- **OS:** Windows
- **OS Version:** Windows 11
- **FreeTube version:** v0.25.2 Beta

## Additional context

Should work for MacOS or Linux builds.  The executable chooser button should filter on exe only on Windows.  And the arguments for yt-dlp should be the same. Although not tested on any thing other than Windows 11.
Update:  Tested on Fedora 43 (xfce) too.


src/renderer/components/ExternalDownloaderSettings.vue
```diff
@@ -0,0 +1,202 @@
...
+        :tooltip="t('Tooltips.External Downloader Settings.yt-dlp Executable Path')"
+        @input="updateYtdlpExecutable"
+      />
+      <div class="buttonPair">
+        <FtButton
+          class="centerButton"
+          :label="t('Settings.External Downloader Settings.Choose Executable')"
+          @click="chooseYtdlpExecutable"
+        />
+        <FtButton
```
radmorecameron: I don't think we should include a download button, I think telling people how to setup yt-dlp belongs in our docs (or it should be a link instead)

src/renderer/components/ExternalDownloaderSettings.vue
```diff
@@ -0,0 +1,202 @@
...
+      <FtInput
+        :placeholder="t('Settings.External Downloader Settings.yt-dlp Executable Path')"
+        :show-action-button="false"
+        :show-label="true"
+        :value="ytdlpExecutable"
+        :tooltip="t('Tooltips.External Downloader Settings.yt-dlp Executable Path')"
+        @input="updateYtdlpExecutable"
+      />
+      <div class="buttonPair">
+        <FtButton
```
radmorecameron: For Linux users, this is a bit useless. (If they install yt-dlp with their package manager, it'll be in their bin folder).

I think a validate button that calls `yt-dlp` with `--version` would be more useful

src/renderer/components/ExternalDownloaderSettings.vue
```diff
@@ -0,0 +1,202 @@
...
+        :tooltip="t('Tooltips.External Downloader Settings.ffmpeg Executable Path')"
+        @input="updateFfmpegExecutable"
+      />
+      <div class="buttonPair">
+        <FtButton
+          class="centerButton"
+          :label="t('Settings.External Downloader Settings.Choose Executable')"
+          @click="chooseFfmpegExecutable"
+        />
+        <FtButton
```
radmorecameron: Ffmpeg is installed by default on most linux distros. We can include a guide in docs for windows/mac users if needed or we can use a link instead.

src/renderer/components/ExternalDownloaderSettings.vue
KerballOne: Linux users are indeed at an advantage because command line programs are nearly always installed in a proper $PATH.  So for them, it is a redundant field.  
A validation feature would be good, and could just populate the field with the binary using `which` and `--version` without needing the user to specify/choose the executable.

src/renderer/components/ExternalDownloaderSettings.vue
KerballOne: Like the yt-dlp executable, the ffmpeg is not always just there, and for licensing, not bundled with yt-dlp.  
In Windows, they are installed separately and neither are included in $PATH.  And even the official installation instructions don't walk the user through that step.
So maybe a validation step for both binaries so if found in $PATH, it will auto populate the field.

src/renderer/components/ExternalDownloaderSettings.vue
KerballOne: Fixed in latest [6e376f commit](https://github.com/FreeTubeApp/FreeTube/commit/6e376f0f4e7b8c916021c3cf6c97d4ed276cd3ab)
Kept the button to link to the yt-dlp readme though, may be replaced with later documentation if you'd like.

src/renderer/components/ExternalDownloaderSettings.vue
KerballOne: Fixed in commit
[6e54b2a](https://github.com/FreeTubeApp/FreeTube/commit/6e54b2aef6e71a5e770aece71c871ef6ebcbd23b)

src/renderer/components/ExternalDownloaderSettings.vue
KerballOne: Fixed in commit
[6e54b2a](https://github.com/FreeTubeApp/FreeTube/commit/6e54b2aef6e71a5e770aece71c871ef6ebcbd23b)

PikachuEXE:
Download dropdown looks strange when there is nothing displayed (just buttons on the left
Inputs of timestamps doesn't look like inputs
Pressing buttons with invalid settings should only get a toast which redirects to settings page on press instead of redirecting users right away
I can download without setting download folder? (It downloads to my FT folder

My custom build: https://github.com/PikachuEXE/FreeTube/actions/runs/31294106398
(I am using `--merge-output-format mp4 --remux-video mp4 -S vcodec:av01:h264,lang,quality,res,fps,hdr:12,acodec:mp4a` for mp4 download)

Can we add a button to list formats and another to allow custom format download (or maybe another PR for that I guess?

KerballOne:
> Inputs of timestamps doesn't look like inputs

Fixed. 
image

> Pressing buttons with invalid settings should only get a toast which redirects to settings page on press instead of redirecting users right away

Done.
image

> I can download without setting download folder? (It downloads to my FT folder

Fixed.  Now the output directory is auto-populated with the user's default Downloads folder.  Still allows user to change.

[4ba728d](https://github.com/FreeTubeApp/FreeTube/commit/4ba728d2c75f443c510eaa4371455905f20c446f)

----

> Download dropdown looks strange when there is nothing displayed (just buttons on the left

Could not replicate.  At what point is there nothing displayed?

> Can we add a button to list formats and another to allow custom format download (or maybe another PR for that I guess?

I thought the objective was to keep options in FT to the minimum needed to just be a frontend wrapper for yt-dlp?  So FT doesn't need to keep up with changes or maintain a list of options?  The defaults just work, but users can go to the yt-dlp README for any and all guidance on custom formats, bitrates, and other options.

absidue (changes requested):
> Now the output directory is auto-populated with the user's default Downloads folder. Still allows user to change.

We cannot use a hardcoded path like that, we always need to prompt the user as in flatpak builds for example the app is only granted access to paths/folders that the user selected in a file picker. You can follow what we do for the screenshot folder (if it is unset or the folder does not exist or freetube doesn't have write access when the user takes their first screenshot we prompt them to pick a new folder).


>> Can we add a button to list formats and another to allow custom format download (or maybe another PR for that I guess?

> I thought the objective was to keep options in FT to the minimum needed to just be a frontend wrapper for yt-dlp? So FT doesn't need to keep up with changes or maintain a list of options? The defaults just work, but users can go to the yt-dlp README for any and all guidance on custom formats, bitrates, and other options.

This pull request already does more that I am personally comfortable merging with the separate audio and video buttons, the timestamp selector, download folder management and ffmpeg management. Part of the reason we got rid of the downloader was that I didn't want to maintain it and consider it out of scope for FreeTube, not just the calls to YouTube but everything around it.
My original suggestion, which you claimed this pull request was based on, was for a single download button on the watch page and just fields for the yt-dlp executable path and args in the settings. FreeTube would then just call the yt-dlp executable with the users args + the video URL and unref the child process so FreeTube is no longer responsible for it. If the user wants to do any customisations they should add their custom args to their yt-dlp config file or in that args field in the FreeTube settings.

efb4f5ff-1298-471a-8973-3d47447115dc:
I agree with absidue, the scope of this PR has been expanded too much. I would be comfortable with the original suggestion 

> was for a single download button on the watch page and just fields for the yt-dlp executable path and args in the settings. FreeTube would then just call the yt-dlp executable with the users args + the video URL and unref the child process so FreeTube is no longer responsible for it. If the user wants to do any customisations they should add their custom args to their yt-dlp config file or in that args field in the FreeTube settings.

PikachuEXE:
> Download dropdown looks strange when there is nothing displayed (just buttons on the left

Sorry, I mean the button placement is off to the left side fo no reason
image

e.g. Share dropdown got centered buttons
image

absidue:
@PikachuEXE That dropdown shouldn't even exist, so layout issues inside it are irrelevant to getting this pull request into a mergable state.

KerballOne:
> I mean the button placement is off to the left side for no reason

Fixed.  See above screenshot in OP.

> We cannot use a hardcoded path like that, we always need to prompt the user as in flatpak builds for example the app is only granted access to paths/folders that the user selected in a file picker. You can follow what we do for the screenshot folder (if it is unset or the folder does not exist or freetube doesn't have write access when the user takes their first screenshot we prompt them to pick a new folder).

Fixed.  It now behaves similar to the Screenshots feature, complete with write perm check.

> This pull request already does more that I am personally comfortable merging with the separate audio and video buttons, the timestamp selector, download folder management and ffmpeg management. Part of the reason we got rid of the downloader was that I didn't want to maintain it and consider it out of scope for FreeTube, not just the calls to YouTube but everything around it.
> My original suggestion, which you claimed this pull request was based on, was for a single download button on the watch page and just fields for the yt-dlp executable path and args in the settings. FreeTube would then just call the yt-dlp executable with the users args + the video URL and unref the child process so FreeTube is no longer responsible for it. If the user wants to do any customisations they should add their custom args to their yt-dlp config file or in that args field in the FreeTube settings.


Acknowledged.  I did originally attempt a barebones implementation based on your original comment.   But that can actually be mostly done already by just overriding any option with the yt-dlp binary.  No PR needed.
Screenshot 2026-08-10 110832
But... the user experience is extremely poor and would not handle the common use cases.  
 
Unfortunately, the user expectations for FreeTube calling an external player, is pretty different than a downloader program like yt-dlp.  
For a player, it is fine to only pass along the video id and the current timestamp.  
For a downloader, the user will often need different options depending on the video.  A separate button for audio is a must since getting songs is a common use case.  And the time span option is critical for users who can't be expected to download entire hours long, gigabytes of video just to get a short snippet.  These are not needed for an external player of course.  

For a proper downloader integration, it's fine for the vast majority of options like file format, encoding, bitrate, etc... to be defined as Global arguments in the settings, the user won't likely need to change them in-between videos.
But per-video dropdown options for audio and time span are needed to meet the minimum viable expected behavior.  

I removed the ffmpeg handling entirely, since you're right, it would add unnecessary management that yt-dlp already does and shows the user if needed.  What remains, are the dropdown options for timespan and audio/video buttons, which doesn't create additional maintenance since it still only passes different arguments when calling the yt-dlp executable.  Fire and forget similar to the exec + args passed for named External Players.  And the UI elements of the dropdown are similar to the Share Dropdown.

Shadorc:
This is just my opinion but it seems a bit overscoped for FreeTube. 
If the issue is that using yt-dlp through CLI is intimidating, there seems to be some pretty good cross-platform GUI for it. If the issue is that it is a few more clicks, I don't think it's worth adding a thousand lines that would need maintainance and that would probably lead to more feature requests.

KerballOne:
> This is just my opinion but it seems a bit overscoped for FreeTube. If the issue is that using yt-dlp through CLI is intimidating, there seems to be some pretty good cross-platform GUI for it.

I would still love a suggestion for a well-maintained, safe yt-dlp GUI frontend that works cross platform.  So for, there is no good candidate that anyone has actually recommended.  I've seen shady ones though.

> If the issue is that it is a few more clicks, I don't think it's worth adding a thousand lines that would need maintainance and that would probably lead to more feature requests.

Point taken, and I have cleaned up some of the lines that may make this PR seem excessive or somehow in need of more care and maintenance than other existing features such as the external players.  This PR adds all the features that were lost with the removal of the built in downloader (#8543), and even adds some common-sense ones that might have been asked for, so very little risk for future feature requests.  Plus, with the included command line argument fields, any additional features won't be for this project, as the user can utilize the robust and feature reach yt-dlp which already has more features than Freetube ever did.

Either way, total lines of code, IMO, are not the best way to determine code maintainability.  The core logic is only 17 lines.  The vast majority of the PR's code is in UI, to make it a good user experience and ensuring the user can do any custom download action in an intuitive way.  Not very fragile, nor a significant maintenance issue.  For comparison, #8543 had less than half the total lines of code as this PR... but with almost 100 lines of critical logic, that feature broke often with YT changes.  Because this PR relies on the robustness of yt-dlp, it need only call it rather than keep up with YT changes. 
A better contrast/compare is the External Players feature.  That needs to maintain data for not one, but a dozen external programs and their cli arguments.  yt-dlp has been very consistent with not breaking backward compatibility over the years of updates.  So it is very safe to have code rely on the few hard-coded args here.  Users can still utilize the full power of yt-dlp, but also the great UX that Freetube provides.  Which is also a great example of why using Freetube as a GUI frontend to initiate downloads is superior to any additional yt-dlp GUI program, or telling non-techy users to learn the command line.

PikachuEXE:
Using https://github.com/ErrorFlynn/ytdlp-interface on windows
macOS & linux = no idea

Issue #9545, opened 2026-07-31
[Bug]: Livestream fails to play due to YouTube enforcing SABR


efb4f5ff-1298-471a-8973-3d47447115dc:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads). **The only official downloads are from the [https://freetubeapp.io](https://freetubeapp.io/#download) website, the [FreeTubeApp/FreeTube GitHub repository](https://github.com/FreeTubeApp/FreeTube/releases) and the [io.freetubeapp.FreeTube flatpak](https://flathub.org/en/apps/io.freetubeapp.FreeTube) from Flathub.**
- [x] I have [searched the issue tracker for **open and closed** issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

Its possible that you need to a few tries to repro this

1. Go to https://youtu.be/X4VbdwhkE10
2. See that player is stuck and throws `TypeError: Cannot read properties of null (reading 'match')`

### Expected Behavior

Stream plays and doesnt throw this error

### Issue Labels

inconsistent behavior

### FreeTube Version

v0.25.1-nightly-7565 Beta

### Operating System Version

Windows 11 IoT Enterprise LTSC 24H2

### Installation Method

.exe

### Primary API used

Local API

### Nightly Build

- [x] I have encountered this bug in the latest **official** [nightly build](https://docs.freetubeapp.io/development/nightly-builds), do not check this box if you are using an unofficial or custom build.

Shadorc:
When it happens, the `manifestUrl` [is set to](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/views/Watch/Watch.js#L758) `hls_manifest_url` but it is undefined.
It then throws [down the line](https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js#L592)

[...]

PR #9607, opened 2026-08-10
 Retrieve potoken challenge data and yt config from the YouTube website HTML

### LINKED ISSUE #9526: [Bug]: Video Freeze around 1 min mark / Reloading player according to SABR request
### Guidelines

### Describe the bug

1. On opening a new video(In this case https://youtu.be/YGwDeY1t6gQ), the video runs until 1.03 or 1,04 and then stops indefinitely. Does not go in SABR reload or buffer futher content, it just stops

Image

2. Noticed it on few other videos as well but the issue is intermittent and does not seem to be related to the video length at all.
3. Manually selecting a later point on the progress bar forces the player into a reload and the video plays normally in its entirety.

Image

    

### Expected Behavior

After any SABR behaviour at the start, the video is expected to play normally or perform a SABR reload and then continue to progress automatically

### Issue Labels

content not loading

### FreeTube Version

v0.25.1 Beta

### Operating System Version

Windows 11 Pro 24H2

### Installation Method

.exe

### Primary API used

Local API

_No response_

_No response_

### Additional Information

_No response_

### Nightly Build

absidue:
## Pull Request Type

- [x] Bugfix

## Related issue

- closes #9526

## Description

**Massive thanks to @LuanRT for helping figure out what YouTube was checking for.** Luan's summary of YouTube's changes is here: https://github.com/LuanRT/BgUtils/pull/44

BotGuard now checks for and validates the `EVENT_ID` field in the config data embedded in the HTML together with the embedded `/att/get` response or doing your own `/att/get` request using the `eacrToken` that is embedded in the page. This pull request implements the extraction of all relevant data from the HTML watch page, so in most cases it should just require a single request now (plus player base.js donwload, with fallbacks to Innertube if the /player or /next responses fail to extract (in case YouTube changes their page and those regexes need changing), it also falls back to calling /att/get with the eacrToken from the HTML page if YouTube returns one of their no-op inline JavaScript interpreters in the page instead of an interpreter URL.

YouTube is more trigger happy with captchas on the HTML pages compared to Innertube so people that were already having playback issues before the PO token changes or are using sketchy/flagged IP addresses will likely have issues, but as we haven't found a way to get the EVENT_ID and eacrToken without extracting them from the HTML, this is our only option to get working PO token generation for the WEB client at the moment.

## Testing

Check that videos play again (live streams are hit and miss, the same as before, because of youtube doesn't always return the HLS URLs #9545).

## Desktop

- **OS:** Windows
- **OS Version:** 11


src/renderer/helpers/api/local.js
```diff
@@ -391,54 +522,43 @@ export async function getLocalSearchContinuation(continuationData) {
  * }>}
  */
 export async function getLocalVideoInfo(id) {
+  // This is a false positive, the variable is definitely used below
```
Shadorc: This has nothing to do with the general purpose of this MR, but I'm curious about this line
I may be missing a typescript behaviour that I don't know about but from my understanding `responseTime` first assignement is indeed useless, as the warning says, because it will be re-assigned before being read either line 538 or line 553. 
Also, `fetchFunction` is only called line 552, so the new assignement line 553 will also overwite the value assigned line 538, no matter if it had early exited or not.

But I may be confused
Anyway, congrats for your (and Luan) work!

PikachuEXE (dismissed):

src/renderer/helpers/api/local.js
absidue: Fixed!

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

kommunarr (approved):

PR #9608, opened 2026-08-10
Feat: Add message on top and at the end of playlist to explain hidden videos

### LINKED ISSUE #9569: [Feature Request]: Indicate when a playlist contains unavailable videos
### Guidelines

### Problem Description

When searching or when opening a playlist, the video count displays the total number of videos in the playlist, even the ones that are unavailable.
This leads to a confusing behaviour: when the playlist is fully loaded, the last index is less than the video count.

YouTube has the same behaviour, but it adds a message at the top of the playlist (rough translation from french, but that should be something like this)
> Videos that are not available have been hidden

Example: https://youtube.com/playlist?list=PLvMVt4c68EeXRV2R1fdieei9HwpuUKXKL 313 videos in total, 247 playable

Image

### Proposed Solution

Add a message, similar to what YouTube does, indicating that some videos have been hidden from the playlist because they are unavailable.

### Alternatives Considered

I don't see any correct alternative other than not changing the current behaviour.
An alternative could be to change the video count to display the available count, but that would deviate from YouTube's behavior and would be confusing.

### Issue Labels

### FreeTube Version

v0.25.1-beta

### Operating System Version

Bazzite

### Installation Method

Flathub

_No response_

### Additional Information

_No response_

Shadorc:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/9569

## Description

**Add a message at the top of the playlist vue to indicate that unavailable videos have been hidden.**
- The detection is done by checking if the "Show unavailable videos" item menu exists in youtube.js response.
image

- This change cannot be implemented for Invidious API because there is no way to determine if videos are hidden in advance without loading the full playlist.
- I've based the UI on `FtNotificationBanner`, but I'm not entirely satisfied with the background color. I've tried various existing theme colors but didn't find something better.
- I've named the UI 'alert' because the corresponding YouTube's component is named `ytd-alert-with-button-renderer`.

**Add a message at the end of the watch video playlist vue to indicate that unavailable videos have been hidden.** 

## Screenshots 

**Local API**
image

image

**Invidious API**
image

## Testing

**Message at the top of the playlist** 
1. Go to `Settings > Preferred API Backend` and chose `Local API`
2. Go to a playlist with unavailable videos (e.g. https://youtube.com/playlist?list=PLvMVt4c68EeXRV2R1fdieei9HwpuUKXKL)
3. Check that a message is displayed
4. Check that the message UI fits well with various resolution (including mobile)
5. Click on the close button
6. Check that the message is removed

**Regressions:**
- Check that the message does not appear in a playlist with no unavailable videos 
- Check that the message does not appear with Invidious selected as `Preferred Backend API`

**Message at the end of the watch video playlist**
1. Go to a playlist with unavailable videos (e.g. https://youtube.com/playlist?list=PLvMVt4c68EeXRV2R1fdieei9HwpuUKXKL)
2. Click on a video
3. Scroll at the bottom of the playlist in the watch page
4. Check that a message is displayed
5. Check that the message UI fits well with various resolution (including mobile)

**Regressions:**
- Check that the message does not appear in a playlist with no unavailable videos 

## Desktop

- **OS:** Bazzite
- **OS Version:**

### DIFF
--- a/src/renderer/views/Playlist/Playlist.vue
+++ b/src/renderer/views/Playlist/Playlist.vue
@@ -55,6 +55,25 @@
       v-if="!isLoading"
       class="playlistItemsCard"
     >
+      <FtFlexBox
+        v-if="showUnavailableVideosAlert"
+        class="alertBox"
+      >
+        <p class="alertLabel">
+          {{ t("Playlist.Unavailable videos are hidden") }}
+        </p>
+        <button
+          class="alertButton"
+          :aria-label="t('Close')"
+          :title="t('Close')"
+          @click="handleCloseAlert"
+        >
+          <FontAwesomeIcon
+            class="alertButtonIcon"
+            :icon="['fas', 'times-circle']"
+          />
+        </button>
+      </FtFlexBox>
       <template
         v-if="shownPlaylistItems.length > 0"
       >
@@ -229,6 +248,7 @@ const channelName = ref('')
 const channelThumbnail = ref('')
 const channelId = ref('')
 const infoSource = ref('local')
+const showUnavailableVideosAlert = ref(false)
 const playlistItems = ref([])
 /** @type {import('vue').ComputedRef<any[] | null>} */
 const tempShownPlaylistItems = ref(null)
@@ -466,6 +486,7 @@ function resetState() {
   channelThumbnail.value = ''
   channelId.value = ''
   infoSource.value = 'local'
+  showUnavailableVideosAlert.value = false
   playlistItems.value = []
   continuationData.value = null
 }
@@ -501,6 +522,7 @@ async function getPlaylistLocal() {
     channelThumbnail.value = result.info.author?.best_thumbnail?.url ?? ''
     channelId.value = result.info.author?.id
     infoSource.value = 'local'
+    showUnavailableVideosAlert.value = result.menu?.items?.some((item) => item.text === 'Show unavailable videos')
 
     store.dispatch('updateSubscriptionDetails', {
       channelThumbnailUrl: channelThumbnail.value,
@@ -1082,6 +1104,10 @@ if (isUserPlaylistRequested.value && searchQueryTextPresent.value) {
 
 function handleResize() {
   forceListView.value = window.innerWidth <= MOBILE_WIDTH_THRESHOLD || window.innerHeight <= PLAYLIST_HEIGHT_FORCE_LIST_THRESHOLD
+}
+
+function handleCloseAlert() {
+  showUnavailableVideosAlert.value = false
 }
 
 onMounted(() => {
--- a/src/renderer/views/Playlist/Playlist.vue
+++ b/src/renderer/views/Playlist/Playlist.vue
@@ -69,8 +69,7 @@
           @click="handleCloseAlert"
         >
           <FontAwesomeIcon
-            class="alertButtonIcon"
-            :icon="['fas', 'times-circle']"
+            :icon="['fas', 'times']"
           />
         </button>
       </FtFlexBox>
--- a/src/renderer/views/Playlist/Playlist.vue
+++ b/src/renderer/views/Playlist/Playlist.vue
@@ -1124,7 +1124,7 @@ onBeforeRouteLeave((to) => {
     store.commit('setCachedPlaylist', {
       id: playlistId.value,
       title: playlistTitle.value,
-      totalVideoCount: videoCount,
+      totalVideoCount: videoCount.value,
       channelName: channelName.value,
       channelId: channelId.value,
       items: sortedPlaylistItems.value,
--- a/src/renderer/views/Playlist/Playlist.vue
+++ b/src/renderer/views/Playlist/Playlist.vue
@@ -1125,6 +1125,7 @@ onBeforeRouteLeave((to) => {
     store.commit('setCachedPlaylist', {
       id: playlistId.value,
       title: playlistTitle.value,
+      totalVideoCount: videoCount,
       channelName: channelName.value,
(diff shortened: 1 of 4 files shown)


Shadorc:
Marked as draft because @efb4f5ff-1298-471a-8973-3d47447115dc would like to expand the feature request

efb4f5ff-1298-471a-8973-3d47447115dc:
Scope is expanded to displaying a message on how many videos are being hidden when user is on the watch page. This is being displayed at the bottom of the playlist

vlc_j2I0q3ZCba

src/renderer/views/Playlist/Playlist.vue
```diff
@@ -55,6 +55,25 @@
...
+        </p>
+        <button
+          class="alertButton"
+          :aria-label="t('Close')"
+          :title="t('Close')"
+          @click="handleCloseAlert"
+        >
+          <FontAwesomeIcon
+            class="alertButtonIcon"
+            :icon="['fas', 'times-circle']"
```
efb4f5ff-1298-471a-8973-3d47447115dc: To make it same as what we use in our other banner and closer to what youtube uses

```suggestion
            :icon="['fas', 'times']"
```

efb4f5ff-1298-471a-8973-3d47447115dc:
> but I'm not entirely satisfied with the background color. I've tried various existing theme colors but didn't find something better

As for this i currently dont have any ideass

PikachuEXE:
I guess you are using the same bg color as tags now? (didn't check code)
Maybe those colors should be improved first... (tags don't look good in dark theme, in black theme it's almost unable to know they are tags

image

However I dunno if tags and "secondary banner" should use the same BG color :P

Shadorc:
I'm using `var(--secondary-card-bg-color)` which is the same as tags, yes. 
And I feel like there should be a new background color for this banner but it feels like a lot of work for this small addition. Also, I would be clueless as to what color to chose for each themes. 
Maybe I could simply remove the background color, and slightly change the UI to have the close button at the left of the text, close to it...
UI is not really my cup of tea 

Edit: I've changed to use `bg-color`, which is now the same as the `watched` background

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

PikachuEXE (dismissed):

src/renderer/views/Playlist/Playlist.vue
```diff
@@ -1099,6 +1124,7 @@ onBeforeRouteLeave((to) => {
     store.commit('setCachedPlaylist', {
       id: playlistId.value,
       title: playlistTitle.value,
+      totalVideoCount: videoCount,
```
caetano-dev:
```suggestion
      totalVideoCount: videoCount.value,
```

src/renderer/views/Playlist/Playlist.vue
```diff
@@ -501,6 +521,7 @@ async function getPlaylistLocal() {
     channelThumbnail.value = result.info.author?.best_thumbnail?.url ?? ''
     channelId.value = result.info.author?.id
     infoSource.value = 'local'
+    showUnavailableVideosAlert.value = result.menu?.items?.some((item) => item.text === 'Show unavailable videos')
```
caetano-dev, line 524: Comparing against the string seems a bit brittle. Is there a more reliable way to check this? This will break if this string ever changes.

Maybe we could do something like `(item) => item.icon_type === 'VISIBILITY' || item.text === 'Show unavailable videos'`? I still have not tested if it works, so I am putting it here as an example.

caetano-dev (changes requested):

src/renderer/views/Playlist/Playlist.vue
Shadorc: Good catch!
I was wondering how come nothing broke, but it is working with this bug. The value is retrieved correctly even though the type isn't right.

src/renderer/views/Playlist/Playlist.vue
Shadorc: I agree that relying on the string is brittle, but a lot of the features rely on these. I don't think we have better options... 
Using `item.icon_type === 'VISIBILITY'` could also cause issue in a scenario where YouTube adds a new item that also has an `icon_type` sets to 'VISIBILITY', all playlists would be considered has having unavailable videos

Shadorc: The `navigationEndpoint` probably contains an information that is a more reliable indicator, but I don't know what it could be. 

```
{
    "type": "MenuNavigationItem",
    "text": "Show unavailable videos",
    "icon_type": "VISIBILITY",
    "endpoint": {
        "type": "NavigationEndpoint",
        "name": "browseEndpoint",
        "payload": {
            "browseId": "VLPLvMVt4c68EeXRV2R1fdieei9HwpuUKXKL",
            "params": "wgYCCAA%3D",
            "nofollow": true,
            "navigationType": "BROWSE_NAVIGATION_TYPE_LOAD_IN_PLACE"
        },
        "metadata": {
            "url": "/playlist?list=PLvMVt4c68EeXRV2R1fdieei9HwpuUKXKL",
            "page_type": "WEB_PAGE_TYPE_PLAYLIST",
            "api_url": "browse"
        },
        "command": {
            "type": "BrowseEndpoint"
        }
    }
}
```

caetano-dev (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

PikachuEXE (approved):

Issue #4291, opened 2023-11-07
[Feature Request]: Select default audio track/language


notpointify:
### Guidelines

- [X] I have searched the issue tracker for [open](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aopen+is%3Aissue) and [closed](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+is%3Aclosed) issues that are similar to the feature request I want to file, without success.
- [X] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the feature request I want to file, without success.
- [X] This issue contains only one feature request.

### Problem Description

While FreeTube does support multiple audio tracks nowadays for videos that have them, currently it defaults to the original language of the video. So if I prefer watching foreign videos in another language, e.g. watching English videos dubbed in Spanish, then I have to manually select the Spanish audio for each video I wanna watch.

### Proposed Solution

Add an option, under either "General Settings" or "Player Settings", to select the default audio track for all videos (a specific language, or just the original video's language).

### Alternatives Considered

Another option could be to use the existing "Locale Preference" setting to select the default audio track instead of adding a new setting, though that might be less "flexible" for the end user.

### Issue Labels

ease of use improvement, improvement to existing feature

absidue:
How should that setting work when the original, dubbed and audio descriptions are all in the same language, like in this video?: https://youtu.be/Kn56bMZ9OE8
On that note audio descriptions exist, which some people would prefer, so I guess you'll need a second setting too.

Atrate:
Considering that Stable Volume appears on all (that I've watched) videos, it'd be nice if it was possible to set Stable Volume as the default.

santaclose:
I'm currently seeing that the original audio is not being chosen by default when watching mental outlaw. It defaults to German for some reason. I think the default should be the original audio? I'm using version v0.22.0 Beta

absidue:
@santaclose YouTube added support for auto-dubbed audio tracks shortly before we released 0.22.0 and unfortunately we only found out about them after the release, so we weren't able to add support for them in time. You'll be pleased to know that we've already added support for them in the development branch so the nightlies already support them and the next release will too :).

mooleshacat:
I just had the same issue watching bonehead truckers video ID: sBWqEuyKyhU

Oddly enough, Invidious will play the English audio, while FreeTube plays the Spanish audio.

I will try the nightly, thank you for your hard work 👍

[...]

PR #9642, opened 2026-08-13
feat: Add default audio selector

burikoni:
## Pull Request Type
- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
open #4291

## Description
Add an option to set the default audio language

## Screenshots
2026-08-12-233205_1407x651_scrot

## Testing
Select a language and open a video with multiple audio tracks. For example, one of MrBeast’s recent videos.

It was also tested with the Invidious API using a locally hosted server, since the online instances either don’t expose the API or are offline.

## Desktop

- **OS:** GNU/Linux
- **OS Version:** Linux debian 6.12.95+deb13-amd64
- **FreeTube version:** v0.25.2 Beta

## Additional context
The list of languages and their codes was obtained using the YouTube API through the [i18nLanguages.list ](https://developers.google.com/youtube/v3/docs/i18nLanguages/list) endpoint, and the values `zh-Hans` for Simplified Chinese and `zh-Hant` for Traditional Chinese were added, as indicated in the `relevanceLanguage` section of [search.list](https://developers.google.com/youtube/v3/docs/search/list).

### DIFF
--- a/src/renderer/components/PlayerSettings/PlayerSettings.vue
+++ b/src/renderer/components/PlayerSettings/PlayerSettings.vue
@@ -103,6 +103,14 @@
         @change="updateDefaultQuality"
       />
       <FtSelect
+        :placeholder="t('Settings.Player Settings.Default Language.Default Language')"
+        :value="defaultAudioLanguage"
+        :select-names="AUDIO_LANGUAGE_NAMES"
+        :select-values="AUDIO_LANGUAGE_VALUES"
+        :icon="['fas', 'language']"
+        @change="updateDefaultAudioLanguage"
+      />
+      <FtSelect
         :placeholder="t('Settings.Player Settings.Video Playback Rate Interval')"
         :value="videoPlaybackRateIntervalString"
         :select-names="PLAYBACK_RATE_INTERVAL_VALUES"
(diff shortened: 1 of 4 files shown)


absidue (changes requested):
The biggest issue with this pull request is that most people will want to keep using the original audio track on most videos and this pull request overrides it to something else every time. So this would be a breaking change that most people wouldn't want.

efb4f5ff-1298-471a-8973-3d47447115dc:
Im not sure if the following is a good idea at all but I would personally be more comfortable with following youtube's flow

That would essentially mean the following:
User that watches a video for the first time and hears original audio track -> user changes audio track to e.g. Spanish -> next video, it checks if Spanish audio track is available, if yes then use it, if no use original. 

Not sure if this is a good idea either but in addition to that we could store last used one track when you close FreeTube. So when you come back to it, it will try to find Spanish first and then switch to original if it cant find it. When close FreeTube it will store original because thats the last one used

burikoni:
hi @absidue I understand what you're saying. It's more a matter of preference; for people who aren't fluent in the original language, the alternative audio is preferable.
Ultimately, it's more about convenience, so you don't have to select the language for each video.

@efb4f5ff-1298-471a-8973-3d47447115dc On YouTube, it depends on the account preference; if you have the language set to Spanish, all videos will default to that language (if the track is available).

absidue:
While it is a matter of preference, the way that the setting is currently implemented is a breaking change that goes against what most FreeTube users probably want, as this pull request always selects a language, it does not provide an option to disable it/use the actual original audio track, which means it is actually an inconvenience for most users as they'll have to manually select the original audio track on every video.
So in its current state this pull request is definitely not mergeable, if it were improved maybe we could consider it.

burikoni:
> While it is a matter of preference, the way that the setting is currently implemented is a breaking change that goes against what most FreeTube users probably want, as this pull request always selects a language, it does not provide an option to disable it/use the actual original audio track, which means it is actually an inconvenience for most users as they'll have to manually select the original audio track on every video. So in its current state this pull request is definitely not mergeable, if it were improved maybe we could consider it.

Thanks for clarifying; I understand the issue you're describing now. Could a "prefer original audio" checkbox be checked by default, such that unchecking it enables the combo box for selecting the alternative audio? That way, it wouldn't break during updates.

absidue:
The dropdown should contain an entry to use the original audio track (explicitly track not language, as videos can also have audio descriptions in the original language which are not the original audio track), which should be the default value for that setting.

absidue:
Also reminds me that you should implement the audio track selection yourself instead of relying on shaka-player's preferred languages setting as you also need to ensure that you don't select an audio description or secondary audio track with this new setting and ideally this setting should only pick a different language if it was uploaded by the person that uploaded the video and avoid the AI/machine learning dubs created by YouTube.

burikoni:
> The dropdown should contain an entry to use the original audio track (explicitly track not language, as videos can also have audio descriptions in the original language which are not the original audio track), which should be the default value for that setting.

Why add "original audio" to the dropdown? It would make sense if that option were set as the default instead of using the checkbox.

> Also reminds me that you should implement the audio track selection yourself instead of relying on shaka-player's preferred languages setting as you also need to ensure that you don't select an audio description or secondary audio track with this new setting and ideally this setting should only pick a different language if it was uploaded by the person that uploaded the video and avoid the AI/machine learning dubs created by YouTube.

It is established that if the preferred audio isn't available, the original track is used—ignoring "Stable Volume" or "Voice Boost" variants; this would also apply to the descriptive audio track.

I don't understand the need to distinguish between user-uploaded dubbing and AI-generated dubbing. It could be a preference setting to avoid loading AI dubbing, but generally speaking, I don't see why they need to be separated.

absidue:
I see no reason to add a checkbox and a dropdown when both things can fit in the same dropdown.

> I don't understand the need to distinguish between user-uploaded dubbing and AI-generated dubbing. It could be a preference setting to avoid loading AI dubbing, but generally speaking, I don't see why they need to be separated.

YouTube's machine generated dubs were hated by quite a few FreeTube users when they were originally introduced and when there were bugs that made FreeTube accidentally select them by default instead of the original audio, users made it very clear that they still hated them.

Issue #9639, opened 2026-08-12
[Bug]: Content not loading


ArmAntenna:
### Guidelines

- [x] I have encountered this bug in the [latest release of FreeTube](https://github.com/FreeTubeApp/FreeTube/releases).
- [x] I have encountered this bug in the [official downloads of FreeTube](https://github.com/FreeTubeApp/FreeTube#official-downloads). **The only official downloads are from the [https://freetubeapp.io](https://freetubeapp.io/#download) website, the [FreeTubeApp/FreeTube GitHub repository](https://github.com/FreeTubeApp/FreeTube/releases) and the [io.freetubeapp.FreeTube flatpak](https://flathub.org/en/apps/io.freetubeapp.FreeTube) from Flathub.**
- [x] I have [searched the issue tracker for **open and closed** issues](https://github.com/FreeTubeApp/FreeTube/issues?q=is%3Aissue+sort%3Arelevance-desc) that are similar to the bug report I want to file, without success.
- [x] I have searched the [documentation](https://docs.freetubeapp.io/) for information that matches the description of the bug I want to file, without success.
- [x] This issue contains only one bug.
- [x] I have read and agree to follow the [rules](https://docs.freetubeapp.io/community/rules/).

### Describe the bug

Videos don’t play on FreeTube but do on Invidious. All videos stuck at 0:00. No SABR activity. This started with some videos and I could resolve by repeatedly trying to reload. Eventually, reloading many multiple of times does not work, except on a rare occasion. 

I have tried: redownloading FreeTube, toggling on/off or changed settings (local vs. invidious API, RSS yes/no, default quality level, Invidious instance, deleted hidden video list), VPN on/off, reset router.

No errors are displayed by FreeTube itself, but errors are being logged in the console itself:

renderer.js:2 
TypeError: Cannot read properties of undefined (reading 'height')
    at new dQ (renderer.js:2:2238407)
    at t.create (renderer.js:2:2346492)
    at renderer.js:2:782754
    at new Ax (renderer.js:2:782836)
    at Object.create (renderer.js:2:783843)
    at renderer.js:2:772047
    at kx.configure (renderer.js:2:772181)
    at $x.configure (renderer.js:2:796978)
    at Z (renderer.js:2:2326253)
    at renderer.js:2:2347668

### Expected Behavior

Thumbnails on subscription page show time, play when start is clicked after SABR

### Issue Labels

content not loading, feature stopped working

### FreeTube Version

[...]

PR #9644, opened 2026-08-14
Fix legacy as default format but unavailable won't fallback to other formats

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

https://github.com/FreeTubeApp/FreeTube/issues/9639

## Description

If default video format set as legacy and YT provides none, the handling is broken (no fallback)
This PR adds a proper fallback (to audio format per existing logic, let me know if it should fallback to dash

## Screenshots 

Will show a toast if legacy format unavailable on start
image

## Testing

- Go to Settings -> Player, set Default Video Format to Legacy Formats
- Play video without legacy format like https://youtu.be/frMR6fwXTiA

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
+++ b/src/renderer/components/ft-shaka-video-player/ft-shaka-video-player.js
@@ -1519,6 +1519,7 @@ export default defineComponent({
 
       /** @type {object[]} */
       const legacyFormats = props.legacyFormats
+      if (legacyFormats.length === 0) { return }
 
       const isPortrait = legacyFormats[0].height > legacyFormats[0].width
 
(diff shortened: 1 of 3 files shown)


absidue (commented):
I haven't had a chance to actually test these changes, so this might be a silly question: IIRC the watch page component controls which format is active, so if the watch page component ensures that the legacy formats cannot be selected when they are unavailable, why do we need changes in the player code too?

PikachuEXE:
I fixed those code and then the watch page still not loading video, the I went to fix watch page
So strictly speaking I think those changes are not needed and I can revert if you want

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
functionality wise seems good to me. ill wait on what absidue has to say on potential changes

absidue (approved):

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

radmorecameron (approved):

PR #9689, opened 2026-08-25
Fix SABR redirect handling not updating SABR URL which is actually used

### LINKED ISSUE #9677: [Bug]: SABR redirects are never followed — the redirect URL is assigned to a property nothing reads, ending in [BAD_HTTP_STATUS: 401]
### Describe the bug

`SabrSchemePlugin` never follows a SABR redirect: the redirect URL is assigned to a property that nothing reads.

https://github.com/FreeTubeApp/FreeTube/blob/9591c1771baef48ba8c0b3fc175f199de57d8fff/src/renderer/helpers/player/SabrSchemePlugin.js#L373-L380

```js
case UMPPartId.SABR_REDIRECT: {
  const sabrRedirect = decodePart(part, SabrRedirect)
  if (!sabrRedirect) break

  currentState.sabrUrl = sabrRedirect.url   // <- CurrentState has no sabrUrl
  shouldRetry = true
  break
}
```

The URL every request actually uses is `currentState.sabrStreamState.sabrUrl` — set at line 642, read at line 338 when the request is built. `CurrentState` (typedef, lines 44–58) has no `sabrUrl` member, so this assignment creates a dead property and the retry goes straight back to the host that just redirected us, which redirects again.

That retry is unbounded. Only the `NEXT_REQUEST_POLICY` path is capped (`cumulativeRetryDueToNextRequestPolicy >= 100`, line 329); the redirect path increments no counter at all. So the player loops on the same URL — roughly a hundred POSTs in a second and a half — until googlevideo cuts the session off with **401**, which reaches the user as a CRITICAL `[BAD_HTTP_STATUS: 401]` on the very first (init) segment.

Videos whose media sits on a host other than the one in `server_abr_streaming_url` are redirected every single time, so those videos fail deterministically and permanently, while everything else plays normally — which is probably why this has stayed quiet.

### Evidence

I hit this on a downstream fork and added logging to the refusal path (the plugin itself is unmodified upstream code). Three occurrences, all identical in shape — 401 on the init segment, seconds into a brand-new session, with the streaming URL nowhere near expiry:

| Date | Request counter at the 401 | Session age | Streaming URL expiry |
| --- | --- | --- | --- |
| 2026-08-16 | 241 | ~2 s | 5 h 59 m left |
| 2026-08-20 | 102, 102, 103 | ~1.5 s | 5 h 59 m left |
| 2026-08-22 | 103, 103, 103 | ~1.4 s | 5 h 59 m left |

The counter is `sabrStreamState.requestNumber`, which starts at 0 per session — so ~100 to 240 requests had already gone out before YouTube refused. The 08-16 sample of 241 also rules out the `NEXT_REQUEST_POLICY` retry as the source, since that path stops itself at 100; `SABR_REDIRECT` is the only other thing that sets `shouldRetry`.

Refetching the streaming data does not help — the fresh session is redirected too, and the redirect is dropped again. The same two video IDs failed on every attempt across a week and never played once.

### Expected Behavior

The redirect is followed: the session continues against the host named in the `SabrRedirect`, and the video plays.

### Suggested fix

One line:

```js
currentState.sabrStreamState.sabrUrl = sabrRedirect.url
```

For reference, the bundled `googlevideo` package does this correctly in both of its own implementations — `SabrStream.js` (`this.serverAbrStreamingUrl = sabrRedirect.url`) and `SabrStreamingAdapter.js`.

It may also be worth bounding the redirect retry the way the NextRequestPolicy retry is bounded, so that a genuine redirect loop reports itself instead of hammering googlevideo into refusing the session. I capped it at five redirects per request downstream, throwing a recoverable error past that, which lets shaka-player retry and otherwise falls through to the legacy formats.

### Issue Labels

Bug

### FreeTube Version

Reproduced on a fork built from `8388d0ce`; the code is unchanged on `development` @ `9591c177`. Present since SABR for VODs landed in #8047.

### Operating System Version

GNU/Linux (x86_64, Electron 43)

### Installation Method

Built from source (.deb)

### Primary API used

Local API

None — the line has been there since #8047.

PikachuEXE:
## Pull Request Type

- [x] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes https://github.com/FreeTubeApp/FreeTube/issues/9677

## Description

SABR set received URL but not to a property used in code

## Screenshots 

N/A

## Testing

No idea since redirect happens rarely but you can try those provided in comments in https://github.com/FreeTubeApp/FreeTube/issues/9677

## Desktop

- **OS:**
- **OS Version:**
- **FreeTube version:**

### DIFF
--- a/src/renderer/helpers/player/SabrSchemePlugin.js
+++ b/src/renderer/helpers/player/SabrSchemePlugin.js
@@ -374,7 +374,7 @@ async function doRequest(
             const sabrRedirect = decodePart(part, SabrRedirect)
             if (!sabrRedirect) break
 
-            currentState.sabrUrl = sabrRedirect.url
+            currentState.sabrStreamState.sabrUrl = sabrRedirect.url
             shouldRetry = true
             break
           }


absidue (approved):
Confirmed that the changes work by watching a video with my VPN set to France, then switching to New Zealand and checking in the devtools network tab that `/videoplayback` requests started going to a different hostname.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

radmorecameron (approved):

PR #9721, opened 2026-08-31
Add Shorten view count numbers using K, M, B suffixes

### LINKED ISSUE #9522: [Feature Request]: Shorten view count numbers using K, M, B suffixes
### Guidelines

### Problem Description

Since views are approximated anyway, the many zeroes at the end just waste space and distract. For example instead of `14,000` just show `14K` and for `3,500,000` show `3.5M`.

### Proposed Solution

Maybe add setting to enable/disable shortening

### Alternatives Considered

.

### Issue Labels

visual improvement

### FreeTube Version

v0.25.1 Beta

### Operating System Version

fedora 44

### Installation Method

Flathub

_No response_

### Additional Information

_No response_

ozrendev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes #9522 

## Description

Adds a helper function to detect if a number (view count) has been rounded
Calls this function when displaying view counts and shortens the number as per `Intl.NumberFormat` formatting if true

## Testing

Search for any video, see that returned search results view count is not shortened as they contain exact numbers
Watch any video, see recommended videos rounded view count have now been shortened

### DIFF
--- a/src/renderer/components/FtListVideo/FtListVideo.vue
+++ b/src/renderer/components/FtListVideo/FtListVideo.vue
@@ -163,7 +163,7 @@
           class="viewCount"
         >
           <template v-if="channelId !== null || channelName !== null"> • </template>
-          {{ t('Global.Counts.View Count', { count: isNumberRounded(viewCount) ? formatNumber(viewCount, { notation: 'compact' }) : parsedViewCount }, viewCount) }}
+          {{ t('Global.Counts.View Count', { count: viewCount % 1000 === 0 ? formatNumber(viewCount, { notation: 'compact' }) : parsedViewCount }, viewCount) }}
         </span>
         <span
           v-if="uploadedTime !== '' && !isLive"
@@ -296,8 +296,7 @@ import {
   showToast,
   toDistractionFreeTitle,
   deepCopy,
-  debounce,
-  isNumberRounded
+  debounce
 } from '../../helpers/utils.js'
 import { deArrowData, deArrowThumbnail } from '../../helpers/sponsorblock.js'
 import thumbnailPlaceholder from '../../assets/img/thumbnail_placeholder.svg'
--- a/src/renderer/components/FtListVideo/FtListVideo.vue
+++ b/src/renderer/components/FtListVideo/FtListVideo.vue
@@ -163,7 +163,7 @@
           class="viewCount"
         >
           <template v-if="channelId !== null || channelName !== null"> • </template>
-          {{ t('Global.Counts.View Count', { count: parsedViewCount }, viewCount) }}
+          {{ t('Global.Counts.View Count', { count: isNumberRounded(viewCount) ? formatNumber(viewCount, { notation: 'compact' }) : parsedViewCount }, viewCount) }}
         </span>
         <span
           v-if="uploadedTime !== '' && !isLive"
@@ -296,7 +296,8 @@ import {
   showToast,
   toDistractionFreeTitle,
   deepCopy,
-  debounce
+  debounce,
+  isNumberRounded
 } from '../../helpers/utils.js'
 import { deArrowData, deArrowThumbnail } from '../../helpers/sponsorblock.js'
 import thumbnailPlaceholder from '../../assets/img/thumbnail_placeholder.svg'
--- a/src/renderer/components/FtListVideo/FtListVideo.vue
+++ b/src/renderer/components/FtListVideo/FtListVideo.vue
@@ -163,7 +163,7 @@
           class="viewCount"
         >
           <template v-if="channelId !== null || channelName !== null"> • </template>
-          {{ t('Global.Counts.View Count', { count: viewCount % 1000 === 0 ? formatNumber(viewCount, { notation: 'compact' }) : parsedViewCount }, viewCount) }}
+          {{ t('Global.Counts.View Count', { count: compactViewCount }, viewCount) }}
         </span>
         <span
           v-if="uploadedTime !== '' && !isLive"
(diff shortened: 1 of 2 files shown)


src/renderer/helpers/utils.js
```diff
@@ -1113,3 +1113,11 @@ export function throttle(func, wait) {
     }
   }
 }
+
+/**
+ * @param {number} number
+ * @returns {boolean}
+ */
+export function isNumberRounded(number) {
+  if (number % 1000 === 0) return true
```
Shadorc: It does not return anything when false, should probably be return number % 1000 === 0 ?

Shadorc: Also, the function checks if the number is rounded to the nearest thousand, it should be part of the name imo, or it should be an additional parameter, because utils functions can be used in a lot of different scenarios

ozrendev: > Also, the function checks if the number is rounded to the nearest thousand, it should be part of the name imo, or it should be an additional parameter, because utils functions can be used in a lot of different scenarios

Yeah, you're right as a utils function it's not very flexible. I think having an additional parameter (equal to 1000 by default?) makes the most sense. 

Although I'm not sure if the function would have uses anywhere else though? Maybe it's better to just call it directly in `FtListVideo`?

Shadorc: On this project, I was told that if a function is used in one class, you make it local, if used in two classes, you duplicate it, and if used in three classes then you make it utils 

So, I would say directly in FtListVideo without parameter

Shadorc:
This is a little bit beyond the initial issue, but I remarked that comment likes have the same issue, they will be displayed as "202000" or "2200" instead of "202k" and "2k2" 
I'm wondering if this should be a separate issue or if this could be part of this one

ozrendev:
> This is a little bit beyond the initial issue, but I remarked that comment likes have the same issue, they will be displayed as "202000" or "2200" instead of "202k" and "2k2" I'm wondering if this should be a separate issue or if this could be part of this one

I think this would be within scope and I'm happy to add it. If we're doing this, then I feel we should make it consistent wherever the number is rounded.

I've also found it in these places, let me know if I missed anywhere:

- Comment likes
- Post likes
- Channel subscriber count under channel thumbnail on search
- Channel subscriber count on main channel view

Then as you suggested, it would make sense to have this as function with an argument for number of zeroes to check for. We would also have to decide at which point it should use short notation. For now it only catches a view count >= 10,000, while technically view counts in the thousands are also rounded.

If we lower this, the chances of a non-rounded, but matching number will be caught. For example in search results, one video may by chance have 1,234,500 views and be affected, while all others around it would not.

Shadorc:
There is something I'm missing, we have view count that are displayed with the same component but sometimes in places where YouTube shorten them and sometimes not, and there is no way to determine that?

I can't really search myself because I don't have access to a computer, but if the previous answer is yes, will it be possible to determine the places where YouTube shorten the counts, and pass a bool to the component to indicate that "this element contains only shorten count" ? 

For the other places where shorten count appear, I would agree to do the same, but I would prefer that a code owner answer.

absidue:
Okay so clearly someone else needs to step in again here and steer the two of you back into the correct direction.

Firstly @Shadorc I already said that saving extra data just for this was a bad idea in the linked issue, not only is it unnecessary because you can just do some math on the number to figure it out but would also mean that we would only display numbers that were saved after this pull request in compact notation. So please lets not go back to that.

Secondly @ozrendev most of that code should not be in the template, it makes it harder to read and also means it runs more often than necessary. Video gets marked as watched, that formatting code gets run again, video gets added or removed from a playlist, you guessed it it runs again, etc. This is a prime candidate for a computed ref.

As for the actual implementation of detecting which numbers should be compacted, which you are both stuck on, maybe I am missing something but, it should be as simple as this pseudo code:
```
if number >= 1000000000 // billion
  if last 8 digits are 0
    return true
  else
    return false
if number >= 1000000 // million
  if last five digits are 0
    return true
  else
    return false
if number >= 1000 // thousands
  if last two digits are 0
    return true
  else
    return false
```

Shadorc:
The possible issue with that was if we have for example three view counts that are not shorten 1234, 1200, 4567, the second one would be shorten. There is only 1% chance that it happens per video, si this could be considered minor enough to be ignored.

ozrendev:
I've made the above changes now. This logic works with way the view count is returned, I just used `switch` as I felt it's easier to read.

> The possible issue with that was if we have for example three view counts that are not shorten 1234, 1200, 4567, the second one would be shorten. There is only 1% chance that it happens per video, si this could be considered minor enough to be ignored.

I think with this approach it just may have to happen. When the view count gets higher, it becomes such a rare chance anyway.

src/renderer/components/FtListVideo/FtListVideo.vue
```diff
@@ -1064,6 +1064,33 @@ function parseVideoData() {
...
+  if (isNumberRounded(viewCount.value)) {
+    return formatNumber(viewCount.value, { notation: 'compact' })
+  } else {
+    return parsedViewCount.value
+  }
+})
+
+function isNumberRounded(number) {
+  switch (number.toString().length) {
+    case 10:
```
Shadorc, line 1077: The most viewed YouTube video is Baby Shark with 17B views, it is 11 numbers so it will fallback to "false" (but currently it seems that YouTube does not shorten it)

Shadorc (changes requested):

src/renderer/components/FtListVideo/FtListVideo.vue
ozrendev: I couldn't find where Baby Shark was suggested (and hence shortened) but have added that case now

src/renderer/components/FtListVideo/FtListVideo.vue
```diff
@@ -1064,6 +1064,34 @@ function parseVideoData() {
   }
 }
 
+const compactViewCount = computed(() => {
```
Shadorc, line 1067: Function name is `compactViewCount` but it does not always return the compact count, maybe `displayViewCount` or `formattedViewCount`?

Shadorc (changes requested):

Shadorc:
There is also a third test case, channel home page and videos tab have rounded views count, which is supported by this MR.

efb4f5ff-1298-471a-8973-3d47447115dc:
fyi, please dont forget to update the PR body with all the latest details so others dont have to go looking for it in the comments. in this case its mostly the additional testcases but the description probably also needs some tweaks

PR #9552, opened 2026-08-02
Fix: Retain pause state of video player upon SABR reload

dkarguth:
## Pull Request Type

- [X] Bugfix
- [ ] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/8977

## Description

Currently upon a SABR reload, the player will resume playback at the timestamp the player was at when the reload was initiated, but the play/pause state of the player is ignored.
This PR adds a mechanism to restore the play/pause state when a SABR reload is executed.

## Screenshots
No visible change

## Testing
Open a video that would normally result in a SABR reload, and immediately pause the playback. When the reload finishes, the player should remain paused.

## Desktop

- **OS:** Windows
- **OS Version:** 10
- **FreeTube version:**

PR #9722, opened 2026-08-31
Feat: Add a reload button to the player video info bar so you can recover after the player crashes

### LINKED ISSUE #1005: Add Reload button

mousetail:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

Closes #1005 

## Description

Currently, sometimes the player gets stuck (session timeout, crash, IP ban, etc.) Right now, the only way to recover from this is to go to your history page, find the video again, then load it. It's not a lot of effort but a bit annoying, especially when the video never actually played so it doesn't show in your history and you need to scroll through the channel page again. 

This adds a button next to the share and format options to reload the player.

## Screenshots 

image

## Testing

This is fairly easy to test. Just open a video, and click the new button and see the page reloads.

## Desktop

- **OS: ** Linux 
- **OS Version:** 7.0.10-1 MANJARO
- **FreeTube version:** Development

## Additional context

I discussed some thing about this in chat


src/renderer/views/Watch/Watch.js
```diff
@@ -1972,9 +1972,7 @@ export default defineComponent({
       this.startNextVideoInPip = uiState.startNextVideoInPip
     },
 
-    async onPlayerReloadRequested() {
-      showToast('Reloading player according to SABR request')
-
+    async saveWatchProgressForReload() {
```
Shadorc, line 1976: This function seems never called?

mousetail: Oops, fixed

radmorecameron (commented):
Haven't tested yet, my initial thought is this seems like a good compromise for the hacky "reload" being removed.

Without having tested this, I feel like our `debounce` function should probably be used for `onManualReloadReloadRequested`?
https://github.com/FreeTubeApp/FreeTube/blob/f7acbc729037869614796915721898f05a0cf530/src/renderer/helpers/utils.js#L1074

src/renderer/views/Watch/Watch.js
```diff
@@ -1991,6 +1989,16 @@ export default defineComponent({
           }
         }
       }
+    },
+
+    async onManualReloadReloadRequested() {
```
caetano-dev: Please, rename this to `onManualReloadRequested`.

caetano-dev (changes requested):

caetano-dev:
> Haven't tested yet, my initial thought is this seems like a good compromise for the hacky "reload" being removed.
> 
> Without having tested this, I feel like our `debounce` function should probably be used for `onManualReloadReloadRequested`?
> 
> https://github.com/FreeTubeApp/FreeTube/blob/f7acbc729037869614796915721898f05a0cf530/src/renderer/helpers/utils.js#L1074

Maybe the `throttle` function would be better, because it executes immediately. `debounce` has to wait after it is called.

src/renderer/views/Watch/Watch.js
mousetail: Fixed

src/renderer/views/Watch/Watch.js
```diff
@@ -1991,6 +1990,16 @@ export default defineComponent({
           }
         }
       }
+    },
+
+    onManualReloadRequested: throttle(async function () {
```
Shadorc, line 1995: Lambda functions are generally preferred in this codebase
```suggestion
    onManualReloadRequested: throttle(async () => {
```

Shadorc:
~Now that manually reloading the player is a valid feature, https://github.com/FreeTubeApp/FreeTube/pull/9552#issuecomment-5317102109 should probably be reconsidered? 
@efb4f5ff-1298-471a-8973-3d47447115dc~
On second thought, it does not seem useful if it is manually triggered

static/locales/en-US.yaml
```diff
@@ -909,6 +909,7 @@ Video:
   Published on: Published on
   Streamed on: Streamed on
   Started streaming on: Started streaming on
+  Reload Player: Reload Player
```
absidue: Refresh page would probably be a more appropriate name as it refreshed everything except the playlist and comments.

src/renderer/views/Watch/Watch.js
mousetail: The function uses `this`

static/locales/en-US.yaml
mousetail: I changed it to `Reload Page`

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -125,6 +125,15 @@
...
           />
+          <FtIconButton
+            class="navIconButton"
+            :title="t('Video.Reload Page')"
+            :icon="['fas', 'sync']"
+            :size="20"
+            :use-shadow="false"
+            theme="secondary"
+            @click="reloadPlayer"
+          />
```
caetano-dev: We can remove the size, since 20 is the default already, remove the unused class name and enable the shadow, which is on by default.

Screenshot to show the shadow difference:
Image

```suggestion
          <FtIconButton
            :title="t('Video.Reload Page')"
            :icon="['fas', 'sync']"
            theme="secondary"
            @click="reloadPlayer"
          />
```

caetano-dev (changes requested):

caetano-dev (dismissed):

Shadorc (dismissed):

efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
* Lets use this icon instead https://fontawesome.com/icons/classic/solid/rotate-right
* Please move the button to sit between Change Media Formats and Share so that options related to the player are grouped

mousetail:
> * Lets use this icon instead https://fontawesome.com/icons/classic/solid/rotate-right
> 
> * Please move the button to sit between Change Media Formats and Share so that options related to the player are grouped

Updated

caetano-dev (dismissed):

Shadorc:
I get this 
image

mousetail:
> I get this image

Fixed

After:

image

src/renderer/main.js
```diff
@@ -214,6 +215,7 @@ library.add(
   faPodcast,
   faQuestionCircle,
   faRandom,
+  faRotateRight,
```
Shadorc: > Please keep the list of constants sorted by name

It should be after `faRetweet`

Shadorc (changes requested):

src/renderer/main.js
mousetail: Fixed

Shadorc (dismissed):

src/renderer/main.js
```diff
@@ -122,7 +123,7 @@ import {
   faVolumeLow,
   faVolumeMute,
   faWifi,
-  faXmark
+  faXmark,
```
efb4f5ff-1298-471a-8973-3d47447115dc: why was this changed? It seems unnecessary to me so please revert it. Also dont force push, use commit instead so its easier to review whats changed

caetano-dev (dismissed):

src/renderer/main.js
mousetail: I force push then I fix a mistake in the previous commit. You can look here: https://github.com/FreeTubeApp/FreeTube/compare/f66d4cf187f570bd1612f31bcae5fef7e9ad6f73..91c6d4bdc57fcb4b711be10ef96a1d987f8a8475 to see what changed

Shadorc (dismissed):

caetano-dev (dismissed):

ArthurKun21:
Couldn't the "Reload" button be actually the same color as refresh in the subscriptions page?

image

efb4f5ff-1298-471a-8973-3d47447115dc:
> Couldn't the "Reload" button be actually the same color as refresh in the subscriptions page?

I dont think thats a good idea because that will make users associate it with a certain behavior even though the underlying functionality is totally different. 

Also that would break the color scheme that is being used by the buttons beside it

ArthurKun21:
it is just that, the placement was last place I would think of for reload.

Was originally gonna propose to add it on the top like similar to the browsers

image

But might seem like a bit of more work, so I just proposed to change the color which is less work to make it more noticeable

mousetail:
> Was originally gonna propose to add it on the top like similar to the browsers
> image
> 
> But might seem like a bit of more work, so I just proposed to change the color which is less work to make it more noticeable

This was discussed in the Matrix, but it doesn't fit that well because reload does something very different for different pages or is not relevant at all for most of them. The UX would be weird and in addition it needs a communication channel between the top bar and the page which breaks the seperation

Shadorc (approved):

PR #9749, opened 2026-09-05
Add "How this was made" disclosure and AI labels

### LINKED ISSUE #9746: [Feature Request]: Label AI content on Watch page
### Guidelines

### Problem Description

In FreeTube there isnt a label to quickly see if a video is made with AI. YouTube has started labeling AI content on the watch page. 

Examples:
* Video, https://youtu.be/7RogQiih3M8
* Short, https://youtu.be/9CaIQ-lNjlw

### Proposed Solution

Provide AI label on watch page.

### Alternatives Considered

None

### Issue Labels

### FreeTube Version

https://github.com/FreeTubeApp/FreeTube/commit/e856e36c11700dea38a855ccd1f4deb9346484b5

### Operating System Version

### Installation Method

.exe

_No response_

### Additional Information

_No response_

caetano-dev:
## Pull Request Type

- [ ] Bugfix
- [x] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue
closes #9746 
## Description
This PR adds a label for AI content in the watch page + "How this was made" disclosures in the description.
The label only shows up if the video has AI. The automatic dubbing and "Captured with a camera" videos show the disclosure only.

## Screenshots 
image

Pasted Graphic 3
Pasted Graphic 4
Pasted Graphic 1

Pasted Graphic 2

## Testing
Open https://www.youtube.com/watch?v=gfjgRHtDa38
Open https://youtu.be/7RogQiih3M8
Open https://youtu.be/BxunALc8r18
and see that the label + description shows up for the AI video. For the others, there should be the description only.

## Desktop
- **OS:** MacOS
- **OS Version:** 27
- **FreeTube version:** v0.25.3 Beta

## Additional context

Initially, I only planned to add the AI tags, but since I was already halfway through adding the other descriptions, I decided to implement both.


efb4f5ff-1298-471a-8973-3d47447115dc (changes requested):
* https://github.com/FreeTubeApp/FreeTube/pull/9749/changes#diff-4eaadcc25bfbd7ab1ba912e36d5e1bcbab346cab59546a970df7546d5fb7a145R895 doesnt seem to be used anywhere but if it was then it begs the question why only translate this but not captured with the camera and other things that can be listed there?
 
* I dont think adding the options to hide these is necessary so i would opt for removing those options

* When hovering the AI label i noticed that the tooltip is reusing the wording of the AI description from how its made. Use whats provided in the aria-label of the AI label instead

* Please put the How this was made section above the license like yt

Image

Image

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -247,6 +265,10 @@ const USING_ELECTRON = process.env.IS_ELECTRON
 
 const { locale, t } = useI18n()
 
+const showAiBadge = computed(() => Boolean(props.contentDisclosures?.hasAI))
```
Shadorc: It seems more common to use `props.contentDisclosures?.hasAI || false`  or `props.contentDisclosures?.hasAI ?? false`

caetano-dev: Suggestion applied.

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -59,9 +84,55 @@ const props = defineProps({
...
+}
+
+const DESCRIPTION_TRANSLATIONS = {
+  'Sounds or visuals were altered or significantly edited.': () => t('Video["Sounds or visuals were altered or significantly edited."]'),
+  'Sounds or visuals were altered or fully generated.': () => t('Video["Sounds or visuals were altered or fully generated."]'),
+  'Sounds or visuals were significantly edited or digitally generated.': () => t('Video["Sounds or visuals were significantly edited or digitally generated."]'),
+  'Sounds or visuals were altered or digitally generated.': () => t('Video["Sounds or visuals were altered or digitally generated."]'),
+  'Audio tracks for some languages were automatically generated.': () => t('Video["Audio tracks for some languages were automatically generated."]'),
+  'The creator used a camera or other recording device to capture this video without altering the sounds or visuals.': () => t('Video["The creator used a camera or other recording device to capture this video without altering the sounds or visuals."]'),
+}
```
caetano-dev, line 116: I added translations for all the possible headers and descriptions. I don't know if there is a more elegant way to do this, so I want to know what you all think.

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -247,6 +265,10 @@ const USING_ELECTRON = process.env.IS_ELECTRON
 
 const { locale, t } = useI18n()
 
+const showAiBadge = computed(() => props.contentDisclosures?.hasAI || false)
+
+const aiTooltipText = computed(() => props.contentDisclosures?.aiBadgeTooltip || 'AI: Content was made with AI')
```
Shadorc: It is only used for `title` and `aria-label`, they should probably be localized
I think you can simply use `t("AI: Content was made with AI")` or is there a benefit to use `props.contentDisclosures?.aiBadgeTooltip` ?

src/renderer/helpers/api/local.js
```diff
@@ -767,6 +767,68 @@ export async function getLocalVideoInfo(id) {
...
+  return aiBadge?.metadataBadgeRenderer?.accessibilityData?.label ?? null
+}
+
+/**
+ * Extracts content disclosures ('How this was made') from YouTube VideoInfo response
+ * @param {import('youtubei.js').YT.VideoInfo} info
+ * @param {object} [nextResponse]
+ * @returns {{ hasAI: boolean, aiBadgeTooltip: string|null, items: Array<{ label: string, description: string, isAI: boolean }> } | null}
+ */
+export function extractLocalContentDisclosures(info, nextResponse = null) {
```
Shadorc: Maybe this kind of parsing could benefit from being implemented into youtube.js directly

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
caetano-dev: The benefit would be to display the same label YouTube provides, in case it changes, but having it localized is more beneficial. I changed it.

src/renderer/helpers/api/local.js
caetano-dev: Hmm I see what you mean, but do you think this is a blocker for this PR?

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -135,7 +206,10 @@ function collapseDescription() {
  */
 function isShortDescription() {
   const descriptionElem = descriptionContainer.value?.$el
-  return descriptionElem?.clientHeight >= descriptionElem?.scrollHeight
+  if (!descriptionElem) {
+    return true
```
Shadorc: Isn't it changing the initial behavior?
If `descriptionElem` is undefined `descriptionElem?.clientHeight >= descriptionElem?.scrollHeight` would be `false`

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -247,6 +265,10 @@ const USING_ELECTRON = process.env.IS_ELECTRON
 
 const { locale, t } = useI18n()
 
+const showAiBadge = computed(() => props.contentDisclosures?.hasAI || false)
+
+const aiTooltipText = computed(() => t('Video["AI: Content was made with AI"]'))
```
Shadorc: This can now be inlined 😄 No need for a function anymore

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
Shadorc: I don't have a better idea
It does not seem too bad but I'm not used to javascript syntax, maybe there is some syntax sugar that could improve this

src/renderer/helpers/api/local.js
Shadorc: It's not a blocker for me, I don't know what is the general FreeTube philosophy for this kind of parsing

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
caetano-dev: Before, `<FtCard>` was only rendered when `shownDescription.length > 0`, so `descriptionElem` was never `undefined`.

Now a video can have disclosures without any description text, so returning `false` would render an "expand description" button and hide the disclosures.

src/renderer/helpers/api/local.js
```diff
@@ -767,6 +767,49 @@ export async function getLocalVideoInfo(id) {
...
+    contentDisclosures: extractLocalContentDisclosures(info),
+  }
+}
+
+/**
+ * Extracts content disclosures ('How this was made') from YouTube VideoInfo response
+ * @param {import('youtubei.js').YT.VideoInfo} info
+ * @returns {{ hasAI: boolean, items: Array<{ label: string, description: string, isAI: boolean }> } | null}
+ */
+export function extractLocalContentDisclosures(info) {
```
Shadorc, line 794: It feels a bit odd that "ai" disclosure has a specific treatment, and that it is specified twice with `isAI` and `hasAI`. 
I feel like this method should only extract the data without any assumption on the final usage, and then it's up to the caller to search if it is AI, or something else.

You could also return an empty array instead of null, it would simplify future usage like `props.contentDisclosures?.items?.length ?? 0) > 0`, but that's a very subjective opinion. I don't know what JS does with empty arrays, if it's allocating memory, then it's probably better to return null.

Previously, I suggested that this parsing could be part of youtube.js, but, in fact, it already does.
If you remove specific parsing of AI disclosure, you could simplify this method with something like this (which you can adapt as you want):

```js
/**
 * Extracts content disclosures ('How this was made') from YouTube VideoInfo response
 * @param {import('youtubei.js').YT.VideoInfo} info
 * @returns { { label: string, description: string }[] | null }
 */
export function extractLocalContentDisclosures(info) {
  if (info?.page?.[1]?.engagement_panels) {

    const structuredDescPanel = info.page[1].engagement_panels.find(
      panel => panel.target_id === 'engagement-panel-structured-description'
    )

    if (structuredDescPanel?.content?.is(YTNodes.StructuredDescriptionContent)) {
      return structuredDescPanel.content.items
        .filterType(YTNodes.HowThisWasMadeSectionView)
        .map((item => ({
          label: item.body_header?.text ?? '',
          description: item.body_text?.text?.replace(/\s*Learn more\.?$/i, '').trim() ?? ''
        })))
    }
  }

  return null
}
```

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -139,6 +210,13 @@ function isShortDescription() {
 }
 
 onMounted(() => {
+  // There is no description, so do not show the controls
+  if (shownDescription.length === 0) {
```
Shadorc, line 235: By curiosity, why is it needed? The else branch should result in the same thing, no?

Shadorc (dismissed):

src/renderer/components/WatchVideoDescription/WatchVideoDescription.css
```diff
@@ -51,3 +51,29 @@
...
+}
+
+.aiDisclosureHeader {
+  font-weight: 600;
+  font-size: 15px;
+  margin-block-end: 4px;
+}
+
+.aiDisclosureBody {
+  font-size: 14px;
```
efb4f5ff-1298-471a-8973-3d47447115dc: Minor nitpick but i would make it the same size as aiDisclosureHeader as 14px seems smallish to me

```suggestion
  font-size: 15px;
```

src/renderer/components/WatchVideoInfo/WatchVideoInfo.vue
```diff
@@ -8,10 +8,24 @@
...
+        >
+          {{ t('Video.Unlisted') }}
+        </div>
+        <div
+          v-if="showAiBadge"
+          class="aiBadge"
+          :title="t('Video.AI: Content was made with AI')"
+          :aria-label="t('Video.AI: Content was made with AI')"
+        >
+          <FontAwesomeIcon :icon="['fas', 'robot']" />
```
efb4f5ff-1298-471a-8973-3d47447115dc: While i personally really like the robot, i know that there are some people in this world that get really confused by fun type of labeling. So lets change it to what yt uses instead to save people's brain from exploding

```suggestion
          <FontAwesomeIcon :icon="['fas', 'circle-info']" />
```

src/renderer/main.js
```diff
@@ -90,6 +90,7 @@ import {
   faQuestionCircle,
   faRandom,
   faRetweet,
+  faRobot,
```
Shadorc: This is not needed anymore

src/renderer/main.js
```diff
@@ -215,6 +216,7 @@ library.add(
...
```
Shadorc: Same

Shadorc (changes requested):

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
caetano-dev: It doesn't result in the same thing because of how `isShortDescription()` evaluates when there is no description.
In this case, `<FtTimestampCatcher>` isn't rendered, so `descriptionContainer.value` is undefined. This makes `isShortDescription()` return `false`, which sets `showFullDescription = false` and `showControls = true`. 

Without this check, videos with disclosures but no description would display the "expand description" button unnecessarily.

Shadorc (dismissed):

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
Shadorc: Thanks for the explanation!

caetano-dev: No problem!

efb4f5ff-1298-471a-8973-3d47447115dc (commented):
I noticed that on YT that it stated 

Info from Truepic in the Captured with camera section

Image

Is that some info we are missing and we should include too?

Edit: Please update your PR body to reflect latest changes

caetano-dev:
I added the disclosure attributions and updated the PR body.
image

efb4f5ff-1298-471a-8973-3d47447115dc (dismissed):

Shadorc (dismissed):

src/renderer/helpers/api/local.js
```diff
@@ -61,6 +61,21 @@ if (process.env.SUPPORTS_LOCAL_API) {
   }
 }
 
+/**
+ * Extends YouTube.js's HowThisWasMadeSectionView to parse attributionText
+ * (e.g. 'Info from Truepic') which is present for verified authentic recordings.
+ */
+class CustomHowThisWasMadeSectionView extends YTNodes.HowThisWasMadeSectionView {
```
Shadorc, line 68: Are you ok if I open an MR to port these changes upstream to youtube.js, or do you prefer to do it yourself?

caetano-dev: Go ahead!

src/renderer/components/WatchVideoDescription/WatchVideoDescription.vue
```diff
@@ -59,9 +90,70 @@ const props = defineProps({
...
+}
+
+const DESCRIPTION_TRANSLATIONS = {
+  'Sounds or visuals were altered or significantly edited.': () => t('Video["Sounds or visuals were altered or significantly edited."]'),
+  'Sounds or visuals were altered or fully generated.': () => t('Video["Sounds or visuals were altered or fully generated."]'),
+  'Sounds or visuals were significantly edited or digitally generated.': () => t('Video["Sounds or visuals were significantly edited or digitally generated."]'),
+  'Sounds or visuals were altered or digitally generated.': () => t('Video["Sounds or visuals were altered or digitally generated."]'),
+  'Audio tracks for some languages were automatically generated.': () => t('Video["Audio tracks for some languages were automatically generated."]'),
+  'The creator used a camera or other recording device to capture this video without altering the sounds or visuals.': () => t('Video["The creator used a camera or other recording device to capture this video without altering the sounds or visuals."]'),
+}
```
absidue: I would prefer if we just showed the original texts from YouTube, rather than this fragile list of translations that replies on YouTube not rewording those strings. Sure we have parsing of translated texts in various places in FreeTube already but in most places it is just single words, so it is less likely to break, the only places we currently check for entire sentences is in error handling to show the user a nicer error message where we fallback to the original error message, I do not think we should base an entire feature on this.

Things break enough in FreeTube, due to YouTube changes as it is, I would prefer (and you probably too once you've been a maintainer long enough) to not go out of our way to implement something in a way that is going to require a bunch of maintainance (be that keeping up with new labels or adapting to text changes on the YouTube side).

src/renderer/components/WatchVideoDescription/WatchVideoDescription.css
```diff
@@ -51,3 +51,34 @@
...
+  opacity: 0.9;
+}
+
+.aiDisclosureAttribution {
+  margin-block-start: 6px;
+}
+
+.aiDisclosureItem + .aiDisclosureItem {
+  margin-block-start: 12px;
+}
```
absidue: Please rename these classes, they are not just for AI stuff.

src/renderer/helpers/api/local.js
```diff
@@ -61,6 +61,21 @@ if (process.env.SUPPORTS_LOCAL_API) {
...
+  constructor(data) {
+    super(data)
+    if (Reflect.has(data, 'attributionText')) {
+      this.attribution_text = Misc.Text.fromAttributed(data.attributionText)
+    }
+  }
+}
+
+Parser.addRuntimeParser('HowThisWasMadeSectionView', CustomHowThisWasMadeSectionView)
+
```
absidue: We should either wait for the YouTube.js update and get rid of this, which would be my preferred approach and how we have always done it in the past or add a big todo label to make sure we get rid of this ASAP, as this should not be on the FreeTube side.

caetano-dev:
Changes applied.

src/renderer/helpers/api/local.js
efb4f5ff-1298-471a-8973-3d47447115dc: I agree, lets draft this PR untill YouTube.js releases their update

Shadorc:
Blocked until youtube.js 18.1.0 is released https://github.com/LuanRT/YouTube.js/pull/1240

PR #9766, opened 2026-09-09
Update to pnpm 11 and enable dedupePeers

absidue:
## Pull Request Type

- [x] Other

## Description

We originally pinned the workflows to pnpm 10 as for a long time dependabot didn't support pnpm 11, however they have supported and been running with pnpm 11 for a few weeks now without any noticable issues, so I think it is time that we pin it to 11 instead. This also enables the `dedupePeers` setting that is new in pnpm 11 which cuts down on some of the information stored in the lock file for peer dependencies which should hopefully reduce conflicts. I chose 11 instead of 12 as dependabot runs 11 but if we think that it should be fine I am also happy to pin to 12 instead.

I also removed the `ci` script in the package.json file, as `pnpm run {x}` runs `pnpm install` internally, so we were running `pnpm install` + `pnpm install --frozen-lockfile` and replaced it with a direct call to `pnpm install --frozen-lockfile` in the workflow files.

## Testing

Successful build workflow:https://github.com/absidue/FreeTube/actions/runs/34990910624
The lint workflow on this pull request

## Desktop

- **OS:** Windows
- **OS Version:** 11


PikachuEXE:
I think we should also migrate to https://github.com/pnpm/setup
https://github.com/FreeTubeApp/FreeTube/pull/9579#issuecomment-5215104549

PikachuEXE:
Also no idea if there are still holes in pnpm 11 support, see https://github.com/dependabot/dependabot-core/issues/14794#issuecomment-5610159226

efb4f5ff-1298-471a-8973-3d47447115dc:
> Also no idea if there are still holes in pnpm 11 support, see [dependabot/dependabot-core#14794 (comment)](https://github.com/dependabot/dependabot-core/issues/14794#issuecomment-5610159226)

Feels like a point of concern to me if it just removes all the security alerts :/

absidue:
dependabot defaults to pnpm 11 at the moment so it is already using pnpm 11, we would need to force it to use pnpm 10 by setting the package manager in the package.json file if we want it to use pnpm 10. They definitely don't support using pnpm 12 like that yet based on the comments in the dependabot threat that pika linked (something about the new pnpm 12 installer not respecting dependabot's proxy settings).

absidue:
Also our project is not affected by that because our pnpm lockfile is a single yaml document, pnpm 11 only switches to a multi document one if you have special settings (e.g. requiring a specific version of pnpm in the package.json), pnpm 12 supposedly always uses that multi document format but i'm using pnpm 12 locally and it hasn't switched the format yet.

efb4f5ff-1298-471a-8973-3d47447115dc:
Okay, convinced :)

> I think we should also migrate to https://github.com/pnpm/setup [#9579 (comment)](https://github.com/FreeTubeApp/FreeTube/pull/9579#issuecomment-5215104549)

Should this be done here or separate PR?

PikachuEXE:
Hmm the only thing I can think of blocking the migration would be osx-x64
> One caveat: pnpm v11 publishes no binary for Intel macOS (darwin-x64); use v12 or newer on Intel macOS runners.

2026-09-11 14_53_02-pnpm_action-setup_ Install pnpm package manager - Vivaldi
2026-09-11 14_53_13-pnpm_setup_ Install pnpm and a JavaScript runtime (Node js, Bun, or Deno) in one

absidue:
The GitHub Actions macOS runners have been using ARM64 CPUs for a while now, so that shouldn't affect use in GitHub Actions, not sure if that is something that would affect local development of FreeTube though.

PikachuEXE:
> not sure if that is something that would affect local development of FreeTube though

Local dev with intel macOS can still install pnpm 11 I guess... or wdym

absidue:
Okay then it doesn't seem like a concern for us.

PikachuEXE:
I can make a separate PR for the migration if you prefer
Otherwise just migrate

absidue:
> I can make a separate PR for the migration if you prefer
Otherwise just migrate

Which migration? What is not done yet in this PR that you want to do in a separate PR?

efb4f5ff-1298-471a-8973-3d47447115dc:
@absidue https://github.com/FreeTubeApp/FreeTube/pull/9766#issuecomment-5609531572

absidue:
As there will almost certainly be conflicts with all the dependabot PRs, I'll rebase this PR/redo the lockfile changes afterwards and do that pnpm setup migration then (looks like we'll want to turn off a bunch of the enabled by default config options in that action, so I'll need to experiment in my fork first).

efb4f5ff-1298-471a-8973-3d47447115dc:
There was allot of activity in the recent hours on v12 support. 1 PR needs to be merged and seems close to me
https://github.com/dependabot/dependabot-core/issues/14794#issuecomment-5532722237

absidue:
Switched to pnpm/setup but only using it to setup pnpm and I also switched from `pnpm run ci` to directly running `pnpm install --frozen-lockfile` in the workflows as `pnpm run {x}` runs `pnpm install` internally so we were previously running `pnpm install` + `pnpm install --frozen-lockfile`.

PikachuEXE:
pnpm/setup has `require-lockfile` option to make it run `pnpm install --frozen-lockfile`

absidue:
As we restore the cache after installing pnpm we cannot use that option. We manage the cache ourselves to have full control over when we ignore it, only read it or read and write to it.

absidue:
`pnpm/setup` writes `pnpm-lockfile-verified-*` entries to the GitHub actions cache even when `cache: false` (which based on their source code is intentional behaviour), I suggest that we stick with `pnpm/action-setup` as that doesn't have that issue, it also doesn't require us to explicitly turn off other things like the automatic `pnpm install` and it fully supports pnpm 11 and pnpm 12.

When pika originally suggested that we switch to `pnpm/setup` the `pnpm/action-setup` action didn't support pnpm 11 or 12 yet, but now it does additionally `pnpm/action-setup` is a more minimal action that does what we want, instead of the magic do everything without letting you control everything action that `pnpm/setup` is.

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Pnpm 12 support just has been added do we want to make the jump or stick with 11?

absidue:
Stick with 11 for now, the last dependabot-core release was before the last pnpm 12 relevant PR was merged in the dependabot-core repo, so we should definitely wait until they have rolled out pnpm 12 support in dependabot before bumping. So I would say get this merged first and then bump to pnpm 12 later once dependabot supports it, rather than leaving this PR open until then and having to deal with the conflicts every week with the dependabot PRs.

efb4f5ff-1298-471a-8973-3d47447115dc (approved):

Shadorc (approved):

PR #9768, opened 2026-09-09
Hoist some v-ifs and use v-else and v-else-if

absidue:
## Pull Request Type

- [x] Performance improvement

## Description

_I recommend turning on Hide Whitespace while reviewing this pull request, as various code blocks have different indentation._

This pull request implements some neglible performance improvements in the form of some v-if, v-else and v-else-if optimisations. Vue always uses tenaries for v-if, so if we have two consecutive v-ifs with inverted conditions, that means we'll get two tenaries that are both `condition ? component : ""`, whereas replacing it with a v-if reduces that to a single tenary without the empty string resulting in less branching. Am I expecting noticable performance improvements, no, but historically my rule has been that every little helps (appart from ones where I have managed to locate a major improvement). Faster doesn't hurt as long as it doesn't break anything.

## Testing

Test that the modified components still work e.g. visiting different types of post,s visiting the history and playlist pages, etc.

## Desktop

- **OS:** Wiondows
- **OS Version:** 11

### DIFF
--- a/src/renderer/components/PlaylistInfo/PlaylistInfo.vue
+++ b/src/renderer/components/PlaylistInfo/PlaylistInfo.vue
@@ -164,14 +164,14 @@
               theme="secondary"
               @click="enterEditMode"
             />
-            <FtIconButton
-              v-if="videoCount > 0 && showPlaylists"
-              :title="$t('User Playlists.Copy Playlist')"
-              :icon="['fas', 'copy']"
-              theme="secondary"
-              @click="toggleCopyVideosPrompt"
-            />
           </template>
+          <FtIconButton
+            v-if="videoCount > 0 && showPlaylists && !editMode"
+            :title="$t('User Playlists.Copy Playlist')"
+            :icon="['fas', 'copy']"
+            theme="secondary"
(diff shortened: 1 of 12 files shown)


src/renderer/components/PlaylistInfo/PlaylistInfo.vue
```diff
@@ -130,44 +130,48 @@
...
+              @click="enableQuickBookmarkForThisPlaylist"
+            />
+            <FtIconButton
+              :title="$t('User Playlists.Edit Playlist Info')"
+              :icon="['fas', 'edit']"
+              theme="secondary"
+              @click="enterEditMode"
+            />
+            <FtIconButton
+              v-if="videoCount > 0 && showPlaylists"
```
radmorecameron: looks like this one didn't have an isUserPlaylist check but now does? (Reviewing on mobile website so can't verify right now)

absidue: Fixed

PikachuEXE (approved):

radmorecameron (approved):

Shadorc (approved):

PR #9791, opened 2026-09-13
Feat: Improve messaging when all channel videos are filtered out

### LINKED ISSUE #9678: [Feature Request]: Improve messaging when all channel videos are filtered out
### Guidelines

### Problem Description

1. Go to https://www.youtube.com/channel/UC2EiGVmCeD79l_vZ204DUSw/videos
2. See empty page with message `This channel does not currently have any videos`
3. Click on `Fetch more results`
4. See more content being loaded

### Proposed Solution

The `This channel does not currently have any videos` should be changed when there is a `Fetch more results` button available 

### Alternatives Considered

None

### Issue Labels

### FreeTube Version

v0.25.2-nightly-7648 Beta

### Operating System Version

### Installation Method

.exe

_No response_

### Additional Information

_No response_

Shadorc:
## Pull Request Type

- [ ] Bugfix
- [X] Feature Implementation
- [ ] Documentation
- [ ] Other

## Related issue

closes https://github.com/FreeTubeApp/FreeTube/issues/9678

## Description

Add a specific message when a channel contains videos but they were all filtered. This can happen if videos are for members only.
I have removed `This channel does not currently have any videos` key because it can't be displayed. Videos tab is not displayed if there are no videos, see https://github.com/FreeTubeApp/FreeTube/blob/development/src/renderer/views/Channel/Channel.vue#L792

## Screenshots 

image

## Testing

1. Go to https://www.youtube.com/channel/UC2EiGVmCeD79l_vZ204DUSw/videos
2. Check that the message is "This channel has videos but none could be displayed, try fetching more"
3. Click on Fetch more results
4. See more content being loaded

## Desktop

- **OS:** Bazzite
- **OS Version:**
- **FreeTube version:** v0.25.3-beta


efb4f5ff-1298-471a-8973-3d47447115dc:
Slightly related, i just noticed that on YouTube, Its possible to select `Public` videos on the videos tab. Would it be possible to directly fetch Public videos for displaying it on the videos tab instead of fetching all and then filtering them?

Edit: What this PR is doing is already a good improvement but its also possible that there are e.g 6 pages full of member only content before public videos are getting displayed. So clicking 6 times on a fetch more button isnt a very good UX

src/renderer/views/Channel/Channel.vue
```diff
@@ -112,7 +112,10 @@
           v-if="currentTab === 'videos' && latestVideos.length === 0"
         >
           <p class="message">
-            {{ $t("Channel.Videos.This channel does not currently have any videos") }}
+            {{ showFetchMoreButton
+              ? $t("Channel.Videos.This channel has videos but none could be displayed, try fetching more")
+              : $t("Channel.Videos.This channel does not currently have any videos")
```
efb4f5ff-1298-471a-8973-3d47447115dc: Dont we filter out tabs when there isnt any content in them? If yes then is it really useful to keep this around?

Shadorc: Indeed, it is hidden (e.g. https://www.youtube.com/@Novideosofficial-n5s) 
It could be removed

Shadorc:
@efb4f5ff-1298-471a-8973-3d47447115dc It can't be done directly, it requires one additional request.
It could be done in `Channel.vue` line 1141, when we detect that no videos were retrieved but more are available.
What do you think?

Code snippet prototype:
```js
if (latestVideos.value.length === 0 && videosTab.has_continuation) {
  const get_public_videos = videosTab.memo.getType(YTNodes.ChipView).filter(item => item.text === "Public")[0].tap_command
  if (get_public_videos.is(YTNodes.NavigationEndpoint)) {
    const result = await get_public_videos.call(videosTab.actions, { parse: true });
    console.log('public videos', result.on_response_received_actions_memo.getType(YTNodes.LockupView))
  }
}
```

src/renderer/views/Channel/Channel.vue
Shadorc: This is filtered here https://github.com/FreeTubeApp/FreeTube/blob/dddbe8cdf41ee14af199e6a5b5311296eb2a7488/src/renderer/views/Channel/Channel.vue#L792

efb4f5ff-1298-471a-8973-3d47447115dc:
Lets wait for the Code Owners to chime in on it

src/renderer/views/Channel/Channel.vue
```diff
@@ -109,10 +109,10 @@
           aria-labelledby="videosTab"
         />
         <FtFlexBox
-          v-if="currentTab === 'videos' && latestVideos.length === 0"
+          v-if="currentTab === 'videos' && latestVideos.length === 0 && showFetchMoreButton"
```
PikachuEXE, line 112: What to show when `!showFetchMoreButton`?

Shadorc: You have the corresponding messages here: https://github.com/FreeTubeApp/FreeTube/pull/9791#discussion_r4000025832
TL;DR: This scenario does not exist

PikachuEXE (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
@PikachuEXE any thoughts on https://github.com/FreeTubeApp/FreeTube/pull/9791#issuecomment-5656524916

PikachuEXE:
I think that belongs to another PR and potentially requires changes to YTJS (better not do too much "internal stuff"

Shadorc:
It indeed could be part of a separate MR. Also, I've missed it, but it is already implemented in ytjs (it is two requests, I would have to dig deeper to only apply the second filter as we don't need the first one, if we chose this):
```js
if (latestVideos.value.length === 0 && videosTab.has_continuation) {
  console.log('public videos', await videosTab.applyFilter('Latest', 'Public'))
}
```

PR #9797, opened 2026-09-13
Use `list` icon instead of `bookmark` to represent Playlists

efb4f5ff-1298-471a-8973-3d47447115dc:
## Pull Request Type

- [x] Change that more opinionated than fact based

## Description

Imo this icon represents what a Playlist is better then the Bookmark one. Also Playlist thumbnails within FreeTube already use this icon to indicate that it is a playlist.

## Testing

- Test how icon looks like in collapsed (default) sidebar view
- Test how icon looks like in expanded sidebar view
- Test how icon looks like on smaller window sizes
- Test how icon looks like in all views when `Theme -> Hide Side Bar Labels` is enabled
- Go to User Playlist and see that new icon is being displayed


radmorecameron:
Heh, I was just looking at the icon and thinking "hmm, should we use a different icon instead of the list icon for podcasts and shows"

radmorecameron (dismissed):

Shadorc:
I prefer this icon too!
While testing, I discovered that small window acts weirdly with the "Hide Side Bar Labels" settings (it works fine in FreeTubeAndroid)

**Not hidden**
image

**Hidden**
image

Update : This only happens when "Expand Side Bar by Default" is enabled

Shadorc (dismissed):

PikachuEXE:
Well should we keep the icon on playlist index page? (red
image

Shadorc (approved):

PikachuEXE (approved):

absidue (approved):

